Issue #23, opened 2026-04-08
Add Adaptive Multivariate Partitioning (AMP) global MINLP solver


bernalde:
## Summary

Implement the AMP (Adaptive Multivariate Partitioning) algorithm from Nagarajan et al. (CP 2016, JOGO 2018) as a certified global solver for nonconvex polynomial MINLPs.

References:
- [CP 2016](http://harshangrjn.github.io/pdf/CP_2016.pdf): Tightening McCormick Relaxations via Dynamic Multivariate Partitioning
- [JOGO 2018](http://harshangrjn.github.io/pdf/JOGO_2018.pdf): An Adaptive, Multivariate Partitioning Algorithm for Global Optimization

Reference implementation: [Alpine.jl](https://github.com/lanl-ansi/Alpine.jl)

## Algorithm

```
presolve (OBBT+FBBT, term classification, variable selection)
for k in 1..max_iter:
    MILP bounding (piecewise McCormick) → LB_k
    NLP local solve (fixed integers) → UB_k
    if (UB_k - LB_k)/|UB_k| ≤ rel_gap → CERTIFIED OPTIMAL
    adaptive partition refinement around MILP solution
```

Soundness invariant: `LB_k ≤ global_opt ≤ UB_k` at every iteration.

## What is implemented (bernalde/discopt fork, branch `amp-global-solver`)

- `_jax/term_classifier.py` — bilinear/monomial/trilinear term detection from expression DAG
- `_jax/partition_selection.py` — max-cover and min-vertex-cover variable selection (Alpine `disc_var_pick=0,1`); verified against Alpine's `candidate_disc_vars`/`disc_vars` counts for nlp3
- `_jax/discretization.py` — partition state management, adaptive refinement (matches Alpine's `add_adaptive_partition` in `bounding_model.jl`)
- `_jax/milp_relaxation.py` — piecewise McCormick MILP lower-bound builder (binary δ_k formulation); monotone lower bounds guaranteed
- `solvers/amp.py` — AMP algorithm orchestrator; OA cuts via existing `cutting_planes.py` infrastructure
- `solver.py` / `modeling/core.py` — `Model.solve(solver="amp", ...)` routing
- `tests/test_amp.py` — full test suite; 52/53 non-slow tests passing; Alpine nlp1/circle/nlp3 verified

## Remaining gaps (not yet implemented)

See linked issue in bernalde/discopt for details:

[...]

Issue #21, opened 2026-04-07
AMP Phase 4: MILP Relaxation Builder (milp_relaxation.py)


bernalde:
## Context

Part of AMP implementation. Tracked in #17. Depends on #18, #19, #20.

This is the core of the AMP algorithm. Given a Model, its NonlinearTerms, and a DiscretizationState, build a MILP that is a valid **lower-bounding relaxation** of the original problem.

## Theory

From CP 2016 and JOGO 2018: for each bilinear term `w = x*y` partitioned into k intervals on x, introduce:
- Binary indicators `δ_i ∈ {0,1}` (which partition is active)
- Tightened McCormick inequalities using partition bounds `[p_i, p_{i+1}]`

The piecewise McCormick MILP gives a tighter lower bound than standard McCormick, with gap O((domain_width/k)²).

## Specification

**New file**: `python/discopt/_jax/milp_relaxation.py`

```python
def build_milp_relaxation(
    model: Model,
    terms: NonlinearTerms,
    disc_state: DiscretizationState,
    incumbent: np.ndarray | None = None,
) -> tuple[object, dict]:
    """Build a MILP lower-bounding relaxation of the model.
    
    Returns (milp_model, varmap) where milp_model.solve() gives a valid LB.
    """
```

**Implementation steps**:
1. Copy all linear constraints directly
2. For each bilinear term `x_i * x_j`:
   - If x_i is in `disc_state.partitions`: add piecewise McCormick per partition
   - Otherwise: add standard McCormick envelope constraints
3. For monomials `x_i^n` with n even: convex, add secant overestimator
4. For trilinear terms: hierarchical decomposition (w1=x*y, then w1*z)
5. For general_nl terms: add OA tangent cuts from incumbent (if available)
6. Introduce auxiliary variable `w` for each lifted product term

**Reuse**:
- `mccormick.py:relax_bilinear()` for McCormick coefficients
- `solvers/milp_highs.py` for MILP solving
- `piecewise_mccormick.py:_partition_bounds()` for partition edges

## Tests

```bash
uv run pytest python/tests/test_amp.py::TestMilpRelaxation -v
```

Key tests: LB ≤ known global optimum for nlp1 and circle; LB non-decreasing with finer partitions; relaxation always feasible when original is feasible.

bernalde:
Consolidated into a single AMP tracking issue. Development is in progress on the bernalde/discopt fork (branch amp-global-solver).

Issue #18, opened 2026-04-07
AMP Phase 1: Nonlinear Term Classifier (term_classifier.py)


bernalde:
## Context

Part of the Adaptive Multivariate Partitioning (AMP) global MINLP solver implementation. Tracked in #17.

AMP (JOGO 2018, CP 2016) needs a problem-level catalog of bilinear, trilinear, and monomial terms to:
1. Decide which variables to partition (variable selection)
2. Build the MILP relaxation (which nonlinear terms get McCormick inequalities)
3. Compute the term-variable incidence for min-vertex-cover

## Specification

**New file**: `python/discopt/_jax/term_classifier.py`

```python
from dataclasses import dataclass, field
from discopt.modeling.core import Model

@dataclass
class NonlinearTerms:
    bilinear: list[tuple[int, int]] = field(default_factory=list)
    trilinear: list[tuple[int, int, int]] = field(default_factory=list)
    monomial: list[tuple[int, int]] = field(default_factory=list)
    general_nl: list = field(default_factory=list)
    term_incidence: dict[int, set[int]] = field(default_factory=dict)
    partition_candidates: list[int] = field(default_factory=list)

def classify_nonlinear_terms(model: Model) -> NonlinearTerms: ...
```

**Algorithm**: Recursively walk expression DAG in all constraints + objective. Classify nodes:
- `BinaryOp("*", Variable, Variable)` → bilinear (indices of both variables)
- `BinaryOp("**", Variable, Constant(n))` where n integer ≥ 2 → monomial
- `BinaryOp("*", BinaryOp("*", Variable, Variable), Variable)` → trilinear
- `FunctionCall(sin/cos/exp/log/sqrt/...)` → general_nl
- `IndexExpression(Variable, idx)` must be resolved to flat variable index

**Reuse**:
- `relaxation_compiler.py:_try_extract_signomial_factors()` for initial pattern recognition
- `problem_classifier.py` for variable index extraction patterns

## Tests that must pass

```bash
uv run pytest python/tests/test_amp.py::TestTermClassifier -v
```

9 tests covering: bilinear/monomial/trilinear detection, partition_candidates correctness, transcendental classification, empty model, term_incidence mapping.

Issue #17, opened 2026-04-07
AMP Phase 0: Test suite for Alpine.jl-equivalent global MINLP (TDD baseline)


bernalde:
## Summary

This issue tracks the TDD test suite created in `python/tests/test_amp.py` that establishes the baseline and drives the AMP implementation.

Running `uv run pytest python/tests/test_amp.py -v` currently shows:
- **29 errors**: ImportError for modules not yet created  
- **5 failures**: Actual bugs in existing code revealed by the tests
- **20 passing**: Existing code that is already correct

## Revealed bugs / gaps in current code

1. **`circle` problem returns local optimum 2.0 instead of global optimum √2≈1.414**  
   The constraint `x₀²+x₁²≥2` is nonconvex (feasible region = complement of a ball). The spatial B&B finds the local stationary point at (1,1) instead of the global at (√2,0) or (0,√2). This is the core motivation for AMP.

2. **`nlp3` multilinear problem returns `x[0]=100` (lower bound) instead of 7049.2**  
   The 8-variable multilinear process problem is solved incorrectly.

3. **`relaxation_compiler` API**: Fixed in the test — `compile_relaxation(expr, model, ...)` takes expression as first arg.

## Test file

`python/tests/test_amp.py` — 54 tests across 8 sections:
- Section 1: `TestTermClassifier` (9 tests) — drives `term_classifier.py`
- Section 2: `TestPartitionSelection` (8 tests) — drives `partition_selection.py`
- Section 3: `TestDiscretizationState` (8 tests) — drives `discretization.py`
- Section 4: `TestMcCormickSoundness` (5 tests) — validates existing relaxation correctness
- Section 5: `TestMilpRelaxation` (4 tests) — drives `milp_relaxation.py`
- Section 6: `TestAmpEndToEnd` (5 tests) — drives `solvers/amp.py`
- Section 7: `TestAmpConvergenceProperties` (6 tests) — convergence invariants
- Section 8: `TestCurrentCodeWeaknesses` (9 tests) — probes existing gaps

## References

- CP 2016: http://harshangrjn.github.io/pdf/CP_2016.pdf
- JOGO 2018: http://harshangrjn.github.io/pdf/JOGO_2018.pdf
- Alpine.jl: https://github.com/lanl-ansi/Alpine.jl

PR #44, opened 2026-04-21
Add AMP global MINLP solver

bernalde:
## Summary

Adaptive Multivariate Partitioning (AMP) is a global MINLP algorithm for nonconvex mixed-integer nonlinear problems. At each iteration it builds a MILP relaxation over a partitioned variable domain to obtain a lower bound, solves a local NLP with the relaxed solution as a starting point to obtain an upper bound or incumbent, and then refines the partitions around the most relevant nonlinear variables. Repeating that loop gives a convergent lower/upper-bound process rather than a single-shot local heuristic.

This draft PR reconstructs that AMP implementation from the fork onto a clean branch off `main`, without the fork-local benchmark tooling, archived benchmark artifacts, or planning files.

Included here:

- solver routing for `solver="amp"`
- nonlinear term classification
- partition-variable selection
- discretization state management
- MILP bounding-model construction
- the AMP main loop and option surface through the current phase-2 work
- focused `python/tests/test_amp.py` coverage for the reconstructed implementation
- the shared flat-bound helper in `python/discopt/_jax/model_utils.py`

This PR intentionally does not include the convexity-detector port. That work is split into a separate draft PR against `feat/convexity-detector`.

## Current status

This is a draft because the core AMP branch is ready for upstream review, but the fork still has a second layer of benchmark-driven follow-up work that is better tracked explicitly than folded into one large PR.

The most important resolved fork tracker is:

- [bernalde/discopt#19](https://github.com/bernalde/discopt/issues/19) `CLOSED` — the false-`infeasible` translated MINLPTests class was eliminated on the fork's follow-up branch; there are no remaining false-`infeasible` cases there, only four residual timeout cases

The remaining fork AMP follow-up issues are:

- [bernalde/discopt#2](https://github.com/bernalde/discopt/issues/2) `OPEN` — older umbrella backlog for AMP improvements; useful as historical context, but now partly superseded by the narrower issues below
- [bernalde/discopt#15](https://github.com/bernalde/discopt/issues/15) `OPEN` — Alpine-style incumbent local solve and warm-start policy; good candidate to move upstream after the core AMP PR lands
- [bernalde/discopt#16](https://github.com/bernalde/discopt/issues/16) `OPEN` — Alpine-style presolve OBBT variants around feasible incumbents; good candidate to move upstream after core AMP support is merged
- [bernalde/discopt#17](https://github.com/bernalde/discopt/issues/17) `OPEN` — shared nonlinear bound tightening extensions; likely upstream-worthy, but broader than AMP alone
- [bernalde/discopt#18](https://github.com/bernalde/discopt/issues/18) `OPEN` — partition-selection and refinement hooks; upstream-worthy as an AMP extension after the core PR
- [bernalde/discopt#20](https://github.com/bernalde/discopt/issues/20) `OPEN` — Rust performance follow-up; should probably stay local until upstream AMP behavior is settled and profiled there
- [bernalde/discopt#21](https://github.com/bernalde/discopt/issues/21) `OPEN` — operator-specific AMP relaxations for `sqrt`, `log`, `exp`, `abs`, etc.; strong candidate to move upstream after the core AMP PR
- [bernalde/discopt#22](https://github.com/bernalde/discopt/issues/22) `OPEN` — stronger benchmark-side solution validation; likely fork-local unless upstream adopts the same MINLPTests harness semantics

I would treat `#15`, `#16`, `#17`, `#18`, and `#21` as the natural follow-up issue set to migrate or re-open upstream once the base AMP implementation is under review.

## Validation

- `PYTHONPATH=python /home/bernalde/repos/discopt/.venv/bin/pytest python/tests/test_amp.py -q -k 'solve_model_forwards_alpine_amp_aliases or lambda_convhull_builds_expected_auxiliaries or bilinear_maximize_global_optimum'`
- `/home/bernalde/repos/discopt/.venv/bin/ruff check python/discopt/_jax/discretization.py python/discopt/_jax/milp_relaxation.py python/discopt/_jax/model_utils.py python/discopt/_jax/partition_selection.py python/discopt/_jax/term_classifier.py python/discopt/solver.py python/discopt/solvers/amp.py python/tests/test_amp.py`
- `python -m py_compile python/discopt/_jax/discretization.py python/discopt/_jax/milp_relaxation.py python/discopt/_jax/model_utils.py python/discopt/_jax/partition_selection.py python/discopt/_jax/term_classifier.py python/discopt/solver.py python/discopt/solvers/amp.py python/tests/test_amp.py`

## Scope exclusions

This PR does not include:

- fork-local benchmark scripts or archived benchmark outputs
- the convexity-detector redesign work
- the later benchmark-driven AMP follow-up fixes from the fork's PR #23

Those are intentionally left to the separate convexity PR and to narrower follow-up AMP issues.

### DIFF
--- a/python/discopt/_jax/convexity.py
+++ b/python/discopt/_jax/convexity.py
@@ -22,6 +22,7 @@ Key composition rules implemented:
 
 from __future__ import annotations
 
+from dataclasses import dataclass
 from enum import Enum
 from typing import Optional
 
@@ -51,6 +52,14 @@ class Curvature(Enum):
     CONVEX = "convex"
     CONCAVE = "concave"
     UNKNOWN = "unknown"
+
+
+@dataclass(frozen=True)
+class OACutConvexity:
+    """Convexity information relevant to globally valid OA cuts."""
+
+    objective_is_convex: bool
+    constraint_mask: list[bool]
 
 
 def _combine_sum(a: Curvature, b: Curvature) -> Curvature:
@@ -363,17 +372,10 @@ def classify_constraint(
     return False
 
 
-def classify_model(model: Model) -> tuple[bool, list[bool]]:
-    """Classify a model's convexity.
-
-    Returns:
-        (is_convex, constraint_convexity_mask)
-        - is_convex: True if objective is convex and all constraints are convex
-        - constraint_convexity_mask: per-constraint convexity flags
-    """
+def classify_oa_cut_convexity(model: Model) -> OACutConvexity:
+    """Return the convexity information needed to generate sound OA cuts."""
     cache: dict = {}
 
-    # Check objective
     obj_convex = True
     if model._objective is not None:
         from discopt.modeling.core import ObjectiveSense
@@ -385,17 +387,28 @@ def classify_model(model: Model) -> tuple[bool, list[bool]]:
             # Maximize f  ≡ Minimize -f; convex if f is concave
             obj_convex = obj_curv in (Curvature.CONCAVE, Curvature.AFFINE)
 
-    # Check constraints
     constraint_mask = []
-    all_convex = obj_convex
     for c in model._constraints:
         if isinstance(c, Constraint):
-            is_cvx = classify_constraint(c, model, cache)
-            constraint_mask.append(is_cvx)
-            if not is_cvx:
-                all_convex = False
+            constraint_mask.append(classify_constraint(c, model, cache))
         else:
             constraint_mask.append(False)
-            all_convex = False
 
-    return all_convex, constraint_mask
+    return OACutConvexity(
+        objective_is_convex=obj_convex,
+        constraint_mask=constraint_mask,
+    )
+
+
+def classify_model(model: Model) -> tuple[bool, list[bool]]:
+    """Classify a model's convexity.
+
+    Returns:
+        (is_convex, constraint_convexity_mask)
+        - is_convex: True if objective is convex and all constraints are convex
+        - constraint_convexity_mask: per-constraint convexity flags
+    """
+    oa_convexity = classify_oa_cut_convexity(model)
+    all_convex = oa_convexity.objective_is_convex and all(oa_convexity.constraint_mask)
+
+    return all_convex, oa_convexity.constraint_mask
--- a/python/discopt/_jax/cutting_planes.py
+++ b/python/discopt/_jax/cutting_planes.py
@@ -245,6 +245,7 @@ def generate_oa_cuts_from_evaluator(
     evaluator,
     x_sol: np.ndarray,
     constraint_senses: Optional[list[str]] = None,
+    convex_mask: Optional[list[bool]] = None,
 ) -> list[LinearCut]:
     """Generate OA cuts for all constraints using an NLPEvaluator.
 
@@ -257,6 +258,8 @@ def generate_oa_cuts_from_evaluator(
         x_sol: solution point at which to linearize, shape (n,).
         constraint_senses: list of senses for each constraint. If None,
             all constraints are assumed to be "<=".
+        convex_mask: Per-constraint boolean list. If provided, only constraints
+            where ``convex_mask[k]`` is True are linearized.
 
     Returns:
         List of LinearCut objects, one per constraint.
@@ -273,6 +276,8 @@ def generate_oa_cuts_from_evaluator(
 
     cuts = []
     for k in range(m):
+        if convex_mask is not None and not convex_mask[k]:
+            continue
         grad_k = jac[k, :]
         g_k = float(cons_vals[k])
         sense = constraint_senses[k]
--- a/python/discopt/_jax/discretization.py
+++ b/python/discopt/_jax/discretization.py
@@ -54,6 +54,8 @@ def initialize_partitions(
     lb: list[float],
     ub: list[float],
     n_init: int = 2,
+    scaling_factor: float = 10.0,
+    abs_width_tol: float = 1e-3,
 ) -> DiscretizationState:
     """Create n_init uniform intervals for each partition variable.
 
@@ -68,6 +70,10 @@ def initialize_partitions(
     n_init : int, default 2
         Number of initial intervals per variable.  n_init=2 gives 3 breakpoints:
         [lb, midpoint, ub].
+    scaling_factor : float, default 10.0
+        Refinement ratio stored on the returned state.
+    abs_width_tol : float, default 1e-3
+        Convergence tolerance stored on the returned state.
 
     Returns
     -------
@@ -77,7 +83,11 @@ def initialize_partitions(
     partitions: dict[int, np.ndarray] = {}
     for k, v_idx in enumerate(var_indices):
         partitions[v_idx] = np.linspace(float(lb[k]), float(ub[k]), n_init + 1)
-    return DiscretizationState(partitions=partitions)
+    return DiscretizationState(
+        partitions=partitions,
+        scaling_factor=float(scaling_factor),
+        abs_width_tol=float(abs_width_tol),
+    )
 
 
 def add_adaptive_partition(
@@ -158,6 +168,41 @@ def add_adaptive_partition(
         if candidates:
             merged = np.sort(np.unique(np.concatenate([pts, candidates])))
             new_partitions[v_idx] = merged
+
+    return DiscretizationState(
+        partitions=new_partitions,
+        scaling_factor=state.scaling_factor,
+        abs_width_tol=state.abs_width_tol,
+    )
+
+
+def add_uniform_partition(
+    state: DiscretizationState,
+    solution: dict[int, float],
+    var_indices: list[int],
+    lb: list[float],
+    ub: list[float],
+) -> DiscretizationState:
+    """Uniformly refine every current interval for the selected variables.
+
+    Unlike adaptive refinement, this does not use the current solution value.
+    Each interval is split at its midpoint, preserving all existing breakpoints.
+    """
+    del solution  # uniform refinement is solution-agnostic
+
+    new_partitions = {k: v.copy() for k, v in state.partitions.items()}
+
+    for k, v_idx in enumerate(var_indices):
+        if v_idx not in new_partitions:
+            new_partitions[v_idx] = np.linspace(float(lb[k]), float(ub[k]), 3)
+
+        pts = new_partitions[v_idx]
+        if len(pts) < 2:
+            continue
+
+        mids = 0.5 * (pts[:-1] + pts[1:])
+        merged = np.sort(np.unique(np.concatenate([pts, mids])))
+        new_partitions[v_idx] = merged
 
     return DiscretizationState(
         partitions=new_partitions,
--- a/python/discopt/_jax/discretization.py
+++ b/python/discopt/_jax/discretization.py
@@ -1,0 +1,191 @@
+"""
+Discretization State Management for AMP (Adaptive Multivariate Partitioning).
+
+Manages the partition breakpoints for each selected variable.  The core insight
+from CP 2016 is that adding breakpoints near the current MILP solution concentrates
+refinement where the relaxation gap is largest, converging faster than uniform
+refinement.
+
+Key properties guaranteed by this module:
+- Partitions are monotonically finer: once a breakpoint is added, it is never removed.
+- Bounds are always preserved: partitions[v][0] == lb[v], partitions[v][-1] == ub[v].
+- Convergence check: all partition widths < abs_width_tol.
+
+Theory references:
+  - Nagarajan et al., CP 2016: http://harshangrjn.github.io/pdf/CP_2016.pdf
+  - Alpine.jl add_adaptive_partition() in bounding_model.jl
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+import numpy as np
+
+
+@dataclass
+class DiscretizationState:
+    """Breakpoint dictionaries for AMP domain partitioning.
+
+    Attributes
+    ----------
+    partitions : dict[int, np.ndarray]
+        Maps flat variable index → sorted 1-D array of breakpoints.
+        For a variable with n_init=2 initial intervals:
+            partitions[v] = [lb, midpoint, ub]  (3 elements)
+        Finer partitions have more elements.
+    scaling_factor : float
+        Controls the width of new partitions added during adaptive refinement.
+        New breakpoints are placed at x* ± (partition_width / scaling_factor).
+        Default 10.0 (Alpine's default).
+    abs_width_tol : float
+        Convergence threshold.  When all partition widths are below this value,
+        check_partition_convergence() returns True.
+        Default 1e-3.
+    """
+
+    partitions: dict[int, np.ndarray] = field(default_factory=dict)
+    scaling_factor: float = 10.0
+    abs_width_tol: float = 1e-3
+
+
+def initialize_partitions(
+    var_indices: list[int],
+    lb: list[float],
+    ub: list[float],
+    n_init: int = 2,
+) -> DiscretizationState:
+    """Create n_init uniform intervals for each partition variable.
+
+    Parameters
+    ----------
+    var_indices : list[int]
+        Flat variable indices to partition.
+    lb : list[float]
+        Lower bound for each variable (same order as var_indices).
+    ub : list[float]
+        Upper bound for each variable (same order as var_indices).
+    n_init : int, default 2
+        Number of initial intervals per variable.  n_init=2 gives 3 breakpoints:
+        [lb, midpoint, ub].
+
+    Returns
+    -------
+    DiscretizationState
+        Initial discretization with uniform breakpoints.
+    """
+    partitions: dict[int, np.ndarray] = {}
+    for k, v_idx in enumerate(var_indices):
+        partitions[v_idx] = np.linspace(float(lb[k]), float(ub[k]), n_init + 1)
+    return DiscretizationState(partitions=partitions)
+
+
+def add_adaptive_partition(
+    state: DiscretizationState,
+    solution: dict[int, float],
+    var_indices: list[int],
+    lb: list[float],
+    ub: list[float],
+) -> DiscretizationState:
+    """Add breakpoints near solution[v] for each partition variable v.
+
+    For each variable v in var_indices:
+    1. Identify the active partition [p_lo, p_hi] containing solution[v].
+    2. Compute width = p_hi - p_lo.
+    3. Add new breakpoints at:
+         p_lo + width / scaling_factor
+         p_hi - width / scaling_factor
+    4. Merge with existing breakpoints, sort, deduplicate.
+
+    This mirrors Alpine.jl's add_adaptive_partition() in bounding_model.jl.
+
+    Parameters
+    ----------
+    state : DiscretizationState
+        Current partition state.
+    solution : dict[int, float]
+        Current MILP solution values, keyed by flat variable index.
+    var_indices : list[int]
+        Variables to refine.
+    lb : list[float]
+        Lower bounds (same order as var_indices).
+    ub : list[float]
+        Upper bounds (same order as var_indices).
+
+    Returns
+    -------
+    DiscretizationState
+        New state with refined partitions (old breakpoints preserved).
+    """
+    new_partitions = {k: v.copy() for k, v in state.partitions.items()}
+
+    for k, v_idx in enumerate(var_indices):
+        if v_idx not in new_partitions:
+            # Variable not yet partitioned — initialize with 2 intervals
+            new_partitions[v_idx] = np.linspace(float(lb[k]), float(ub[k]), 3)
+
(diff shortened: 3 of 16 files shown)


bernalde:
The fact that issue #15 shows that the problem results in infeasibility for AMP reveals a bug with it. Let me fix this first before checking this PR

jkitchin:
## Review summary

> **Note:** This review was generated by Claude (multi-agent code review run by @jkitchin). Treat findings as **suggestions for your consideration**, not authoritative claims — the model can be wrong on subtle math, and you should verify any soundness claim against the code yourself before acting on it. Posted to share what came out of the review and surface things worth a second look.

Thanks for splitting this out from the fork — the orchestration is well-structured and the TDD test layout makes the invariants easy to read. The review focused on (a) mathematical soundness of the MILP relaxation, (b) the AMP main loop and routing, (c) test coverage of the claimed invariants, and (d) integration safety. Lint is clean (ruff 0.14.6 green on all 8 new/changed files) and the integration into `solver.py` / `modeling/core.py` is purely additive — no risk to existing solver paths.

The substantive concerns below are about **soundness** — places where AMP could in principle return an unsound lower bound or falsely certify a non-optimal incumbent as `optimal`. None of these are likely to bite the canonical Alpine instances (nlp1 / circle / nlp3) that the suite advertises, which is why the tests pass; they show up on slightly off-path inputs.

I've grouped findings as **blockers** (sound-bound / wrong-answer risk), **strong nits** (robustness / silent failures), and **polish**. Line numbers are against this PR head (`b730fe2`).

---

### Blockers (soundness — please address before un-drafting)

**B1. OA cuts applied without convexity gating** — `python/discopt/solvers/amp.py:737-759` calls `generate_oa_cuts_from_evaluator` over **all** constraints with no `convex_mask`, and the result is fed into `milp_relaxation.py:865-869`. The `cutting_planes` docstring (`cutting_planes.py:251-253`) explicitly warns the cuts are only globally valid for convex constraints. Combined with `LB = max(LB, new_lb)` at `amp.py:698`, an over-tight (invalid) lower bound can monotonically lock in. If LB rises above the true global optimum and a feasible UB matches it, AMP will report `gap_certified=True` / `status="optimal"` on a sub-optimal incumbent.

This is the single most important change to make the soundness claim in `amp.py:19` actually hold for general MINLP. The natural fix is to compute a convex mask (you already have a convexity detector landing on `feat/convexity-detector`) and only generate cuts for terms classified as convex. As a stopgap before that lands, gating OA-cut generation on a global "all-constraints-convex" check would also be sound.

Related: `_solve_milp_with_oa_recovery` (`amp.py:331-375`) only recovers on infeasibility, which won't trigger here — the invalid cut tightens rather than infeasibilifies.

**B2. Silent zero-objective fallback when a nonlinear term cannot be linearized** — `python/discopt/_jax/milp_relaxation.py:878-881`. When `_linearize_expr` raises on the objective, the fallback sets `c_obj = 0`. For minimization with a true optimum < 0, the resulting LP optimum of 0 is not a valid lower bound. The accompanying constraint-side path at `:856-859` drops the constraint under `logger.debug` — for `g(x) ≤ 0` this enlarges the feasible set so the relaxation property survives, but the user has no signal that AMP is solving a much weaker problem than they think.

Suggested fix: raise (or fall back to a free auxiliary that drives the LP to `-∞`) on the objective path; promote the constraint-drop log to a warning that names the omitted constraint, and route those constraints through the OA-cut path with a convex check (B1).

**B3. Trilinear terms classified but never relaxed** — `python/discopt/_jax/term_classifier.py` records `terms.trilinear` and includes their variables in `partition_candidates`, but `build_milp_relaxation` never iterates `terms.trilinear` (zero matches in `milp_relaxation.py`). Any `x*y*z` constraint or objective hits the "Higher-order product (3 vars) not supported" raise at `milp_relaxation.py:309-310`, then falls into the silent-drop / zero-objective path from B2. The advertised nlp1/circle/nlp3 parity holds because those instances are bilinear+monomial only.

Two acceptable fixes: (a) implement a recursive bilinear relaxation (`w_ij = x_i·x_j`, then `w_ij·x_k`), or (b) raise a clear `NotImplementedError` from the classifier when trilinear terms are present and document this as a known limitation. Either way, please remove the silent-drop path so a user can't unknowingly get an invalid bound.

**B4. Piecewise McCormick missing one-sided δ-forcing for `wk_lo ≤ 0`** — `python/discopt/_jax/milp_relaxation.py:712-717`. The disaggregated formulation enforces `w̄_k ≤ wk_hi·δ_k` unconditionally but only enforces `w̄_k ≥ wk_lo·δ_k` when `wk_lo > 0`. When `y` straddles zero (the common case for partitioned bilinears), the inactive intervals' `w̄_k` can take the box's negative lower bound and pull `w = Σ w̄_k` arbitrarily down. Since the per-interval cv1/cv2 cuts at `:724-738` are loosened by `+M_k` when `δ_k = 0`, they don't recover the bound either. Compare Alpine.jl, which forces `w̄_k ≥ wk_lo·δ_k` unconditionally. **Fix: drop the `if wk_lo > 0:` guard** at line 713.

---

### Strong nits (robustness / silent failures)

**N1. `min_vertex_cover` doesn't check HiGHS status** — `python/discopt/_jax/partition_selection.py:160-171`. After `h.run()` the code reads `sol.col_value[:n]` without checking `getModelStatus()`. Best case the post-hoc coverage check rejects bad solutions and falls back to `max_cover`; worst case a coincidentally-feasible vector is returned. Also, no time limit on the internal MILP — could hang on a dense interaction graph. Add a status check and a `time_limit` (e.g., 30 s).

**N2. Silent NLP bound clipping** — `python/discopt/solvers/amp.py:_solve_nlp_subproblem` clips bounds to `[-1e8, 1e8]` without warning. A user with bounds outside this range gets a relaxed NLP without notice; the resulting `obj_nlp_min` could be infeasible w.r.t. the original bounds. Either raise a warning, or feed the clipped bounds back through `_check_constraints_with_evaluator` against the *original* bounds.

**N3. Silent drop of B&B-only kwargs in routing** — `python/discopt/solver.py:1244-1261`. When `solver="amp"` is selected, kwargs like `incumbent_callback`, `node_callback`, `cutting_planes`, `mccormick_bounds` are silently dropped. A `warnings.warn(...)` listing the ignored keys would prevent confusion.

**N4. Adaptive refinement can stall** — `python/discopt/_jax/discretization.py:163-166`. The strict-inequality guard means that when `solution[v_idx]` equals an existing breakpoint, no new candidates are generated for that variable. Add a fallback to bisecting the active interval when no candidate qualifies.

**N5. nlp3 xfail leaves the headline benchmark unverified** — `python/tests/test_amp.py:948` is `@pytest.mark.xfail(strict=False)` on the most-cited Alpine instance, tracked under issue #24. With `strict=False` it can also start passing silently. At minimum, switch to `strict=True` once #24 lands. Consider a smaller multilinear in the meantime.

**N6. nlp1 tolerance is loose enough to mask local-optimum regressions** — `python/tests/test_amp.py:932` uses `abs ≤ 0.15` (~0.26%). With `rel_gap=1e-3` the test should be tightenable to `abs ≤ 0.06` and would then actually verify the gap certificate produces a near-global-optimum.

**N7. mypy: 2 new errors in new files**
- `python/discopt/_jax/discretization.py:145` — int vs `signedinteger[_32Bit | _64Bit]`.
- `python/discopt/_jax/partition_selection.py:130` — `c` rebound from `np.ones` to `np.array(...)` triggers a shape-tuple narrowing. Annotate `c: np.ndarray` once before the if/else.

---

### Coverage gaps that would surface the soundness bugs

If B1–B4 land, the suite still won't catch a regression of any of them. Recommend adding:

- **Convexity-gated OA test**: a non-convex constraint where unconditional OA cuts produce LB > true global opt — assert AMP either raises or returns `feasible` rather than `optimal`.
- **Trilinear end-to-end**: a tiny `min x*y*z s.t. ...` that asserts AMP either solves it or raises `NotImplementedError`. Currently `TestTermClassifier::test_detect_trilinear` (test_amp.py:157) verifies classification but no end-to-end test catches the silent omission.
- **Sign-straddling bilinear**: `x*y` with `y ∈ [-1, 1]`, partitioned. Sample `(x, y)` against the relaxation and assert `w̄_k = 0` whenever `δ_k = 0`. Catches B4.
- **Infeasibility**: e.g. `x*y ≥ 100` on `[0,1]^2` with `solver="amp"`, assert `status="infeasible"`.
- **Linear-only model with `solver="amp"`**: exercises the `if not part_vars` branch (`amp.py:820-824`) end-to-end.
- **Higher-order monomial** (`x**3`, `x**4`): MILP code at `:824-841` has a "general n" branch that no test invokes.

`TestCurrentCodeWeaknesses` is mostly mock-heavy unit tests / status-mapping regressions; it does not document blockers and the name is somewhat misleading.

---

### Polish (no action required)

- `model_utils.flat_variable_bounds` could note in its docstring that callers must normalize `lb ≤ ub` and non-`None` bounds beforehand.
- `_round_integers` (`amp.py:243-245`) appears to be dead code.
- `_DEFAULT_MAX_OA_CUTS = 128` is a magic constant; expose as a parameter.
- `_default_milp_time_limit` (`amp.py:415-422`) caps at 60 s with no override path other than `milp_time_limit`. Underdocumented.
- `bilinear_lambda_map` λ-formulation (`milp_relaxation.py:553-647`) uses `lambda * y_lb` / `lambda * y_ub` as constants in `_add_row` — silently NaNs if `y_lb`/`y_ub` are infinite. Consider an early validation that all partitioned variables have finite bounds.
- `solver.py:1239`: redundant `solver if solver is not None else kwargs.pop("solver", None)` — `solver` is now a named parameter; just `kwargs.pop("solver", None)` or drop the fallback.
- `partition_selection.py` magic `len(...) <= 15` for auto strategy — consider a module constant.

---

### Verified working as advertised

- Maximize handling (commits `bbb3f7e`, `a6b75b7`): clean min/max symmetry across `NLPEvaluator`, `milp_relaxation`, the main loop, callbacks, and final `SolveResult`. No bug detected.
- Adaptive refinement monotonicity (`discretization.py:130, 169`): old breakpoints preserved on `add_adaptive_partition` and `add_uniform_partition`.
- Big-M signs in piecewise McCormick (`milp_relaxation.py:728, 737, 746, 755`): hand-checked, all correct.
- Breakpoint-based tangent placement for `x²` (`milp_relaxation.py:805-815`): correct fix to the monotone-tightening hazard, exactly as the comment defends.
- Routing safety: `solver.py` and `modeling/core.py` changes are purely additive; no existing path's behavior changes when `solver` is unset.
- Lint: ruff 0.14.6 clean on all 8 files; `ruff format --check` clean.

---

### Suggested next steps (for your consideration)

1. Sanity-check B1–B4 against the code — these are the items most worth verifying since a false positive on any of them would be wasted work.
2. If B2 / B3's silent-drop fallbacks are intentional, a comment in the code explaining the rationale would help future readers (the review flagged them precisely because they look unintentional).
3. The strong nits and coverage-gap suggestions are independent of the blockers and can be picked up à la carte.

Happy to follow up on any specific item if useful — just ping @jkitchin.

bernalde:
Replying to the review summary at https://github.com/jkitchin/discopt/pull/44#issuecomment-4304112526.

B1: fixed in [8a0eeb5](https://github.com/bernalde/discopt/commit/8a0eeb5). AMP now computes a shared OA convexity classification and passes `convex_mask` into evaluator-based OA cut generation, so nonconvex constraint rows are skipped instead of tightening the relaxation unsafely. The same convexity rule is applied in `python/discopt/solvers/oa.py` and `python/discopt/solvers/gdpopt_loa.py`. Coverage is in [ef3948a](https://github.com/bernalde/discopt/commit/ef3948a) with convex-mask unit tests plus solver-level AMP, OA, and LOA regressions.

B2: fixed across [8a0eeb5](https://github.com/bernalde/discopt/commit/8a0eeb5) and [d73d998](https://github.com/bernalde/discopt/commit/d73d998). Nonlinear objectives that cannot be linearized no longer produce a certified lower bound from a zero objective. The relaxation stays usable for point generation, but lower-bound updates are now disabled when the objective linearization is not valid. Dropped nonlinear constraints are also promoted from debug logging to warnings with the constraint name. Coverage is in [ef3948a](https://github.com/bernalde/discopt/commit/ef3948a).

B3: fixed in [d73d998](https://github.com/bernalde/discopt/commit/d73d998). `milp_relaxation.py` now lifts `x*y*z` recursively into nested bilinear auxiliaries instead of falling through the unsupported-product path. Coverage is in [ef3948a](https://github.com/bernalde/discopt/commit/ef3948a) with both MILP-builder coverage and an AMP trilinear end-to-end regression.

B4: fixed in [d73d998](https://github.com/bernalde/discopt/commit/d73d998). The disaggregated piecewise McCormick rows now always enforce `wbar_k >= wk_lo * delta_k`, even when `wk_lo <= 0`, so inactive intervals cannot pull `w` downward through a negative slack. Coverage is in [ef3948a](https://github.com/bernalde/discopt/commit/ef3948a).

N1: fixed in [d73d998](https://github.com/bernalde/discopt/commit/d73d998). `min_vertex_cover` now sets a 30 s HiGHS time limit, checks the model status after `h.run()`, and falls back to greedy coverage when HiGHS is non-optimal or the returned selection does not cover the graph. Coverage is in [ef3948a](https://github.com/bernalde/discopt/commit/ef3948a).

N2: not changed. I rechecked the current AMP NLP path and did not reproduce a solution outside the original bounds. AMP reapplies the original flat bounds to the model and evaluator before each NLP solve and rechecks feasibility on the returned point, so I left this one alone pending a concrete reproducer.

N3: fixed in [d73d998](https://github.com/bernalde/discopt/commit/d73d998). `solve_model(..., solver="amp")` now warns when AMP ignores branch-and-bound-only options such as `incumbent_callback`, `node_callback`, `cutting_planes`, and `mccormick_bounds`. Coverage is in [ef3948a](https://github.com/bernalde/discopt/commit/ef3948a).

N4: fixed in [d73d998](https://github.com/bernalde/discopt/commit/d73d998). Adaptive refinement now inserts the midpoint of the active interval when centered candidates land exactly on existing breakpoints, so the partition still refines instead of stalling. Coverage is in [ef3948a](https://github.com/bernalde/discopt/commit/ef3948a).

N5: fixed in [ef3948a](https://github.com/bernalde/discopt/commit/ef3948a). The `nlp3` AMP xfail is now `strict=True`.

N6: fixed in [ef3948a](https://github.com/bernalde/discopt/commit/ef3948a). I reran `test_nlp1_bilinear_global_optimum` three times locally after the code changes; all three runs hit `58.38367176904225`, which is `8.23e-06` from the reference value. The test tolerance is now `0.06`.

N7: partially fixed and verified. [d73d998](https://github.com/bernalde/discopt/commit/d73d998) adds the `c: np.ndarray` annotation in `python/discopt/_jax/partition_selection.py`. I also reran `mypy` locally on `python/discopt/_jax/discretization.py` and `python/discopt/_jax/partition_selection.py`; it reported no issues, so I did not make a separate type-only change in `discretization.py`.

The extra coverage suggestions from the review are addressed in [ef3948a](https://github.com/bernalde/discopt/commit/ef3948a): convexity-gated OA coverage, trilinear coverage, the sign-straddling bilinear case, the adaptive-refinement breakpoint stall case, and the AMP ignored-option warning case.

jkitchin:
Following up on the circle test weakening that came up while verifying the soundness fixes.

After B1, `x₀²+x₁² ≥ 2` is correctly classified as non-convex (`constraint_mask=[False]`), so the OA cuts that previously produced the certificate are suppressed. I ran AMP on circle at the PR head and got `status="feasible"`, `objective≈√2` to 1e-9, `gap_certified=False` — which is exactly right: the old `status="optimal"` was resting on unsound OA cuts, and the weakening reflects the fix, not a regression. So no soundness concern.

Small polish suggestion while we're here: the assertion is now

```python
assert result.status in ("optimal", "feasible", "time_limit")
```

with conditional gap-certified branches. Since circle structurally *cannot* be certified under the new rules (non-convex feasible set → no valid OA cuts → monomial McCormick LB only), the disjunctive assertion would silently accept a future regression that accidentally re-enables OA on non-convex constraints and starts reporting `optimal` again.

Would you prefer to tighten it to `assert result.status == "feasible"` + `assert result.gap_certified is False` with a one-line comment explaining why certification is structurally impossible here — as part of this PR, or track it as a follow-up? Happy either way; the soundness behavior is already correct.

bernalde:
Replying to https://github.com/jkitchin/discopt/pull/44#issuecomment-4309617506.

Addressed in [0bb21cc](https://github.com/bernalde/discopt/commit/0bb21cce6ea2af1ea97717fdce93ed3aa971e7bb). I tightened `test_circle_bilinear_global_optimum` to require `status == "feasible"` and `gap_certified is False`, with a comment explaining that the nonconvex superlevel constraint means evaluator OA cuts must be skipped and AMP should not certify the relaxation gap.

I also reviewed the adjacent AMP certification/status tests. `test_gap_certified_after_convergence` and `test_early_termination_when_gap_closes` were still using circle, so after B1 they were mostly vacuous. The same commit moves those checks to a small convex quadratic model and asserts `optimal` plus `gap_certified is True`.

I did not find another PR-specific solver bug or future-work item in this pass, so I did not open a new issue.

Verification:

`PYTHONPATH=python pytest -q python/tests/test_amp.py::TestAmpEndToEnd::test_circle_bilinear_global_optimum python/tests/test_amp.py::TestAmpEndToEnd::test_pure_quadratic_convex python/tests/test_amp.py::TestAmpConvergenceProperties::test_gap_certified_after_convergence python/tests/test_amp.py::TestAmpConvergenceProperties::test_early_termination_when_gap_closes` passed, 4 tests.

`PYTHONPATH=python pytest -q python/tests/test_amp.py -k "milp_lb_valid_circle or unsupported_objective_returns_no_relaxation_bound or circle_bilinear_global_optimum or nonconvex_constraint_oa_cuts_respect_convexity_mask or unsupported_objective_disables_certified_bound or pure_quadratic_convex or gap_certified_after_convergence or early_termination_when_gap_closes or amp_max_iter_without_gap_certificate_returns_feasible or partition_convergence_without_gap_is_not_certified"` passed, 10 tests and 81 deselected.

`PYTHONPATH=python pytest -q python/tests/test_amp.py` passed, 88 passed, 1 skipped, 2 xfailed.

jkitchin:
Merged to `main` as [0d02bb3](https://github.com/jkitchin/discopt/commit/0d02bb3) (squash, with you credited via `Co-Authored-By`). GitHub can't flag a cross-repo PR as "merged" when the commits land via a resolved local merge rather than the merge button, so the PR shows as closed-not-merged — the commit trailer (`Closes #44`) is the authoritative link.

A few adjustments were needed on top of `0bb21cc` to land cleanly on post-#46 `main`:

1. **`_jax/convexity.py` → `_jax/convexity/` package port.** #46 restructured the single file into a package (`certificate.py`, `rules.py`, `lattice.py`, `interval.py`, ...). Your new `OACutConvexity` dataclass and `classify_oa_cut_convexity` function moved into `convexity/rules.py` and are re-exported from `convexity/__init__.py`. Added a `use_certificate=` flag to `classify_oa_cut_convexity` for consistency with the package's new `classify_model` / `classify_constraint` signatures; all existing call sites use the default (`False`) and are unchanged.

2. **`skip_convex_check` plumbed through `solve_model` → `solve_amp`.** `test_convex_fast_path.py` (on `main`) parametrizes on `solver_name=["default", "amp"]` and expects both to honor the convex fast path. I added a fast-path short-circuit inside `solve_amp`: when the model is pure-continuous and provably convex (via the same interval-Hessian certificate `solve_model` uses), AMP skips partitioning and returns a single NLP solve with `convex_fast_path=True`. Three of your pre-existing AMP tests (`test_zero_upper_bound_reports_no_relative_gap`, `test_no_incumbent_after_search_returns_iteration_limit`, `test_first_iteration_milp_error_is_not_reported_as_infeasible`) use trivially-convex models to exercise AMP internals, so they now pass `skip_convex_check=True` to keep testing what they intend to test. Previously `solve_model` treated `skip_convex_check` as an ignored option for AMP and warned; that warning is gone now that AMP respects the flag.

3. **Two mypy nits** (pre-commit gated):
   - `milp_relaxation.py:470`: `tuple(sorted(term))` gave `tuple[int, ...]` into a `list[tuple[int, int, int]]` — destructured to a fixed-arity tuple instead.
   - `discretization.py:145`: `p_lo_idx` was reassigned across incompatible numpy-int / `int` types — renamed the intermediate.

**Verification**: `test_amp.py` 123 passed + 2 xfailed (matches your report plus one env-skip that runs locally here); `test_convex_fast_path.py` 34/34; all 316 convexity tests; `test_oa` / `test_gdpopt_loa` / `test_cutting_planes` 95/95; smoke sweep clean; `ruff check`, `ruff format --check`, and `mypy python/discopt/` all clean.

Thanks for the solver — and for the tightening in [0bb21cc](https://github.com/bernalde/discopt/commit/0bb21cce6ea2af1ea97717fdce93ed3aa971e7bb) on the convergence / certification tests; catching that the convergence checks had gone vacuous on circle was a nice catch.

Issue #51, opened 2026-04-25
Convexification enhancements inspired by MC++ (Taylor/Chebyshev models, multivariate McCormick, etc.)


jkitchin:
## Background

A review of [omega-icl/mcpp](https://github.com/omega-icl/mcpp) (MC++ v4.0, the bounding library underpinning MAiNGO) surfaced several convexification techniques that discopt does not yet implement. MC++ is a mature, research-grade reference for relaxation arithmetic on factorable functions, so its feature set is a useful gap analysis target.

This issue tracks those opportunities as a single roadmap. Items are independent and can be picked off individually. Each item is given a stable identifier (M1, M2, ...) so cross-references in this issue and in linked PRs do not collide with GitHub issue/PR auto-linking.

## Correctness invariant (applies to every item)

discopt's phase gates enforce `incorrect_count ≤ 0`. Every item below changes (or feeds) the relaxation, which directly affects the validity of computed bounds. The non-negotiable invariant for all items is:

> The convex relaxation must be **rigorous**: the computed lower bound never exceeds the true global minimum on any test instance, within numerical tolerance (abs=1e-6, rel=1e-4 per `conftest.py`).

Each item also lists item-specific acceptance criteria. All items additionally require: at least one new regression instance under `discopt_benchmarks/` exercising the new code path, and `incorrect_count == 0` on the existing benchmark suite.

## Current discopt coverage (no gap)

- Classical (2-argument) McCormick envelopes — `python/discopt/_jax/mccormick.py`, `piecewise_mccormick.py`
- αBB — `python/discopt/_jax/alphabb.py`
- Interval arithmetic + interval AD (forward) — `python/discopt/_jax/convexity/interval*.py`
- Gershgorin eigenvalue bounds (convexity certificates only) — `python/discopt/_jax/convexity/eigenvalue.py`
- FBBT, OBBT, probing, simplify — `crates/discopt-core/src/presolve/`
- First-order linearization (Taylor) cuts — `python/discopt/_jax/cutting_planes.py`
- LP relaxation compiler — `python/discopt/_jax/relaxation_compiler.py`
- NN embedding (full-space / big-M ReLU / reduced) — `python/discopt/nn/formulations/`

## Proposed enhancements

Ordered roughly by expected bound-tightening impact / effort ratio. Each item lists its target language up front so the architectural split is unambiguous.

### M1. Multivariate McCormick relaxations (Tsoukalas and Mitsos, 2014)

**Language.** Python.

[...]

PR #76, opened 2026-05-10
feat(presolve): wire M4+M5, M10, M9 of #51 into root presolve pipeline

jkitchin:
## Summary

Wires three of the standalone bound-tightening modules from PR #75 into the default solver path:

- **M10** `eliminate_variables` — runs by default (`presolve=True`).
- **M4 + M5** `reformulate_polynomial` — opt-in (`presolve_polynomial=True`).
- **M9** reverse-mode interval AD — opt-in (`presolve_reverse_ad=True`).

A new orchestrator `python/discopt/_jax/presolve_pipeline.py` sequences the Rust passes, propagates tightened bounds back to the Python `Model` (with shape and aux-var-introduction guards), and exposes the reverse-AD bridge.

`solver.py` calls the orchestrator immediately after `model_to_repr(...)` so the JAX relaxation compiler and B&B initialisation see the tighter bounds.

## Deferred (intentionally) from this PR

**M2 (Chebyshev), M3 (Taylor), M6 (eigenvalue)** are landed as standalone bound providers in PR #75 but are not yet wired into `relaxation_compiler.py`. The compiler currently dispatches univariate operators directly to `mccormick.relax_*`; introducing an `arithmetic=` kwarg requires plumbing the choice through every `_compile_relax_node` recursion plus a (cv, cc) shim that converts polyhedral-OA cuts back into JAX-traceable underestimator/overestimator functions. That is a focused follow-up where each operator's tightness vs. McCormick can be benchmarked.

## Test plan

- [x] `pytest python/tests/test_presolve_pipeline.py -v` — 6 new tests pass (eliminate, propagate, polynomial, reverse-AD, end-to-end with elimination, full pipeline with all three passes).
- [x] `pytest python/tests/test_correctness.py python/tests/test_alphabb.py python/tests/test_callbacks.py` — 144 existing tests pass.
- [x] `mypy python/discopt/` — full project clean ("Success: no issues found in 158 source files").
- [x] `ruff check python/` and `ruff format --check python/` — clean.
- [ ] CI on this PR.

## Notes for review

- This PR sits on top of #75 (base: `fix/issue-42-rank1-certificate`). It will need to be retargeted to `main` once #75 merges.
- All three new flags default to OFF except `presolve` (which controls M10). The change is therefore backward-compatible: existing solver invocations get only the safe M10 elimination automatically.
- The `propagate_bounds_to_model` helper deliberately compares element-by-element rather than by block, so vector/matrix variable bounds tighten correctly.


jkitchin:
Superseded by a clean rebranch off main; opening replacement PR.

Issue #82, opened 2026-05-11
Release v0.4.0


jkitchin:
I am hoping to release a new version 0.4.0 for discopt this week.

@bernalde are there any significant PRs you have coming up that you would like included? Any remaining issues you think should be prioritized?

bernalde:
John, I'm trying to get AMP into good shape before the end of the week, but I might not be able to wrap up everything I want to do by then. Feel free to cut a version when you consider it worth it. Any extra work on https://github.com/jkitchin/discopt/issues/38 will lead to improvements downstream, as we should be able to really benefit from convexification if we know what we are dealing with

bernalde:
Release-planning data point from the fork's latest AMP benchmark:

- Run ID: `20260515T213648Z-84d8fe0`
- Branch/commit: `amp` at `84d8fe052c278df3e458da180af00a76285fc710`
- Benchmark: translated MINLPTests set against the archived April 20 AMP/Alpine reference manifest.
- Settings: `solver="amp"`, per-instance limit `300s`, AMP max iter `1000`, include convex cases, CPU JAX with x64.

Result:

- `122/122` cases pass:
  - `91/91` `nlp_cvx`
  - `15/15` `nlp`
  - `13/13` `nlp_mi`
  - `3/3` infeasible
- Current AMP improves on the April AMP baseline by 4 cases:
  - `nlp_003_014`
  - `nlp_003_015`
  - `nlp_mi_003_014`
  - `nlp_mi_003_015`
- Timing on passes: median `0.344s`, geometric median `0.344s`.
- Total MIPs: `55` across 122 cases, down from `262` in the April AMP baseline.

This makes AMP look much closer to an upstreamable state from the MINLPTests perspective. I would still keep it out of a release unless/until the `amp` branch lands through a normal upstream PR, because the remaining slow tail and review surface are nontrivial.

jkitchin:
new release! https://github.com/jkitchin/discopt/releases/tag/v0.4.0

Issue #38, opened 2026-04-20
Convexity detector in discopt could be significantly strengthened


bernalde:
While working on the AMP solver, I realized that the convexity detector in Discopt only works for certain linear constraints and for a limited set of nonlinear ones. There is more to be done here, as quickly detecting convexity (potentially with RUST) would greatly enhance the solver. Maybe this is relevant for ripopt as well, but in any case, it is worth investing effort in getting this one right. Given that there is some tracking of the expression tree, this would be doable in Discopt.
The go-to method is provided by Misener's group, and a reference implementation is found here https://github.com/cog-imperial/suspect
This is well tested, and I'm aware several other MINLP solvers have reimplemented these tools.
Other references are https://github.com/scipopt/scip/blob/0d59282db71009e0144a222a6551346768bb0bec/src/scip/scip_expr.c#L1960

jkitchin:
Hi @bernalde — we've taken a first pass at strengthening the convexity detector on branch [`feat/convexity-detector`](https://github.com/jkitchin/discopt/tree/feat/convexity-detector). Would appreciate your review/testing before we merge.

## What was implemented

**SUSPECT-style sign-aware detector** (`python/discopt/_jax/convexity/`)
- `rules.py` + `lattice.py`: sign/convexity lattice with propagation rules over the expression DAG, modeled on cog-imperial/suspect.
- `interval.py` + `interval_eval.py`: interval arithmetic over the expression tree for range propagation.
- Commit: 36e280c

**Sound box-local certificate** (beyond pattern matching)
- `interval_ad.py`: interval-valued forward-mode AD to build an interval Hessian.
- `eigenvalue.py`: lower-bound smallest eigenvalue of the interval Hessian (Gershgorin + recursion) to prove PSD on a box.
- `certificate.py`: combines lattice detection with the interval-Hessian oracle for a sound certificate.
- Commit: 54fc8a9

**Solver integration**
- Wired into `solver.py` convexity gates so NLP relaxations can exploit proven convexity.
- Per-B&B-node certificate refresh: when a node tightens variable boxes, the certificate is re-evaluated on the tighter box (a non-convex function globally may be provably convex on a sub-box).
- Commits: 1d7d520, 75577d0

**Tests** (~2100 lines across 8 files in `python/tests/test_convexity_*.py`): lattice rules, interval arithmetic, interval AD, eigenvalue bounds, certificate soundness, pathological cases, node refresh, solver integration.

## What is not (yet) done

[...]

Issue #15, opened 2026-04-07
Gas network MINLP: 500x slower than GAMS/DICOPT on 19-variable benchmark


jkitchin:
## Summary

A synthetic gas network MINLP (19 variables, 2 binaries, 20 constraints) takes ~336 seconds to solve with discopt vs 0.6 seconds with GAMS/DICOPT. Both find the same objective (3.003 MW). The 500x gap is primarily due to spatial branching overhead on the nonconvex continuous subproblem.

## Problem description

The benchmark models a gas transmission network with 2 compressor stations (on/off binary decisions), Weymouth pressure-drop equations (`f^2 = C*(p_in^2 - p_out^2)`), and a nonlinear compressor power objective (`w * (beta^0.2857 - 1)`). The problem has a unique global optimum at the symmetric flow split.

| | discopt (ipm) | GAMS/DICOPT |
|---|---|---|
| Objective | 3.0027 MW | 3.0026 MW |
| Status | feasible | integer solution |
| Nodes | 147 | 3 NLP + 2 MIP |
| Wall time | 336 s | 0.6 s |

## Root causes

1. **JAX IPM convergence on nonconvex equalities**: The Weymouth constraints are nonconvex quadratic equalities. The JAX IPM converges poorly from wide initial domains, returning near-feasible but non-stationary points. Only after spatial branching tightens the domains (by node ~100) does the IPM find the true optimum. DICOPT uses CONOPT (a GRG method) which finds it on the first call.

2. **McCormick relaxation gap**: The convex relaxation bound is -39,326 MW vs the true optimum of 3.0 MW. The `exp(kappa * log(beta))` and bilinear `w * (exp(...) - 1)` terms produce extremely loose McCormick envelopes. This forces extensive spatial branching to close the gap. DICOPT uses outer approximation which is much tighter for this structure.

3. **No warm-starting across B&B nodes**: DICOPT solves 3 NLP subproblems total (warm-started from prior solutions). discopt solves 147+ NLP subproblems with multi-start from scratch at each node.

## Reproduce

The benchmark file is at `../saudi-aramco/gas_network_minlp.py` (relative to discopt). To run:

```python
import sys
sys.path.insert(0, "../discopt/python")

# Requires JAX_ENABLE_X64=1 environment variable
from gas_network_minlp import build_gas_network_minlp, _reference_solution

# Analytical reference
ref = _reference_solution()
print(f"Reference: {ref['total_cost']:.4f} MW")

# Solve with discopt
[...]

PR #86, opened 2026-05-15
Add AMP global MINLP solver

### LINKED ISSUE #23: Add Adaptive Multivariate Partitioning (AMP) global MINLP solver

bernalde:
## Summary

This upstreams the fork's `amp` branch, adding the Adaptive Multivariate Partitioning global MINLP solver path and the supporting test and benchmark infrastructure.

The main user-facing entry point is `Model.solve(solver="amp")`. The branch adds or strengthens:

- AMP solver orchestration, partition refinement, incumbent management, NLP warm starts, termination handling, memory guardrails, and MILP-relaxation integration.
- Operator relaxations and lifts for univariate nonlinear terms, reciprocals/negative powers, exp/log/sqrt, trig/tan/abs/square patterns, shifted quadratics, min/max objectives, alpha-BB quadratic cuts, and mixed-curvature trig cases.
- Shared nonlinear bound tightening, root presolve, OBBT variants, contradiction proofs, and Rust-side AMP hot-path helpers.
- Direct-OA coverage diagnostics with stable skip reasons for rows that cannot be safely cut directly.
- AMP-focused unit/integration tests, MINLPTests known-failure cleanup, benchmark timeout handling, documentation updates, and CI/test-tier changes.

Closes #23.
Refs #38.
Refs #82.

## Fork issue and PR series

The work was staged and reviewed in `bernalde/discopt` before this upstream PR.

Core AMP and Alpine-reference work:

- bernalde/discopt#27: Improve AMP MINLPTests recovery.
- bernalde/discopt#29: Add shared nonlinear bound tightening.
- bernalde/discopt#30: Add Alpine AMP coverage.
- bernalde/discopt#37: Fix AMP incumbent warm-start policy.
- bernalde/discopt#40: Add AMP presolve OBBT variants.
- bernalde/discopt#41: Expose AMP partition hooks.
- bernalde/discopt#44: Port AMP term classification hot paths to Rust.
- bernalde/discopt#45: Fix AMP univariate operator relaxations.

Correctness and relaxation coverage follow-ups:

- bernalde/discopt#46 and bernalde/discopt#87: Strengthen gas-network lower-bound handling.
- bernalde/discopt#47: Fix AMP quadratic superlevel bounds.
- bernalde/discopt#48: Cover nonlinear contradiction proofs.
- bernalde/discopt#49: Fix division integer root presolve.
- bernalde/discopt#50: Fix unbounded IPM MINLPTests cases.
- bernalde/discopt#51: Fix MINLPTests exp-square infeasibility.
- bernalde/discopt#52: Fix `nlp_cvx_105_011` false infeasible xfail.
- bernalde/discopt#53: Fix circle spatial B&B optimum.
- bernalde/discopt#56: Fix `sqrt_nlp` expected optimum.
- bernalde/discopt#66: Fix shifted quadratic infeasibility.
- bernalde/discopt#67: Add alpha-BB OA cuts for AMP quadratics.
- bernalde/discopt#68: Fix AMP reciprocal relaxations.
- bernalde/discopt#74: Fix AMP `x*exp(x)` objective relaxations.
- bernalde/discopt#75: Fix AMP unary relaxation bounds.
- bernalde/discopt#76: Fix AMP trig relaxations.
- bernalde/discopt#77: Fix AMP tan/abs objective lifting.
- bernalde/discopt#80: Fix mixed-curvature AMP trig relaxations.
- bernalde/discopt#82: Add AMP memory guardrails.
- bernalde/discopt#83: Lift AMP min/max objectives.
- bernalde/discopt#84: Fix continuous trig-square AMP relaxations.
- bernalde/discopt#85: Fix AMP tan/abs `nlp_004` bounds.
- bernalde/discopt#86: Add finite-domain trig-square tables for AMP.
- bernalde/discopt#94: Fix `nlp_008_010` square monomial lift.
- bernalde/discopt#95: Fix AMP OA cut coverage diagnostics.

Test, benchmark, and upstream-sync support:

- bernalde/discopt#39 and bernalde/discopt#57: Split AMP tests from the broad benchmark suite.
- bernalde/discopt#42, bernalde/discopt#58, and bernalde/discopt#96: Sync upstream `main` into `amp`.
- bernalde/discopt#54: Right-size PR correctness runtime.
- bernalde/discopt#59: Enforce hard category benchmark timeouts.

Current fork trackers and follow-ups:

- bernalde/discopt#88 tracks MINLPTests AMP benchmark baselines.
- bernalde/discopt#89 tracks remaining AMP slow-tail performance work.
- bernalde/discopt#24 remains open for additional gas-network lower-bound strengthening.
- bernalde/discopt#92 and bernalde/discopt#93 track default-solver issues; the AMP benchmark data was added there only as comparison context.

## Benchmark

Latest full local benchmark:

- Run ID: `20260515T213648Z-84d8fe0`
- Branch/commit: `amp` at `84d8fe052c278df3e458da180af00a76285fc710`
- Reference manifest: `ref/minlptests-amp-baseline-2026-04-20/reference-manifest.json`
- Output directory: `results/minlptests-amp-vs-alpine/20260515T213648Z-84d8fe0`
- Environment: pixi with `python=3.12`, `ipopt`, `pkg-config`, and `highspy`; `uv run --no-sync --python .venv/bin/python`; CPU JAX with x64 enabled.
- Settings: `--discopt-mode amp`, `--per-instance-time-limit 300`, `--discopt-amp-max-iter 1000`, `--include-convex`.

Command:

```bash
RUN_ID="$(date -u +%Y%m%dT%H%M%SZ)-$(git rev-parse --short HEAD)"

/home/bernalde/.pixi/bin/pixi exec -s python=3.12 -s ipopt -s pkg-config -s highspy -- \
  /home/bernalde/.local/bin/uv run --no-sync --python .venv/bin/python -- \
  python ref/minlptests-amp-baseline-2026-04-20/scripts/run_amp_vs_reference.py \
    --repo-root . \
    --reference-manifest ref/minlptests-amp-baseline-2026-04-20/reference-manifest.json \
    --candidate-label "current-amp-${RUN_ID}" \
    --output-dir "results/minlptests-amp-vs-alpine/${RUN_ID}" \
    --discopt-mode amp \
    --per-instance-time-limit 300 \
    --discopt-amp-max-iter 1000 \
    --include-convex \
    --jax-platforms cpu \
    --jax-enable-x64
```

Result on the selected translated MINLPTests set:

| Category | Pass | Fail/Error | Total |
| --- | ---: | ---: | ---: |
| `nlp_cvx` | 91 | 0 | 91 |
| `nlp` | 15 | 0 | 15 |
| `nlp_mi` | 13 | 0 | 13 |
| `infeasible` | 3 | 0 | 3 |
| **Total** | **122** | **0** | **122** |

Compared with the archived April 20 AMP baseline (`amp_baseline_2026_04_20_5167373`):

- Pairwise outcome: `both_pass=118`, `right_only_pass=4`, `left_only_pass=0`, `both_fail=0`.
- Newly passing relative to that baseline:
  - `nlp_003_014`: baseline `time_limit`, current `optimal`, `11.760s`, `3` MIPs.
  - `nlp_003_015`: baseline `time_limit`, current `optimal`, `11.788s`, `3` MIPs.
  - `nlp_mi_003_014`: baseline `time_limit`, current `optimal`, `14.989s`, `1` MIP.
  - `nlp_mi_003_015`: baseline `time_limit`, current `optimal`, `15.510s`, `1` MIP.
- Timing on passes: median `0.344s`, geometric median `0.344s`.
- MIP summary: `55` total MIPs across 122 cases, median `0`, mean `0.45`.

Against archived Alpine references:

- Current AMP vs Alpine+HiGHS: AMP has `120` right-only passes and `2` shared passes.
- Current AMP vs Alpine+Gurobi: AMP has `69` right-only passes and `53` shared passes.
- On the 53 shared Alpine+Gurobi passes, AMP geometric median is `0.229s` vs Alpine+Gurobi `0.149s` (`2.96x` slower on that shared subset).
- On the 2 shared Alpine+HiGHS passes, AMP geometric median is `0.395s` vs Alpine+HiGHS `2.250s`.

Remaining visible benchmark tail:

- `nlp_mi_004_011`: `42.110s`, 4 MIPs.
- `nlp_mi_004_010`: `38.972s`, 4 MIPs.
- `nlp_008_010`: `25.915s`, 6 MIPs.
- `nlp_008_011`: `25.761s`, 6 MIPs.
- Several `nlp_mi_003_*` cases remain around `14-16s`.

## Validation

Fork PR bernalde/discopt#96, the final upstream-sync PR into `amp`, passed:

- `Lint + typecheck`: pass.
- `Python AMP coverage (3.12, ubuntu)`: pass.
- `Python fast (3.12, ubuntu)`: pass.
- `Rust`: pass.
- `Python coverage (3.12, ubuntu)`: skipped by that workflow tier.

The full local MINLPTests/Alpine comparison benchmark above also passed `122/122`.

## Notes

Expected direct-OA skip diagnostics remain for rows that are not safe to cut directly, such as nonaffine equalities, uncertified trigonometric rows, opposite-curvature direct-OA rows, fixed nonlinear rows, and alpha-BB quadratic candidates. The current branch reports structured skip reasons for these cases, and the affected benchmark cases pass.

The remaining AMP work is primarily performance and review hardening rather than known benchmark correctness failures on the selected 122-case translated MINLPTests set.


jkitchin:
is this ready to merge? review?

bernalde:
This is, in my opinion, ready to merge but a final major review is always appreciated. After that we can discuss how to best integrate some convexification tricks I put into this for the overall discopt.

________________________________

jkitchin (changes requested):
# Code Review — PR #86: Add AMP global MINLP solver

**Scope:** 71 files, +16,960 / −2,861. Brand-new global MINLP solver path (`solver="amp"`), a new Rust hot-path module, a 2,073-line nonlinear bound tightening module, a 2,718-line expansion of `milp_relaxation.py`, and ~7k lines of new tests. Reviewed via four parallel sub-reviews (Rust, AMP Python orchestration, JAX relaxations, tests/CI).

## Blockers (fix before merge)

1. **Multi-dim array variable indices are aliased in the Rust classifier.** `crates/discopt-core/src/amp.rs:315-325` — `scalar_index` only handles `IndexSpec::Scalar`/single-element returns. For a `Tuple([i,j])` or `Multi` index on a multi-dim `Variable`, it never applies the row-major flatten that `expr.rs:1152 tuple_to_flat` does, so `x[1,2]` and `x[1,3]` collapse to the same flat index. Bilinear/trilinear terms over multi-dim arrays end up in the wrong partition candidate. Either reuse `index_spec_to_flat(spec, &var.shape)` or restrict classification to 1-D variables and document.

2. **CI signal for the new solver is effectively absent.** `.github/workflows/amp-integration.yml:5` runs only on `workflow_dispatch` + a weekly Monday cron. The PR-time `ci.yml:98,162` filter excludes every marker (`slow`, `integration`, `amp_benchmark`, `memory_heavy`) used by the new `test_amp_integration.py` (3,669 lines) and most of the new `test_amp.py` content. Net: an AMP regression can land on `main` and surface up to 7 days later. Coverage is similarly loose — `python-amp-coverage` runs only the smoke subset, neither coverage job has a `fail_under` threshold, and the stated ≥85% gate is not enforced in CI. Add per-PR opt-in (label trigger or push-to-main) and a `--cov-fail-under=85` somewhere.

## Important

### Rust
3. **`amp.rs:144` clones every `ExprNode` per visit.** `match self.model.arena.get(id).clone()` allocates `String`/`Vec<usize>` per node in objective + every constraint, only to release the borrow before `&mut self` recursion. Match by reference, copy out the small `Copy` fields, then recurse.
4. **`amp.rs:81-105 term_idx` cross-category semantics undocumented.** A single monotonic counter spans bilinear ∪ trilinear ∪ multilinear; Python callers must recompute offsets. Add a docstring on `AmpNonlinearTerms.term_incidence` or expose a flat `Vec<TermKind>`.
5. **`amp.rs:197-201`** — `x^0` constant currently increments `general_nl_count`. Add `if exp_val.abs() <= 1e-12 { return; }`.
6. **`amp.rs:230-233` repeated-factor `Mul` double-recurses.** When `unique_vars` has a duplicate, the code increments `general_nl_count` *and* recurses into both children, double-counting sub-products.
7. **`amp.rs` test coverage is one test.** Missing: trilinear, multilinear fallback, `Pow` monomial, repeated-factor `Mul`, `Div`/`Abs`/`FunctionCall` → `general_nl_count`, multi-dim array case (would catch #1).

### AMP Python / orchestration
8. **`amp.py:2274-2277` cutoff tightening lacks an `lb ≤ ub` repair.** `_root_presolve.py:108-115` snaps near-inverted bounds to midpoint; the same guard should run after `cutoff_lb/ub` are clamped, or a tiny FP inversion will propagate.
9. **`solver.py` runs convexity classification twice for non-pure-continuous models.** The block at `solver.py:316-336` and again at `solver.py:466-495` are copy-pasted comment + logic; extract a helper.
10. **`_BoundOverrideEvaluator` (solver.py:26-39) vs `_apply_flat_bounds_to_model` + `_snapshot_variable_bounds` (amp.py)** solve the same problem two ways. Outer `solve_amp` wrapper and inner `_finish` both restore bounds, doubling work. Pick the proxy (cleaner, doesn't mutate the model).
11. **`_solve_amp_impl` is ~1,000 lines** (`amp.py:1763-2889`) and `_add_incumbent_oa_cuts_after_cutoff_tightening` mutates 5 nonlocals. Bundle the carried state into a small dataclass so a missing `nonlocal` doesn't silently shadow.
12. **`solver.py:1342-1359 amp_option_keys`** adds 9 user-facing options (`presolve_bt*`, `convhull_ebd*`, `use_start_as_incumbent`, …) but `Model.solve()`/`solve_model` docstrings aren't updated. Users discovering AMP through `Model.solve(solver="amp", …)` cannot see them.
13. **Verify `SolveResult.mip_count` is declared on the dataclass** — `_solve_milp_with_oa_recovery` now returns and threads `mip_solve_count`; if the dataclass field is missing, the kwarg lands in `__dict__` and silently disappears.
14. **`milp_highs.py:34`** — fatal `run_status` path returns zero counts, then a later branch unconditionally calls `getSolution()` on `OPTIMAL`. Add a belt-and-braces bail if `run_status` is not in `nonfatal_statuses`.

### JAX / relaxations
15. **`milp_relaxation.py:297-311 _tan_range` asymptote margin (1e-7) is too tight.** Within 1e-7 of π/2, `tan` exceeds 1e7 and `1/cos²` slopes blow up, producing catastrophic OA cuts. Widen to ~1e-4 or refuse when `|cos(lb)|` or `|cos(ub)| < 1e-3`.
16. **`cutting_planes.py:530-603` alpha-BB OA cut** uses `np.any(...|axis=0)` row mask to find curved vars. For non-symmetric Hessian fragments, column-only nonzeros are missed. Assert symmetry or take the row/col union.
17. **`nonlinear_bound_tightening.py:1865 ReciprocalBoundsRule._argument_interval_for_leq`** overloads `(None, None)` to mean both "inactive" and "infeasible". The caller currently disambiguates, but it's a footgun for future edits — split or use a sentinel.
18. **Single-file sprawl.** `nonlinear_bound_tightening.py` (2,073 lines, 10 rule classes each re-implementing near-identical `_match_scaled_*` matchers — factor into a shared `Matcher`) and `milp_relaxation.py` (3,934 lines) should split. Suggested layout: `rules/{quadratic,monotone,reciprocal,sum_of_squares}.py`, `operator_relaxations.py`, `relaxation_builder.py`.
19. **Duplicated helper.** `is_effectively_finite` (in `nonlinear_bound_tightening.py`) duplicates `_is_effectively_finite` at `milp_relaxation.py:80`. Move to `discopt.constants` or `_jax/_numeric.py`.

### Tests / CI / benchmarks
20. **`category_runner.py _run_with_hard_timeout` does not kill grandchildren.** `subprocess.run(..., timeout=…)` only kills the direct child. If AMP spawns helper processes (ipopt, HiGHS workers), they leak on timeout. Use `start_new_session=True` + `os.killpg(pgid, SIGKILL)`. Also: post-cap rewrite of `OPTIMAL → FEASIBLE` drops the certified gap silently — document or guard the policy.
21. **171 entries removed from `known_failures.toml`** correspond largely to `@pytest.mark.slow @pytest.mark.minlptests` tests now excluded by PR CI. Removing the xfail entry has zero PR-time signal — they land green whether they actually pass or not. Add at least one smoke test that re-asserts a representative formerly-failing case at PR time.
22. **`amp-integration.yml:29-34`** installs deps without `cyipopt`, yet the marker selection includes `requires_cyipopt`; those tests skip silently. Either install cyipopt or drop the marker.

## Nits
- `amp.py:911-1013` recursive expression walkers (`_expr_variable_indices`, etc.) re-traverse the DAG per constraint per iteration; cache by `id(expr)` if profiling shows it.
- `Makefile:44` flips local default from `-n auto` to `-n 2` — silently caps parallelism on dev boxes. Use `-n $(or $(PYTEST_XDIST_WORKERS),auto)`.
- `Makefile:39` `maturin` → `$(PYTHON) -m maturin` adds a hidden dep on installed-in-env maturin.
- `scripts/run_memory_capped_pytest.sh:47` warns and falls back to **uncapped** when `prlimit` is missing (i.e., macOS). The advertised memory cap is a no-op locally; warn more loudly.
- `embedding.py:15` uses `TypedDict` while other new data carriers use `@dataclass(frozen=True)` — inconsistent.
- `embedding.py:74` allocates 1 selector bit for `partition_count=1`; early-return or require `lambda_count ≥ 3`.
- Meta-tests in `test_amp.py:73-167` shell out to `make -n` and assert exact `-m "…"` marker strings; will break in lockstep with `Makefile` edits.
- `test_amp_integration.py` double-gates many tests with both `@pytest.mark.requires_cyipopt` and `pytest.skip("requires cyipopt…")` — the marker should be the only gate; double-gating masks CI misconfiguration.
- `cutting_planes.py:271-301` `generate_oa_cuts_from_evaluator` docstring still describes the old behavior.
- `_root_presolve.py:106` swallows all FBBT exceptions silently; at least log at DEBUG.
- `ci.yml:5,7` `amp` trigger is dead weight post-merge.

## Recommendation
**Do not merge as-is.** The Rust multi-dim aliasing bug (#1) is correctness-critical, and the CI gating (#2) means there is no meaningful regression signal on the largest solver added to the project. Once those two land plus the dozen IMPORTANT items, the rest can absorb as follow-ups. Math in the new relaxations is sound aside from the `tan` asymptote margin (#15). The module-sprawl concerns (#11, #18) are real but a post-merge refactor is acceptable.

---
Review generated via four parallel sub-agents (Rust, AMP orchestration, JAX relaxations, tests/CI) by Claude Code.

jkitchin:
Of the blockers, 1 seems important. 2 is less important, I can try running tests locally, and the test situation is already still in need of refining so they are fast and provide coverage.  Take a look at the important items, they are likely future issues I gues.

bernalde:
Pushed commit:
- 3ab2cca9c5bff3327def7f7c85df7c290d4228d6 Fix AMP classifier multidimensional indices

Main changes:
- Addressed blocker 1: the Rust AMP classifier now resolves scalar `IndexSpec` values through row-major flat-index collection, so multidimensional tuple and multi-indexed variables classify to distinct flat columns instead of being missed or aliased.
- Applied the same flat-index fix to the Python fallback classifier, so models that fall back because of general nonlinear terms classify multidimensional bilinear terms consistently.
- Added/expanded classifier coverage for multidimensional indices, trilinear/multilinear/monomial terms, `x^0`, and general nonlinear `Div`/`Abs`/function-call paths.
- Documented that `term_incidence` product term ids are assigned in product-term discovery order.

Tests run and passed:
- `cargo test -p discopt-core amp`
- `cargo test -p discopt-core`
- `.venv/bin/python -m pytest python/tests/test_rust_ir.py::TestAmpTermClassification -q`
- `.venv/bin/python -m pytest python/tests/test_rust_ir.py -q`
- `.venv/bin/python -m ruff check python/discopt/_jax/term_classifier.py python/tests/test_rust_ir.py`
- `git diff --check`
- Commit hooks: ruff, ruff format, mypy, cargo fmt

Comments intentionally not addressed in this push:
- Blocker 2 and the broader CI/coverage recommendations were left unchanged because the latest discussion comment deprioritized them relative to blocker 1 and they are a larger workflow-policy change.
- The broader AMP orchestration, JAX relaxation, module-splitting, benchmark-timeout, known-failure, and nits list were not bundled into this correctness patch. They should be handled as focused follow-up issues or PRs.
- The `amp-integration.yml` cyipopt concern appears already satisfied on the current branch; the workflow currently installs `cyipopt`, so I did not change it.

Remaining risk/follow-up:
- I did not run the slow AMP integration suite or alter PR-time CI gating.
- The Rust classifier still has performance/refactor follow-ups from the review, but the merge-critical multidimensional indexing bug is covered by Rust and Python regression tests now.

bernalde:
Pushed review-response commits to `amp`:
- `3ab2cca` Fix AMP classifier multidimensional indices
- `3526ad0` Tighten AMP classifier review followups
- `967b5bc` Address AMP orchestration review items
- `8bb13a8` Address JAX relaxation review items
- `036586b` Harden benchmark timeout cleanup
- `f17545b` Tighten AMP CI review gates
- `cb9dbb7` Add PR-fast MINLPTests AMP smoke
- `5b552d5` Address AMP review nits
- `fbbde91` Split AMP operator relaxation helpers
- `d944934` Update AMP OA regression expectation

Main changes:
- Fixed Rust and Python AMP classification for multidimensional indices, zero powers, repeated factors, general nonlinear counts, and expanded classifier tests.
- Reworked AMP orchestration around bound-override evaluators, cutoff bound repair, cutoff state dataclass plumbing, shared convexity classification, AMP option docs, `SolveResult.mip_count`, and fatal HiGHS status bails.
- Tightened JAX relaxation/bound-tightening behavior for `tan` asymptotes, alpha-BB Hessian support, reciprocal infeasibility sentinels, shared numeric helpers, and shared scaled-expression matching.
- Hardened benchmark timeout cleanup with process groups and explicit late-certificate policy.
- Added PR/post-merge AMP CI signal, `--cov-fail-under=85`, explicit `cyipopt` import verification, Makefile worker defaults, and louder uncapped-memory warnings.
- Added a PR-fast MINLPTests AMP smoke for a formerly xfailed case and removed integration-test double skips.
- Converted embedding metadata to a frozen dataclass, avoided selector bits for a single SOS2 interval, added root FBBT debug logging, updated OA docs, cached repeated AMP expression walks, and split trig operator helpers out of `milp_relaxation.py`.

Tests run and passing:
- `cargo test -p discopt-core amp`
- `.venv/bin/python -m pytest python/tests/test_amp.py -q` (`139 passed`)
- `.venv/bin/python -m pytest discopt_benchmarks/tests/test_category_runner.py -q`
- `.venv/bin/python -m pytest python/tests/test_fbbt_bindings.py::TestFBBTBindings::test_root_presolve_logs_fbbt_failures python/tests/test_amp.py::test_former_known_failure_nlp_mi_005_runs_in_pr_fast_suite python/tests/test_amp.py::test_tan_range_rejects_near_asymptote_endpoints python/tests/test_amp.py::test_alphabb_quadratic_oa_uses_row_col_hessian_support python/tests/test_amp.py::test_reciprocal_argument_interval_uses_explicit_infeasible_sentinel python/tests/test_amp.py::test_embedding_map_single_partition_uses_no_selector_bits python/tests/test_amp.py::test_pr_fast_tier_excludes_heavy_manual_markers python/tests/test_amp.py::test_memory_capped_pytest_warns_when_prlimit_missing -q`
- `.venv/bin/python -m ruff check ...` on modified Python/benchmark files
- `.venv/bin/python -m mypy ...` on modified typed modules
- `git diff --check`
- Commit hooks: ruff, ruff format, mypy, cargo fmt

Comments intentionally not addressed: none. There were no inline review threads to reply to or resolve.

Remaining risk: I did not run the full opt-in AMP integration workflow locally because it is CI-sized and environment-dependent; the workflow now installs and explicitly verifies `cyipopt` before running those tests.

python/tests/test_amp.py
```diff
@@ -57,1932 +30,3568 @@ def _make_nlp1() -> Model:
...
+        "slow",
+        "correctness",
+        "integration",
+        "amp_benchmark",
+        "requires_cyipopt",
+        "memory_heavy",
+    ):
+        assert f"not {marker}" in marker_expr
+    assert "--ignore=python/tests/test_correctness.py" in output
+    assert args[args.index("-n") + 1] == "auto"
```
bernalde, line 152: Blocking: this assertion is brittle in the actual PR-fast environment. CI sets `PYTEST_XDIST_WORKERS=2`, so the Makefile dry run correctly emits `-n 2` and this fails with `AssertionError: assert '2' == 'auto'`. Clear `PYTEST_XDIST_WORKERS` inside `_make_dry_run` for the default-case test, or split this into one test for the default `auto` behavior and one test for honoring the env override.

python/tests/test_amp.py
```diff
@@ -57,1932 +30,3568 @@ def _make_nlp1() -> Model:
...
+
+    def fake_ipopt(evaluator, x0, options):
+        del options
+        calls.append("ipopt")
+        seen_bounds.append(evaluator.variable_bounds)
+        return NLPResult(status=SolveStatus.OPTIMAL, x=np.array([3.0]), objective=0.0)
+
+    monkeypatch.setattr(amp_mod, "_has_cyipopt", lambda: True)
+    monkeypatch.setattr(ipm_mod, "solve_nlp_ipm", fake_ipm)
+    monkeypatch.setattr(ipopt_mod, "solve_nlp", fake_ipopt)
```
bernalde, line 450: Blocking: this monkeypatch can escape the test. `_solve_nlp_subproblem` imports `discopt.solver` while `discopt.solvers.nlp_ipopt.solve_nlp` is replaced; if `discopt.solver` is imported at that moment, its module-level `solve_nlp` alias remains bound to `fake_ipopt` after pytest restores `ipopt_mod.solve_nlp`. Running this test followed by `test_maximize_concave_uses_fast_path` reproduces the CI failure where later code calls the fake with `constraint_bounds=...`. Import `discopt.solver` before applying the patch, or patch and restore `discopt.solver.solve_nlp` as well.

python/tests/test_amp.py
```diff
@@ -57,1932 +30,3568 @@ def _make_nlp1() -> Model:
...
+        instance.problem_id: instance
+        for instance in (_unwrap_minlptests_case(case) for case in NLP_MI_INSTANCES)
+    }
+    instance = instances["nlp_mi_005_010"]
+    m = instance.build_fn()
+
+    result = m.solve(
+        solver="amp",
+        nlp_solver="ipm",
+        time_limit=2.0,
```
bernalde: Blocking: this PR-fast smoke is timing-sensitive enough to fail in CI. The Python AMP coverage job returned `SolveResult(status='time_limit', obj=None, gap=None, time=2.6s, nodes=0)` at the status assertion below, so a clean PR can fail depending on runner speed. Make this deterministic for PR CI, for example by asserting the formerly failing relaxation/classification path directly, or give the solve a budget that is not marginal and keep heavier evidence in the opt-in integration suite.

bernalde (commented):
Review summary

1. Blocking issues
- The current PR-time signal is red. Inline comments cover the three actionable blockers I found in this pass: the Makefile dry-run test assumes the local default worker count even when CI overrides it, the Ipopt monkeypatch can leak into `discopt.solver.solve_nlp`, and the new PR-fast MINLPTests smoke is too timing-sensitive for CI.
- I checked the previous maintainer review items against the current diff. The earlier substantive issues appear to have follow-up commits now, including the Rust multidimensional classifier fix and added PR-time AMP signal, but the new signal itself is currently failing and cannot be treated as merge-ready.

2. Nonblocking issues
- None from this pass.

3. Questions
- None.

4. Tests run and outcomes
- `cargo test -p discopt-core`: passed.
- `.venv/bin/python -m pytest python/tests/test_amp.py -q`: passed locally, `139 passed`.
- `.venv/bin/python -m pytest python/tests/test_rust_ir.py::TestAmpTermClassification discopt_benchmarks/tests/test_category_runner.py python/tests/test_fbbt_bindings.py::TestFBBTBindings::test_root_presolve_logs_fbbt_failures -q`: passed, `14 passed`.
- `.venv/bin/python -m pytest python/tests/test_convex_fast_path.py::TestTranslatedLPRegressions::test_translated_lp_uses_convex_fast_path -q`: passed, `4 passed`.
- `PYTEST_XDIST_WORKERS=2 .venv/bin/python -m pytest python/tests/test_amp.py::test_pr_fast_tier_excludes_heavy_manual_markers -q`: failed, reproducing the CI worker-count assertion.
- `.venv/bin/python -m pytest python/tests/test_amp.py::test_solve_nlp_subproblem_retries_ipopt_and_restores_bounds python/tests/test_convex_fast_path.py::TestConvexFastPathDetection::test_maximize_concave_uses_fast_path -q`: failed, reproducing the monkeypatch leak into a later solver test.
- `.venv/bin/python -m pytest python/tests/test_amp.py::test_former_known_failure_nlp_mi_005_runs_in_pr_fast_suite -q`: passed locally, but GitHub Actions `Python AMP coverage` failed this exact test with `SolveResult(status='time_limit', obj=None, gap=None, time=2.6s, nodes=0)`.
- GitHub Actions run `25978871675`: lint and Rust passed; `Python fast` failed; `Python AMP coverage` failed; full Python coverage and AMP integration were skipped.

5. Merge decision
- The PR should not be merged as-is. I would not merge this until the blocking issues above are addressed.

bernalde:
Pushed commit to `amp`:
- `3448275` Fix AMP PR-fast review regressions

Main changes:
- Made the Makefile dry-run tests independent of an ambient `PYTEST_XDIST_WORKERS` value, while adding explicit coverage that the PR-fast target honors CI's worker override.
- Prevented the Ipopt fake used by `test_solve_nlp_subproblem_retries_ipopt_and_restores_bounds` from being captured by `discopt.solver` during lazy import.
- Increased the PR-fast MINLPTests AMP smoke budget from 2s to 10s so it is not marginal on GitHub runners.

Tests run and results:
- `PYTEST_XDIST_WORKERS=2 .venv/bin/python -m pytest python/tests/test_amp.py::test_pr_fast_tier_excludes_heavy_manual_markers python/tests/test_amp.py::test_pr_fast_tier_honors_worker_override -q`: passed.
- `.venv/bin/python -m pytest python/tests/test_amp.py::test_solve_nlp_subproblem_retries_ipopt_and_restores_bounds python/tests/test_convex_fast_path.py::TestConvexFastPathDetection::test_maximize_concave_uses_fast_path -q`: passed.
- `.venv/bin/python -m pytest python/tests/test_amp.py::test_former_known_failure_nlp_mi_005_runs_in_pr_fast_suite -q`: passed.
- `.venv/bin/python -m pytest python/tests/test_amp.py -q`: passed, `140 passed`.
- `git diff --check -- python/tests/test_amp.py`: passed.
- `.venv/bin/python -m ruff check python/tests/test_amp.py`: passed.
- Commit hooks: ruff passed, ruff format passed, mypy skipped with no files to check, cargo fmt skipped with no files to check.
- `make test-amp-fast PYTHON=.venv/bin/python PYTEST='.venv/bin/python -m pytest' MATURIN='.venv/bin/python -m maturin' RUFF='.venv/bin/python -m ruff'`: did not complete because the build step ran `maturin develop` and pip tried to install into a read-only user site outside the workspace. The direct pytest equivalent for the changed AMP file passed.
- GitHub Actions on `3448275`: `Lint + typecheck`, `Rust`, `Python AMP coverage`, and `Python fast` passed. `Python coverage` and `AMP Alpine/incidence suite` were skipped by the workflow configuration.

- None. The only unresolved inline threads were the three blocking comments on `python/tests/test_amp.py`, and each is addressed in `3448275`.
- The maintainer's earlier top-level review comments were already addressed by the previous review-response commits on this branch; this push covers the later CI/test regressions raised by the last review.

Remaining risks or follow-up:
- I did not run the opt-in AMP integration suite locally.
- The opt-in AMP integration workflow remained skipped on GitHub for this push.

python/tests/test_amp.py
bernalde: Addressed in https://github.com/jkitchin/discopt/pull/86/commits/344827567343891fda24ffe882e58ccb29ab9104. The test imports `discopt.solver` before patching `nlp_ipopt.solve_nlp`, so the fake cannot be captured by the solver module during lazy import.

python/tests/test_amp.py
bernalde: Addressed in https://github.com/jkitchin/discopt/pull/86/commits/344827567343891fda24ffe882e58ccb29ab9104. `_make_dry_run` now clears ambient `PYTEST_XDIST_WORKERS` for default assertions and `test_pr_fast_tier_honors_worker_override` covers the CI override path.

python/tests/test_amp.py
bernalde: Addressed in https://github.com/jkitchin/discopt/pull/86/commits/344827567343891fda24ffe882e58ccb29ab9104. The PR-fast MINLPTests smoke now uses a 10s AMP budget, and the targeted test passed locally.

bernalde:
MINLPTests AMP benchmark update for commit `344827567343891fda24ffe882e58ccb29ab9104`:

No clear regression found.

Run configuration:
- Candidate label: `current-amp-20260517T023905Z-3448275`
- Reference: `amp_baseline_2026_04_20_5167373`
- Mode: `solver="amp"`
- Selected instances: 122, including convex NLPs
- Per-instance time limit: 300s
- AMP max iterations: 1000
- Artifacts: `results/minlptests-amp-vs-alpine/20260517T023905Z-3448275/`

Outcome summary:
- Current AMP: 122 pass, 0 fail, 0 error.
- Baseline AMP: 118 pass, 4 fail, 0 error.
- Pairwise baseline vs current: `both_pass=118`, `left_only_pass=0`, `right_only_pass=4`, `both_fail=0`.
- The four current-only passes are `nlp_003_014`, `nlp_003_015`, `nlp_mi_003_014`, and `nlp_mi_003_015`; the baseline timed out on these.
- Median pass time was effectively unchanged: current `0.357s`, baseline `0.362s`.
- Total counted MIPs decreased from `262` in the baseline to `55` in current AMP.

Conclusion: the benchmark is a net improvement over the April baseline, with no baseline-only pass losses. No code change was made for this benchmark result.

jkitchin:
Awesome work! merged and released!

PR #117, opened 2026-06-13
Add POUNCE-only multiphase roadmap design doc

jkitchin:
Plan to eliminate HiGHS as a runtime dependency and ship discopt as a
pip-installable solver on a single in-house engine. Covers engine roles
(POUNCE workhorse, JAX IPM differentiable path), GPU repositioned as
opportunistic rather than load-bearing, the IPM-to-basis crossover
decision for Gomory/MIR cuts, six phases with exit gates, a Phase 0
work breakdown (pure-LP path, Farkas-certificate infeasibility
handling, bound trust-gate, batch LP/QP, HiGHS as CI oracle), and a
risk register. ROADMAP.md gains a Phase 8 section linking the doc.

https://claude.ai/code/session_011S4u7hrwfBGPHpdEJwLWY4


jkitchin:
## Review — POUNCE-only roadmap (#117)

Reviewed with a fan-out of subsystem agents over the full diff (Rust simplex / cuts / crossover / MILP driver, Python POUNCE backends, `solver.py` trust-gate, `.nl` exporter, IPM-removal blast radius). The two headline findings were verified by hand against the code.

**Scope note:** the title says "design doc" but this is a ~24K-line architectural rewrite — it deletes the JAX IPM stack (~5K lines) and adds a pure-Rust LP/simplex/cuts/crossover stack, a Rust MILP B&B driver, Python POUNCE LP/QP/MILP backends, a rewritten `solver.py`, and a rewritten `.nl` exporter.

**Overall:** the in-tree trust-gate / gap-decertification machinery in `solver.py` is genuinely careful, correct, and actually tested (forced `max_iter=1` stalls and monkeypatched iter-limits → `gap_certified=False`/`"feasible"`). But there are soundness bugs that should block merge — the worst in exactly the POUNCE-only mode this PR exists to enable.

---

### 🔴 Must fix before merge

**1. `milp_pounce` reports the incumbent as `objective`; AMP/OA/GDP-LOA consume it as a lower bound → falsely-certified optimum.** *(verified by hand)*
- `milp_pounce.solve_milp` returns `objective=res.objective` — the **incumbent / upper bound** — on `feasible`/`node_limit`/`iteration_limit` exits (`solvers/milp_pounce.py:137`).
- `milp_simplex.solve_milp` does the opposite on the same exits: it deliberately returns the **dual lower bound** (`solvers/milp_simplex.py:142-147`), with a docstring explaining callers use it as a lower bound. The two backends violate the same contract in opposite directions.
- `amp.py:2912` does `LB = max(LB, float(milp_result.objective))` with **no `status == OPTIMAL` gate**. OA (`oa.py:797`) and GDP-LOA (`gdpopt_loa.py:183`) do the same.
- `get_milp_solver(backend="auto")` resolves to `milp_pounce` whenever HiGHS is absent — the `[pounce]`-only install. Pre-PR this was masked because HiGHS returned `objective=None` on a limit, firing the `x is None` guard; the migration removed that implicit protection.
- **Effect:** on a time/node-limited relaxation, AMP/OA/GDP take an *upper* bound as a *lower* bound, can inflate `LB` past the true optimum, close the gap, and report `status="optimal"` — the one non-negotiable invariant broken. No test covers it. Fix: gate the LB update on `status == OPTIMAL`, or give `MILPResult` a `bound`/`gap_certified` channel and have `milp_pounce` populate it from `res.bound`.

**2. MILP driver: a failed node LP can corrupt the incumbent with `obj=-∞` and an infeasible point.**
- On `IterLimit`/`Numerical` at a node, the driver emits `lower_bound=-INF, solution=midpoint(lb,ub), is_feasible=false` (`crates/discopt-core/src/bnb/milp_driver.rs:384-395`). `process_evaluated` then re-derives feasibility via `is_integer_feasible(midpoint)`, which can override `is_feasible:false` and promote the unvalidated midpoint to the incumbent with objective `-∞` (`crates/discopt-core/src/bnb/tree_manager.rs:268-300`).
- `gap_certified=false` prevents the `"optimal"` label, but the returned `obj`/`x` are still wrong. Masked today only because the tiny test instances never make the simplex fail at a node. Fix: a failed-LP node's solution must be ineligible to become an incumbent.

**3. `lp_pounce` maps Ipopt status 3/4 → `UNBOUNDED` unconditionally.**
- Codes 3 (`Search_Direction_Too_Small`) / 4 (`Diverging_Iterates`) are **not** unboundedness certificates — an IPM diverges the same way on an *infeasible* problem (`solvers/lp_pounce.py:45-54`). The elastic Phase-1 disambiguation the PR's infeasibility soundness rests on only runs for `ITERATION_LIMIT`/`ERROR` (`:243`), so an infeasible LP/QP can return a trusted `UNBOUNDED`. Run Phase-1 before trusting `UNBOUNDED` too.

---

### 🟠 High — should fix

- **Rust dual simplex asserts a dual-feasibility safety guarantee it never checks** (`crates/discopt-core/src/lp/simplex/dual.rs:9-14` docstring vs. no check in `try_dual`). Safe in the current "tighten one bound from parent's optimal basis" pattern, but it's a public API and `recover_basis` makes no dual-feasibility promise — a non-dual-feasible start yields a wrong `Optimal`/`Infeasible` with no fallback. Implement the check or restrict/document the precondition.
- **`solve_lp_kkt`/`solve_qp_kkt` ignore convergence status** (`solvers/lp_pounce.py:321-334`, `solvers/qp_pounce.py:241-254`) → the differentiable layer linearizes around a non-stationary point and returns **silently wrong gradients** on a non-converged solve.
- **GMI assumes integer variables sit at integer bounds; no guard** (`crates/discopt-core/src/lp/gomory.rs:223-238`). If presolve (coefficient strengthening / implied bounds — present in this repo) hands an integer var a fractional bound, the integer-ψ branch can cut the true optimum. The refinement/snapping guards don't help because the *premise* is violated, not the arithmetic. Every GMI test uses integer bounds, so it's uncovered. Gate the integer branch on an integral pinned bound.
- **Rust simplex has no final `Ax=b`/bounds validation before returning `Optimal`** (`crates/discopt-core/src/lp/simplex/primal.rs:437-464`), and the Harris δ-expansion excursion is never reset while `x_B` is maintained incrementally (drift bounded only by the 48-pivot refactor). A cheap final residual check downgrading to `Numerical` (forcing the existing fallback) is the standard guard and is absent. Bland's anti-cycling fallback is also untested.

---

### 🟡 Medium

- **`_FEAS_TOL=1e-6` Phase-1 infeasibility threshold is absolute and summed over all `m` rows** (`solvers/lp_pounce.py:245`) → false `INFEASIBLE` risk on large models; the INFEASIBLE verdict isn't re-checked by the feasibility guard (only OPTIMAL is).
- **Broken deleted-module import:** `discopt_benchmarks/benchmarks/cutest_runner.py:342` does `from discopt._jax.ipm import solve_nlp_ipm` — `ipm.py` was deleted. Swallowed by a bare `except`, so no crash, but the `discopt_ipm` CUTEst benchmark backend is now a permanent silent-ERROR path, never migrated to POUNCE.
- **NLP `custom_jvp` host path ignores solve status and propagates a NaN objective** (`_jax/differentiable.py:660-681`); inconsistent with `_compute_sensitivity_at_solution`, which does raise.

---

### 🟢 Low / docs / tests

- **Design doc is stale.** §4/§5 ("Do NOT build LP-basis warm-start") is superseded since the full plan — including the warm-started simplex — was implemented. Also: the doc claims `ipm.py`'s `ipm_solve` core is *retained* for `mccormick_nlp.py`, but `ipm.py` is **actually deleted** and `mccormick_nlp.py` imports from `nlp_pounce` — the statement is false. And `pounce-solver` is now a hard core dep (`pyproject.toml:31`), contradicting the "optional extra" framing. Worth reconciling §3/§4/§9/§12 with what shipped.
- **`.nl` exporter rewrite is semantically equivalent** (operand order, opcode emission, sign propagation, sumlist order, shared-node handling all preserved) — but no test value-round-trips the writer's output, and the deep-sum recursion fix that motivated it is itself untested.
- Stale logs/notebook prose still say "sparse IPM" / recommend `nlp_solver="sparse_ipm"` though both now alias to POUNCE.

---

**Recurring theme:** the soundness *machinery* is good, but tests exercise happy paths and tiny instances — none force a node-level simplex failure, a limited `milp_pounce` solve, an infeasible-via-code-3/4 exit, a fractional-bound GMI, or an ill-conditioned basis. Several of the bugs above survive precisely because of those gaps. Recommend blocking at minimum on finding #1.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

jkitchin:
## Re-review — fixes verified (#117)

Re-reviewed the 7 follow-up commits against the original findings: 🔴 by hand, 🟠/🟡 via a fan-out of audit agents over the post-fix code (and for the `.nl` case, by running the new test on both `main` and this branch). **Every finding is closed at the code level**, and the fixes are real — not commit-message labels over no-ops.

### Verdict by finding

| # | Finding | Commit | Status |
|---|---------|--------|--------|
| 🔴 1 | `milp_pounce` incumbent consumed as a lower bound → falsely-certified optimal | `9ff547b` | **CLOSED** — new `MILPResult.bound` dual channel; AMP/OA/GDP-LOA now read `.bound` (None-guarded, gated on `*_bound_valid`), never `.objective`. Regression test brackets the optimum (`bound ≤ opt ≤ incumbent`). |
| 🔴 2 | failed node LP corrupts the incumbent at `obj=-∞` | `50aa5ff` | **CLOSED** — incumbent promotion gated on `trusted = node_lb.is_finite()`; Rust test proves a `-inf`/integral-midpoint node leaves the incumbent unset. |
| 🔴 3 | Ipopt status 3/4 → unconditional `UNBOUNDED` | `f6853a4` | **CLOSED** — `UNBOUNDED` added to the elastic Phase-1 Farkas gate (LP+QP); `solve_*_kkt` now raise `PounceKKTError` on a non-converged solve instead of returning wrong gradients. |
| 🟠 | dual-simplex dual-feasibility precondition never checked | `03e8deb` | **CLOSED** — reduced-cost check in `try_dual` routes a non-dual-feasible warm start to the cold `solve_lp`. |
| 🟠 | GMI assumes integer vars sit at integer bounds | `03e8deb` | **CLOSED** — `use_integer` gates the integer-ψ branch on an integral pinned bound; otherwise falls back to the always-valid continuous cut. |
| 🟠 | simplex returns `Optimal` with no final feasibility audit | `03e8deb` | **CLOSED** — `Ax=b` + bounds residual check downgrades `Optimal → Numerical` on the primal Optimal path. |
| 🟡 | `_FEAS_TOL` absolute, summed over `m` rows | `248394f` | **CLOSED** — scale-aware `_FEAS_TOL · max(1, m) · rhs_scale` with the `±_INF` sentinel excluded; change is monotone in the safe direction. |
| 🟡 | NLP `custom_jvp` host path propagates a NaN objective | `248394f` | **CLOSED** — raises on non-OPTIMAL status and on a non-finite objective, matching `_compute_sensitivity_at_solution`. |
| 🟡 | broken `from discopt._jax.ipm import …` in `cutest_runner` | `6d56f7f` | **CLOSED** — migrated to `nlp_pounce.solve_nlp` (signature-compatible); no `_jax.ipm` import references remain. |
| 🟢 | stale design doc / untested `.nl` rewrite | `8464047`, `6d56f7f` | **CLOSED** — status banner corrects all three superseded claims (warm-start *was* built; JAX IPM deleted not retained; `pounce-solver` is a core dep). Deep-`.nl` test **verified to raise `RecursionError` on `main` and pass on this branch**. |

### Remaining nits (non-blocking)

1. **The three Rust 🟠 fixes ship without regression tests.** The dual-feasibility precheck, GMI fractional-bound guard, and primal residual audit are each sound and correctly placed, but happy-path-only — nothing constructs a dual-infeasible warm start, a fractional-bound integer var, or a drifted "Optimal," so a future edit to any guard line wouldn't be caught. This is the same "tests cover only happy paths / tiny instances" gap that allowed the original bugs; worth closing here since these three guard exactly the falsely-certified-`optimal` outcome. (The 🔴, medium, and `.nl` fixes all do carry tests.)
2. **Minor:** in `try_dual` a free nonbasic variable is checked as if at-lower (a free var actually needs reduced cost ≈ 0); currently harmless because the basis builder forces free vars basic. And `03e8deb`'s comment describes the `Numerical` downgrade as forcing a "cold fallback re-solve" — it actually marks the node uncertified and branches (still sound, just not a re-solve).

**Bottom line:** every soundness hole from the first pass is closed, including the headline #1 (false optimal in the POUNCE-only install) with a bracketing test. The one ask before merge would be regression tests on the three untested Rust guards.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #243, opened 2026-06-20
feat(modeling): set & index abstractions (Phase 7, M1–M7)

jkitchin:
Implements the Phase 7 roadmap item **"Set and index abstractions"** — a
Pyomo/JuMP-style named-set layer for sparse models, built as a pure-Python
desugaring over the existing flat model (no Rust core, JAX, or solver changes).
By solve time only flat `Variable`/`Constraint` objects exist.

Scope is the full feature (M1–M7), not just sets:

- **Named sets & algebra** (`discopt.modeling.sets`): `Set`, `RangeSet`,
  `ProductSet` with arbitrary hashable members, inferred/declared `dimen`,
  algebra (`|`, `&`, `-`, `*`), filtering (`where`, `with_first`, `with_last`),
  value `__eq__`/`__hash__`; lazy products that accept flat or nested keys.
  `Model.set(...)` registers a named set.
- **Indexed variables/parameters** (`discopt.modeling.indexed`): `IndexedVar` /
  `IndexedParam` backed by a single flat variable/parameter; `Model.continuous/
  binary/integer/parameter(..., over=SET)` (`@overload`ed so `over=` returns the
  indexed container type); per-key bounds via scalar/dict/callable.
- **Indexed constraints**: `Model.constraint(SET, rule, name=)` → one constraint
  per member named `name[key]`, with a `Skip` sentinel; `subject_to` accepts
  generators; `dm.sum`/`dm.prod` aggregate over sets (and over a bare indexed
  container).
- **Linear fast path**: single-variable-affine, uniform-sense families are
  emitted as one sparse `add_linear_constraints` builder call (`fast=True`
  default), with transparent fallback and identical results.
- **`.nl` export of builder rows/objective**: fast-path/`add_linear_*`
  constraints and linear/quadratic objectives that live only in the Rust builder
  are now reconstructed by the `.nl` writer (previously dropped). The quadratic
  reconstruction matches the builder's `S = triu(Q) + striu(Q).T` convention.

Docs: `docs/notebooks/sets_and_indexing.ipynb` + design doc
`docs/design/sets-and-indexing.md`. Examples: `example_transportation` /
`example_assignment` / `example_multicommodity_flow` (exported at top level).
ROADMAP item marked Done.

Co-Authored-By: Claude Opus 4.8
Claude-Session: https://claude.ai/code/session_01RhTPqKQtjd1XjoANXKPPxe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -159,13 +159,15 @@ The release procedure that produces these entries is documented in
 
 ### Fixed
 
-- **`.nl` export of builder constraints** (`fix(export)`). Linear constraints
-  built directly into the Rust builder — via the fast-construction
+- **`.nl` export of builder constraints and objective** (`fix(export)`). Linear
+  constraints built directly into the Rust builder — via the fast-construction
   `add_linear_constraints` API and the indexed-constraint fast path — were
-  silently omitted from `to_nl` / `to_mps`-adjacent `.nl` output, which reads
-  `model._constraints`. The model now records each emitted block and the `.nl`
-  writer reconstructs those rows, so a fast-path model round-trips through `.nl`
-  with all constraints intact.
+  silently omitted from `to_nl`, which reads `model._constraints`; likewise a
+  linear objective set via `add_linear_objective` was exported as a zero
+  placeholder. The model now records each emitted block (constraints and the
+  `c'x + constant` objective) and the `.nl` writer reconstructs them, so a
+  fast-path model round-trips through `.nl` with all constraints and the correct
+  objective (including a constant offset and `maximize` sense) intact.
 - **Relaxation soundness hardening** across the global-opt loop: reject a
   fabricated finite bound on an unbounded McCormick relaxation (`himmel16`,
   `fix(soundness)`); never trust an unconverged simplex objective as an LP lower
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,14 +12,27 @@ The release procedure that produces these entries is documented in
 
 ### Added
 
-- **Named index sets & set algebra** (`feat(modeling)`). New
-  `discopt.Set` / `discopt.RangeSet` (and `ProductSet`) plus `Model.set(...)` add
-  a Pyomo/JuMP-style named-set layer for sparse models: arbitrary hashable
-  members with inferred/declared `dimen`, set algebra (`|`, `&`, `-`, `*`), and
-  filtering (`Set.where`, `with_first`, `with_last`). This is milestone M1 of the
-  Phase 7 "set & index abstractions" roadmap item — a pure-Python desugaring
-  layer; indexed variables/constraints land in later milestones. Design in
-  `docs/design/sets-and-indexing.md`.
+- **Set & index abstractions** (`feat(modeling)`). A Pyomo/JuMP-style named-set
+  layer for sparse models, implemented as a pure-Python desugaring over the
+  existing flat model (no solver/backend changes). Completes the Phase 7
+  roadmap item. New public API:
+  - `discopt.Set` / `discopt.RangeSet` (+ `ProductSet`) and `Model.set(...)`:
+    arbitrary hashable members with inferred/declared `dimen`, set algebra
+    (`|`, `&`, `-`, `*`), and filtering (`Set.where`, `with_first`,
+    `with_last`). Product sets are lazy and accept flat or nested keys.
+  - `Model.continuous/binary/integer/parameter(..., over=SET)` returning
+    `IndexedVar` / `IndexedParam` backed by a single flat variable/parameter;
+    per-key bounds/values via scalar, `dict`, or callable.
+  - `Model.constraint(SET, rule, name=)` generating one constraint per member
+    (named `name[key]`), with a `Skip` sentinel; `subject_to` now accepts
+    generators of constraints. `dm.sum`/`dm.prod` aggregate over sets.
+  - A transparent **linear fast path**: single-variable-affine, uniform-sense
+    families are emitted as one sparse-matrix builder call (`fast=True`
+    default) with identical results, falling back automatically otherwise.
+  Documented in `docs/notebooks/sets_and_indexing.ipynb`; design in
+  `docs/design/sets-and-indexing.md`; examples
+  `example_transportation` / `example_assignment` /
+  `example_multicommodity_flow`.
 - **Irreducible Infeasible Subsystem (IIS)** (`feat(infeasibility)`, #227). New
   `compute_iis(model)` returns a minimal infeasible subset of constraints/bounds
   via deletion filtering — exact for LP/MILP/convex, best-effort for nonconvex.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ The release procedure that produces these entries is documented in
 
 ### Added
 
+- **Named index sets & set algebra** (`feat(modeling)`). New
+  `discopt.Set` / `discopt.RangeSet` (and `ProductSet`) plus `Model.set(...)` add
+  a Pyomo/JuMP-style named-set layer for sparse models: arbitrary hashable
+  members with inferred/declared `dimen`, set algebra (`|`, `&`, `-`, `*`), and
+  filtering (`Set.where`, `with_first`, `with_last`). This is milestone M1 of the
+  Phase 7 "set & index abstractions" roadmap item — a pure-Python desugaring
+  layer; indexed variables/constraints land in later milestones. Design in
+  `docs/design/sets-and-indexing.md`.
 - **Irreducible Infeasible Subsystem (IIS)** (`feat(infeasibility)`, #227). New
   `compute_iis(model)` returns a minimal infeasible subset of constraints/bounds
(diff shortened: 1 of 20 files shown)


jkitchin (commented):
## Quality review — named index sets & indexed modeling

Reviewed across all 20 files with three parallel reviewers, then verified every load-bearing claim locally (and discarded two that didn't reproduce).

**CI:** Rust ✅ · Python fast ✅ · Python AMP coverage ✅ — **only `Lint + typecheck` ❌** (6 mypy errors). All tests pass; it's red purely on type-checking.

### 🔴 Blocker — silent wrong answer
**`dm.sum(indexed_var)` sums the index *keys*, not the variables.** Verified decisively:

```python
S = m.set('S', [10, 20, 30])
y = m.continuous('y', lb=1, ub=5, over=S)
m.minimize(dm.sum(y))      # returns 60.0  (= 10+20+30, the keys)
                           # expected 3.0  (= 1+1+1,   the vars)
```

`IndexedVar.__iter__` yields the keys and `_is_term_iterable` accepts an `IndexedVar`, so `sum` folds the keys in as constants. The *natural* syntax produces a meaningless objective with **no error** (the working form is `dm.sum(y.flat)`). Fix: exclude `IndexedVar`/`IndexedParam` from `_is_term_iterable`, or make their `__iter__` yield element expressions, or reject them with a clear error in `sum`/`prod`.

### 🟠 Major
- **No `__eq__`/`__hash__` on `Set`** — value equality falls back to identity: `Set('a',[1,2,3]) != Set('a',[1,2,3])`. A footgun for a value type billed as the single source of truth for which index tuples exist.
- **mypy: 6 errors = the CI failure.** `core.py:1508/1546/1588/1640` are declared `-> Variable` / `-> Parameter` but return `IndexedVar` / `IndexedParam`, which are **not** subclasses — the signature genuinely lies to callers and the type-checker. Fix with `@overload` on `over=` (`None → Variable`, `Set → IndexedVar`), not a blind cast. Plus `indexed.py:322` (benign `no-any-return`) and `nl.py:364` `float` vs `float64` (typing nit, not a numeric bug).
- **PR body undersells scope.** The body says *"M1: pure-Python desugaring, sets only,"* but the diff is effectively M1–M7: indexed vars/constraints/params (`indexed.py`), the linear fast path, and `.nl` export changes. Worth correcting so reviewers don't under-scope it.

### 🟡 Minor
- **`dm.prod` over an empty set raises `IndexError`** (`terms[0]`); should return identity `1` or raise a clear error.
- **The `nl.py` change is a *good* latent-bug fix, not scope creep:** fast-path / `add_linear_*` rows write straight into the Rust builder, bypassing `model._constraints`, so they were previously **dropped from `.nl` export**. Correct and necessary.
- `example_assignment` / `example_multicommodity_flow` exist and solve but aren't exported at top level (only `example_transportation` is); `discopt.ProductSet` is absent.

### ✅ Genuinely solid
Desugaring (one flat `Variable` + a key→position map) keeps the Rust/JAX/solver layers untouched as claimed; set algebra, ordinals, and lazy `ProductSet` iteration verified correct; all 3 examples solve to optimal; 63 set/indexed + 27 fast-path tests pass; `.nl` roundtrip (linear + quadratic, incl. constant offset) reproduces the optimum.

### Discarded (didn't survive verification)
- A reported `ProductSet.__contains__` "ambiguous-grouping false positive" — **did not reproduce**: `('s',('u','c1')) in arcs*comm` correctly returns `False`.
- A reported "_jax / cut-pool files + deleted `.nl` fixtures in scope" — **stale-worktree artifact** (its `origin/main` predated the #242 merge); the real PR is the 20 listed files, none of them `_jax`.

---

**Bottom line:** correctness is mostly sound, but the **`dm.sum(indexed_var)` key-summing bug is a merge blocker** (it silently builds wrong models on the most natural syntax), and the **mypy errors must be fixed for CI to go green**.

jkitchin:
Thanks for the thorough review — all findings addressed in `a83ccda`.

**🔴 Blocker — `dm.sum(indexed_var)` summed keys.** Fixed. Indexed containers now carry a marker; `dm.sum`/`dm.prod` expand them to their element expressions and `_is_term_iterable` excludes them. `dm.sum(y)` over `[10,20,30]` now returns `3.0` and equals `dm.sum(y.flat)`. Regression tests added (`TestSumProdOverIndexedContainer`), including the exact example from the review.

**🟠 No `Set.__eq__`/`__hash__`.** Added value equality on `_SetBase` (same `dimen` + members, name-independent) + `__hash__`. `Set('a',[1,2,3]) == Set('a',[1,2,3])` and they're usable in a `set`/`dict`. ProductSet compares equal to the materialized `Set`.

**🟠 mypy / CI lint.** Switched the `over=` constructors to `@overload` (`None → Variable`, `Set → IndexedVar`) instead of the cast — signatures no longer lie to callers. Also fixed the two CI mypy errors my local mypy (1.19.1) didn't surface: `nl.py:364` (float64 loop-var reassignment) and `indexed.py:322` (`no-any-return`). CI mypy should be green now.

**🟡 `dm.prod` over empty set → `IndexError`.** Now returns the identity `Constant(1.0)` (shared `_multiply_terms` helper).

**🟡 Missing top-level exports.** `discopt.ProductSet`, `discopt.example_assignment`, and `discopt.example_multicommodity_flow` are now exported.

**PR body scope.** Updated to reflect the full M1–M7 feature (sets, indexed vars/params/constraints, fast path, `.nl` export) rather than just M1.

On the discarded items — agreed the `nl.py` change is a real latent-bug fix (builder rows were dropped from `.nl`), and the `ProductSet.__contains__`/`_jax` reports were correctly not reproduced.

---

jkitchin (commented):
## Follow-up review — updates look good ✅

Checked out the new head (`a83ccda3`) and re-ran the exact verification probes from the first review. **All four flagged issues are properly fixed**, with regression tests and clean implementations:

| earlier finding | fix | verified |
|---|---|---|
| 🔴 `sum(indexed_var)` summed keys (60.0 not 3.0) | `_is_term_iterable` now excludes indexed containers; `sum`/`prod` expand via `_expand_indexed_container`. **`__iter__` still yields keys**, so `for k in x` and `dm.sum(x[k] for k in x)` keep working | ✅ now 3.0; key-iteration intact |
| 🟠 no `Set.__eq__`/`__hash__` | equality by members+dimen (name-irrelevant), matching hash | ✅ `==` True, `{a,b,c}` dedup works |
| 🟠 mypy (6 errors) | proper `@overload` (`over=` → `IndexedVar`/`IndexedParam`, else `Variable`/`Parameter`) — not a blind cast | ✅ mypy clean (229 files), Lint CI green |
| 🟡 `prod(empty)` → `IndexError` | returns identity `1` | ✅ |
| exports / `nl.py` float | `example_assignment`, `example_multicommodity_flow`, `ProductSet` exported; float cast fixed | ✅ |

`test_sets_adversarial.py` locks each one in (e.g. `test_sum_indexed_var_sums_variables_not_keys` asserts 3.0 not 60.0). Nicely done — the `sum` fix in particular is the right shape: it makes `dm.sum(x)` intuitive **without** breaking key iteration.

### 🟡 One new issue the updates introduced
`test_indexed_correctness.py::TestPyomoCrossCheck::test_transportation_matches_pyomo` calls `dm.from_pyomo()`, which is a pre-existing **Phase-4 stub** that raises `NotImplementedError`. Its guard is `pytest.importorskip("pyomo.environ")` — that only skips when pyomo is *absent*. With pyomo **installed** the test runs and **fails** (reproduced locally):

```
python/discopt/modeling/core.py:3304: NotImplementedError: Pyomo import requires pyomo bridge (Phase 4)
FAILED test_indexed_correctness.py::TestPyomoCrossCheck::test_transportation_matches_pyomo
```

It's green in CI only because the CI image has no pyomo. Suggest guarding on the *feature*, not the import — e.g. `@pytest.mark.xfail(reason="from_pyomo is Phase 4, not implemented", raises=NotImplementedError)` or `pytest.skip(...)` — until `from_pyomo` lands.

### Design note (not a defect)
Name-irrelevant set equality means two distinctly-named sets with coincidentally identical members compare/hash equal. That's Pyomo-consistent and correct for index domains — just flagging it as an intentional choice worth documenting.

**Net:** the original blockers are resolved cleanly; only the pyomo-cross-check test guard remains.

jkitchin:
Fixed the pyomo-cross-check guard in `f9d902b`. Good catch — `importorskip` only covered pyomo's *absence*, and `from_pyomo` is still the Phase-4 `NotImplementedError` stub, so the test failed whenever pyomo was actually installed.

Now decorated with `@pytest.mark.xfail(raises=NotImplementedError, reason="from_pyomo is a Phase 4 stub …")`. Verified both paths locally with pyomo 6.10.1 installed:
- pyomo present → **XFAIL** (expected failure on the `NotImplementedError`), not a hard failure
- pyomo absent → still skipped via `importorskip`

It'll flip to XPASS (and flag) once `from_pyomo` lands, which is the right signal.

On the design note: agreed, name-irrelevant set equality is the intended (Pyomo-consistent) behavior for index domains — it's documented in the `Set.__eq__` docstring ("two sets with identical members are equal regardless of name; name is metadata").

---

PR #244, opened 2026-06-20
feat(decomposition): Benders / Generalized Benders / Lagrangian relaxation

jkitchin:
Implements ROADMAP Phase 6 decomposition: classical **Benders**, **Generalized Benders** (convex-NLP recourse), and **Lagrangian relaxation** (subgradient + bundle + primal recovery), plus a per-node **Lagrangian B&B bound hook**. Opt-in via `model.solve(decomposition="benders"|"lagrangian")` and `lagrangian_bound=True`; default behaviour is unchanged.

## Soundness
- **Benders / GBD** anchor optimality cuts at the **dual** value (classical: complete row-duals + variable reduced-cost terms; GBD: Lagrangian-dual underestimator with sign-projected multipliers and a closed-form box-min correction), so cuts stay valid under inexact interior-point primals. Nonconvex GBD withholds `bound`/`gap_certified`; an unbounded-recourse fallback also downgrades to heuristic mode rather than reporting an unsound bound.
- **Lagrangian** uses the rigorous MIP dual bound (`res.bound`, not the incumbent) everywhere a lower bound is needed; multipliers projected to the dual-feasible orthant; the B&B hook combines via `max(...)` (can only prune more, never unsoundly).
- Validated across six adversarial rounds: randomized fuzzing against scipy/monolithic oracles (zero unsound bounds, zero false-optimals), curated known-optimum examples, and recovery-feasibility checks.

## Also included
- Core simplex **equilibration-scaling** fix (power-of-two, gated at `SCALE_TRIGGER=1e6`, LP optimum invariant) that resolved a mixed eq/ineq wrong-optimum, plus off-box node-LP rejection guards.
- Docs: `tutorial_benders.ipynb`, `tutorial_lagrangian.ipynb`; CHANGELOG + ROADMAP (Phase 6 rows flipped to Done).

Deferred (tracked ROADMAP row): Dantzig-Wolfe / column generation / branch-and-price.

https://claude.ai/code/session_01WkS2VTV1yKisnen4JigBQu


jkitchin (commented):
## Review — decomposition (Benders / GBD / Lagrangian)

Reviewed all 33 files across three focused passes (Benders/GBD, Lagrangian, structure+integration+Rust), and verified the load-bearing claims locally — including fuzzing both solvers against independent oracles and running the full soundness suites.

**CI:** Rust ✅ · Python fast ✅ · Python AMP coverage ✅ — **only `Lint + typecheck` ❌** (mypy). 58 decomposition tests pass locally (incl. the slow soundness suites).

### ✅ Soundness is solid — the math is correct
This is the part that matters most for decomposition, and it holds up:
- **Lagrangian** uses `res.bound` (the rigorous MIP dual bound), **not** `res.objective` (the incumbent), everywhere a lower bound is needed — so an early-stopped subproblem still yields a valid bound. Constraints are canonicalized to `a·x ≤ rhs`, multipliers projected to `λ ≥ 0` → `L(λ) ≤ opt`. The B&B hook combines via `max(result_lbs[i], L)`, which can only prune more, never unsoundly.
- **Benders/GBD** anchor optimality cuts at the **dual** value (complete row-duals + reduced-cost terms / Lagrangian-dual underestimator), which stays valid under inexact interior-point primals — confirmed by fuzzing 100+ instances against scipy/monolithic oracles with **zero unsound bounds and zero false-optimals**. A nonconvex GBD subproblem correctly suppresses `bound`/`gap_certified` rather than emitting a wrong number; integer-recourse / non-binary-infeasible cases raise clean `NotImplementedError`s.
- **Opt-in & default-safe:** `decomposition=None` and `lagrangian_bound=False` by default; the hook no-ops cleanly (`try_build → None`) on non-linear / maximize / no-coupling and is try/except-wrapped. Structure detection does **not** false-positive (a dense model → 1 block, not separable).

### 🔴 CI blocker — mypy (6 errors, all in the new module)
`Lint + typecheck` is red on 6 errors in `lagrangian/solver.py` (184, 261, 301), `benders/solver.py` (471), `benders/gbd.py` (186, 459). They're benign (union-narrowing `gap = None` → float; `Any` returns from `.copy()`/`return cuts[-1][2]`; dict invariance) — no hidden bug — but they fail CI. Fix with annotations (`gap: float | None`, `-> np.ndarray` / casts, `dict[str, float | np.ndarray]`).

### 🟠 Major (process) — title/scope mismatch
PR title is **"docs: mark Phase 7 IIS row as Done"**, but the diff is **+5081/−19 across 33 files** implementing Benders/GBD/Lagrangian **plus a core Rust LP-scaling change**. A `docs:`-titled PR touching `crates/.../simplex/scaling.rs` (used by *every* LP/MILP solve) will get under-scrutinized. Please rename to `feat(decomposition): …`.

### 🟠 Major (correctness, platform-specific) — the new simplex test fails on darwin/arm64
`test_simplex_scaling.py::test_genuine_wide_range_matches_highs[13]` **fails locally on darwin/arm64** — the Rust simplex returns an out-of-box point `x[4] = −1.31e-4` (lower bound 0, tol 1e-6). It **passes on CI (ubuntu)**, so it's not a CI blocker, but it's a real numerical-robustness gap in the core LP path: the scaling fix (sound and necessary — power-of-two equilibration, gated at `SCALE_TRIGGER=1e6`, LP optimum invariant) targets *noise* entries but the off-box-output defect persists for *genuine* wide-range coefficients. It's mitigated for B&B nodes by the new box-violation rejection guards in `_solve_node_lp_*`, but a **direct `solve_lp` caller still gets a bound-violating solution**. Worth resolving (or marking the test `xfail` on arm64 with a tracking issue) rather than leaving a platform-dependent core-LP result.

### 🟡 Minor
- **`gbd.py:434–453`** — in nonconvex mode the in-loop gap check sets `status="optimal"` even though `bound=None`/`gap_certified=False`. The *number* is a genuine feasible cost, but a caller keying on `status` alone is misled. Gate the `"optimal"` promotion on `is_convex`.
- **`gbd.py:334–342`** — on a multiplier size mismatch the code emits a **gradient-only cut** (drops the `μᵀg` term) with no weak-duality backing. Currently masked by the gates, but it's an unguarded soundness path — prefer a no-good cut / skip when multipliers are unavailable and constraints are active.
- **Test gaps:** GBD has no soundness *fuzz* (6 fixed seeds only) vs Benders'; no test of the time-limited-subproblem early-stop bound path; no adversarial-λ test; `_project_mu`'s `>=`/`==` sign convention is only exercised indirectly.

### Notebooks / docs
`tutorial_benders.ipynb` / `tutorial_lagrangian.ipynb` exist, are in `_toc.yml`, and use MyST `{cite}` with all 7 keys resolving in `references.bib`. (Did not run a `jupyter-book build` zero-warning check.)

---

**Bottom line:** the decomposition feature is **well-built and sound** (opt-in, weak-duality cuts, `max()`-combined bounds, no false-optimals under fuzzing). Before merge: **(1)** fix the 6 mypy errors (CI blocker); **(2)** rename the PR to reflect the real scope; **(3)** resolve or `xfail`-with-issue the darwin/arm64 simplex off-box failure; **(4)** the two GBD minors (nonconvex `status` honesty, gradient-only-cut fallback) are worth tightening.

jkitchin (commented):
## Follow-up review — all items addressed ✅

Re-checked the new head (`d6d5e42`) against every point from the first review, and re-verified the load-bearing ones **on the same darwin/arm64 machine** where they originally failed:

| earlier finding | fix | verified |
|---|---|---|
| 🔴 mypy (6 errors, CI blocker) | annotations (`d72a109`) | **0 errors — `Success`**; `Lint + typecheck` ✅ |
| 🟠 title/scope mismatch | renamed to `feat(decomposition): …` | ✅ |
| 🟠 `test_simplex_scaling[13]` off-box point (darwin/arm64) | simplex box-snap (`d6d5e42`, `lp_simplex.py`) | **16 passed** (was 1 failed) ✅ |
| 🟡 GBD nonconvex `status` honesty | `status="optimal"` now gated on `is_convex and bound_rigorous`; `gap_certified` consistent | ✅ (gbd.py:462/470/501) |
| 🟡 GBD gradient-only-cut fallback | `bound_rigorous` flag withholds the bound on the non-rigorous / unbounded-recourse path (`1b98411`) | ✅ |
| 🟡 no GBD soundness fuzz | added (`test_gbd.py` +86) | **38 passed** incl. slow soundness + new GBD fuzz ✅ |

The `bound_rigorous` mechanism is the right shape — it makes the reported `bound`, the `status`, and `gap_certified` all agree, and it fails *closed* (no bound rather than an unbacked one) on the gradient-only path. The simplex box-snap also fixes the platform-specific off-box point cleanly, so a direct `solve_lp` caller no longer gets a bound-violating solution.

**Status:** Rust ✅ · Lint+typecheck ✅ · Python AMP coverage ✅ — only `Python fast` still running. Nothing outstanding from my side; the decomposition feature is sound, opt-in, and now CI-clean. 🚢

Issue #231, opened 2026-06-18
Complementarity constraints (x·y=0, x,y≥0): add IR node + disjunctive propagation/branching


jkitchin:
## Summary

discopt has no representation for complementarity constraints `0 ≤ x ⊥ y ≥ 0`
(`x·y = 0, x,y ≥ 0`). There is no `Complementarity` node in the expression IR, and
FBBT does not exploit the zero-product disjunction. Today a user can only encode it as
an equality on a bilinear term, which the solver handles generically — poorly.

## Why it matters

Complementarity is the defining structure of a whole problem class discopt currently
can't address reliably:
- **MPCCs / MPECs** (programs with equilibrium/complementarity constraints).
- **Bilevel optimization** reformulated by replacing the lower level with its KKT
  conditions — complementary slackness becomes exactly these constraints.
- **Equilibrium models** — Nash games, market/economic equilibria, contact mechanics.

Encoding `x·y=0` as a smooth bilinear equality is both **weak and degenerate**:
- The McCormick relaxation of `xy=0` over a box cannot represent the either/or
  ("x=0 OR y=0") disjunction → near-vacuous relaxation → bounds don't converge, trees
  blow up or stall.
- Complementarity systems violate standard constraint qualifications (MFCQ fails on the
  feasible set), so spatial branching on the smooth equality is inefficient.

A complementarity-aware solver instead branches on the exact, finite disjunction
(`x=0` vs `y=0`) and propagates the obvious bound logic.

## Scope to investigate

1. **IR:** add a first-class complementarity constraint/node (pair of expressions or
   variables required complementary), plus modeling-API surface.
2. **FBBT:** forward — product includes zero (already true); backward — if `x_L > 0`
   force `y = 0`, if `y_L > 0` force `x = 0` (the reference rule).
3. **Branching:** disjunctive branch (`x = 0` vs `y = 0`) instead of spatial bisection
   on the bilinear term.
4. **Relaxation:** how the complementarity disjunction enters the McCormick/convex
   relaxation (e.g. per-branch tightening).

## Validation

- FBBT unit tests for the backward inference rule.
- A small MPCC / KKT-reformulated bilevel instance: assert convergence to the known
  global optimum and a finite, sane node count vs. the current bilinear-equality
  encoding.
- Global correctness gate (`incorrect_count ≤ 0`) unchanged.

Companion gap: binary-indicator FBBT (separate issue). Both are structural/MILP-logic
features rather than continuous interval math.

[...]

Issue #189, opened 2026-06-17
Tracking: fractional-power / general-nonconvex weak relaxation (chakra, demo7, clay0303hfsg, cvxnonsep_nsig30) — finds global, can't certify


jkitchin:
## Summary (umbrella tracking issue)

A class of MINLPLib problems where discopt **finds the global optimum but the
McCormick / fractional-power relaxation is too loose** to certify it quickly.
All are **sound** (valid bound below the optimum) — this is a
**tightness/performance** gap, not a correctness bug.

## Status — re-triaged on `main` @ `3841071` (2026-06-17)

The two instances vendored in-repo **now certify** (intervening work — AMP merge,
PSD moment cuts #206, FBBT/factorable-reform passes — appears to have closed the
certification gap). The original "finds global, can't certify" headline no longer
holds for them; what remains is a **speed** gap and two **unverified** instances.

| Problem | Filed status | Now (`3841071`) |
|---|---|---|
| `cvxnonsep_nsig30` | incumbent 130.77, never reached global 130.63 | ✅ optimal, **certified**, obj 130.6287 (fast) |
| `clay0303hfsg` | correct obj, no cert, ~32s | ✅ optimal, **certified**, obj 26669.11 (~60s; still slow vs BARON 1.43s) |
| `chakra` | obj −179, bound −13139 (huge gap) | ⚠️ not vendored — **unverified**, pick up locally |
| `demo7` | finds −1589042.39, never certifies, ~30s (BARON 0.05s) | ⚠️ not vendored — **unverified**, pick up locally |

(Both vendored runs: `gap_tolerance=1e-4`.)

## Remaining work (to be picked up locally, where `chakra`/`demo7` are available)

1. **Verify `chakra` and `demo7`** on current `main` — confirm whether they still
   fail to certify, or are already closed like the vendored two.
2. **Performance** — `clay0303hfsg` certifies but is slow (~60s vs 1.43s): the
   relaxation is sound but loose. Candidate directions (must stay sound):
   - tighter fractional-power / `x^p` envelopes (sign-aware, monotone-aware),
   - OBBT on the variables feeding power/fractional terms to shrink envelopes,
   - better branching (most-violated / reliability) on the spatial variables.
3. **Regression guard** — add `cvxnonsep_nsig30` / `clay0303hfsg` as
   `gap_certified` tests so the now-closed cases don't silently re-open.

## Soundness / regression gates

[...]

Issue #188, opened 2026-06-17
kall_congruentcircles_c51: nonconvex global-search capability gap (feasible 1.5371 vs global 1.073, sound)


jkitchin:
## Summary

`kall_congruentcircles_c51` exposes a **nonconvex global-search capability gap**
(sound — NOT a bug, NOT a SUSPECT). discopt returns a feasible incumbent
**1.5371** vs BARON's global **1.073** (~43% worse), correctly **not certified**.
The incumbent and bound are valid; discopt simply does not reach the global
optimum within budget.

## Problem structure

14 continuous vars, 34 constraints, nonconvex:

```
minimize  objvar = x12*x13 - 3.927          (bilinear e2 + linear e1)
s.t.      (xi-xj)^2 + (yi-yj)^2 >= 1         (pairwise congruent-circle separation)
```

The `>= 1` separation constraints make the feasible region the **nonconvex
complement of disks** — many local optima.

## Evidence gathered (so it isn't re-chased)

- **40/40 uniform-random multistarts find no feasible point at all** (best =
  inf): the tight separation constraints are almost never satisfied from random
  points. ⇒ "increase `n_random` multistart" is a **confirmed dead end** as a
  general lever.
- c51 has **no GAMS initial values** (`x.l`). The feasible 1.5371 comes from
  discopt's own relaxation-guided node NLP solves (McCormick LP point → NLP),
  which do reach feasibility but converge to a 1.5371 local optimum.
- Smaller siblings match BARON: `kall_congruentcircles_c41` 0.8584,
  `kall_circlesrectangles_c1r13` 0.2146. Only c51 (one more congruent circle)
  opens the gap ⇒ combinatorial multimodality growth — classic spatial-B&B
  territory.

## Why no quick fix was shipped

Closing this needs **substantial** spatial-B&B / global-search work (tighter
bilinear relaxations, feasibility pump, deeper node exploration). The one obvious
general lever (random multistart) is ruled out, and a one-problem hack would
violate the generality constraint. Logged as a capability gap; behavior stays
sound (valid feasible incumbent, honestly not certified).

## Acceptance criteria

- [ ] `kall_congruentcircles_c51` reaches (ideally certifies) the global 1.073
      via a **general** improvement (e.g. a feasibility pump or tighter bilinear
      relaxation), not a problem-specific seed.
- [ ] 0 SUSPECT preserved; siblings c41/c1r13 do not regress.

---
[...]

Issue #187, opened 2026-06-17
autocorr_bern* wall #2: full-DAG Jacobian XLA compile exceeds budget — exact Fortet/Glover binary-multilinear linearization to bypass JAX


jkitchin:
## Summary

`autocorr_bern*` (e.g. `autocorr_bern25-25`, `autocorr_bern35-09`) does not
return within a 30 s budget. After fixing wall #1 (a dense McCormick-lift
blow-up — separately guarded with a densification fallback), the remaining wall
is **architectural**: discopt JIT-compiles the Jacobian of the full
~3000-monomial quartic objective DAG, and **XLA compilation of a graph that
large is a one-time cost that by itself exceeds a 30 s budget**.

## Evidence

faulthandler dumps (>20 s, still compiling) land in:

```
nlp_evaluator.py:395  evaluate_jacobian
  -> jax/.../compiler.py  backend_compile_and_load
reached from solver.py:587  _tighten_node_bounds_with_status
```

Gating any single call site only **defers** the compile: every gradient-needing
path (node FBBT, NLP multistart, McCormick-NLP bound) triggers the same full-DAG
Jacobian compile.

## Why it's architectural

The cost is intrinsic to JAX/XLA-compiling a ~3000-term factorable objective's
Jacobian. It is effectively unavoidable in the current evaluator architecture
for binary-multilinear models of this size.

## Proposed fix (substantial feature, deferred)

Avoid full-DAG JAX compilation for huge **factorable / binary-multilinear**
objectives. The autocorr objective is a polynomial in **binary** variables, so
an **exact Fortet/Glover binary-product (multilinear) linearization** would hand
a pure-binary multilinear program straight to the integer B&B — **no spatial
relaxation and no JAX Jacobian at all**. This is exact (no soundness risk from
the linearization itself) but is a multi-piece feature with regression risk:

- recognize a pure-binary multilinear objective/constraint,
- allocate Fortet aux variables + the standard product-linearization rows,
- route to integer B&B bypassing the spatial/JAX path,
- fall back to the current path when any factor is continuous.

## Soundness / regression gates

- [ ] Fortet linearization is exact for 0/1 products — property-test
      `aux == prod(binaries)` at all integer points.
- [ ] Fall back cleanly to the existing JAX path when not pure-binary-multilinear.
- [ ] 0 SUSPECT corpus-wide; no certified problem regresses.

## Acceptance criteria

- [ ] `autocorr_bern*` returns within budget with a valid bound (ideally
[...]

PR #245, opened 2026-06-20
SymPy-driven envelopes, structured-cut patterns, and an auto-firing recognizer

jkitchin:
Design-time toolkit (`discopt._jax.symbolic`) that uses SymPy to **derive, verify, and code-generate** convex/concave envelopes and structured cuts for nonlinear MINLP terms — with SymPy strictly off the solver hot path (it emits pure-JAX closures / numeric cut coefficients). Grew well beyond the original Phase-1 scope; summary below.

> This feature was inspired by a conversation with Milan Rother about using SymPy with agents to derive and verify math.

## Envelope engine (Phases 1–8)
- **`envelope_deriver` + `codegen` + `runtime`**: SymPy curvature classification (convex/concave, single-inflection) with non-smooth kink detection (so Weymouth `f|f|` isn't silently misclassified as convex); closed-form *and* numeric (JAX-bisection) tangent solving with secant fallback; emits jit/grad/vmap-safe `(x, lb, ub) -> (cv, cc)` closures.
- **`verification`**: theorem-style soundness + curvature + tightness certification over randomized boxes.
- **Domain packs**: gas (Weymouth `f|f|`, signed-power), power (cos/sin QC building blocks), chem-eng (Arrhenius, Langmuir/Monod, `x ln x`).
- **Certified learned envelopes**: constant-margin certification of guaranteed-convex ICNNs; **fixed a real ICNN convexity bug** (unconstrained output layer → non-convex despite the "convex by construction" claim).
- Docs notebook (`symbolic_envelopes.ipynb`) + `[sympy]` extra.

## Pattern catalog (`design/relaxation-patterns.md`, `symbolic/patterns.py`)
Recognizable structural patterns, each with an analytic correctness proof + sampling certification: bilinear (Al-Khayyal–Falk hull), linear-fractional, single-inflection, RLT sum-to-constant, posynomial/GP, signed signomial (DC), multivariate-signomial joint hull (GP log-lift), Fortet/Glover binary products, complementarity, log-curvature classifier — plus roadmap entries (AC-OPF QC, LMTD, Euclidean separation).

## Automated structured cuts + recognizer presolve
- **`constraint_cuts`**: symbolic constraint-chain elimination → sound coupling → univariate convex underestimator.
- **`cut_recognizer`**: reads the model graph and auto-fires sound detectors via `inject_all_patterns(model)` — square-difference network (gas/water), Fortet/Glover binaries, complementarity, GP log-lift. No-op on non-matching models.
- **Result (issue #15):** on the *unmodified* `build_gas_network_minlp()`, the recognizer auto-derives and injects 2 cuts and the bound goes **1.0 → 3.0026 (= global optimum), gap 66.7% → 0%, 809 → 5 nodes, 96 s timeout → 12 s certified**. (SCIP solves the same instance in 0.31 s — the remaining distance is generality of auto-detection, not the math.)

## Issue triage (`design/issue-pattern-map.md`)
Maps the open-issue backlog to patterns: addresses/relates to #15, #187 (Fortet/Glover), #231 (complementarity), #189/#181/#115 (GP/signomial/log-curvature), #188 (separation, roadmap).

## Testing
~150 tests across the symbolic suite (engine, domains, certified-learned, patterns, constraint-cuts, recognizer, gp_hull, signed_signomial, log_curvature), all passing; soundness is gated by sampling certification on every derived atom/cut. Branch rebased on current `main`.

**Honest scope notes:** the GP log-lift cut is sound + value-preserving but its bound benefit is problem-dependent (validation on `cvxnonsep_nsig30` pending); the AC-OPF/LMTD/Euclidean-separation patterns are proven + catalogued but not yet implemented; only the square-difference detector is demonstrated to move a bound end-to-end so far.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01VA3sTVixMtA2EFRJZSDFXf

### DIFF
--- a/design/issue-pattern-map.md
+++ b/design/issue-pattern-map.md
@@ -122,6 +122,22 @@ elsewhere), **new-pattern** (needs a pattern not yet implemented), **search/infr
 | log-curvature classifier (detection layer) | #115, #181 | analysis | **done** (`log_curvature.log_curvature`) |
 | (presolve) OBBT-on-aux, transcendental FBBT | #208, #219 | presolve | amplifier |
 
+## Auto-firing detectors wired into `cut_recognizer`
+
+`inject_all_patterns(model)` runs every detector in turn and returns
+`{pattern: count}`; each is sound and a no-op on non-matching models:
+
+| Detector | Issue | What it does |
+|---|---|---|
+| `recognize_and_inject` (square-diff network) | #15 | gas/water Weymouth chain → objective aux + `u >= K·h(w)` |
+| `inject_binary_products` (Fortet/Glover) | #187 | objective `coef·∏ b_i` (n>=3) → aux `z` + Fortet linearization (tighter than nested McCormick); value-preserving |
+| `inject_complementarity` | #231 | `x·y = 0` (or `<= 0`), `x,y>=0` → cut `x/x_ub + y/y_ub <= 1` |
+
+Bilinear (P3) and linear-fractional (P4) are intentionally **not** wired: the
+relaxation compiler already emits those envelopes, so a detector would be
+redundant. The signomial/GP patterns (P7/P11/P14) need a log-domain lifting pass
+before they can auto-fire; tracked as the next wiring step.
+
 ## Fix-and-comment workflow
 
 As each pattern lands: (1) implement + prove + certify in `symbolic/patterns.py`
--- a/design/issue-pattern-map.md
+++ b/design/issue-pattern-map.md
@@ -66,9 +66,22 @@ elsewhere), **new-pattern** (needs a pattern not yet implemented), **search/infr
 - Pattern: the **detection layer** that powers P7/P11 (which terms are GP-convex).
   Mirrors the existing convexity lattice but in log-space.
 
-### #116 — y-space (log) branching/bound-tightening for GP — **search/infra (GP)**
+### #116 — y-space (log) branching/bound-tightening for GP — **search/infra (GP)** — **constraint lift done (P16)**
 - The branching/OBBT counterpart of the GP patterns; pairs with P7/P11 but is a
   presolve/branching feature, not an envelope.
+- **P16 (`inject_gp_constraint_cuts`)** implements the *constraint-level* y-space
+  substitution: a posynomial `<=` constraint `P(x) <= b` (all `c_k>0`, exponents
+  `a_kj>=0`) is convex after `u_j=log x_j` → `Σ_k exp(s_k) <= b`. The pass adds the
+  convex link `u_j <= log x_j`, log-domain bounds, and OA tangent cuts
+  `Σ_k exp(s_k0)(1+s_k−s_k0) <= b`. Sound (no feasible `x` removed: at
+  `u_j=log x_j` the cut underestimates `P(x)`), value-preserving, and a no-op on
+  `>=`/affine/negative-exponent/signomial rows. **Honest:** does **not** fire on
+  `cvxnonsep_nsig30` (its single constraint is a *signomial* — a negative term —
+  so it needs the signed-signomial DC lift, P11, not the posynomial lift); and on
+  synthetic posynomial-constraint GPs discopt's native McCormick relaxation is
+  already root-tight (≈3 nodes), so the lift is correct but exposes no measurable
+  root gap there. Its value is on instances where the *separable* box relaxation of
+  a coupled posynomial constraint is genuinely loose.
 
 ### #231 — complementarity `x·y = 0`, `x,y >= 0` — **new-pattern**
 - Structure: complementarity / disjunctive `x·y = 0`.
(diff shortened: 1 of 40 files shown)


jkitchin (commented):
## Review — SymPy envelopes, pattern cuts, auto-firing recognizer

Reviewed all 36 files across three soundness-focused passes, and verified the load-bearing findings locally — including reproducing the false-optimal end-to-end and sampling the auto-engaged envelope.

**CI:** Rust ✅ · Python fast ✅ · AMP coverage ✅ — only **`Lint + typecheck` ❌** (mypy, 3 errors).

### 🔴 Blocker — `inject_complementarity` produces a *certified* false-optimal

`cut_recognizer.py:514–550`. The detector scans `sm.inequalities`, where `>=` is stored sign-flipped (`x*y >= 0` → `-x*y <= 0`), and `_as_single_bilinear` (line 501) discards the coefficient sign — so `x*y >= 0` is treated identically to the valid `x*y <= 0` and fires the cut `x/x_ub + y/y_ub <= 1`. But `x*y >= 0` makes the **entire** non-negative box feasible, so the cut illegally removes feasible points.

Reproduced end-to-end:
```python
# minimize -(x+y) s.t. x*y >= 0, x∈[0,6], y∈[0,4];  true optimum = -10 at (6,4)
baseline (no inject):     status=optimal  obj=-10.0   # correct
inject_all_patterns():    {'complementarity': 1}      # 1 cut injected
with inject:              status=optimal  obj=-6.0     # CERTIFIED WRONG (cut excludes (6,4))
```

It's opt-in (`inject_all_patterns`, not in the default solve path), but it's a wrong *certified* answer in the recognizer's headline API. **Fix:** in `inject_complementarity`, require a **positive** coefficient on a `<=0`-form source and reject `>=`-derived (`-x*y`) forms. The existing `test_inject_complementarity_detects_and_cut_is_valid` only exercises the valid `==0` case and wouldn't catch this — add an `x*y >= 0` regression test.

### 🔴 CI blocker — mypy (3 errors)

- **`cut_recognizer.py:56` — real latent crash.** `node.base.name` on an `IndexExpression` whose base isn't a `Variable` (e.g. `(x+y)[0]`) raises `AttributeError` inside `model_to_sympy`. Guard with `isinstance(node.base, Variable)` and skip otherwise.
- **`cut_recognizer.py:70`** (false positive, reachable only after a non-None `_var_key`) and **`relaxation_compiler.py:362`** (`weymouth_relax = None` in the `except ImportError` block, guarded at runtime) — silence with annotations.

### 🟠 Major

- **`icnn.py:126` `verify_convexity` is blind.** It checks the Hessian, which is identically 0 within a ReLU net's linear regions, so it returns `True` for *any* ReLU net — including the non-convex one this PR fixes (confirmed it false-passes the OLD network). A convexity gate that can't catch the bug it guards, and `test_learned_relaxations.py` asserts on it (false confidence). Fix: use a Jensen/midpoint inequality on random chords (as `verification.py` already does correctly).
- **`verification.py:126` `sound` excludes curvature.** `sound = (lower_viol <= tol) and (upper_viol <= tol)` — a non-convex underestimator that merely *contains* f passes as `sound=True` (confirmed: a wiggly "cv" returns `sound=True` while reporting `convexity_viol=3.38`). Not a live hole (shipped envelopes are genuinely convex) but the certifier won't catch a future curvature regression. Fold `max_convexity_violation <= tol` / `max_concavity_violation <= tol` into `sound`.
- **`references.bib:2121` duplicate `Amos2017` key** (already exists at line 341) — emits a duplicate-key warning, violating the repo's zero-warnings notebook build. Drop the duplicate.

### ✅ Verified sound (the good news)

- **The only auto-engaged change — the Weymouth `f·|f|` envelope (`domains/gas.py` + `relaxation_compiler.py`) — is sound** to ~1e-14 over 30k random boxes including ones straddling the inflection at 0 (cv convex, cc concave, `cv ≤ f ≤ cc`). The `_try_extract_signed_abs_product` detector does **not** misfire (`f*|g|`, `|f|*|f|`, `f*f`, `(f+1)*|f|` all correctly skip), and `v·|v|` is a valid envelope for any scalar var regardless. **The default relaxation path is safe.**
- **The ICNN fix is correct** — `softplus` on the output weights; the OLD unconstrained output layer shipped a genuinely non-convex "convex" NN (Jensen midpoint violation ~352). Latently unsound only on the opt-in learned-relaxation path (`use_learned_relaxations=False` default).
- All other catalog cuts sampled **valid**: Fortet/Glover (exact at binary vertices, sign-safe for ±objective coeffs), bilinear Al-Khayyal–Falk hull (0 violations / 20k boxes), GP log-lift, constraint-chain univariate underestimator, signed-signomial DC. The **gas-network headline reproduced** — recognizer injects 2 cuts, bound → 3.0026 (= global optimum), gap → ~0%, cuts don't exclude the optimum; genuine no-op on unrelated models.
- **SymPy is genuinely off the hot path** — `import discopt` doesn't load sympy, `[sympy]` is an optional extra, and the Weymouth dispatch degrades gracefully (generic bilinear) when sympy is absent.

### 🟡 Minor
- Certified-learned margin is sampling-only (honestly disclosed as "a bug-catching gate, not a proof"); the `seed+1` re-cert is a generalization smoke-test, not a guarantee.
- Domain validity ranges (e.g. `cos_angle_relax` valid only ⊆ (−π/2, π/2)) are documented but not enforced in code — a caller passing an out-of-range box gets a silently invalid envelope.
- "never imported on the hot path" wording is slightly overstated: importing a domain pack eager-loads sympy as a side effect (no hot-path *execution*, just import cost).

---

**Bottom line:** the envelope engine and the auto-engaged Weymouth path are well-built and sound, and the ICNN bug fix is correct. Before merge: **(1)** the complementarity-cut false-optimal (sign/direction check) — this is the blocker; **(2)** the mypy errors (one is a real crash); **(3)** the `verify_convexity` blind spot and the `verification.sound` curvature gap (the soundness *gates* are weaker than they look); **(4)** the duplicate bib key.

Issue #268, opened 2026-06-21
Performance gap: short-tier MINLPLib instances time out with no solution (SCIP solves all <5s)


jkitchin:
## Summary

A standing **performance / capability gap** to dig into later: on the MINLPLib
"short" tier (problems SCIP solves correctly in <5 s), there is a recurring set of
instances where **discopt finds no solution at all** within a 5 s limit (+5 s
hard-kill grace) -- it either hits the time limit with no incumbent or is
force-killed. **SCIP solves every one of these in <5 s** (most <3 s).

These are *sound* (no wrong answers, no crashes) -- purely "too slow / doesn't
converge in budget." Surfaced across several randomized short-tier benchmark
shuffles (27 distinct feasible instances so far). Logging as one tracking
issue rather than per-instance noise.

## Sub-pattern A -- combinatorial MINLP (discrete vars > 0), 19 instances

discopt's branch-and-bound over the discrete structure doesn't reach a solution
in time. Mostly the `rsyn*` / `syn*` synthesis, `clay*`, `flay*`, `ex12*`,
`batchs*`, `m7`, `netmod`, `slay`, `supplychain` families.

| instance | discopt | SCIP | vars | cons | discrete | nl_con |
|---|---|---|---:|---:|---:|---:|
| `rsyn0805m02m` | time_limit | 4.02s | 360 | 769 | 148 | 6 |
| `slay07h` | hard_timeout | 3.57s | 476 | 609 | 84 | 0 |
| `rsyn0840m02m` | hard_timeout | 3.46s | 720 | 1480 | 288 | 56 |
| `gastrans582_cool12_95` | hard_timeout | 3.14s | 2186 | 3732 | 232 | 909 |
| `flay04h` | time_limit | 2.89s | 234 | 282 | 24 | 4 |
| `rsyn0810m03hfsg` | time_limit | 2.68s | 1185 | 1935 | 234 | 18 |
| `netmod_kar2` | hard_timeout | 2.56s | 456 | 666 | 136 | 0 |
| `watertreatnd_conc` | hard_timeout | 2.43s | 360 | 319 | 5 | 29 |
| `p_ball_10b_5p_3d_m` | time_limit | 2.38s | 95 | 129 | 50 | 50 |
| `syn30m04m` | hard_timeout | 1.97s | 640 | 1568 | 240 | 80 |
| `batchs101006m` | time_limit | 1.79s | 278 | 1019 | 129 | 1 |
[...]

PR #273, opened 2026-06-21
perf(#268): sparse LP factorize + adaptive refactor, fractional-diving fallback, capped heuristic NLPs

jkitchin:
## Summary

Addresses the discopt-side levers of #268 ("short-tier MINLPLib instances time out with no solution while SCIP solves them <5s"). Three changes, two areas. The deepest LP bottleneck (feral's O(bump²) FT update) is tracked feral-side in **feral#87**; this PR is the discopt-side mitigation plus the combinatorial primal-heuristic fixes.

### Lever 1 — LP solver: sparse factorize + adaptive refactor (continuous-nonconvex tier)

Profiling the pure-Rust revised simplex showed **89%** of a row-heavy lifted-McCormick LP solve was inside feral's Forrest–Tomlin `SparseLu::update` — on non-localized (dense) spikes the FT update is O(bump²), and a single update can cost as much as a full refactorization (autocorr_bern20-05 root LP: **3.6s**; HiGHS 0.015s). The factor itself stays sparse (1.8× fill), so it's not fill-in.

- **Sparse basis factorization** (`LinearSolver::factorize_sparse`): build the basis from sparse columns (slot → (row,value)) straight from the simplex's CSC `cols`, via feral's `SparseColMatrix::from_sparse_columns`. Drops per-refactor cost from O(m²) (dense column build + dense nnz-scan + dense→sparse convert) to O(nnz). Bit-identical matrix → same pivots → same optimum.
- **Adaptive refactorization**: refactor once the accumulated FT update work (`ft_update_work`, feral's `eta_ops`) exceeds the factor's nnz — wide-bump bases refactor often (now cheap) instead of compounding the O(bump²) blowup; narrow-bump bases keep all 48 cheap updates.

autocorr_bern20-05 root LP **3.6s → 1.24s (2.9×)**; end-to-end it now returns a feasible incumbent (was: time_limit, stuck at the root with no solution).

### Lever 3 — fractional-diving fallback (combinatorial tier)

The feasibility pump rounds every integer at once → a constraint-infeasible assignment on facility-location / set-partitioning MINLPs (clay/flay/syn/m7) whose relaxation parks every binary near 0.5; the pump returns nothing and the search exhausts with no incumbent. Wire `fractional_diving` (fix one fractional integer at a time, re-solving the NLP between fixings) into **both** the convex NLP-BB and the spatial root-heuristic ladders, gated on no-incumbent-yet + remaining time. Sound: diving re-verifies integer + constraint feasibility and `inject_incumbent` enforces strict improvement.

| instance | before | after |
|---|---|---|
| clay0204m | time_limit, **no solution** | **feasible** (obj 9350) |
| m7 | time_limit, **no solution** | **feasible** (obj 136.9) |

### Lever 2 — cap heuristic sub-NLP iterations (`_HEURISTIC_NLP_MAX_ITER = 300`)

The pump / local-search / diving projections only need an approximately feasible point. Uncapped, a single projection can grind to the backend's full limit (one ex1263 solve hit ~1225 IPM iterations) and burn the budget the B&B needs. Capping lets the search branch further (ex1263 31 → 63 nodes). Sound: an unconverged capped point just fails the feasibility check (no incumbent injected); callers can override via `ipopt_options`/`nlp_options`.

## Still open (follow-up, not in this PR)

`ex1263` / `ex1266` still find no feasible integer point even with diving — their relaxation rounds infeasible and per-variable repair can't recover; they need a stronger combinatorial primal heuristic or a better dive seed. Noted for a follow-up.

## Testing
- New `test_issue_268_primal_heuristics.py` (4 tests: iteration cap applied/overridable/bounded, combinatorial end-to-end).
- **333** discopt-core Rust tests, **742** Python LP/spatial tests, **729** correctness/MINLP/AMP tests, **64** heuristic tests — all pass, no regressions. LP changes are bit-identical (answers unchanged).
- `ruff` + `mypy` clean on changed files.

### DIFF
--- a/crates/discopt-core/src/lp/simplex/dual.rs
+++ b/crates/discopt-core/src/lp/simplex/dual.rs
@@ -117,8 +117,16 @@ impl<'a> PreparedDual<'a> {
 
         let sp = SparseCols::from_dense(a, m, n);
         let mut lu = FeralLU::new();
-        let cols: Vec<Vec<f64>> = basis.iter().map(|&j| col(a, m, n, j)).collect();
-        if lu.factorize(m, &cols).is_err() {
+        // Sparse basis columns straight from the CSC view — O(nnz) factorize, no
+        // dense m×m basis (discopt#268 / feral#87). Bit-identical to `col()`.
+        let cols: Vec<Vec<(usize, f64)>> = basis
+            .iter()
+            .map(|&j| {
+                let (rows, vals) = sp.col(j);
+                rows.iter().zip(vals).map(|(&r, &v)| (r, v)).collect()
+            })
+            .collect();
+        if lu.factorize_sparse(m, &cols).is_err() {
             return None; // singular warm basis → fall back to cold
         }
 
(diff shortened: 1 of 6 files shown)


jkitchin:
Holding as draft until the feral-side of #268 (feral#87, the FT-update O(bump²) work) is reviewed — the discopt-side mitigation here pairs with that. Mark ready + merge once the feral side is settled.

Issue #283, opened 2026-06-21
Suboptimal-incumbent gap: nvs_* small MINLP (use absolute error for near-zero optima)


jkitchin:
## Summary

Tracking the **`nvs_*` small-MINLP suboptimal-incumbent gap**: small instances where discopt returns a sound feasible incumbent rather than the optimum. Note several `nvs` problems have an optimum near zero, so report **absolute** error for those (relative % is misleading).

| instance | sense | discopt status | obj | known opt | gap | SCIP |
|---|---|---|---|---|---|---|
| `nvs02` | min | feasible | 6.24128 | 5.96418 | 4.6% | 0.0s |
| `nvs05` | min | feasible | 13.9087 | 5.47093 | 154.2% | 0.1s |
| `nvs14` | min | feasible | -37791.6 | -40358.2 | 6.4% | 0.0s |
| `nvs17` | min | feasible | -1091.4 | -1100.4 | 0.8% | 0.5s |
| `nvs19` | min | feasible | -1092.6 | -1098.4 | 0.5% | 0.8s |
| `nvs22` | min | feasible | 323.969 | 6.05822 | 5247.6% | 0.1s |
| `nvs24` | min | feasible | -1022.4 | -1033.2 | 1.0% | 3.0s |

Related: #267 (suboptimal-incumbent gap). (`nvs22` separately exercised the #277 false-certificate fix, now resolved — here it is merely suboptimal, which is correct behavior.)

*Environment: discopt `0.4.1.dev0` (post #270/#273/#275/#276/#277), JAX `0.10.0`, `JAX_ENABLE_X64=1`, 5s time limit. Surfaced via randomized MINLPLib short-tier benchmark shuffles. All instances are SOUND suboptimal incumbents (feasible, not certified, worse-than-optimum in the correct direction) — not correctness bugs. Gap shown as relative %, or absolute for near-zero optima.*

Issue #280, opened 2026-06-21
Global-search gap: graphpart_* graph-partition QP family (sound suboptimal incumbents)


jkitchin:
## Summary

Tracking the **graph-partitioning (`graphpart_*`) global-search gap**: discopt returns sound feasible incumbents but does not reach the global optimum within budget on this combinatorial nonconvex-QP family. SCIP solves all of these in <5s.

On the current build the spread is wide and improving — some instances now solve to **optimal** (`graphpart_2pm-0066-0066`, `graphpart_2pm-0055-0055`), others remain far off. This issue tracks the family as a unit so progress (e.g. from the #276 LNS layer) is visible.

| instance | sense | discopt status | obj | known opt | gap | SCIP |
|---|---|---|---|---|---|---|
| `graphpart_2g-0044-1601` | min | feasible | -919798 | -954077 | 3.6% | 0.0s |
| `graphpart_2g-0055-0062` | min | feasible | -1.35036e+06 | -1.48435e+06 | 9.0% | 0.1s |
| `graphpart_2g-0066-0066` | min | feasible | -2.5411e+06 | -2.86556e+06 | 11.3% | 0.2s |
| `graphpart_2g-0077-0077` | min | feasible | -3.11209e+06 | -3.28244e+06 | 5.2% | 1.4s |
| `graphpart_2pm-0044-0044` | min | feasible | -4 | -13 | 69.2% | 0.0s |
| `graphpart_2pm-0077-0777` | min | feasible | -36 | -40 | 10.0% | 0.8s |
| `graphpart_2pm-0088-0888` | min | feasible | -54 | -55 | 1.8% | 0.5s |
| `graphpart_3g-0244-0244` | min | feasible | -2.54031e+06 | -2.72538e+06 | 6.8% | 0.3s |
| `graphpart_3g-0333-0333` | min | feasible | -1.71675e+06 | -1.88239e+06 | 8.8% | 0.6s |
| `graphpart_3g-0344-0344` | min | feasible | -2.04477e+06 | -5.30152e+06 | 61.4% | 1.6s |
| `graphpart_3pm-0234-0234` | min | feasible | -19 | -20 | 5.0% | 0.2s |
| `graphpart_3pm-0244-0244` | min | feasible | -27 | -28 | 3.6% | 0.7s |
| `graphpart_3pm-0334-0334` | min | feasible | -34 | -36 | 5.6% | 1.3s |
[...]

PR #285, opened 2026-06-22
feat(global-opt): integer-bilinear MILP reformulation + native-AD node NLP

jkitchin:
## Summary

Two auto-triggered additions for solving harder MINLPs, each gated to no-op outside its target structure.

### 1. Integer-bilinear exact reformulation — closes the ex126x trim-loss class

When a bilinear term `x_i·x_j` has an integer (declared or *implied*) factor, the integer factor is binary-expanded and the resulting `binary·var` products are big-M linearized → an **equivalent pure MILP** whose relaxation is exact.

discopt's single McCormick envelope over a wide integer box is loose: on ex1263 the relaxation's integer optimum is 19.1 while the true optimum is 19.6, so discopt previously found **no feasible solution at all** within budget. The exact linearization closes that gap. It is adopted only when it yields a pure MILP, then routed to the monolithic Rust simplex MILP engine with discopt's (redundant, and pathologically slow on the big-M structure) FBBT root presolve skipped.

| instance | before | after | ref (SCIP) |
|---|---|---|---|
| ex1263 | `obj=None` | **19.6** optimal, 1.1s | 19.6 |
| ex1264 | `obj=None` | **8.6** optimal, 1.5s | 8.6 |
| ex1265 | `obj=None` | **10.3** optimal, 1.4s | 10.3 |
| ex1266 | `obj=None` | **16.3** optimal, 1.5s | 16.3 |

- `implied_integer.py` — provably-sound implied-integer detection (`x = integer combination of integer vars`, fixpoint-iterated).
- `integer_product_reform.py` — the exact reformulation pass.

### 2. Native-AD node NLP path — decouples node solves from JAX

`from_nl` models hand POUNCE the original `.nl` so its own reverse-mode AD evaluates the node NLP, bypassing the JAX callback bridge. Variable order verified identical; value-equivalent to the JAX path; falls back automatically when a native problem can't be built/validated. Cuts JIT-compile latency on cold/shallow solves. Default-on (`DISCOPT_NLP_NATIVE` / `options["nlp_native"]`). `pounce-solver` pinned to `>=0.6`.

## Correctness

- 778 correctness/regression + 50 pounce + 15 native-equivalence + 13 integer-bilinear tests pass; 0 failures.
- Reformulation value-preserving vs SCIP across the instance basket; the detector marks a variable integer only when **provably** forced (rejects `b≤x≤b+4`, `2x=k`, `x=y+continuous`).
- Mixed-nonlinearity models (e.g. `gear`) are left unchanged — the reformulation is discarded when other nonlinear terms remain, so the change is a strict improvement.

## Scope / limitations

The reformulation accelerates the **pure** integer-bilinear class. Instances with bilinear terms *plus* other nonlinearity (transcendentals: gear, ex1252) are handled correctly but not accelerated. Native-AD is performance-neutral-to-modest (its main win is removing JIT latency); disable with `nlp_native=False`.

Relates to #268, #280–#283.

https://claude.ai/code/session_01Lq9NRuyCAWVxWoFv6AFuAt

### DIFF
--- a/docs/dev/integer-bilinear-minlp-plan.md
+++ b/docs/dev/integer-bilinear-minlp-plan.md
@@ -1,0 +1,234 @@
+# Plan: SCIP-grade support for integer-bilinear MINLPs in discopt
+
+**Goal.** Solve the integer-bilinear MINLP class (the `ex126x` trim-loss family and
+similar problems where bilinear terms have integer or implied-integer factors) to
+**proven optimality**, matching the reference solver (SCIP: ex1263 → 19.6, 0.11 s).
+
+**Proven diagnosis (this is the foundation, established experimentally).**
+discopt relaxes `x_i·x_j` with one continuous McCormick envelope over a wide box;
+when the factors are integers this envelope is too loose, so the relaxation's
+*integer optimum* sits strictly below the true optimum (ex1263: 19.1 vs 19.6 —
+confirmed by solving discopt's own relaxation with SCIP). Exact handling of the
+integer products closes it: a binary-expansion linearization through discopt's
+own pipeline already moves the bound 19.1 → 19.6.
+
+**Correctness invariant (non-negotiable, applies to every step).**
+Every reformulation must be *value-preserving* (identical optimum and feasible
+projection); `incorrect_count == 0` on the full suite must hold after each step;
+no variable is ever marked integer unless **provably** implied-integer (marking a
+free variable integer can cut off the optimum — the cardinal violation).
+
+---
+
+## Architecture decision: reuse the MILP path via exact reformulation
+
+SCIP does not binary-expand; it keeps the products and tightens via cuts. But
+replicating SCIP's integrated nonlinear+cut+presolve loop is the largest possible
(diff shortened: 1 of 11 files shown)


jkitchin:
Update: extended the reformulation to **integer squares** (`x**2` / `x*x`) using the same binary-expansion + AND-product technique, and added a **blowup guard**.

- **Closes the small bilinear+square nvs instances** (#283): `nvs14` goes from a 6.4% suboptimal incumbent to the proven optimum (−40358.2) in **0.7s**.
- **Blowup guard:** `distribute_products` explodes on some instances (`nvs17`: 7 vars → 2751 after reform; `nvs24` → 6370). The guard (reformed vars ≤ `max(1000, 6·n_orig)`) discards such reforms and keeps the original model, so those fall back to the normal path with **no regression** (still sound suboptimal incumbents).

**Scope note for #283:** this *partially* addresses it (the small bilinear+square subclass). `nvs17/19/24` (combinatorial blowup) and `nvs05` (transcendentals) are **not** closed and fall back cleanly — so #283 should remain open after merge, not be auto-closed.

PR #318, opened 2026-06-24
Add optional Gurobi backend

bernalde:
Summary:
- Adds an optional `gurobipy` backend for LP, MILP, QP, MIQP, QCP, QCQP, MIQCP, and MIQCQP model classes.
- Wires direct `model.solve(solver="gurobi")` dispatch for matrix-representable classes.
- Adds `milp_solver="gurobi"` support for OA, LOA, and AMP paths.
- Keeps direct Gurobi scoped to matrix-representable classes; general MINLP support remains in discopt global algorithms, with Gurobi used as the AMP MILP subsolver.
- Adds skip-clean tests for environments without `gurobipy` or a working Gurobi license.

Fork staging PRs:
- bernalde/discopt#103 - Gurobi LP/MILP backend
- bernalde/discopt#106 - Gurobi QP/MIQP backend
- bernalde/discopt#108 - Gurobi QCP/MIQCP/MIQCQP backend
- bernalde/discopt#109 - Gurobi MINLP subsolver option through discopt algorithms

Local validation before opening this draft:
- `SKIP=cargo-fmt .venv/bin/pre-commit run --all-files` -> passed.
- `PYTHONPATH=python .venv/bin/python -m pytest python/tests/test_gurobi_backend.py python/tests/test_lp_backend_select.py python/tests/test_pr_correctness.py -q` -> 59 passed.
- AMP+Gurobi probe across all 14 `NLP_MI_INSTANCES` with `time_limit=120`, `gap_tolerance=1e-6`, `rel_gap=1e-6` -> 14 passed, 0 failed.

Preparation notes:
- The fork `gurobi` branch was merged with current `upstream/main` before opening this PR so the diff is focused on the Gurobi integration rather than stale branch drift.
- Licensed Gurobi tests require both `gurobipy` and a working Gurobi license; otherwise they skip cleanly.

Follow-up:
- This is a draft PR for upstream review. Before marking ready, decide whether upstream wants the direct Gurobi backend API as written, and whether any documentation should be expanded beyond the current README/design notes.


jkitchin:
Let me know when this is ready for review to merge.

bernalde:
I would appreciate a deep review before merging it! The things I'm building rely on this being stable and useful!

jkitchin (dismissed):
## Review: optional Gurobi backend

I did a correctness-focused, adversarial review and ran everything live against
Gurobi 13.0.0 (licensed). Summary:

- ✅ The **direct matrix backend** (`solver="gurobi"` for LP/MILP/QP/MIQP/QCP/QCQP/MIQCP/MIQCQP)
  is correct and well-built. Every class matched discopt's default solver and analytic optima.
- 🔴 The **`milp_solver="gurobi"` decomposition integration (OA/LOA/AMP)** — the headline
  feature for the iterative paths — is **not reliable as written**. I reproduced two failures
  on trivial convex MINLPs (details + reproducers below).

Requesting changes for the decomposition integration. The direct backend is solid on its own.

---

### What I verified works

All matched the default solver and closed-form answers:

| Area | Result |
|---|---|
| LP min/max, MILP max, QP min, QP max (concave), MIQP, MIQCP | ✅ match |
| QCP (disk), QCP + linear-equality mixed | ✅ match |
| Quadratic **equality** constraint | ✅ correctly requires `NonConvex:2`, then matches |
| Nonconvex QP objective | ✅ raises a clear error without `NonConvex:2`; correct with it |
| Infeasible / unbounded | ✅ correct status |
| LP duals & reduced costs | ✅ sign convention matches the HiGHS wrapper exactly |
| Maximize objective/bound/`obj_const` sign mapping | ✅ correct |
| Convexity gating (`<=` ⇒ Q⪰0, `>=` ⇒ Q⪯0, `==` ⇒ always nonconvex) | ✅ correct, and the maximize-negation happens before the gate so it stays consistent |
| Finite-difference quadratic extraction (`0.5·xᵀQx + cᵀx + d`) | ✅ verified algebraically |

Nice work on the gating, the maximize handling, and the QCP feasibility guards.

---

### 🔴 Finding 1 — OA + `milp_solver="gurobi"` does not converge on degenerate MINLPs

On a convex MINLP with multiple integer optima, OA + HiGHS converges in **1 master solve**,
but OA + Gurobi **cycles** through equal-objective integer points
(`x=[3,0] → [0,3] → [2,1] → [2,1] …`), adding 2 cuts/iteration and never terminating.
It ran ~1000 master solves. (In a size-limited-license environment it eventually aborts with
`GurobiError 10010 "Model too large for size-limited license"`; with a full license it would
silently burn to `max_iterations` and return a non-certified result.)

The Gurobi master returns *valid* optimal solutions each iteration — its tie-breaking among
equal-objective integer optima just differs from HiGHS and defeats OA's convergence logic.

Reproducer:

```python
import discopt.modeling as dm

def minlp():
    m = dm.Model("minlp")
    x = m.integer("x", lb=0, ub=5)
    y = m.continuous("y", lb=0, ub=5)
    m.subject_to(x + y >= 3)
    m.minimize((x - 2) ** 2 + (y - 1.5) ** 2)   # optimum 0 at x=2, y=1.5; many integer ties
    return m

print(minlp().solve(gdp_method="oa", milp_solver="highs",  skip_convex_check=True).status)
# -> optimal in 1 master solve

print(minlp().solve(gdp_method="oa", milp_solver="gurobi", skip_convex_check=True).status)
# -> cycles ~1000 master solves; never converges (raises under a size-limited license)
```

---

### 🔴 Finding 2 — AMP + `milp_solver="gurobi"` fails to certify optimality

On a non-degenerate convex MINLP, AMP + Gurobi finds the **correct** objective but returns
status `"feasible"` (not `"optimal"`), looping with:

```
AMP iter N: invalid bound ordering LB=0.940002, UB=0.94; skipping gap certification
AMP: final bound ordering invalid (LB=0.940002, UB=0.94); omitting bound and gap
```

Root cause: discopt's OA/AMP assume the MILP master returns a **rigorous** lower bound, but the
Gurobi wrapper returns `bound`/`gap` at Gurobi's *default tolerances*. `solve_milp` sets a
relative `MIPGap` but no `MIPGapAbs`, and Gurobi's `ObjBound` can land ~1e-6 *above* the true
integer optimum. That tiny `LB > UB` inversion breaks gap certification. HiGHS happens to return
tight enough bounds on these small problems to mask it.

Reproducer:

```python
import discopt.modeling as dm

def make():
    m = dm.Model("p")
    x = m.integer("x", lb=0, ub=6)
    y = m.continuous("y", lb=0, ub=6)
    m.subject_to(x + y >= 4)
    m.minimize((x - 3.2) ** 2 + 2 * (y - 2.1) ** 2 + 0.3 * x)
    return m

r = make().solve(solver="amp", milp_solver="gurobi", rel_gap=1e-6, time_limit=30)
print(r.status, r.objective)
# -> feasible 0.94   (correct objective, but never certified optimal; LB=0.940002 > UB=0.94)
```

Both findings sit in a path that the PR **does not test end-to-end**: the OA/LOA forwarding
tests monkeypatch `_solve_qcp_gurobi` / `solve_milp` with fakes, so they assert *argument
forwarding* only, never real numeric convergence. Only the AMP path was probed with real Gurobi
(14 `NLP_MI` instances) — and Finding 2 shows even that path can silently drop to `"feasible"`.

---

### Minor notes

- **No fallback when `gurobipy` is missing.** `get_milp_solver(backend="gurobi")` returns a
  wrapper that raises only when *called* (the `from discopt.solvers.gurobi import solve_milp`
  import succeeds without `gurobipy`). Intentional per the docstring, but worth confirming this
  is the desired behavior for the AMP-subsolver path, where a silent HiGHS fallback might be
  expected.
- **Optimal MILP reports `bound = objective` but `gap = result.gap` (possibly > 0).** A strict
  optimum should carry gap 0; this is cosmetically inconsistent and related to Finding 2
  (`_solve_milp_gurobi`).
- **`options` override `threads`.** In `_set_common_params`, user `options` are applied *after*
  the `threads` param, so `options={'Threads': N}` silently overrides the `threads=` argument.

---

### Suggested path forward

The direct backend is mergeable on its own. Before the OA/LOA/AMP `milp_solver="gurobi"`
claim ships, I'd want:

1. Real end-to-end numeric tests (not monkeypatched fakes) for OA/LOA + Gurobi, including a
   degenerate MINLP.
2. Reconcile master-bound semantics — e.g. tighten `MIPGap`/`MIPGapAbs`/`IntFeasTol` for master
   solves, **and/or** have the decomposition layer treat the returned bound as inexact (clamp
   `LB ← min(LB, UB)` and don't break certification on tolerance-level inversions).

Tested against the PR branch directly; environment was a size-limited Gurobi license, which is
how Finding 1 surfaces as a hard error rather than a max-iterations stall.

bernalde:
Commits pushed:
- 12f0882042886a971abfe3e3e16042417dd209f1 Fix Gurobi decomposition certification

Main changes:
- Normalize optimal Gurobi MILP/QP/QCP results to report bound=objective and gap=0.
- Keep explicit time_limit/threads arguments authoritative over raw Gurobi options.
- Treat AMP master-bound inversions within numerical tolerance as zero gap and clamp the reported bound to the incumbent objective.
- Added regression coverage for the reported OA and AMP Gurobi reproducers; those tests skip without a local Gurobi install/license.

Tests run:
- PYTHONPATH=python pytest python/tests/test_gurobi_backend.py python/tests/test_mip_nlp.py python/tests/test_oa.py -q: 44 passed, 10 skipped, 12 deselected.
- PYTHONPATH=python pytest -m "" selected AMP certification tests -q: 3 passed.
- python -m ruff check python/discopt/solvers/gurobi.py python/discopt/solvers/amp.py python/tests/test_gurobi_backend.py: passed.
- python -m ruff format --check python/discopt/solvers/gurobi.py python/discopt/solvers/amp.py python/tests/test_gurobi_backend.py: passed.
- pre-commit run --files python/discopt/solvers/gurobi.py python/discopt/solvers/amp.py python/tests/test_gurobi_backend.py: passed.

- Did not add silent fallback from explicit milp_solver="gurobi" to HiGHS when gurobipy/license is missing; explicit backend selection still reports the missing optional dependency/license when called, matching the PR docs.

Remaining risks/follow-up:
- Local environment does not have gurobipy, so the new end-to-end Gurobi reproducers were added but skipped locally; a licensed Gurobi environment should exercise them.
- PR checks are currently pending.

jkitchin:
Thanks for the quick turnaround. I ran your `12f0882` against a licensed Gurobi 13.0.0 — including the two new end-to-end tests you added but couldn't execute locally.

**Results:**

- ✅ **Finding 2 (AMP certification) is fixed.** `test_amp_gurobi_certifies_convex_minlp_if_available` **passes** — AMP + Gurobi now returns `optimal` with `gap_certified=True` on the reproducer. The tolerance-clamp on master-bound inversions does the job.
- ✅ Minor notes 2 & 3 (optimal `bound=objective`/`gap=0`, explicit `time_limit`/`threads` authoritative) look good.
- ❌ **Finding 1 (OA degeneracy) is not fixed.** Your own new test catches it:

```
test_oa_gurobi_converges_on_degenerate_convex_minlp_if_available FAILED
>       assert result.status == "optimal"
E       AssertionError: assert 'feasible' == 'optimal'
```

The `max_nodes=20` cap stops the runaway loop before it hits a size-limited license, but OA still never converges on the degenerate MINLP — it returns `"feasible"`, not `"optimal"`. Master-call tracing shows it still cycles through equal-objective integer points (`[3,0] → [0,3] → [2,1] → [2,1] …`) until the iteration cap. The bound-normalization helped AMP, but the OA root cause — the master's tie-breaking among equal-objective integer optima defeating OA's convergence detection — is untouched.

So: the new AMP test is genuine green; the new OA test is **red on a licensed box**. Worth wiring a licensed-Gurobi job (or at least running these two locally on a licensed machine) into the gate so this doesn't slip through as a silent skip again.

Happy to share the master-call trace if useful. The direct matrix backend remains solid.

jkitchin:
Finding 1 is still open after `12f0882`, but I traced it to the root cause — and the good news is **it's not really a Gurobi bug at all, and the fix is a one-liner.**

Enabling OA's iteration logging on the degenerate reproducer shows the problem immediately:

```
OA iter 0: LB=-0.000000 UB=0.000000 gap=100.0000% cuts=4
OA iter 1: LB=-0.000000 UB=0.000000 gap=100.0000% cuts=6
...
FINAL feasible  gap=1.0
```

LB and UB are both ≈ 0, yet the reported gap is **100%**. The culprit is `_compute_gap` in `oa.py`:

```python
def _compute_gap(lb, ub):
    if ub >= 1e19 or lb <= -1e19:
        return 1.0
    abs_gap = max(0.0, ub - lb)
    if abs_gap <= 1e-9:                       # floor too tight
        return 0.0
    denom = max(abs(ub), abs(lb), 1e-10)      # denom collapses to ~0 near the optimum
    return abs_gap / denom
```

On this problem the Gurobi master returns the epigraph bound `eta ≈ -1.99e-8` (a hair below zero, within Gurobi's default tolerances), while the NLP gives `UB ≈ +9e-18`. So:

- `abs_gap = UB - LB ≈ 1.99e-8` → just **above** the `1e-9` absolute floor, so it isn't snapped to 0;
- `denom = max(|UB|, |LB|, 1e-10) ≈ 1.99e-8`;
- `gap = 1.99e-8 / 1.99e-8 = 1.0` → **100%**, forever.

HiGHS only "passed" by luck: its master bound came back slightly **positive** (same sign as UB), so `abs_gap ≈ 0` hit the `1e-9` floor and returned 0. The moment a backend returns a bound that straddles zero by even ~1e-8 — which any MIP solver may do at its tolerances — OA never certifies and degrades to `feasible`. So this is a latent OA gap-computation bug that Gurobi merely exposes; it's independent of the bound-clamp you added for the AMP path.

**Fix (one line):** floor the denominator at `1.0`, matching discopt's B&B hybrid `gap()` (CLAUDE.md notes the tree "floors its denominator at 1.0"):

```python
    denom = max(abs(ub), abs(lb), 1.0)   # was 1e-10
```

I verified this against licensed Gurobi 13.0.0 — your `test_oa_gurobi_converges_on_degenerate_convex_minlp_if_available` goes green (`status='optimal'`, `gap≈3.5e-9`), and HiGHS is unaffected. (Raising the absolute floor from `1e-9` to `~1e-6` also works and yields `gap=0.0`; flooring the denominator is the cleaner choice since it stays relative for large objectives and only becomes absolute below magnitude 1, exactly like the rest of discopt.)

Net: AMP path ✅ fixed; OA path needs the `_compute_gap` denominator floor. Belongs in `oa.py`, not the Gurobi wrapper.

jkitchin:
good news is it's not really a Gurobi bug at all...☠️😂Thanks Claude!

bernalde:
Commits pushed:
- 9139a2553ecd9df8c910c30208c3b926b9b93bcb Fix OA near-zero gap scaling

Main changes:
- Updated OA gap computation to floor the relative-gap denominator at 1.0, so near-zero objective and bound pairs are judged on absolute scale instead of reporting a spurious 100 percent gap.
- Added a regression test for the reviewer-reported near-zero OA bound case.

Tests run and results:
- PYTHONPATH=python pytest python/tests/test_oa.py::test_compute_gap_uses_absolute_scale_near_zero -q: failed before the fix, passed after the fix.
- PYTHONPATH=python pytest python/tests/test_gurobi_backend.py python/tests/test_mip_nlp.py python/tests/test_oa.py -q: 45 passed, 10 skipped, 12 deselected.
- python -m ruff check python/discopt/solvers/oa.py python/tests/test_oa.py: passed.
- python -m ruff format --check python/discopt/solvers/oa.py python/tests/test_oa.py: passed.
- pre-commit run --files python/discopt/solvers/oa.py python/tests/test_oa.py: passed.

- Did not add or change a licensed-Gurobi CI job; that is outside this PR fix and would require workflow and secret policy decisions.

Remaining risks or follow-up:
- Local environment still does not have gurobipy, so the licensed Gurobi end-to-end OA test remains skipped locally. The code path is covered by the added unit regression and by the reviewer licensed-machine reproduction.

jkitchin (approved):
Approving. Both blocking findings are resolved and verified end-to-end against a licensed Gurobi 13.0.0:

- **Finding 2 (AMP certification)** — fixed in `12f0882`; AMP + Gurobi returns `optimal` with `gap_certified=True`.
- **Finding 1 (OA near-zero gap)** — fixed in `9139a25` via the `_compute_gap` denominator floor; OA + Gurobi now converges to `optimal` on the degenerate reproducer, and HiGHS is unaffected.

Verification on my licensed box:
- `test_oa_gurobi_converges_on_degenerate_convex_minlp_if_available` and `test_amp_gurobi_certifies_convex_minlp_if_available` both pass (these skip without a license).
- Full suite `test_gurobi_backend.py + test_oa.py + test_mip_nlp.py + test_amp.py`: **216 passed**.
- The direct matrix backend (LP/MILP/QP/MIQP/QCP/QCQP/MIQCP/MIQCQP) matched discopt's default solver and analytic optima across every class.

One non-blocking follow-up (your call, fine for a separate issue): the licensed-Gurobi end-to-end tests skip in CI without a Gurobi job, so they're currently only exercised on a licensed machine. Worth tracking so they don't silently regress.

Thanks for the fast, clean turnarounds.

PR #320, opened 2026-06-24
Move OA selection to mip-nlp solver

### LINKED ISSUE #319: Move OA MINLP selection out of gdp_method
## Problem

Discopt currently uses `gdp_method="oa"` to select outer approximation for MINLP solves. That is conceptually different from GDP reformulation choices such as `big-m`, `hull`, and `mbigm`. OA, ECP, and related MindtPy-style methods are MINLP solution strategies, while `gdp_method` should describe how disjunctive constraints are reformulated before the chosen solver runs.

This is especially confusing now that discopt has both GDP logic-based OA paths and MINLP decomposition/OA paths.

## Proposed migration

- Add a solver-family selector for MIP-NLP decomposition methods, e.g. `solver="mip-nlp"`.
- Select the concrete algorithm with a method option, e.g. `mip_nlp_method="oa"`, `"ecp"`, `"goa"`, `"roa"`, `"fp"`, or `"lp_nlp_bb"`.
- Keep `gdp_method` for GDP reformulation only.
- Preserve `gdp_method="oa"` temporarily as a deprecation-warning alias for `solver="mip-nlp", mip_nlp_method="oa"`.
- Keep GDP logic-based OA, currently `gdp_method="loa"`, separate from MINLP OA.

## Motivation

This makes the API reflect the algorithmic layers more accurately and leaves room for a MindtPy-inspired MIP-NLP decomposition solver family without overloading GDP reformulation names.

bernalde:
## Summary

Closes #319.

This moves OA selection toward the solver-family API described in the issue:

- adds `solver="mip-nlp"` as a first-class MINLP decomposition selector
- adds `mip_nlp_method` / `mip_nlp_options`
- routes `mip_nlp_method="oa"` and `"ecp"` through the existing OA implementation
- keeps `gdp_method="oa"` as a deprecation-warning compatibility alias
- leaves `gdp_method` to describe GDP reformulation choices for the new route

The broader MindtPy-style methods (`goa`, `roa`, `fp`, `lp_nlp_bb`) are reserved by name and raise `NotImplementedError` until they are implemented.

## Validation

- `PYTHONPATH=python python -m ruff check python/discopt/solvers/mip_nlp.py python/discopt/solver.py python/discopt/modeling/core.py python/tests/test_mip_nlp.py python/tests/test_oa.py`
- `PYTHONPATH=python pytest python/tests/test_mip_nlp.py python/tests/test_oa.py -q`

Note: local commit hooks were skipped because `pre-commit` is not installed in this environment; the targeted Ruff and pytest checks above pass.

### DIFF
--- a/python/discopt/modeling/core.py
+++ b/python/discopt/modeling/core.py
@@ -2782,6 +2782,11 @@ class Model:
             ``convhull_ebd_encoding``, ``use_start_as_incumbent``,
             ``obbt_at_root``, ``obbt_with_cutoff``, ``alphabb_cutoff_obbt``,
             and ``obbt_time_limit``.
+            Use ``solver="mip-nlp"`` to select the MIP-NLP decomposition
+            family. Current implemented ``mip_nlp_method`` values are ``"oa"``
+            and ``"ecp"``; ``"goa"``, ``"roa"``, ``"fp"``, and
+            ``"lp_nlp_bb"`` are reserved until their dedicated implementations
+            land.
         validate : bool, default False
             If True, run Examiner-style KKT validation on the returned
             point and attach the :class:`~discopt.validation.ExaminerReport`
--- a/python/discopt/solver.py
+++ b/python/discopt/solver.py
@@ -2179,6 +2179,8 @@ def solve_model(
     solver : str or None, default None
         Optional global-solver selector. Use ``"amp"`` to dispatch to
         Adaptive Multivariate Partitioning instead of branch-and-bound.
+        Use ``"mip-nlp"`` to dispatch to the MIP-NLP decomposition family
+        (OA/ECP now; GOA/ROA/FP/LP-NLP-BB are reserved method selectors).
         Use ``"gp"`` to dispatch to the geometric-programming fast path:
         the model is checked for GP structure (posynomial/monomial
         objective and constraints over strictly-positive continuous
@@ -2202,6 +2204,13 @@ def solve_model(
         ``disc_abs_width_tol``, ``convhull_formulation``, ``convhull_ebd``,
         ``convhull_ebd_encoding``, ``use_start_as_incumbent``, ``obbt_at_root``,
         ``obbt_with_cutoff``, ``alphabb_cutoff_obbt``, and ``obbt_time_limit``.
+    solver="mip-nlp" options
+        The MIP-NLP backend accepts ``mip_nlp_method`` and
+        ``mip_nlp_options``. Current implemented methods are ``"oa"`` and
+        ``"ecp"``; ``"goa"``, ``"roa"``, ``"fp"``, and ``"lp_nlp_bb"`` raise
+        ``NotImplementedError`` until their dedicated implementations land.
+        Existing OA options ``equality_relaxation``, ``ecp_mode``, and
+        ``feasibility_cuts`` may be passed as top-level aliases.
 
     Returns
     -------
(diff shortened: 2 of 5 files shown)


jkitchin:
A review highlighted: 

3. [python/discopt/solver.py](https://github.com/jkitchin/discopt/blob/claude/eager-knuth-7s7heo/python/discopt/solver.py) (new mip-nlp route, ~2400) — GDP-reformulation asymmetry between the two "equivalent" OA routes.
The new solver="mip-nlp" route calls reformulate_gdp(model, method="big-m") before solving, but the deprecated gdp_method="oa" alias does not reformulate (it calls solve_mip_nlp on the raw model, matching the old solve_oa path). So for a model containing disjunctions, the documented-as-equivalent old and new routes produce materially different problems. Worth either reformulating in both routes or documenting that the deprecated alias deliberately preserves the no-reformulation legacy behavior. (reformulate_gdp is a no-op when no GDP constraints exist, so models without disjunctions — including all the new tests — are unaffected, which is why this slipped past the suite.)

Notes (not blocking)
The mip-nlp route warns only about leftover **kwargs; named solve_model options it ignores (e.g. threads, strategy, skip_convex_check) are dropped silently, unlike the AMP route which explicitly enumerates ignored named options. Minor consistency gap.

Can you take a look and let me know when you think this is ready to merge?

bernalde:
Commits pushed:
- f5e57df Address mip-nlp GDP routing review

Main changes:
- Made the deprecated `gdp_method="oa"` alias reformulate GDP models with `big-m` before dispatching to `solve_mip_nlp`, matching the explicit `solver="mip-nlp", mip_nlp_method="oa"` route.
- Updated the deprecation warning to say that `gdp_method="oa"` is interpreted as `big-m` for GDP reformulation in this compatibility path.
- Added MIP-NLP warnings for non-default named `solve_model` options that the route ignores, including `threads`, `strategy`, and `skip_convex_check`.
- Added a regression test covering GDP reformulation for both the new MIP-NLP route and the deprecated OA alias.

Tests run:
- `PYTHONPATH=python pytest python/tests/test_mip_nlp.py::test_mip_nlp_and_deprecated_oa_alias_reformulate_gdp -q` failed before the fix, then passed after.
- `PYTHONPATH=python pytest python/tests/test_mip_nlp.py -q`: 5 passed.
- `PYTHONPATH=python python -m ruff format --check python/discopt/solver.py python/tests/test_mip_nlp.py`: passed.
- `PYTHONPATH=python python -m ruff check python/discopt/solver.py python/tests/test_mip_nlp.py`: passed.
- `PYTHONPATH=python python -m mypy python/discopt/solver.py python/tests/test_mip_nlp.py`: passed.
- `PYTHONPATH=python pytest python/tests/test_mip_nlp.py python/tests/test_oa.py -q`: 12 passed.
- `PYTHONPATH=python pytest python/tests/test_mip_nlp.py python/tests/test_gdp.py python/tests/test_oa.py -q`: 114 passed, 22 deselected.

- None.

Remaining risks or follow-up:
- This keeps the MIP-NLP route on the explicit GDP-to-MIP reformulation path. Native disjunctive or logic-based GDP handling should remain a separate solver/design path rather than being implied by this OA compatibility alias.

bernalde:
@jkitchin One deeper design question from #319 that I think is worth settling before we build further on this API: should discopt treat disjunctive solvers as a separate axis from MIP/MINLP solvers?

My current mental model is:

- `gdp_method` should answer how a disjunctive model is handled: either reformulate disjunctions into an algebraic MIP/MINLP with `big-m`, `hull`, `mbigm`, or `auto`, or solve the disjunctive form natively with a logic-based method.
- `solver` / `mip_nlp_method` should answer how the resulting algebraic MIP/MINLP is solved. OA belongs in this category in general: it is a MIP-NLP decomposition method, although the same OA idea can be generalized to disjunctive programs as logic-based OA.
- Logic-based OA for GDPs should probably live with native disjunctive solving, not as a GDP reformulation option.

This PR chooses the reformulation path for the deprecated alias: `gdp_method="oa"` means OA solver selection plus `big-m` GDP reformulation for compatibility. Does that match how you want to separate these axes long term, or would you prefer a different naming split for native disjunctive solvers?

jkitchin:
Agreed on the two-axis split — that's the contract we'll build on, and I've merged this PR since it's correct and unblocks the API.

- **`gdp_method`** = how a disjunctive model is *handled*: reformulate to algebraic MIP/MINLP (`big-m`, `hull`, `mbigm`, `auto`) **or** solve the disjunctive form natively (logic-based methods).
- **`solver` / `mip_nlp_method`** = how the resulting algebraic MIP/MINLP is *solved* (OA, ECP, and the reserved `goa`/`roa`/`fp`/`lp_nlp_bb`).

The native axis already lives on `gdp_method` today, which is the proof that this split is the existing convention rather than a new one: `gdp_method="loa"` dispatches to `solve_gdpopt_loa` (`solver.py:2699`). So logic-based OA is already a native-disjunctive option, not a `mip_nlp_method`. This PR's deprecated `gdp_method="oa"` → "OA solver + big-M reformulation" is consistent with that — it's reformulation + algebraic OA, not native solving.

Two decisions we'll treat as the agreed contract for follow-ups:

1. **Native + `mip-nlp` ⇒ raise.** A native-disjunctive `gdp_method` (e.g. `loa`) combined with `solver="mip-nlp"` is contradictory (one says "solve natively," the other "reformulate to algebraic and decompose"), so it should raise rather than silently reformulate.
2. **`goa` ≠ `gloa`.** `goa` stays a `mip_nlp_method` (generalized/global OA on the algebraic MINLP); global *logic-based* OA (`gloa`) belongs on the native-disjunctive axis under `gdp_method`. They must not be aliased to each other.

Captured both in #323 so they aren't lost. Thanks for pushing on the design here.

---

PR #328, opened 2026-06-27
Recognize convex exp-reciprocal products and shifted affine powers

### LINKED ISSUE #327: Recognize convex exp-reciprocal products and shifted affine powers
## Context

While working on the MIP-NLP GOA path in the fork, we hit a convex benchmark that discopt could not certify through its existing expression recognition/relaxation machinery.

Downstream context:

- Fork issue: bernalde/discopt#118
- Draft implementation PR: bernalde/discopt#129

The relevant benchmark shape is the Pyomo MindtPy `MINLP5_simple` fixture. It is a convex MINLP, but two expression forms were not handled by discopt's current classifiers/relaxers:

- `exp(2*x) * y**(-4)`, with `y > 0`;
- `(y - 6)**4`, a shifted affine even integer power.

## Proposed Classifier/Relaxation Improvements

Add a conservative product-pattern recognizer for expressions of the form

```text
exp(affine) * prod_i g_i(x)**a_i
```

where every `g_i` is affine, strictly positive on its box, and every `a_i <= 0`.

This is convex because it can be written as

```text
exp(affine + sum_i a_i * log(g_i(x)))
```

where `a_i <= 0` makes `a_i * log(g_i(x))` convex over `g_i(x) > 0`, and `exp` is convex and nondecreasing.

Also extend the composite power relaxation gate so non-bare affine integer powers such as `(y - 6)**4` are lifted instead of falling through to an unsupported generic power expression. Even integer powers of affine expressions are convex even when the affine base crosses zero.

## Why Split This Out

These changes are useful independent of the downstream GOA work. They improve discopt's ability to recognize and relax convex expression forms, while the fork's larger PR uses the improved machinery to certify convex MINLPs through OA before falling back to the nonconvex global-relaxation path.

## Validation Target

The focused PR should include:

- a convexity unit test for `exp(2*x) * y**(-4)` with `y >= 1`;
- a power-relaxation regression for a shifted even integer power such as `(x - 1)**4`;
- lint/format checks on the touched files.

bernalde:
## Summary

Closes jkitchin/discopt#327.

This adds two small, conservative improvements to discopt's convexity and relaxation machinery:

- recognize `exp(affine) * prod_i g_i(x)**a_i` as convex when each `g_i` is affine and strictly positive on its bounds and each exponent `a_i <= 0`;
- lift non-bare affine integer powers with exponent `p >= 3`, and classify even affine integer powers as convex even when the affine base crosses zero.

These forms came up while testing the downstream GOA/MIP-NLP work in bernalde/discopt#118 and bernalde/discopt#129, specifically on a MindtPy MINLP5-style convex fixture containing `exp(2*x) * y**(-4)` and `(y - 6)**4`.

## Rationale

The product rule is conservative. The expression

```text
exp(affine) * prod_i g_i(x)**a_i
```

can be written as

```text
exp(affine + sum_i a_i * log(g_i(x)))
```

For `g_i(x) > 0` and `a_i <= 0`, each `a_i * log(g_i(x))` is convex, and composition with the nondecreasing convex `exp` preserves convexity.

For shifted even integer powers, `(affine)**(2k)` is convex over the full real line, so the relaxation gate should not require the affine base to be nonnegative.

## Tests

- `python -m ruff check python/discopt/_jax/convexity/patterns.py python/discopt/_jax/milp_relaxation.py python/tests/test_convexity.py python/tests/test_power_certification.py`
- `python -m ruff format --check python/discopt/_jax/convexity/patterns.py python/discopt/_jax/milp_relaxation.py python/tests/test_convexity.py python/tests/test_power_certification.py`
- `git diff --check`
- `PYTHONPATH=python pytest python/tests/test_convexity.py::TestConvexExpressions::test_exp_affine_times_reciprocal_power -q`
- `PYTHONPATH=python pytest python/tests/test_power_certification.py::test_shifted_even_integer_power_certifies -q -m correctness`
- `pre-commit run` via the commit hook

The correctness test emitted the existing local JAX persistent-cache warning because `/home/bernalde/.cache/discopt/jax-cache` is read-only in this environment; the test passed.

### DIFF
--- a/python/discopt/_jax/convexity/patterns.py
+++ b/python/discopt/_jax/convexity/patterns.py
@@ -14,6 +14,9 @@ shapes where a dedicated proof exists:
   a homogeneous PSD quadratic.
 * **Perspective of exp: ``y * exp(x / y)`` with ``y > 0``**
   (:func:`classify_product_pattern`).
+* **Exp times reciprocal powers: ``exp(affine) * prod_i g_i(x)**a_i``**
+  with each ``g_i`` affine, strictly positive, and ``a_i <= 0``
+  (:func:`classify_product_pattern`).
 * **Weighted geometric mean ``prod_i x_i^{a_i}``** with ``x_i >= 0``,
   ``0 <= a_i <= 1``, and ``sum_i a_i == 1`` (:func:`classify_product_pattern`).
 * **Norm ``sqrt(x^T Q x)`` with Q PSD**
@@ -644,6 +647,9 @@ def classify_product_pattern(
 
     * ``y * exp(x / y)`` with ``y > 0`` and ``x`` affine — perspective of
       ``exp``: CONVEX.
+    * ``exp(affine) * prod_i g_i(x)**a_i`` where every ``g_i`` is affine and
+      strictly positive and every ``a_i <= 0``: CONVEX because it is
+      ``exp(affine + Σ a_i log(g_i(x)))`` with a convex exponent.
     * **Signomial monomial** ``c * prod_i base_i ** a_i`` with every base
       affine, classified by exponent sign pattern on the positive orthant
       (Boyd & Vandenberghe §3.1.5):
@@ -674,9 +680,8 @@ def classify_product_pattern(
                 ):
                     return Curvature.CONVEX
 
-    # Signomial monomial: c * prod_i base_i ** a_i, each base affine.
-    # Peel the leading scalar constant, then classify the product of powers
-    # by its exponent sign pattern (see the function docstring). A negative
+    # Peel the leading scalar constant. The remaining product is inspected by
+    # the exp-times-reciprocal-power and signomial recognizers below. A negative
     # constant flips the curvature.
     const, core = _split_const_mul(expr)
     if core is None or abs(const) <= 1e-12:
@@ -685,6 +690,37 @@ def classify_product_pattern(
     _flatten_product(core, factors)
     if len(factors) < 2:
         return None
+
+    exp_factor_count = 0
+    power_factors: list[tuple[Expression, float]] = []
+    exp_power_candidate = True
+    for factor in factors:
+        if (
+            isinstance(factor, FunctionCall)
+            and factor.func_name == "exp"
+            and len(factor.args) == 1
+            and classify_expr(factor.args[0], model, cache) == Curvature.AFFINE
+        ):
+            exp_factor_count += 1
+            continue
+        extracted = _extract_power_factor(factor)
+        if extracted is None:
+            exp_power_candidate = False
+            break
+        base, exponent = extracted
+        if classify_expr(base, model, cache) != Curvature.AFFINE:
+            exp_power_candidate = False
+            break
+        power_factors.append((base, exponent))
+
+    tol = 1e-10
+    if (
+        exp_power_candidate
+        and exp_factor_count >= 1
+        and all(exp <= tol for _base, exp in power_factors)
+        and all(_affine_strictly_positive(base, model) for base, _exp in power_factors)
+    ):
+        return Curvature.CONCAVE if const < 0 else Curvature.CONVEX
 
     parsed: list[tuple[Expression, float]] = []
     for factor in factors:
@@ -699,7 +735,6 @@ def classify_product_pattern(
     exps = [exp for _, exp in parsed]
     bases = [base for base, _ in parsed]
     total = sum(exps)
-    tol = 1e-10
 
     base_curv: Optional[Curvature] = None
     if (
--- a/python/discopt/_jax/milp_relaxation.py
+++ b/python/discopt/_jax/milp_relaxation.py
@@ -3439,7 +3439,9 @@ def _should_claim_composite(expr: Expression, model: Model, n_orig: int) -> bool
     * ``(base)**p`` with a *non-bare* base and either a non-integer constant
       exponent or an integer exponent ``p >= 3``.  A bare variable base is a
       monomial / fractional-power column, and integer squares of non-bare bases
-      stay on the dedicated square lift.
+      stay on the dedicated square lift.  Non-affine integer-power bases are
+      only claimed here; they still need certified curvature before any
+      envelope is emitted.
     """
     if isinstance(expr, FunctionCall) and len(expr.args) == 1:
         op_info = _univariate_arg(expr)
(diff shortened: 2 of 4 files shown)


jkitchin (changes requested):
## Review summary: request changes

Thanks for splitting these out — the convexity-recognizer half is clean, but the relaxation-gate half introduces a **soundness regression** that violates the repo's non-negotiable correctness invariant, so this needs a fix before it can land.

### 🔴 Blocker: even-power branch relaxes non-convex expressions as convex

The new branch in `_affine_base_power_curvature` (`milp_relaxation.py`):

```python
if p_is_int and p_int >= 2 and p_int % 2 == 0:
    return "convex"
```

returns `"convex"` for *any* even integer power **without verifying the base is affine** — but the function's contract (and name) assumes an affine base. `(affine)**(2k)` is convex over ℝ; `(non-affine)**(2k)` is not.

Combined with the widened gate in `_should_claim_composite` (`if p == int(p): return int(p) >= 3`), non-affine single-variable bases like `sin(x)`, `cos(x)`, `tanh(x)` now get **claimed** as composites and analytically certified convex. Unlike the odd/non-integer branches — which guard on the `base_lo` sign, so a non-affine base's inconclusive interval falls through to the rigorous `_subdivision_curvature` — the even branch short-circuits with **no guard at all**, skipping subdivision entirely.

**Confirmed end-to-end (regression vs. `main`):**

```python
# min sin(x)**4  over x in [pi/2 - 1, pi/2 + 1]
# box is centered on a local MAX (f = 1); true minimum on the box ~= 0.0852 (endpoints)
```

| branch | objective | dual bound |
|--------|-----------|------------|
| `main` | 0.0852 ✓ | 0.0852 ✓ (sound — falls back safely) |
| this PR | **1.0** ✗ | **1.0** ✗ (cuts off the true optimum) |

The bogus convex envelope places a tangent at the interior maximum (`aux >= 1` everywhere), so the solver returns a wrong optimum *and* an invalid dual bound (`1.0 > 0.0852`). This is exactly the failure mode `CLAUDE.md` calls out: *"Correctness is non-negotiable … Every phase gate enforces `incorrect_count <= 0`."*

Why the PR's own test doesn't catch it:

- `(x - 1)**4` passes only because its base is **affine** — it never exercises the missing guard.
- `(x**2 - 1)**4` happens to stay safe by accident: `distribute_products` rebuilds the inner `x**2` node, breaking the `id()`-keyed wiring so the relaxation is silently dropped. **Function-call bases (`sin`, `cos`, `tanh`) are left intact by `distribute_products`, so they wire straight through to the unsound cut.**

**Suggested fix:** make the even-power branch (ideally the whole `_affine_base_power_curvature`, to honor its name/docstring) require a proven-affine base before returning a verdict; otherwise return `None` and let `_subdivision_curvature` rigorously certify (it correctly abstains on `sin(x)**4`). Equivalently, gate `_should_claim_composite`'s new integer `p >= 3` case on the base being affine. Either way, please add a regression test with a **non-affine** even-power base (e.g. `min sin(x)**4` on the box above) asserting `r.bound <= true_min + tol` — the current tests only cover affine bases.

### ✅ Correct: `convexity/patterns.py` exp-times-reciprocal-power recognizer

This change is sound and conservative. The recognizer for `exp(affine) * prod_i g_i(x)**a_i` guards every premise: each `exp` argument is `AFFINE`; each power base is `AFFINE` **and** `_affine_strictly_positive`; all exponents `<= tol` (≈ ≤ 0); requires `exp_factor_count >= 1`; and flips CONVEX↔CONCAVE on a negative leading constant. The math (`exp(affine + sum a_i log g_i)`, convex exponent composed with convex-nondecreasing `exp`) holds even with shared variables across factors, and the early `tol` hoist doesn't affect the later signomial branch. No issues.

### Minor

`_should_claim_composite` now also claims **odd** integer powers `p >= 3` of non-affine bases (e.g. `sin(x)**3`). Those are *safe* because the odd path falls through to subdivision — but a one-line comment noting the even/odd asymmetry would help a future reader avoid "simplifying" the even branch to match.

---

**Bottom line:** the `patterns.py` half is mergeable as-is; the `milp_relaxation.py` half needs an affine-base guard on the even-power branch plus a non-affine-base regression test.

bernalde:
Addressed the requested changes from review `4586017800`.

- Commits pushed: `0fb76ee fix: guard affine power relaxation shortcut (#327)`.
- Main changes: `_affine_base_power_curvature` now first proves the power base is affine before returning any analytic affine-power curvature verdict. Non-affine integer-power bases may still be claimed for composite handling, but they must pass the rigorous subdivision curvature path before an envelope is emitted. Added a comment documenting that split.
- Regression coverage: added a direct guard test for `sin(x)**4` abstaining from the affine shortcut and an end-to-end test for `min sin(x)**4` on `[pi/2 - 1, pi/2 + 1]` asserting the objective reaches the true endpoint minimum and the dual bound remains sound.
- Tests run: reproduced the pre-fix regression locally (`objective=1.0`, `bound=1.0`, true minimum about `0.0852`); `python -m ruff check python/discopt/_jax/milp_relaxation.py python/tests/test_power_certification.py`; `git diff --check`; targeted pytest for the two new tests; `PYTHONPATH=python pytest python/tests/test_power_certification.py -q`; `PYTHONPATH=python pytest python/tests/test_power_certification.py::test_shifted_even_integer_power_certifies -q -m correctness`; `python -m ruff format --check python/discopt/_jax/milp_relaxation.py python/tests/test_power_certification.py`; `PYTHONPATH=python pytest python/tests/test_factorable_reform.py python/tests/test_monomial_lp_bound.py python/tests/test_relaxation_theorems.py -q`; `pre-commit run --files python/discopt/_jax/milp_relaxation.py python/tests/test_power_certification.py`.
- Comments intentionally not addressed: none.
- Remaining risks/follow-up: hosted CI for `0fb76ee` is pending. Local pytest still emits the existing read-only JAX persistent-cache warnings in this environment.

jkitchin (approved):
Approving — the soundness blocker is resolved.

Commit `0fb76ee` adds the affine-base guard to `_affine_base_power_curvature` (non-affine bases return `None` and fall through to rigorous subdivision), exactly as needed, plus regression tests including the `sin(x)**4` repro.

Adversarial verification done on my side, all clean:
- 96 `base**p` curvature verdicts vs numerical `f''` — 0 unsound; guard accepts only genuinely-affine bases and conservatively rejects everything else.
- `exp·∏gᵢ^aᵢ` classifier vs sampled Hessians — every near-miss (positive exponent, sign-crossing base, non-affine base/arg) correctly abstains.
- 60 randomized + multivar end-to-end solves — dual bound never exceeds the brute-force global minimum.
- Multivar path has no analogous shortcut (rigorous subdivision).

Pre-existing note (not blocking, unrelated to this PR): `test_unguarded_power_instances_certify[st_e11]` fails on a clean `main` locally too; it passes in CI here. Worth a separate look against `main`.

Issue #330, opened 2026-06-27
Performance: MINLP/MIQP wall time is dominated by Python/JAX per-solve overhead, not the solver core (3–35× SCIP on easy instances)


jkitchin:
## Context & evidence

On easy MINLP/MIQP instances, discopt is 3–35× slower than SCIP — but the solver **core is
idle**. Measured on bundled minlplib `.nl` instances (this machine, SCIP 10.0.2), solving via
`from_nl(...).solve(time_limit=60, gap_tolerance=1e-4)` and reading the
`rust_time / jax_time / python_time` split off the `SolveResult`:

```
instance   discopt wall   rust% / jax% / py%   nodes   SCIP wall   slowdown
ex1221       0.73 s        0.0 /  6.6 / 93.3      3      0.09 s       8.3x
ex1226       0.60 s        0.0 /  2.6 / 97.4      3      0.04 s      16.9x
nvs03        0.10 s        0.1 / 43.3 / 56.6      5      0.03 s       3.0x
nvs06        0.42 s        0.0 / 14.3 / 85.7      3      0.04 s       9.6x
alan         0.33 s        0.1 / 97.0 /  2.9     21      0.09 s       3.8x
dispatch     0.20 s        0.0 / 16.9 / 83.1      3      0.04 s       5.5x
ex1225       1.11 s        0.0 /  2.6 / 97.4      5      0.03 s      34.5x
```

Every instance solves in ≤21 nodes, so this is **not** a branching/node-count problem and it
is **not** the Rust solver core (rust ≈ 0% everywhere). The gap is **fixed per-solve overhead**
in the Python + JAX layers. This is a different problem from the pure-MILP node-efficiency
issue (Regime B) and the sparse-LP data-layer issue (Regime C); a faster solver core does
nothing for these instances.

## Root-cause hypothesis — CONFIRM BY PROFILING FIRST (do not assume)

The dominant cost on small/easy MINLPs is one-time setup that isn't amortized over the (tiny)
search:
- **JAX:** tracing/compilation of the relaxation/evaluator on every solve, possibly per node,
  with no compile-cache reuse across solves of the same structure (alan: 97% jax).
- **Python:** model → `lp_data` translation, `classify_nonlinear_terms`, relaxation
  construction, dispatch, and result assembly in `solver.py` (ex1225: 97% python).

Which of the two dominates varies by instance, so profile both before changing anything.

## Step 1 — Reproduce & profile (required before any change)

[...]

PR #333, opened 2026-06-27
Fix per-solve overhead on easy instances (issue #330)

jkitchin:
## Summary

This PR addresses the per-solve orchestration overhead that dominates wall time on easy MINLP/MIQP instances (issue #330). The fix has two parts: (1) route `from_nl` models through a fast numeric Rust-repr extractor instead of the expensive per-primitive eager-JAX autodiff fallback for QP/LP data extraction, and (2) short-circuit redundant root primal heuristics once the optimum is certified at the root node.

### 1. Fast QP/LP Data Extraction for `from_nl` Models
- **`python/discopt/_jax/problem_classifier.py`**: Added fast numeric repr probe in `extract_lp_data()` and `extract_qp_data()` dispatchers before falling back to autodiff. For `from_nl` models (where the algebraic DAG walk cannot run), this skips the per-primitive eager-JAX compilation storm and uses the Rust `ModelRepr` evaluator directly—orders of magnitude faster on small instances.
- The fast path is attempted first and only falls back to autodiff on exception, preserving correctness while eliminating unnecessary XLA compiles.

### 2. Root-Optimality Short-Circuit for Primal Heuristics
- **`python/discopt/solver.py`**: Added `_root_optimum_proven()` check that certifies optimality at the root node by comparing the batch minimum relaxation bound (a valid global dual bound at iteration 0) against the incumbent. Once proven, all remaining root primal heuristics are skipped as provably-wasted work.
- Applied to three heuristic entry points: main subnlp block, subnlp loop over relaxation candidates, and enumerate-binary-seeds subnlp.
- Restricted to `iteration == 0` to ensure the batch minimum is a valid global bound; on harder instances the root gap stays open and every heuristic runs as before.

### 3. Measurement & Regression Tests
- **`discopt_benchmarks/perf/overhead_bench.py`** (new): Micro-benchmark that solves the 7 flagged overhead-regime instances (each ≤21 nodes, rust% ≈ 0) and prints a table with wall time breakdown (rust/jax/python %), XLA compile count, node count, and slowdown vs SCIP 10.0.2 reference. Supports `--json` output and `--no-scip` to drop the reference column. Not a CI gate; exists for on-demand before/after measurement.
- **`discopt_benchmarks/tests/test_overhead_330.py`** (new): Three regression tests pinning correctness and overhead wins:
  - `test_repr_qp_extraction_matches_autodiff()`: Verifies the fast repr extractor matches autodiff bit-for-bit on Q, c, A_eq for `alan` (the flagship MIQP where algebraic walk fails).
  - `test_extract_qp_data_prefers_repr_for_from_nl()`: Confirms the dispatcher returns the fast repr result for `from_nl` models.
  - `test_alan_solve_does_not_recompile_per_node()`: Asserts the solve completes with <30 XLA compiles (vs ~100 pre-fix) and certifies the correct optimum.

## Implementation Details

- The fast repr extractor (`_extract_qp_data_from_repr`, `_extract_lp_data_from_repr`) was already present; this PR simply routes `from_nl` models through it before autodiff.
- The root-optimality check uses the same gap test as `_gap_converged` (absolute tolerance + relative tolerance) to ensure consistency.
- The short-circuit is conservative: it only fires on instances already solved at the root (trivially easy ones); harder instances see no regression because the root gap remains open.
- SCIP reference times (SCIP 10.0.2) are baked into the benchmark so the slowdown ratio is reproducible without a SCIP install.

https://claude.ai/code/session_01338eFiTASK1Xswrpmptdsz

### DIFF
--- a/discopt_benchmarks/perf/overhead_bench.py
+++ b/discopt_benchmarks/perf/overhead_bench.py
@@ -57,8 +57,45 @@ INSTANCES: list[OverheadInstance] = [
 ]
 
 
-def run(gap_tolerance: float = 1e-4) -> list[dict]:
-    """Solve every overhead instance and return one result dict per instance."""
+def scip_wall(nl_path: str, time_limit: float, gap_tolerance: float = 1e-4) -> float | None:
+    """Solve a ``.nl`` instance with SCIP (PySCIPOpt) and return its wall time, or
+    ``None`` when PySCIPOpt is not installed. Lets the slowdown ratio be measured
+    on the *same* machine instead of relying on the baked-in #330 reference."""
+    try:
+        from pyscipopt import Model as _ScipModel
+    except Exception:
+        return None
+    import time
+
+    m = _ScipModel()
+    m.hideOutput()
+    m.readProblem(nl_path)
+    m.setRealParam("limits/gap", gap_tolerance)
+    m.setRealParam("limits/time", time_limit)
+    t0 = time.perf_counter()
+    m.optimize()
+    return time.perf_counter() - t0
+
+
+def run(gap_tolerance: float = 1e-4, live_scip: bool = False) -> list[dict]:
+    """Solve every overhead instance and return one result dict per instance.
+
+    When ``live_scip`` is set and PySCIPOpt is installed, the SCIP wall is
+    measured on this machine; otherwise the #330 reference (SCIP 10.0.2) is used.
+    """
+    # Warm the JAX/XLA stack once on a throwaway solve so the first measured
+    # instance is not charged the one-time import + cold-compile cost (which
+    # otherwise shows up as a ~15 s outlier on whichever instance runs first).
+    _first = next(
+        (i for i in INSTANCES if os.path.exists(os.path.join(DATA_DIR, f"{i.name}.nl"))), None
+    )
+    if _first is not None:
+        measure_solve(
+            os.path.join(DATA_DIR, f"{_first.name}.nl"),
+            time_limit=_first.time_limit,
+            gap_tolerance=gap_tolerance,
+        )
+
     rows: list[dict] = []
     for inst in INSTANCES:
         path = os.path.join(DATA_DIR, f"{inst.name}.nl")
@@ -67,11 +104,15 @@ def run(gap_tolerance: float = 1e-4) -> list[dict]:
         rec = measure_solve(path, time_limit=inst.time_limit, gap_tolerance=gap_tolerance)
         wall = rec.wall_time or 0.0
         row = rec.to_json()
-        row["scip_wall"] = inst.scip_wall
+        _scip = None
+        if live_scip:
+            _scip = scip_wall(path, inst.time_limit, gap_tolerance)
+        row["scip_wall"] = inst.scip_wall if _scip is None else _scip
+        row["scip_source"] = "reference" if _scip is None else "live"
         row["rust_pct"] = 100.0 * rec.rust_time / wall if wall else 0.0
         row["jax_pct"] = 100.0 * rec.jax_time / wall if wall else 0.0
         row["python_pct"] = 100.0 * rec.python_time / wall if wall else 0.0
-        row["slowdown_vs_scip"] = wall / inst.scip_wall if inst.scip_wall else None
+        row["slowdown_vs_scip"] = wall / row["scip_wall"] if row["scip_wall"] else None
         rows.append(row)
     return rows
 
@@ -103,10 +144,17 @@ def main() -> int:
     ap.add_argument("--gap", type=float, default=1e-4, help="relative gap tolerance")
     ap.add_argument("--json", type=str, default=None, help="write per-instance records as JSONL")
     ap.add_argument("--no-scip", action="store_true", help="drop the SCIP reference column")
+    ap.add_argument(
+        "--live-scip",
+        action="store_true",
+        help="measure SCIP wall on this machine via PySCIPOpt (else use #330 reference)",
+    )
     args = ap.parse_args()
 
-    rows = run(gap_tolerance=args.gap)
+    rows = run(gap_tolerance=args.gap, live_scip=args.live_scip)
     print(format_table(rows, with_scip=not args.no_scip))
+    if rows and any(r.get("scip_source") == "live" for r in rows):
+        print("(SCIP wall measured live on this machine)")
 
     if args.json:
         with open(args.json, "w") as fh:
--- a/discopt_benchmarks/perf/overhead_bench.py
+++ b/discopt_benchmarks/perf/overhead_bench.py
@@ -1,0 +1,120 @@
+"""Per-solve orchestration-overhead micro-bench (issue #330).
+
+On easy MINLP/MIQP instances discopt's wall time is dominated by **fixed
+Python + JAX per-solve overhead**, not the Rust solver core — every instance
+below certifies in ≤21 nodes with ``rust% ≈ 0``, so a faster core does nothing
+for them. This micro-bench is the committed reproduction of that measurement:
+it solves the flagged instances via ``from_nl(...).solve(...)``, reads the
+``rust_time / jax_time / python_time`` split (plus XLA compile count and node
+count) straight off the :class:`~discopt_benchmarks.perf.measure.PerfRecord`,
+and prints a table with the slowdown vs SCIP.
+
+It is *not* a CI gate (the slow-search panel + soundness tripwire in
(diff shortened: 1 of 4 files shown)


jkitchin:
## Measured against #330 — fixes the JAX half, not the Python half

Ran the issue's own warm-loop methodology (`measure_solve`, with XLA compile counting) on this branch vs `main`, same machine/session. **Objectives identical on all 7 instances** — no correctness change. Same-session apples-to-apples:

| instance | main wall | PR wall | main compiles | PR compiles | main ×SCIP | PR ×SCIP |
|---|---|---|---|---|---|---|
| **alan** | 0.415s | **0.041s** | **102** | **0** | 4.6× | **0.5×** |
| ex1221 | 1.143s | 1.333s | 7 | 7 | 12.7× | 14.8× |
| ex1226 | 0.622s | 0.627s | 5 | 5 | 15.5× | 15.7× |
| nvs03 | 0.117s | 0.115s | 5 | 5 | 3.9× | 3.8× |
| nvs06 | 0.460s | 0.449s | 5 | 5 | 11.5× | 11.2× |
| dispatch | 0.267s | 0.231s | 5 | 5 | 6.7× | 5.8× |
| ex1225 | 1.265s | 1.285s | 5 | 5 | 42.2× | 42.8× |

### What this PR fixes ✅
- **The JAX recompile-storm root cause.** `alan` (the jax-bound instance, 97% jax in #330) drops from **102 → 0 XLA compiles, 0.415s → 0.041s (~10×), and 4.6× → 0.5× SCIP** — now faster than SCIP. This is the fast `from_nl` Rust-repr extraction path doing exactly what it should.
- Correctness preserved (same optima/certification across the set; verified separately on diverse MINLP/MIQP/multi-disjunct-GDP instances, all identical to `main`).
- The committed micro-bench deliverable (`overhead_bench.py` with the rust/jax/python + SCIP split) is in place.

### What this PR does *not* fix ❌
- **The Python-orchestration root cause** — the *other* dominant cost #330 explicitly calls out. The python-bound heavy hitters are essentially unchanged: **ex1225 (97% py, still 42.8× SCIP)**, **ex1226 (15.7×)**, **ex1221 (14.8×)**, **nvs06 (11.2×)**. Five of seven instances remain at 6–43× SCIP, outside #330's 3–5× acceptance target.
- The root-optimality short-circuit (change #2) doesn't move wall time measurably on this set — the heuristics it skips were cheap relative to total overhead here.

### Recommendation
Merge as-is — it's a clean, regression-free win on the JAX half with no correctness impact — but **keep #330 open**. The acceptance bar ("move the ≤21-node set into ~3–5× SCIP") is met only for `alan`/`nvs03`; the `model → lp_data → solve → result` Python overhead on `ex1225`-style instances is the remaining work. Suggest re-scoping #330 to that Python-bound regime (or filing a follow-up) rather than treating this PR as closing it. The PR body uses "addresses," not a `Closes #330` keyword, so merging won't auto-close — good.

(Note on absolute numbers: these warm-loop walls differ slightly from #330's original table — expected, since that table was a separate measurement run; the PR-vs-main *delta* above is what's controlled.)

jkitchin:
## Update — `5943b00` addresses the Python half too (supersedes my earlier "Python half not fixed" note)

My earlier comment measured the PR *before* the heuristic-budget commit and concluded the Python-orchestration half of #330 was untouched. **`5943b00` (success-weighted improver budget) changes that** — re-measured this branch vs `main`, same machine/session:

| instance | main wall | PR wall | main ×SCIP | PR ×SCIP | Δ |
|---|---|---|---|---|---|
| ex1225 | 1.265s | 0.450s | 42.2× | **15.0×** | −64% |
| ex1226 | 0.622s | 0.229s | 15.5× | **5.7×** | −63% |
| ex1221 | 1.143s | 0.806s | 12.7× | 9.0× | −29% |
| dispatch | 0.267s | 0.191s | 6.7× | 4.8× | −28% |
| alan | 0.415s | 0.041s | 4.6× | 0.5× | (JAX path) |
| nvs03 | 0.117s | 0.102s | 3.9× | 3.4× | ~flat (already fast) |
| nvs06 | 0.460s | 0.406s | 11.5× | 10.1× | ~flat (relaxation-bound) |

So the python-bound heavy hitters I'd called unaddressed are now substantially cut — `ex1225` 42×→15×, `ex1226` into the target band. **4/7 are now at/near #330's 3–5× SCIP target** (alan, nvs03, dispatch, ex1226); `ex1221`/`ex1225` are improved but still above, and `nvs06` is relaxation-bound (a different regime, expected flat).

### Correctness of the gating — verified, not just claimed
The budget gates only **improver**-role calls (incumbent already exists); finder-role and the lighter lattice searches stay ungated, so B&B remains exhaustive. I checked the load-bearing claim ("identical certified optima ON vs OFF") empirically across **all 79 bundled minlplib instances**, `DISCOPT_HEUR_BUDGET=1` vs `=0`:

- **Every certified-`optimal` result is identical ON vs OFF.** ✅
- Exactly one instance differs — `st_e04` — where *both* modes are time-limited `feasible` (uncertified); budget-ON is actually slightly *closer* to the true optimum (5197.49 vs 5198.33; opt≈5194.87). Favorable noise, not a regression.
- All 4 overhead regression tests pass, including the new `test_heuristic_budget_cuts_solves_without_changing_optimum`.

Also re-confirmed the earlier soundness sweep (diverse MINLP/MIQP/multi-disjunct GDP) stays bit-identical to `main`, and the root-optimality short-circuit is sound (iteration-0 batch = single root → valid global bound).

### Net
Both halves of #330 are now genuinely addressed: JAX recompile storm (alan 102→0 compiles) **and** Python improver overhead (ex1225 −64%), with no correctness or large-instance regression. Good to merge. On #330 itself: much closer to the acceptance bar (4/7 in band vs 2/7 before) — suggest leaving it open for the residual `ex1225`/`ex1221` python overhead rather than auto-closing, but that's a judgment call.

jkitchin:
## Real same-box SCIP numbers (PySCIPOpt / SCIP 10.0)

Installed PySCIPOpt and wired live SCIP into the micro-bench (`overhead_bench.py --live-scip`, warm JAX before timing). Measured budget **OFF vs ON** on the same machine, gap=1e-4:

| instance | discopt OFF | discopt ON | SCIP | OFF ×SCIP | **ON ×SCIP** | Δ wall |
|---|---|---|---|---|---|---|
| **alan** | 0.087s | 0.082s | 0.115s | 0.8× | **0.7×** | fixed (faster than SCIP) |
| ex1221 | 0.623s | 0.402s | 0.002s | 339× | 233× | −33% |
| ex1226 | 1.403s | 0.704s | 0.006s | 234× | 114× | −51% |
| ex1225 | 3.998s | 1.322s | 0.002s | 1768× | 586× | −67% |
| dispatch | 0.938s | 0.898s | 0.007s | 139× | 133× | −4% |
| nvs03 | 0.242s | 0.252s | 0.003s | 91× | 93× | ~0 |
| nvs06 | 10.15s | 10.0s | 0.030s | 340× | 337× | ~0 (Regime C) |

Objectives identical OON vs OFF on every solvable instance; adversarial soundness suite + minlplib optimality suite both pass.

### Reading the numbers honestly
- **alan is fully fixed** — 0.7× SCIP, *faster than SCIP on the same box*, 100→0 XLA compiles. It's the one instance whose cost isn't POUNCE NLP solves, and it lands exactly where #330 wants.
- The heuristic budget gives large, controlled cuts on the python-bound instances: **ex1225 −67%, ex1226 −51%, ex1221 −33%**.
- **The absolute ×SCIP here is not representative.** This box runs SCIP in 2–7 ms (~10× faster than the issue's 30–90 ms machine) while discopt's POUNCE has no optimized BLAS (~10× slower than normal); the two compound to ~100×. That's why the ratios read 100–1700× rather than the issue's 3–35×. The `alan` row (sub-1× SCIP) shows what the others would approach if the per-NLP-solve cost weren't crippled here.

### Status vs #330's acceptance bar
Both **root causes** #330 named are fixed (JAX recompile-storm + in-solver heuristic over-firing), with correctness preserved. But the **3–5× target is not confirmed** for ex1221/ex1226/ex1225 — extrapolated to a representative box they likely sit ~7–15×, not 3–5×. The residual is the per-evaluation JAX dispatch (jacobian/hessian ~150 µs × thousands of calls) plus the `ils`/`box` solves left ungated to protect nvs05/nvs23. Suggest keeping #330 open for that last leg (evaluator-dispatch rework), which I'm looking at next. nvs06 is explicitly a different regime (relaxation LP / Regime C).

---

PR #334, opened 2026-06-28
perf(milp): make the pure-MILP simplex engine fast on sparse problems

jkitchin:
## Summary

The Rust-internal pure-MILP branch-and-cut driver was **8–74× slower than SCIP on sparse set-covering models**, collapsing at the root node. Profiling **refuted the original "dense data layer" hypothesis** — the CSC rebuild / dense working matrix cost <0.1s. The real bottlenecks were dense linear algebra and engine mismatches at the root. This PR lands five profiled, validated changes.

## Changes

1. **Sparse-LU Gomory separation** — `separate_gomory` built dense `m×m` `B`/`Bᵀ` and ran **O(m³)** Gaussian elimination *per fractional integer basic variable*. Now a single feral sparse-LU factorization + ftran/btran (sparse iterative refinement preserves the GMI numerics). **Separation: 17s → 12ms** on sc2000x800.

2. **Warm-start diving** — the root continuous-repair dive did up to `n_int` *cold* primal solves. Now warm-starts each integer-fix from the running basis (one bound change → a few dual pivots), with the existing cold fallback preserving big-M robustness. **DiveRepair: ~3s → ~15ms.**

3. **feral re-pin (`a5c789f`, issue #89)** — the FT-update `u_above` reindex was O(m³) on near-dense factors → now O(m²), plus a true per-update cost counter. Latent-robustness win; a validated no-op on these sparse-factor bases.

4. **`sb_node_budget` 1024 → 128** — strong branching every node was ~2.7× the node-solve cost. Concentrating it early (reliability branching) cut sc2000x800 **~22%** with no regression on dense knapsack.

5. **Dual-slack root solve** — the cold *primal* is the wrong engine for covering LPs (10644 pivots, phase-2 degenerate stalling). The all-slack basis is **dual-feasible** (c≥0 ⇒ dⱼ=cⱼ≥0) and primal-infeasible — the **dual simplex's** home turf: **sc2000 root in 1384 pivots / 0.12s vs 3.85s (32×)**. `solve_lp_warm` re-verifies dual feasibility and falls back to the cold primal when the slack basis doesn't qualify, so the result is **always the same optimum** — safe by construction, a no-op on LPs without a complete slack basis.

## Engine bench (sparse set-covering, identical instances)

| instance | before | after |
|---|---|---|
| sc500x250 | 8.16s | **0.68s** |
| sc1000x500 | 11.5s | **2.16s** |
| sc2000x800 | 14s (never finished) | **10.2s optimal** |
| sc4000x1500 | never returned | reaches the tree (feasible incumbent) |

Dense multidimensional knapsack (regime B): **no wall-time regression**. The warm-start dual simplex math is unchanged (`primal.rs`/`dual.rs`/`linsolve.rs` untouched).

## Validation

- `cargo test -p discopt-core`: **340 tests green**; clippy + fmt clean.
- Benchmarks committed: `examples/milp_bench.rs` (dependency-free engine bench), `discopt_benchmarks/bench_milp_sparse.py` (set-cover + knapsack vs SCIP), `examples/feral_update_cost.rs` (feral #89 reproducer).

## Follow-ups (not in this PR)

- **sc4000x1500** clears the root now but the B&B tree is large — needs better cuts/branching.
- The Python `_SIMPLEX_MILP_BUDGET_CAP_S = 10.0` engine budget cap now **masks the faster engine** for hard instances (they still fall back to HiGHS). Raising it is the highest-leverage *product* lever next, with the pathological-reformulation guard in mind.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,7 +78,7 @@ checksum = "91622ff5e7162018101f2fea40d6ebf4a78bbe5a49736a2020649edf9693679e"
 [[package]]
 name = "feral"
 version = "0.11.2"
-source = "git+https://github.com/jkitchin/feral?rev=a9cea82f99c996551805f1b2d145a750c5f6bbca#a9cea82f99c996551805f1b2d145a750c5f6bbca"
+source = "git+https://github.com/jkitchin/feral?rev=a5c789f92d928446b2373af434268adb800c2f10#a5c789f92d928446b2373af434268adb800c2f10"
 dependencies = [
  "feral-amd",
  "feral-amf",
@@ -95,7 +95,7 @@ dependencies = [
 [[package]]
 name = "feral-amd"
 version = "0.2.1"
-source = "git+https://github.com/jkitchin/feral?rev=a9cea82f99c996551805f1b2d145a750c5f6bbca#a9cea82f99c996551805f1b2d145a750c5f6bbca"
+source = "git+https://github.com/jkitchin/feral?rev=a5c789f92d928446b2373af434268adb800c2f10#a5c789f92d928446b2373af434268adb800c2f10"
 dependencies = [
  "feral-ordering-core",
 ]
@@ -103,7 +103,7 @@ dependencies = [
 [[package]]
 name = "feral-amf"
 version = "0.2.1"
-source = "git+https://github.com/jkitchin/feral?rev=a9cea82f99c996551805f1b2d145a750c5f6bbca#a9cea82f99c996551805f1b2d145a750c5f6bbca"
+source = "git+https://github.com/jkitchin/feral?rev=a5c789f92d928446b2373af434268adb800c2f10#a5c789f92d928446b2373af434268adb800c2f10"
 dependencies = [
  "feral-ordering-core",
 ]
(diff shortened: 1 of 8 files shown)


jkitchin:
### Budget-cap follow-up: measured, decided **not** to raise it

The PR description flagged that `_SIMPLEX_MILP_BUDGET_CAP_S = 10.0` might mask the now-faster engine. I measured the engine directly (np.random set-cover, the bench instances) and the premise does **not** hold:

| instance | engine @ 10s | engine @ 120s | HiGHS fallback |
|---|---|---|---|
| sc500x250 | optimal, 7.84s | optimal, 7.91s | ~0.5s |
| sc1000x500 | node_limit (no incumbent) | optimal, **19.95s** | ~1.5s |
| sc2000x800 | node_limit, 1 node | node_limit, **3 nodes in 120s** (stuck) | ~14s |

On these harder covering draws the **HiGHS fallback is faster than the engine**, so the cap is already making the right call (bound the engine, defer to the faster fallback). **Raising it would regress** sc1000/sc2000. The flat Python-vs-SCIP numbers were the fallback winning, not the cap masking the engine.

So the cap stays at 10s. The data instead points to two real (separate) levers, not pursued here:

1. **No-incumbent early-bail** — the ~10s spent before falling back is wasted on no-incumbent cases (sc1000/sc2000/#291 nvs12). Aborting once the engine is clearly unproductive would cut that to ~2-3s. Tradeoff: in HiGHS-free mode the fallback is slower, so this needs care.
2. **Engine node-solve speed on hard covering instances** — the np.random sc2000 does ~40-55s/node (dive + strong-branch probes falling back to slow cold solves). The deeper fix.

Net: this PR's engine wins stand on their own (Rust bench); the budget cap is correctly tuned and unchanged.

PR #343, opened 2026-06-28
Reject non-finite LP data before HiGHS

### LINKED ISSUE #342: Reject non-finite LP data before calling HiGHS
## Summary

While testing bernalde/discopt#129, the PR-fast suite exposed a native crash in the LP path. The crash is not specific to GOA: `extract_lp_data` can return LP matrices/vectors containing NaN values, and `python/discopt/solvers/lp_highs.py` can pass those values into HiGHS. In the observed case, the process segfaulted at `h.run()` instead of returning a Python exception or falling back to another path.

## Observed failure

The hosted run failed while executing:

- `python/tests/test_dae_fd_mol.py::TestMOLCollocation::test_mol_collocation_solves`

Root-cause inspection showed invalid/non-finite data coming from the LP extraction path before the call into HiGHS.

## Expected behavior

Discopt should not pass NaN-valued LP data into HiGHS. It should reject invalid LP extraction results in Python and keep a final defensive validation in the HiGHS adapter so bad inputs fail with a clear `ValueError` rather than risking a native crash.

## Proposed fix

- Validate `LPData` returned by `extract_lp_data` and reject non-finite objective, constraint, and RHS data.
- Allow infinite bounds where they are semantically valid, but reject NaN bounds.
- Add a final finite-value guard in the HiGHS adapter before `Highs.run()`.
- Add a regression test proving NaN matrix data is rejected before HiGHS is called.

Related downstream PR carrying the temporary fix: bernalde/discopt#129.

bernalde:
## Summary

Reject invalid LP data before it reaches native HiGHS calls.

This adds two layers of protection:

- `extract_lp_data` validates returned `LPData` and rejects non-finite objective, equality matrix, RHS, and objective constant data, while still allowing infinite bounds and rejecting NaN bounds.
- `solve_lp` performs a final defensive finite-value check on objective, constraint, and RHS arrays before calling `Highs.run()`.

The issue was found while testing a downstream GOA branch, but the bug is in the general LP path and can crash the Python process if NaN matrix data is passed into HiGHS.

Closes #342.

## Tests

- `PYTHONPATH=python pytest python/tests/test_dae_fd_mol.py::TestMOLCollocation::test_mol_collocation_solves python/tests/test_lp_highs.py::TestDimensionMismatch::test_non_finite_matrix_rejected_before_highs -q -vv --tb=short`
- `PYTHONPATH=python pytest python/tests/test_lp_highs.py -q --tb=short`
- `python -m ruff check python/discopt/_jax/problem_classifier.py python/discopt/solvers/lp_highs.py python/tests/test_lp_highs.py`
- `python -m ruff format --check python/discopt/_jax/problem_classifier.py python/discopt/solvers/lp_highs.py python/tests/test_lp_highs.py`
- `git diff --check upstream/main...HEAD`
- `pre-commit run --files python/discopt/_jax/problem_classifier.py python/discopt/solvers/lp_highs.py python/tests/test_lp_highs.py`

Local pytest emitted the existing read-only JAX persistent-cache warnings in this environment. The tests passed.

## Branch Hygiene

- Base branch: `jkitchin/discopt:main`.
- Head branch: `bernalde/discopt:fix/lp-nonfinite-highs-342`.
- Source branch point: `jkitchin/discopt:main@60e3fe3`.
- Stacked status: not stacked; this branch contains one LP-specific commit on upstream `main`.
- Related downstream PR waiting on this split: bernalde/discopt#129.

### DIFF
--- a/python/discopt/_jax/problem_classifier.py
+++ b/python/discopt/_jax/problem_classifier.py
@@ -116,6 +116,24 @@ class LPData(NamedTuple):
     obj_const: float = 0.0  # constant term in objective
 
 
+def _validate_lp_data(data: LPData) -> LPData:
+    """Reject LP data that is unsafe to pass to native solver backends."""
+    finite_fields = {
+        "c": np.asarray(data.c),
+        "A_eq": np.asarray(data.A_eq),
+        "b_eq": np.asarray(data.b_eq),
+    }
+    for name, arr in finite_fields.items():
+        if not np.all(np.isfinite(arr)):
+            raise _NotLinearError(f"LP extraction produced non-finite {name}.")
+    if not np.isfinite(float(data.obj_const)):
+        raise _NotLinearError("LP extraction produced non-finite objective constant.")
+    for name, arr in (("x_l", np.asarray(data.x_l)), ("x_u", np.asarray(data.x_u))):
+        if np.any(np.isnan(arr)):
+            raise _NotLinearError(f"LP extraction produced NaN {name}.")
+    return data
+
+
 class QPData(NamedTuple):
     """Standard-form QP: min 0.5 x'Qx + c'x + d s.t. A_eq x = b_eq, bounds."""
 
@@ -1170,12 +1188,12 @@ def extract_lp_data(model: Model) -> LPData:
     _builder = getattr(model, "_builder", None)
     if _builder is not None:
         try:
-            return _extract_lp_data_from_repr(model)
+            return _validate_lp_data(_extract_lp_data_from_repr(model))
         except Exception:
             pass
 
     try:
-        return extract_lp_data_algebraic(model)
+        return _validate_lp_data(extract_lp_data_algebraic(model))
     except (_NotLinearError, Exception):
         pass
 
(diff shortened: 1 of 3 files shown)


jkitchin:
Heads up: it looks like this work landed in parallel with #339 ("kill the CI-blocking HiGHS hang"), which is already merged and is the base of this branch (`main@60e3fe3`). #339 added a `_reject_nonfinite` guard in `solve_lp` that runs unconditionally before `passModel`/`run()` and already covers everything the `lp_highs.py` changes here add:

- objective `c` (no Inf) — `_reject_nonfinite("objective (c)", ...)`
- column bounds, NaN only — `_reject_nonfinite("column lower/upper bounds", ..., allow_inf=True)`
- row bounds (`b_ub`/`b_eq`), NaN only — `_reject_nonfinite("row lower/upper bounds", ...)`
- constraint matrix (`A_ub`/`A_eq`) via `csc.data` (no Inf) — `_reject_nonfinite("constraint matrix", ...)`

So the new `_require_finite` calls and the `bounds contain NaN values` check can't fire on anything the existing guard would have let through, and the new test (`test_non_finite_matrix_rejected_before_highs`) exercises that redundant layer — the same input is already rejected on `main`, just with a different message. The original issue (#342) therefore seems largely resolved by #339.

The one piece here not covered by #339 is the `_validate_lp_data` wrapper in `problem_classifier.py`, which validates extracted `LPData` before it ever reaches `solve_lp`. That's genuine defense-in-depth, but note it changes behavior: a non-finite result from an extraction path now raises `_NotLinearError` and silently falls through to the next extraction method, rather than surfacing downstream. Worth deciding explicitly whether fall-through or fail-fast is wanted.

Suggestion: either close this in favor of #339, or narrow it to just the `problem_classifier.py` `_validate_lp_data` layer (dropping the `lp_highs.py` changes and their test as redundant).

---

bernalde:
Agreed. After checking current `main`, #339 already covers the original crash/hang path: it rejects non-finite LP data before HiGHS and also makes the repr LP extractor decline the vector-valued MOL case so autodiff handles it.

The remaining `_validate_lp_data` wrapper in this PR would be a broader extractor-policy change, and I do not want to land that without a separate failing case and an explicit fail-fast vs fallback decision.

I am closing this PR in favor of #339.

Issue #356, opened 2026-06-30
Pure-Rust standalone node numerics: remove HiGHS + SciPy from the core solve path


jkitchin:
## Goal

Make the spatial-B&B **node numerics fully pure-Rust / standalone** — no SciPy optimizers and no HiGHS in the core MINLP solve path. This issue tracks the remaining work and records hard-won intel from a first pass so the next attempt doesn't repeat dead ends.

## Done (PR #355)

SciPy removed from both per-node **bound** paths:
- **Convex-objective bound** (`solver._convex_objective_lower_bound`) — was 3 random-restart `scipy.optimize` L-BFGS-B solves + a cohort-sized `n_vars ≤ 50` cap. Now a closed-form/deterministic **box-QP**: recover the constant Hessian, warm-start at the box-projected Newton point `clip(H⁻¹(−g))`, refine with deterministic FISTA, take the supporting hyperplane (exact at `argmin_box`, valid at any iterate), margin-guarded. No cap, any size. (commit `34c8a70`)
- **αBB bound** (`solver._compute_alphabb_bound`) — was 3 random-restart `scipy.optimize` L-BFGS-B solves on an objective-only oracle (SciPy finite-differenced the gradient). Now a deterministic **supporting hyperplane of the convex underestimator `L`** using the evaluator's analytic derivatives + the perturbation's closed-form derivatives, FISTA for the anchor, convexity guard, margin-guarded. Removes the **last `scipy.optimize.minimize` use in `solver.py`** (import dropped). Also strictly more rigorous than before (the hyperplane is sound at any iterate; the old code trusted the optimizer's reported min). (commit `c4d5051`)

Both validated: convex-QP cohort (nvs17/19/23) still certifies; 117 αBB-related soundness/behavior tests green.

### A. Infeasible / "optimal-too-high" / unbounded cross-checks → pure-Rust (the `mccormick_lp` guards)

`MccormickLPRelaxer.solve_at_node` has four soundness guards that re-verify a fast-Rust-simplex verdict with `milp.solve(backend="auto")` — which is **HiGHS-first when `highspy` is installed** (`lp_backend.get_milp_solver`). The four sites (false-infeasible, non-optimal, ill-conditioned-optimal, unbounded-nonlinear-col) are the load-bearing guards that prevent false certificates.

[...]

PR #357, opened 2026-06-30
Issue #356: Pure-Rust LP certificates for sound spatial-B&B bounds

jkitchin:
## Summary

Implements rigorous certificate side-channels from the Rust simplex solver to enable sound spatial-B&B node fathoming without external HiGHS/POUNCE cross-checks. Two complementary certificates are added:

1. **Neumaier–Shcherbina safe lower bound** from row duals: a rigorous under-estimate of the LP optimum that holds at any conditioning, never falsely fathoming feasible nodes.
2. **Verified Farkas dual ray**: an independently certified proof of infeasibility that rigorously fathoms empty polytopes without a second solve.

This closes the #355-review gap where the incremental McCormick path dropped the cold path's independent verification guard, and eliminates the dominant per-node cost of re-verifying infeasible verdicts with HiGHS.

### Rust Simplex (Primal & Dual)
- **`crates/discopt-core/src/lp/simplex/primal.rs`**: Capture row duals `y = B⁻ᵀ c_B` on optimal solves and Farkas ray on infeasible solves; unscale certificate vectors back to original space after equilibration.
- **`crates/discopt-core/src/lp/simplex/dual.rs`**: Extract row duals on optimal and Farkas ray on infeasible; unscale certificates post-solve.
- **`crates/discopt-core/src/lp/simplex/scaling.rs`**: Add `unscale_dual()` and `unscale_ray()` to map scaled certificates back to original constraint/variable space (critical for ill-conditioned LPs where the certificate matters most).
- **`crates/discopt-core/src/lp/simplex/mod.rs`**: Extend `LpSolve` struct with `dual` and `ray` fields; document their interpretation by status.
- **`crates/discopt-python/src/lp_bindings.rs`**: Update Python bindings to surface the new certificate vectors.

### Python Certificate Verification
- **`python/discopt/solvers/milp_simplex.py`**:
  - Add `LpWarmCert` NamedTuple to hold `safe_bound` and `farkas_certified` flags.
  - Implement `_safe_lp_lower_bound_std()`: Neumaier–Shcherbina safe bound from free-sign multipliers, with magnitude-scaled margin to dominate float64 rounding error.
  - Implement `_farkas_certified_std()`: Verify a Farkas ray candidate by checking if the `c=0` safe bound exceeds zero (proving infeasibility).
  - Extend `solve_lp_warm_std()` with optional `return_cert` parameter to return `(result, basis, cert)` triple.

### Incremental McCormick Integration
- **`python/discopt/_jax/incremental_mccormick.py`**: Update `solve_assembled_full()` to return 5-tuple `(status, bound, x, basis, farkas_certified)` and document the safe-bound semantics.
- **`python/discopt/_jax/mccormick_lp.py`**: 
  - Trust an incremental `infeasible` verdict directly when `farkas_certified=True` (no re-verify needed).
  - Fall back to equilibration re-verify only when the Farkas ray did not verify (ill-conditioned candidate).
  - Update `_reverify_incremental_infeasible()` to extract and prefer a fresh Farkas certificate from the equilibrated re-solve.

### Relaxation & Backend Updates
- **`python/discopt/_jax/milp_relaxation.py`**: Add `safe_bound` and `farkas_certified` fields to `MilpRelaxationResult`; populate them from the warm-started simplex path.
- **`python/discopt/solvers/lp_backend.py`**: Update docstring to reflect that matrix-MILP now defaults to the self-hosted Rust

https://claude.ai/code/session_0154SAWZh7TggKvFaPkXd3t5

### DIFF
--- a/crates/discopt-core/src/expr.rs
+++ b/crates/discopt-core/src/expr.rs
@@ -1043,9 +1043,7 @@ impl ExprArena {
                         a0.max(a1)
                     }
                     MathFunc::Prod => self.reduction_values(args, x).iter().product(),
-                    MathFunc::Norm1 => {
-                        self.reduction_values(args, x).iter().map(|t| t.abs()).sum()
-                    }
+                    MathFunc::Norm1 => self.reduction_values(args, x).iter().map(|t| t.abs()).sum(),
                     MathFunc::Norm2 => self
                         .reduction_values(args, x)
                         .iter()
@@ -1451,9 +1449,7 @@ impl ModelRepr {
                         a0.max(a1)
                     }
                     MathFunc::Prod => self.reduction_values(args, x).iter().product(),
-                    MathFunc::Norm1 => {
-                        self.reduction_values(args, x).iter().map(|t| t.abs()).sum()
-                    }
+                    MathFunc::Norm1 => self.reduction_values(args, x).iter().map(|t| t.abs()).sum(),
                     MathFunc::Norm2 => self
                         .reduction_values(args, x)
                         .iter()
--- a/crates/discopt-core/src/lp/simplex/dual.rs
+++ b/crates/discopt-core/src/lp/simplex/dual.rs
@@ -39,6 +39,12 @@ pub fn solve_lp_warm(lp: &LpView<'_>, b: &[f64], start: &Basis, opts: &SimplexOp
             let view = scaled.view();
             let mut sol = solve_lp_warm_scaled(&view, scaled.b(), start, opts);
             scaled.unscale_x(&mut sol.x);
+            // Map the certificate vectors back to the original space too (see the
+            // matching note in `primal::solve_lp`): the duals/Farkas ray are in
+            // scaled coordinates and must align with the unscaled A/b the caller
+            // checks against.
+            scaled.unscale_dual(&mut sol.dual);
+            scaled.unscale_ray(&mut sol.ray);
             sol
         }
         None => solve_lp_warm_scaled(lp, b, start, opts),
@@ -292,6 +298,14 @@ impl<'a> PreparedDual<'a> {
                         None => {
                             dvec = recompute_reduced_costs(&mut lu, sp, c, n, &basis)?;
                             if dual_feasible(n, &stat, l, u, &dvec, tol) {
+                                // Row duals `y = B⁻ᵀ c_B` for the safe dual bound;
+                                // empty (caller falls back) if the btran fails.
+                                let mut y: Vec<f64> = basis.iter().map(|&j| c[j]).collect();
+                                let dual = if lu.btran(&mut y).is_ok() {
+                                    y
+                                } else {
+                                    Vec::new()
+                                };
                                 return Some(assemble(
                                     n,
                                     m,
@@ -304,6 +318,8 @@ impl<'a> PreparedDual<'a> {
                                     u,
                                     LpStatus::Optimal,
                                     pivots,
+                                    dual,
+                                    Vec::new(),
                                 ));
                             }
                             // Exact reduced costs reveal dual infeasibility (a drifted
@@ -342,6 +358,10 @@ impl<'a> PreparedDual<'a> {
                 cand = build_candidates(n, &stat, l, u, &alpha_r, &dvec, to_lower, tol);
                 if cand.is_empty() {
                     xb = recompute_basic_values(&mut lu, sp, b, n, l, u, &stat)?;
+                    // Farkas dual ray: the leaving row's `ρ = eᵣᵀ B⁻¹`. With the
+                    // ratio test finding no entering column, `ρ` (or `−ρ`) is a
+                    // direction along which `bᵀy` beats the box-max of `(Aᵀy)ᵀz`,
+                    // certifying primal infeasibility — the caller verifies it.
                     return Some(assemble(
                         n,
                         m,
@@ -354,6 +374,8 @@ impl<'a> PreparedDual<'a> {
                         u,
                         LpStatus::Infeasible,
                         pivots,
+                        rho.clone(),
+                        Vec::new(),
                     ));
                 }
             }
@@ -665,6 +687,8 @@ fn assemble(
     u: &[f64],
     status: LpStatus,
     pivots: usize,
+    dual: Vec<f64>,
+    ray: Vec<f64>,
 ) -> LpSolve {
     let mut x = vec![0.0; n];
     for j in 0..n {
@@ -690,6 +714,8 @@ fn assemble(
             basic_vars,
         },
         iters: pivots,
+        dual,
+        ray,
     }
 }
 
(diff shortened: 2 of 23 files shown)


jkitchin:
## Review: certificate work is sound; one blocking bound-coverage regression

Reviewed locally on branch `claude/issue-356-ilicuj` (release build) plus a static read of the Rust certificate extraction. Summary: **the certificate architecture is sound and well-engineered, but the HiGHS removal regresses bound *coverage* on ill-conditioned lifted LPs — confirmed by a local main-vs-branch comparison — and that regression is invisible to PR-fast CI.**

### ✅ What's sound (keep it)

- **Rust certificate extraction — sound across the board** (static review):
  - Dual `y = B⁻ᵀ c_B` is computed correctly (`primal.rs:670-696`, `dual.rs:303-308`) and sign-aligned with the Python N–S bound (`rc = c − Aᵀy`, no negation needed; weak duality gives `g(y) ≤ truth` for any `y`).
  - The Farkas ray is **independently re-verified** in Python (`_farkas_certified_std` → `g₀(y) > 0`), checked against the original `A/b/lb/ub`. A wrong ray can only *fail to certify* (forcing fallback); it cannot manufacture a false infeasibility proof.
  - **Unscaling is mathematically exact**: `unscale_dual` does `y = R ŷ` (`scaling.rs:155-162`), and `ĝ(ŷ) = g(Rŷ) = g(y)` is invariant under the equilibration — so the certificate is correct precisely on the ill-conditioned LPs where it matters. `unscale_ray` uses the column transform, correct for the primal ray space.
  - Every failure path returns an **empty** certificate (`Vec::new()`), never garbage → Python reports `safe_bound=None` / `farkas_certified=False` and the consumer falls back safely. No false-certificate path found.
- **It correctly closes the #355 infeasible-trust gap** — the verified Farkas ray replaces the cold path's dropped independent cross-check, without a second solve.
- **The `scipy.linprog(method="highs") → POUNCE` separator swap is sound** (`edge_concave.py`, `multilinear_separation.py`): it takes only the dual *slope* from POUNCE and **recomputes the intercept from the vertices** with a validity check — neatly sidestepping POUNCE's dual sign/scale convention. 42 convexity/SOC soundness tests pass.
- Local tests green: **95 passed** (`test_lp_certificates` + simplex + backend-select); **42 passed** (convexity/SOC separators).
- The rust-analyzer "missing `dual`/`ray` field" diagnostics on `primal.rs:672/686`, `dual.rs:457` are **stale LSP noise** — `cargo check` compiles clean; every real `LpSolve { … }` construction sets the new fields.

### 🔴 Blocking: bound-coverage regression on ill-conditioned lifted LPs

`python/tests/test_bucket2_sound_bounds.py` (unmodified by this PR) regresses:

| | result |
|---|---|
| **main** (rebuilt, `-m "not slow"`) | **30 passed** |
| **this branch** (rebuilt, same selection) | **10 failed**, 86 passed |

All 10 failures are on the conditioning-sensitive instances — **nvs05, nvs20, nvs22, chance, st_e36** — with the assertion:

```
AssertionError: [nvs05] objective dropped — no root bound
assert res.lower_bound is not None
```

This is **not unsoundness** (`None` is never a false certificate), but it is a real **robustness regression**: with HiGHS removed, the Neumaier–Shcherbina safe bound returns `None` on these ill-conditioned lifted LPs (the fast Rust simplex is less robust there than HiGHS was), and the HiGHS fallback that previously produced a finite bound is gone — so the root bound is **dropped**. These instances would lose certification or get much weaker bounds, and **`nvs05` was a recent win (#354)** that this regresses.

### ⚠️ PR-fast CI does not catch this

The regressing tests are `@pytest.mark.correctness`, which the PR-fast and coverage jobs exclude (`-m "... and not correctness ..."`) — only 3/32 of that file run in PR CI. So this PR can go **green on PR CI while silently regressing** these bounds; it only surfaces in a full/nightly correctness run or the benchmark phase-gate.

### Recommendation

The certificate work should land — it's sound and is the right architecture for #356. Before merge, please:

1. **Add a pure-Rust bound fallback for the dropped-bound cases.** When `safe_bound` is `None`, fall back to a finite pure-Rust bound (the interval/αBB bound already in-tree, or POUNCE's IPM — more robust than the fast simplex on these ill-conditioned systems) rather than letting `lower_bound` drop to `None`. Goal: restore `test_bucket2_sound_bounds` to green.
2. **Run the `correctness`-marked bucket2 tests in this PR's gate** (or a documented pre-merge full run) so the regression can't slip through green PR-fast CI.

Non-blocking: the safe bound's `_NS_MARGIN_REL = 1e-9` is a heuristic float-error bound, not directed-rounding — acceptable (mirrors `obbt._ns_safe_lp_lower_bound`) but worth a one-line doc note that "rigorous" is "almost certainly correct" rather than provable for very large dense systems (~n ≳ 1e7 terms).

jkitchin:
## Re-review: the bound-coverage regression is FIXED (rigorously). Remaining blockers are mechanical.

Re-reviewed locally on the updated branch (`88a515b`, release rebuild). The blocking finding from my earlier review is resolved.

### ✅ The bound-coverage regression is fixed — and the fix is more rigorous than the snap

`b6efd8e` ("rigorous FBBT safe bound for free/unbounded lifted columns") fixes the nvs05/nvs20/nvs22/chance/st_e36 root-bound drop. The mechanism is rigorous, not heuristic: it recovers a finite **valid** box for the free/unbounded lifted columns via **FBBT on the equality system `A_std z = b`**. FBBT bounds contain the whole feasible set, so the box-min identity `g(y) <= p*` stays sound while becoming finite — no reduced-cost snapping involved.

**Verified locally: `test_bucket2_sound_bounds.py -m "not slow"` → 30 passed** (was 10 failed at my last review).

### ✅ Soundness is clean across the (now larger) scope

The PR grew since the last review — it now also defaults the **LP** path to the Rust simplex, plus sparse equilibration and a CSC-native warm solve. The full soundness gauntlet stays green: **87 passed** across `test_lp_certificates`, `test_gear4_false_infeasible`, `test_hda_false_infeasible`, `test_unbounded_relaxation_soundness`, `test_cross_term_sqrt_soundness`, `test_nested_division_soundness`, `test_nvs16_soundness`, `test_adversarial_recent_fixes`, `test_alphabb_bound_soundness`, `test_obbt_soundness`, `test_bucket2_composition_soundness`. Combined with the earlier static verification of the Rust certificate extraction (Farkas ray independently verified; N–S unscaling `y = Rŷ` exact), the soundness story holds.

### 🔴 Not mergeable as-is — three mechanical blockers (none are soundness)

1. **Conflicts with `main`** now that #361 (QP HiGHS removal) has merged → needs a rebase. Small surface: `python/discopt/solvers/lp_backend.py` and `python/tests/test_lp_backend_select.py` (both backend-routing defaults).
2. **CI red, all fixable:**
   - Lint: `ruff format` on `python/discopt/_jax/convexity/linear_context.py` (one file).
   - Rust: `cargo fmt` on `crates/discopt-core/src/expr.rs:1043` (a `Norm1` match arm) — Rust tests themselves pass (346).
   - **Python fast — the one real item:** `test_amp_simplex_backend.py::TestAmpSimplexCertificationParity::test_same_certificate_as_default[_bilinear]` runs **42s solo** on the new sparse/CSC simplex path, so it times out (>120s) under xdist parallel load. This is not a flake — it's a genuine perf characteristic of the new path. Needs a speedup, or a budget/marker adjustment, before CI is green.

### ⚠️ Coordination with #361 (worth reconciling)

#357's FBBT fix and #361's now-merged `rc_snap_tol` snap both address the **same** free-column safe-bound problem, by different mechanisms:
- **#357 — FBBT** in the safe-bound path: rigorous (FBBT box contains the feasible set), and it's what makes bucket2 green here.
- **#361 — `rc_snap_tol` reduced-cost snap** in `obbt._ns_safe_lp_lower_bound`: heuristic, **default-off**, currently inert.

They live in different functions, so they coexist after rebase — but the rigorous FBBT fix likely makes the snap **redundant** for this purpose. Please reconcile (keep FBBT as the primary; decide whether the snap stays as an off-by-default fallback or is removed). This also affects tracking issue **#363** (which assumed the snap would be the enablement for this exact regression — FBBT supersedes that, so #363 should be updated or closed once this lands).

### Verdict

The substantive blocker is **resolved with a rigorous fix**, and soundness is clean across the expanded scope — strong progress. What's left is mechanical: **rebase on main + `ruff format` + `cargo fmt` + the 42s `_bilinear` timeout** (the only genuine perf item). I'd re-run the `correctness`-marked bucket2 tests explicitly in the merge gate, since PR-fast still excludes them.

PR #360, opened 2026-06-30
Raise pytest memory cap for JAX tests

bernalde:
## Summary

- Raise the default pytest virtual-address cap from 16 GB to 32 GB for PR-fast, coverage, and AMP coverage runs.
- Document when to keep the cap versus disabling it for the single-process AMP integration suite.
- Add regression coverage for the wrapper default and make the dry-run worker-count test independent of an outer `PYTEST_XDIST_WORKERS` override.

## Root cause

The capped pytest wrapper uses `prlimit --as`, which limits virtual address space rather than resident memory. Modern JAX/XLA can reserve more than 16 GB of virtual address space during compilation, so the old default could trigger native aborts such as `std::bad_alloc` or exit 134 even when actual RAM use was not the problem.

## Validation

- `PYTHONPATH=python python -m pytest python/tests/test_amp.py::test_memory_capped_pytest_wrapper_dry_run python/tests/test_amp.py::test_memory_capped_pytest_wrapper_default_allows_jax_virtual_memory python/tests/test_amp.py::test_memory_capped_pytest_warns_when_prlimit_missing python/tests/test_amp.py::test_pr_fast_tier_excludes_heavy_manual_markers python/tests/test_amp.py::test_pr_fast_tier_honors_worker_override -q` - 5 passed
- `PYTHONPATH=python scripts/run_memory_capped_pytest.sh python -m pytest python/tests/test_affine_decision_rule.py python/tests/test_alphabb.py python/tests/test_alphabb_bound_soundness.py python/tests/test_amp.py python/tests/test_amp_milp_backend_failure_recovery.py -q --tb=short -rs` - 225 passed, 15 deselected
- On the development integration branch before extracting this patch: default capped `make test-all` passed with 4500 passed, 65 skipped, 1172 deselected, 2 xfailed.


jkitchin (approved):
Reviewed and approve. Correct root cause: `prlimit --as` caps virtual address
space (not RSS), and JAX/XLA reserves large virtual address space during
compilation, so the 16 GB cap could abort with `std::bad_alloc` / exit 134.
Raising to 32 GB is the right fix.

- `16384 → 32768` applied consistently across ci.yml (×3), Makefile, and the
  wrapper default; `--as=34359738368` == 32768 × 1024² (verified).
- README `PYTEST_MEMORY_LIMIT_MB=0` escape hatch for the single-process AMP
  integration suite is a good addition.
- Worker-count dry-run test hardened against an outer PYTEST_XDIST_WORKERS.
- Verified locally: the 5 wrapper tests pass; CI green. CI/infra only, no solver
  logic touched.

Issue #145, opened 2026-06-15
Solver soundness: false-optimal under large-coefficient trilinear-equality + integer B&B


jkitchin:
## Summary

The spatial + integer B&B can **certify a false optimum** (`gap_certified=True`
with an objective ~4 orders of magnitude worse than the true optimum) on a model
that combines a **trilinear-equality constraint**, **integer variables**, and a
**large linking coefficient**. The reported dual `bound` equals the suboptimal
incumbent, which means an invalid (too-high) node lower bound pruned the subtree
containing the true optimum.

This violates the prime soundness invariant (a valid dual bound must never exceed
the true optimum; "optimal" must never be reported at a suboptimal incumbent).

**No current MINLPLib corpus instance triggers it on `main`** — the ratio terms
that would create this structure are soundly *dropped* from the relaxation today
(they return `bound=None`, never a false bound). It surfaced while prototyping the
issue #138 "lift `y = N/D` to the bilinear equality `y*D = N`" reformulation: that
rewrite produces exactly this structure, so it is a blocker for #138 and a latent
landmine for any user model of this shape.

## Deterministic repro (pure `dm.Model`, no `.nl`)

```python
import os
os.environ["JAX_PLATFORMS"] = "cpu"
os.environ["JAX_ENABLE_X64"] = "1"
import discopt.modeling as dm

# y is pinned to the integer ratio (x0*x1)/(x2*x3) by a trilinear equality;
# the objective minimizes |T - 1e6*y| encoded with two nonneg slacks x4, x5.
m = dm.Model("g")
x0 = m.integer("x0", lb=12, ub=60); x1 = m.integer("x1", lb=12, ub=60)
x2 = m.integer("x2", lb=12, ub=60); x3 = m.integer("x3", lb=12, ub=60)
x4 = m.continuous("x4", lb=0.0, ub=2e5); x5 = m.continuous("x5", lb=0.0, ub=2e5)
y  = m.continuous("y", lb=0.04, ub=25.0)   # exact interval of (x0*x1)/(x2*x3)
m.minimize(x4 + x5)
m.subject_to(y * x2 * x3 - x0 * x1 == 0)            # defining: y = (x0*x1)/(x2*x3)
m.subject_to(1e6 * y + x4 - x5 == 144279.32477276)  # large linking coefficient
r = m.solve(time_limit=60, gap_tolerance=1e-4)
print(r.status, r.objective, r.bound, r.gap_certified)
```

[...]

PR #361, opened 2026-06-30
Make the QP path HiGHS-free and remove qp_highs (issue #359)

### LINKED ISSUE #359: QP seam still HiGHS-first: route get_qp_solver off HiGHS (follow-up to #356)
## Context

In #356 the LP/MILP matrix-solver routing was made HiGHS-free by default:

- `get_milp_solver(backend="auto")` → simplex-first (Rust warm-started B&B), HiGHS/POUNCE as fallbacks.
- `get_lp_solver()` / `get_exact_dual_lp_solver()` → the pure-Rust simplex, which now exposes `dual_values`/`reduced_costs` from the optimal basis in HiGHS's convention (validated equal to HiGHS on non-degenerate LPs, and strong-duality-valid on degenerate ones).

**The one remaining HiGHS-first default is the QP seam:** `get_qp_solver(prefer_pounce=False)` in `python/discopt/solvers/lp_backend.py` still orders `(_qp_highs, _qp_pounce)`.

## Why it was left out of #356

1. **No Rust QP engine.** The self-hosted Rust simplex solves LPs only; there is no quadratic-objective counterpart, so there is no exact, fast, Rust-native QP to put first (the way the LP/MILP simplex is).
2. **POUNCE's QP duals are not rigorous.** The only non-HiGHS QP backend is the POUNCE IPM (`qp_pounce`). POUNCE returns the analytic center of the optimal face; its objective normally matches but can drift on ill-conditioned problems while still reporting `OPTIMAL` (issue #145), and its duals are interior-point (not vertex) multipliers. So a blind flip to POUNCE-first risks the same unsound over-tightening / wrong-dual hazards #356 was careful to avoid for the LP path.

## What "done" looks like

A genuinely HiGHS-free QP default, without regressing soundness. Options to evaluate:

- **Option A — POUNCE-first with a validity guard.** Default `get_qp_solver` to `(_qp_pounce, _qp_highs)` but, where QP duals/reduced costs are consumed, verify them (KKT residual / strong-duality / sign check) and degrade gracefully on failure — the same "verify a certificate, don't trust the solve" pattern used for the McCormick separators and LP safe bound in #356. Need to inventory which `get_qp_solver` consumers read duals vs only the primal/objective; primal-only consumers can flip immediately.
- **Option B — a pure-Rust QP.** A bounded-variable QP (active-set, reusing the existing simplex basis machinery, or a small dense IPM) that returns exact vertex/KKT duals. Larger effort; the clean long-term answer.

## Acceptance

- `get_qp_solver()` no longer prefers HiGHS by default.
- QP-dual consumers stay sound (validated against HiGHS on a battery; soundness suites green).
- `prefer_pounce` semantics preserved.

## Pointers

- `python/discopt/solvers/lp_backend.py` — `get_qp_solver`, `_qp_highs`, `_qp_pounce` (and the module docstring, which already notes the QP carve-out).
- `python/discopt/solvers/qp_pounce.py`, `qp_highs.py` — the two QP backends.
- #356 (parts A/B/C + the LP-dual exposure) for the LP/MILP precedent and the validation approach.

jkitchin:
Closes #359.

Routes the QP/MIQP path entirely onto discopt's own engines and removes the HiGHS QP backend, completing the HiGHS-free default the LP/MILP routing got in #356. Per direction on the issue, this goes past the literal acceptance ("`get_qp_solver` no longer prefers HiGHS") to **no HiGHS QP routing at all** — POUNCE is the QP engine.

## What changed

**QP routing is HiGHS-free**
- `_solve_qp`: POUNCE first, then the JAX QP IPM as a last resort — HiGHS removed from the order. `prefer_pounce` is retained for call-site compatibility but no longer selects a backend, so the deprecated `nlp_solver="ipm"` alias no longer opts back into HiGHS for QPs.
- Convex MIQP dispatch: dropped the `_solve_qp_highs` short-circuit; routes straight to the self-hosted B&B (`_solve_miqp_bb`, POUNCE node relaxations).
- `get_qp_solver` is POUNCE-only and raises (no HiGHS fallback) when POUNCE is unavailable.
- QP/MIQP dual recovery (`_mip_recover_relaxation_duals`) uses POUNCE for the QP case (LP/MILP recovery still uses `lp_highs`, untouched — that's separate MILP work).

**Soundness guard for the POUNCE-first default (#145 drift)**
- QP duals/reduced costs are reported, never consumed for bound tightening (OBBT/DBBT read the LP oracles), so the only hazard is a drifted objective on an unconverged interior-point solve.
- Added `QPResult.kkt_error`, surfaced from POUNCE (prefers the unscaled M1-fixed residual from pounce#174 when present, else 0.6.0's scaled `final_kkt_error`).
- `_solve_qp_matrix` now rejects a non-stationary "optimal" (`kkt_error > 1e-6`) in addition to the existing primal-feasibility check, degrading to the next engine. Measured healthy solves report ~1e-9 (incl. a 1e6-conditioned Q), so the guard has ~3 orders of margin and does not misfire.

**No-rescue tracking (accepted POUNCE-only, made observable)**
- The JAX QP IPM fallback is weak (returns `iteration_limit` even on easy QPs), so with HiGHS gone POUNCE is effectively the only working QP engine. Rather than build a speculative rescue, a POUNCE non-result is logged at WARNING with greppable markers `qp-pounce-no-result` (POUNCE available but failed/guard-rejected) and `qp-pounce-unavailable` (not installed) so the frequency can be measured.

**Removed the qp_highs backend**
- Deleted `python/discopt/solvers/qp_highs.py` and `_solve_qp_highs`; dropped the `_qp_highs` selector factory and dangling docstring cross-references.

**OBBT N–S reduced-cost snap (first commit)**
- Opt-in reduced-cost snap on `_ns_safe_lp_lower_bound` (`rc_snap_tol`, default 0.0 = exact legacy behaviour, all current OBBT callers unchanged). Lets the Neumaier–Shcherbina safe bound consume an interior-point (POUNCE) dual without dropping to `None` on free columns — a column with an infinite binding bound must have reduced cost exactly 0 in any dual-feasible point, but an IPM reports a small residual and `rc * inf` blows the bound to `-inf` (confirmed on nvs05/nvs22/st_e36). Snapping `|rc| <= rc_snap_tol` to 0 only where the binding bound is infinite recovers the KKT value; a genuine nonzero reduced cost exceeds the tol and still yields `None`, so the tol also guards real unboundedness. Rigorous once `rc_snap_tol` is tied to POUNCE's unscaled dual-infeasibility residual (pounce#174); documented as a heuristic until that wheel ships. Analysis recorded on pounce#173.

## Tests
- Renamed `test_qp_highs.py` → `test_qp_solve.py`, keeping the engine-agnostic end-to-end QP/MIQP correctness coverage (now exercising POUNCE) and dropping the HiGHS-only direct/fallback/ipm cases.
- `test_qp_pounce.py`: replaced the HiGHS dual/objective cross-checks with an independent SLSQP (scipy) oracle.
- Seam tests assert no `qp_highs` engine exists, the `ipm` alias is HiGHS-free, both guards (feasibility + KKT-residual) reject bad results at the `_solve_qp_matrix` level, and the tracking marker fires.
- Selector test asserts `get_qp_solver` raises with no POUNCE (no HiGHS fallback).
- Solver-free unit tests for the N–S snap: default-off drops the trapped bound, snap recovers a finite sound bound, the tol does not mask genuine unboundedness, no-op on finite bounds.
- `test_p1_milp_bb_soundness.py` updated; `Makefile` test path updated; design docs + the minlp-solver-expert skill swept to the HiGHS-free routing.

All QP/MIQP/OBBT/convex/soundness suites pass locally; ruff + format clean; full test collection succeeds (4331 tests).

## Notes / follow-ups
- `lp_highs`/`milp_highs` are intentionally untouched — full HiGHS removal on the LP/MILP side is separate (the DBBT dual oracle still needs a rigorous source; tracked via pounce#173/#174).
- `rc_snap_tol` stays a documented heuristic until a POUNCE release ships #174's unscaled residual; the QP KKT guard already auto-prefers that residual when present.
- `CHANGELOG.md` left as historical record.

## Commits
1. `eeb6b41` — opt-in N–S reduced-cost snap
2. `a1da83f` — HiGHS-free QP routing + KKT-residual guard
3. `a2285ba` — no-rescue tracking marker
4. `44778c7` — `qp_highs` removal
5. `f4f9509` — doc sweep


jkitchin:
## Review: clean, sound-by-design, no regression — good to merge (two notes)

Reviewed locally on `claude/issue-359-review-7ncpr8` (Python-only on current `main`'s Rust, so no rebuild). A much tidier change than #357.

### ✅ Sound / verified

- **The N–S reduced-cost snap is correctly default-off** (`rc_snap_tol=0.0`, gated by `if rc_snap_tol > 0.0`) → inert for every current caller. Confirmed by the soundness suite: **47 passed** (`test_obbt_soundness` + `test_bucket2_sound_bounds` + `test_convex_objective_bound` + `test_alphabb_bound_soundness`), with `test_bucket2_sound_bounds` staying at **30 passed**. The snap *logic* is honestly a heuristic and labeled as such — snapping `|rc| <= tol → 0` on an infinite-bound column overestimates that term (`-inf → 0`), which is exact only because an *optimal*-status solve forces the free column's true KKT `rc` to 0 — and the docstring calls out the rigorous path (pounce#174's unscaled dual-inf residual). Landing it off, with solver-free unit tests, is the right way to stage it.
- **QP KKT-residual guard** (`_QP_KKT_RESIDUAL_TOL = 1e-6`, ~3 orders of margin over the observed healthy ~1e-9) rejects a non-stationary "optimal" and degrades to the next engine — sound direction. And per the routing docstring, QP duals/reduced costs are reported but **never consumed for bound tightening** (OBBT/DBBT read the LP oracles), so the only hazard is a drifted objective, which the feasibility + KKT guards catch.
- **POUNCE-only QP/MIQP correctness** (local sweep, `time_limit=60`): `st_miqp1`=281, `st_miqp2`=2, `st_miqp3`=−6, `st_miqp4`=−4574 (all match known optima), `meanvarx`/`ex1221`/`ex1226` solve to certified optima. No coverage regression on the cohort.
- **92 passed** — the PR's own tests (`test_ns_safe_bound_snap`, `test_qp_solve`, `test_qp_pounce`, `test_qp_backend_seam`, `test_lp_backend_select`, `test_miqp_nonconvex_dispatch`, `test_p1_milp_bb_soundness`).
- `qp_highs` removal is clean — the only remaining mentions are explanatory docstrings in `lp_backend.py`.

### ⚠️ Two notes (non-blocking)

1. **POUNCE-only QP has no robust fallback.** The JAX QP IPM last resort is weak (returns `iteration_limit` on easy QPs). The cohort is fine, but a QP POUNCE can't solve would silently degrade vs. the old HiGHS path. Handling it as *observability* (`qp-pounce-no-result` WARNING marker) rather than a speculative rescue is the right call — just worth actually watching the marker frequency post-merge, since this is the one place coverage could quietly drop.
2. **Don't enable `rc_snap_tol` with a fixed tol in a soundness-critical bound path** until pounce#174's rigorous residual lands (or with careful gating). Off now, so no concern for this PR — flagging so it isn't switched on casually. When the LP/MILP HiGHS removal (#356/#357) needs it to recover the dropped nvs05/nvs22/st_e36 bounds, that enablement is exactly where the heuristic→rigorous distinction becomes load-bearing.

### Context

For reference, #357 (the LP/MILP-side certificate work) currently has a confirmed bound-coverage regression — 10 `@pytest.mark.correctness` bucket2 soundness tests drop their root bound to `None` on nvs05/nvs22/st_e36, invisible to PR-fast CI (correctness marker excluded). #361's snap is the eventual *fix* for that class, staged here safely off — good sequencing. (Detail on #357.)

Net: approve from a soundness standpoint. The change is honest about its tradeoffs and stages the risky piece (the snap) in the off position.

Issue #413, opened 2026-07-03
fable review of submodules


jkitchin:
Fable reviews

**▶ Master priorities + Opus fix-loop protocol: [`docs/dev/review-execution-plan.md`](https://github.com/jkitchin/discopt/blob/claude/dae-module-review-bdybrr/docs/dev/review-execution-plan.md)** (see the pinned comment below for the condensed version).

Planning docs (pre-existing):
- [ ] certification https://github.com/jkitchin/discopt/blob/claude/dae-module-review-bdybrr/docs/dev/certification-gap-plan.md (underway)
- [ ] correctness https://github.com/jkitchin/discopt/blob/claude/dae-module-review-bdybrr/docs/dev/correctness-issues.md
- [ ] performance https://github.com/jkitchin/discopt/blob/claude/dae-module-review-bdybrr/docs/dev/performance-plan.md (revisit after fixes land)

Check for correctness, thoroughness, performance issues. Each review ends with a detailed plan for Opus (verify → fix → validate, per-finding IDs, repros, acceptance criteria).

**Checkboxes below = review written. Fix progress is tracked by the `STATUS:` banner at the top of each review doc and the ledger in the execution plan.** All submodule reviews are complete:

- [x] nn [docs/dev/nn-module-review.md](https://github.com/jkitchin/discopt/blob/claude/dae-module-review-bdybrr/docs/dev/nn-module-review.md) — **P0** scaling → unsound big-M
- [x] dae [docs/dev/dae-module-review.md](https://github.com/jkitchin/discopt/blob/claude/dae-module-review-bdybrr/docs/dev/dae-module-review.md) — **P0** Neumann sign, integral 2×, silent no-dynamics
- [x] modeling [docs/dev/modeling-module-review.md](https://github.com/jkitchin/discopt/blob/claude/dae-module-review-bdybrr/docs/dev/modeling-module-review.md) — **P0** vector-constraint aggregation, from_nl binary, cross-model aliasing
- [x] mo [docs/dev/mo-module-review.md](https://github.com/jkitchin/discopt/blob/claude/dae-module-review-bdybrr/docs/dev/mo-module-review.md) — NBI axis (k≥3); sense handling verified correct
- [x] gp [docs/dev/gp-module-review.md](https://github.com/jkitchin/discopt/blob/claude/dae-module-review-bdybrr/docs/dev/gp-module-review.md) — **P0** builder-row blind spot; gap_certified mislabel
[...]

PR #492, opened 2026-07-05
perf/fix(mo): MO5 compile-cache sweep objectives + MO6-10 tail (#413)

jkitchin:
Closes the mo-module review tail (issue #413): **MO5** (P2 perf) and **MO6–MO10** (smaller, by-inspection). MO1–MO4 landed in prior PRs (#480 and earlier). Only `python/discopt/mo/` is touched on the code side — mo is a pure orchestration layer over `Model.solve()` and is **off the B&B solve path**, so the solver engine (`solver.py`, `_jax/`, Rust) is byte-for-byte unchanged.

### MO5 — repeated JAX recompilation in sweeps (P2 perf, bound-neutral) — FIXED
`evaluate_expression` recompiled (Expression-DAG walk + JAX trace) each objective on **every** call: k per accepted point plus k² for the payoff table, of the same unchanging expressions. **Confirmed:** 26 recompiles on an 11-point bi-objective weighted-sum sweep.

Fix: new `ObjectiveEvaluator` (`utils.py`) compiles each objective **once** via the parameter-agnostic `compile_expression_params` and evaluates against the model's current parameter snapshot each call. Threaded through `_collect_objectives_at_x`, `_payoff_matrix`, `nadir_point`, and `lexicographic_payoff` so all five sweeps compile each objective exactly once.

**Verification (bound-neutral):**
- Legacy `compile_expression` calls on the sweep objective-eval path: **26 → 0**; `compile_expression_params` == k (one per objective).
- Fronts **BIT-IDENTICAL** before vs after across all five methods (weighted_sum, epsilon_constraint, weighted_tchebycheff, NBI, NNC) for k=2 and k=3 toys — max abs diff **0.0**.
- Isolated objective-eval microbenchmark: 32.5 ms → 9.5 ms for 400 evals (**3.4×**). Solves dominate real wall time, so this matters most for cheap LP/QP subproblems, as the review notes.

### MO6 — `weighted_sum(ideal=...)` discarded the ideal — FIXED
Baseline recomputed both ideal and anchors (1 discarded `ideal_point` solve) even when the supplied `ideal` sufficed. Now anchors are solved only when actually needed to estimate the nadir (`normalize=True` and no `nadir`); with `normalize=False` or an explicit `nadir` the supplied ideal is honored with **0** anchor solves.

### MO7 — no overall time budget — FIXED
All five sweeps now take `total_time_limit=`; on expiry the sweep stops and returns the **partial** front with `method` tagged `".../truncated"`. A warning fires when the subproblem grid exceeds ~200 solves.

### MO8 — brittle positional constraint cleanup — FIXED
Every scalarizer's `del model._constraints[saved_n:]` is replaced with identity-based removal of exactly the rows each scalarizer added, so a row appended by other code during a solve survives.

### MO9 — dead `senses_list` arg — FIXED
Removed from `_payoff_matrix`; signature is now `(model, objectives, anchors, evaluator=None)`.

### MO10 — AMP drops parameter-carrying constraints — CONFIRMED, DEFERRED
Re-confirmed live (`AMP: omitting constraint _mo_nbi_eq_1 … Cannot linearize Parameter`). The behavior is **sound** (dropping rows only loosens the relaxation) but **bound-changing** — the real fix (substitute the `Parameter`'s current value into the MILP relaxation) alters what the B&B sees and needs the CLAUDE.md §5 differential-bound protocol. Per the plan §7 Phase 3 it is design-first and out of scope for this by-inspection tail PR. Tracked for the solver-integration round; does not gate closing the mo review.

## Cert-baseline neutrality
Exact by construction: no code outside `python/discopt/mo/` changed, and nothing on the B&B solve path imports `discopt.mo`. `incorrect_count == 0`; node_count / certified objective unchanged.

## Tests run
- New: `python/tests/test_mo_tail_mo5_mo10.py` — 14 smoke tests. **Fails-before/passes-after** verified: MO5 recompile count (26 baseline → asserts 0), MO6 ideal recompute (1 baseline → asserts 0). MO5 bit-identity, MO7 budget/tag/warn, MO8 identity cleanup + foreign-row survival, MO9 signature.
- `pytest -m smoke` (full, excluding the unrelated `test_nn_formulations.py` torch-import segfault — a stale-venv/torch env issue, not this change): **544 passed, 1 skipped**.
- mo module suite (`test_mo_*`, fast): **59 passed**. Slow solver-backed mo suite: running at PR time; result to follow in a comment.
- `ruff check` / `ruff format --check` / `mypy` clean on `python/discopt/mo/`.

https://claude.ai/code/session_019GsMQNrw4gayMFzEcySThS


jkitchin:
Slow solver-backed mo suite result: the full `pytest -m "slow or not slow"` over `test_mo_scalarization.py`, `test_mo_nbi.py`, `test_mo_pareto.py`, `test_mo_indicators.py`, `test_mo_augmecon2.py`, `test_mo_tail_mo5_mo10.py` completed **green (exit 0, no failures/errors)** under `JAX_PLATFORMS=cpu JAX_ENABLE_X64=1`. The suite is long (~25+ min) because the solver-backed k=3 NBI/ε-constraint tests each run full global solves — a pre-existing characteristic, unchanged by this PR (MO5 is bound-neutral: it only removes redundant JAX recompiles; fronts are bit-identical before/after).

Issue #557, opened 2026-07-08
Engine: FT LU-update fails on McCormick dense columns → refactor-per-pivot is the ~13× BARON gap


jkitchin:
## Summary

The remaining ~13× wall gap vs BARON on jointly-proved global-opt instances is now root-caused to a single engine-level cause: **the in-house dual simplex's product-form (FT) LU basis-update fails on the wide dense columns produced by McCormick lifting, forcing a from-scratch refactorization on nearly every pivot.** This tracks the engine-level work needed to close it — all larger than a single bound-neutral PR, hence a tracking issue rather than an immediate task.

## Evidence (THRU-3/4/5, all merged)

Session diagnosis chain (see `docs/dev/thru3-node-decomp-2026-07-07.md`, `thru4-cut-inheritance-2026-07-07.md`, `thru5-base-node-solve-2026-07-07.md`):

- **Not** Python orchestration (~2%, PYFIX-1), **not** cut strength (CUT-1 NO-GO), **not** coverage (G3 kill), **not** bounding-envelope (G4 kill), **not** heuristic volume (fixed by ILS cap + governor).
- **THRU-3:** per-node separation loops (`_separate_univariate_square` + `_separate_psd`) were a dense-QP-specific drag; both off → nvs24 9→309 nodes (36× ceiling), bound ~unchanged.
- **THRU-4:** root-cut-pool inheritance recovered 2–5× on the dense-QP class (nvs19 gained a cert) but is throughput-neutral broadly — stays default-OFF.
- **THRU-5 (the floor):** with separation off, the base per-node solve is **>98% the Rust simplex call**; the dense-rebuild / marshaling hypotheses were falsified (matrix already sparse, density ≈0.01, built once, warm-basis threaded). Per-node LP wall is dominated by **refactorization** (nvs24 55.8%, nvs19 47.2%, st_e36 66.2%), and refactorizations are **89–100% `RefacFtFail`**: the product-form (FT) LU update returns `Err` on the wide dense-column lifted-McCormick basis → full refactor. On st_e36 that is **once per pivot (42,143 refactorizations)**. The warm dual re-solve is attempted every node but rejected ~90% of the time (falls through to cold primal). Net: O(m²)-per-pivot where a working update is O(m), with m≈4456 on nvs24.

[...]

PR #573, opened 2026-07-10
perf(lp): #557 — density-aware dense/sparse LU route for wide-McCormick node bases (flag-gated)

jkitchin:
## Summary

Root-cause + fix for the Forrest–Tomlin refactor **storm** that dominated per-node LP cost on dense-McCormick spatial-B&B nodes (THRU-5: refactorization 55–66% of node-LP wall, `RefacFtFail` ~100%). Closes the engine core of #557. **Two commits that ship together:** the density-aware LU route (flag-gated) + the feral#112 bump it depends on for bound-closing.

**The storm is a dense/sparse routing artifact in discopt.** `FeralLU::factorize` forced feral's **dense** LU for every basis with `m ≤ 256`. The wide lifted-McCormick node bases land at `m ≈ 200–256` but are only ~4% dense; forcing them dense paid an O(m³) refactor **and** drove the dense FT update's `TinyPivot` refactor-every-pivot storm. feral's **sparse** LU factors them ~13× faster and its sparse FT update commits on ~every pivot. (This is why ENGINE-1/2 #570/#571 and the cadence experiment missed it — they profiled the *sparse* FT path while the runtime storm was in the *dense* path these bases were mis-routed to.)

## The change

1. **`8598a8f6`** — `crates/discopt-core/src/lp/simplex/linsolve.rs` (+91/−2): `route_dense_decision()`, flag-gated by `DISCOPT_LU_DENSITY_ROUTE`. **OFF (default): the exact historical expression — byte-identical routing.** ON: for `16 < m ≤ 256`, defer to feral's own density gate (`nnz·4 ≥ m·m`); tiny and truly-dense-small bases stay dense; `m > 256` unchanged. Two unit tests pin OFF=historical, ON=diverts-the-sparse-band-only.
2. **`5ce15161`** — feral bumped to git rev `660224dc` (the feral#112 compensated-accumulation fix). **Required, not optional:** a global50 gap-instance smoke showed that on plain feral 0.12, route-ON leaves st_e36's dual bound **stuck at the root value** (−304.5) → status drops `optimal → feasible`. The storm-elimination is feral-independent; the **bound-closing is not** — it needs #112's more accurate sparse factorization. With the bundle, st_e36 is `optimal` (bound −246.019) at **1.69×**.

## Measurements (route ON vs OFF, bundled build)

- **Storm:** st_e36 `RefacFtFail` 42337 → ~3–5k (**8–13×↓**); nvs24 2996 → 263 (**11×↓**). Residual fails are genuinely-dense small bases that correctly stay dense.
- **Wall:** st_e36 24.3s → **14.4s (1.69×, `optimal`, 153 nodes identical)**; st_e07 6.5s → **0.84s (7.75×)**; nvs06 1.55×; st_e13 neutral. Broad 35-instance sweep otherwise neutral, **no real regression** (sub-100ms "regressions" were jitter; neutral on min-of-5).
- **Honest scope note:** the fast June-18 global50 gap rows (st_e38, st_test1, st_miqp1/2/4) were already closed by other work since that baseline — the route is inert on them. The route's wins are the storm-bound class (st_e36, st_e07, nvs06, …).

## Correctness (bound-changing-numerics regime, CLAUDE.md §5)

A different factorization of the same LP ⇒ acceptance bar is **status-preserved + node-count-neutral + in-tolerance objective**, `incorrect_count = 0`:

- **10-instance deterministic cert panel (bundled build, OFF vs ON): all 10 `optimal → optimal` (zero certification losses), node_count byte-identical 10/10**, objectives 8/10 byte-identical + 2 last-bit drifts (ex1225 5.3e-12, st_e36 2.8e-14).
- Broad sweep: 0 node_count drift on all 27 optimal pairs; worst objective drift st_e07 (abs 5.96e-5, rel 1.4e-7 — well inside the abs 1e-4 / rel 1e-3 gate; present OFF and ON, unchanged in kind by the route).
- **Suites (bundled, route ON):** `cargo test -p discopt-core` 430/0 · `pytest -m smoke` 633/1skip/0 · adversarial 10/0 · clippy clean. Default-OFF byte-identity reconfirmed (st_e36: optimal/−246.0/153/storm 42337 — identical to baseline; the feral bump alone changes nothing without the route).

## Rollout

Ships **default-OFF** (`DISCOPT_LU_DENSITY_ROUTE=1` to enable). Graduation to default-ON is a separate step after consecutive-nightly soak, per the §5 bound-changing protocol.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,9 +77,8 @@ checksum = "91622ff5e7162018101f2fea40d6ebf4a78bbe5a49736a2020649edf9693679e"
 
 [[package]]
 name = "feral"
-version = "0.12.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef91918cdb92a5ea18d2a40563014b84102ec41ebf621fe4a5c0d9d337d59326"
+version = "0.13.0"
+source = "git+https://github.com/jkitchin/feral?rev=660224dc929953072582603a15c7cb4c7bf0592c#660224dc929953072582603a15c7cb4c7bf0592c"
 dependencies = [
  "feral-amd",
  "feral-amf",
@@ -96,8 +95,7 @@ dependencies = [
 [[package]]
 name = "feral-amd"
 version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ead2f064dbcef0e0d3110202e9d6d7c293c1dffe6da5964513a20c5967e4d560"
+source = "git+https://github.com/jkitchin/feral?rev=660224dc929953072582603a15c7cb4c7bf0592c#660224dc929953072582603a15c7cb4c7bf0592c"
 dependencies = [
  "feral-ordering-core",
 ]
@@ -105,8 +103,7 @@ dependencies = [
 [[package]]
 name = "feral-amf"
 version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a3ea6dacabfa085e7f40080c894da245d9db60e9211f1f8ab08d715d6d0f5a29"
+source = "git+https://github.com/jkitchin/feral?rev=660224dc929953072582603a15c7cb4c7bf0592c#660224dc929953072582603a15c7cb4c7bf0592c"
 dependencies = [
  "feral-ordering-core",
 ]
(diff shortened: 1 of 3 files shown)


jkitchin:
**Correction incoming — do not merge yet.** A global50 gap-instance smoke surfaced a bound-dependency the earlier "feral-independent" verdict missed:

- On plain feral 0.12 + route, **st_e36 drops `optimal → feasible`** (dual bound stuck at the loose root −304.5; incumbent correct, bound valid, but optimality no longer certified). The **storm-elimination is feral-independent**, but the **bound-closing needs feral#112's compensated factorization** — with #112 + route, st_e36 stays `optimal` (bound −246.02) at ~1.7×.
- Also: the fast `st_e*/miqp*` global50 gaps (st_e38, st_test1, st_miqp1/2/4) were **already closed by other perf work since the June-18 baseline** — the route is inert on them (they never hit the storm). st_e36 is the live case among the smoked instances.

So this PR is being updated to **bundle the feral#112 bump** (route flag-gated as before; #112 factorization always-on), with the cert panel re-run on the bundled build to confirm no other instance silently loses its certificate. Marking draft until that lands.

Issue #606, opened 2026-07-10
MILP B&B node-efficiency gap: 81% degenerate pivots, 24x nodes vs HiGHS (split from #598)


jkitchin:
## In-house MILP B&B node-efficiency gap: 459 nodes / 22,833 pivots (81% degenerate) where HiGHS needs 19 nodes / one call

Split from #598 (whose certification half is FIXED in #604 — gap-closed solves now certify `optimal`; failed node LPs are soundly accounted via parent-bound floors + branching, and the untrusted-placeholder false-incumbent hole is closed).

**What remains — the efficiency half (measured on the AMP multi4N MILP relaxation, `_make_alpine_multi4n(n=2, exprmode=1)`):**
- discopt in-house MILP B&B: **459 nodes, 22,833 simplex pivots, 81% degenerate dual pivots**, RefacCap=443
- scipy/HiGHS on the identical matrix: **19 nodes, one call, 0.24 s** (~24× node ratio)

The dominant signatures are (a) heavy dual-simplex degeneracy (81% of pivots make no progress — anti-degeneracy: perturbation/EXPAND-style bound shifting is the standard cure) and (b) node counts far above what the relaxation strength justifies (branching quality / node presolve on the partitioned-lift structure).

Per DECOMP-1 (`docs/dev/certification-effort-decomposition-2026-07-10.md`), this is Lever B/C on the MILP-relaxation path; the broader panel's dominant cost is Lever A (relaxation looseness), so this is queued behind the A-class work unless MILP-relaxation solves show up hot in profiles.

jkitchin:
**Closing after a Step-0 re-measurement (2026-07-15): the probe class this issue measured no longer exists on the default path.**

Re-ran the DECOMP-1 `amp_multi4n` lane at `main@7f190df` (full record: `docs/dev/issue-606-step0-remeasurement-2026-07-15.md` on branch `claude/issue-606-relevance-bemk1r`):

[...]

Issue #178, opened 2026-06-16
Incremental dual simplex inner loop (supersedes cross-node factorization carry-over)


jkitchin:
## Summary

Profiling (see the comment below) showed that **cross-node factorization carry-over — the original subject of this issue — is not worth it**: the per-node initial factorization it would eliminate is only ~7% of the work on `ex1252`. The ~92% lives in the **dual-simplex inner loop**, which recomputes too much per iteration. This issue is repointed at that.

## The actual hot spot

`PreparedDual::run_dual` (`crates/discopt-core/src/lp/simplex/dual.rs`) recomputes, **every iteration**:

- `x_B = B⁻¹(b − Σ_{nonbasic} Aⱼ xⱼ)` — a full scatter over all nonbasic columns **plus a full `ftran`**; and
- `y = B⁻ᵀ c_B` (and per-`j` reduced costs `d_j = c_j − yᵀAⱼ` via `sp.dot`).

For `ex1252` (437×530, ~2,246 dual pivots) this dominates: `reopt` = 624 ms of a 589 ms-wall / summed-thread profile, vs 29 ms for the initial factorize.

## Proposed work — incrementalize the dual loop

Maintain state across pivots with rank-1 updates instead of recomputing from scratch, mirroring how the **primal** simplex already maintains `x_B` incrementally (`primal.rs`):

1. **Incremental `x_B`.** After a dual pivot, update `x_B` along the pivot column rather than rebuilding the RHS and re-`ftran`-ing. Refresh exactly on each refactorization to bound floating-point drift (same discipline as the primal path).
2. **Incremental reduced costs.** Maintain `d_j` across pivots via the pivot row `ρ = e_rᵀB⁻¹` (already computed for the ratio test): `d_j ← d_j − (d_q/α_rq)·α_rj`. Avoid the per-iteration `y = B⁻ᵀc_B` + full pricing scan; refresh on refactorization.
3. **Dual steepest-edge / Devex pricing (optional, follow-on).** Once `d_j` is incremental, a cheap reference-weight pricing for the leaving variable can cut iteration counts; correctness is unaffected (it only changes which improving row leaves).

## Soundness / guardrails

- Pure performance change. Keep the existing dual-feasibility precondition, the cold-primal fallback on any difficulty, and the final feasibility audit. Incremental quantities are refreshed on every refactorization, so drift is bounded exactly as in the primal solver.
[...]

PR #614, opened 2026-07-11
p3(#99): exact dual steepest-edge pricing — sound, flag-gated (DISCOPT_DUAL_DSE), measured KILL on the #606 pathology

jkitchin:
### TL;DR
The task premise — that **DSE pricing and BFRT are both MISSING** from the in-house dual simplex — is **false**. Both landed in commit `299c28eb` (#178, *"dual Devex pricing + bound-flipping ratio test"*): the dual loop already prices with **Devex** (a steepest-edge *approximation*, Goldfarb–Reid reference weights) and already runs the **bound-flipping long-step dual ratio test (BFRT)**. The one genuine delta was **Devex → exact DSE**. I implemented exact DSE, proved it sound, and measured it on the reproduced #606 pathology: **it regresses** (2.5x wall, RefacCap 3.4x), so it ships **behind `DISCOPT_DUAL_DSE`, default-OFF, and the default stays OFF.** Recorded as falsification **F11**.

This is the "partial, sound, verified > big and unverified" outcome the task anticipated — the verified result is a KILL.

### What the reference-guide row got wrong
`docs/dev/reference-solver-guide.md` §2 listed *Dual DSE* and *BFRT* as **MISSING** — a keyword miss (it scanned for "steepest-edge"/"DSE", not the "Devex" that implements the approximation). Corrected the §2 matrix rows and Lever 1 with `file:symbol` evidence.

### Reproduced pathology (deterministic, repeated — matches DECOMP-1 exactly)
Repro: `_make_alpine_multi4n(n=2, exprmode=1)` -> `build_milp_relaxation` -> `.solve()` under `DISCOPT_PROFILE=1`.

| metric | DSE OFF (Devex, = DECOMP-1 baseline) | DSE ON (exact) |
|---|---|---|
| objective | -26.82204470195638 | -26.822044701956543 (LP-bound-neutral ~1e-13) |
| status | optimal | optimal |
| `DualWarmSolves` | 761 | 634 |
| `DualDegeneratePivots` | 18511 (81% of dual pivots) | 16710 (-10%) |
| `RefacCap` | 443 | 1495 (3.4x) |
| `Phase1Pivots` (primal) | 21471 | 75178 (3.5x) |
| `NodeLpSolve` | 459 @ 1694 ms | 321 @ 4239 ms (2.5x slower) |

DSE-OFF reproduces the DECOMP-1 figures (`DualWarmSolves=761`, `DualDegeneratePivots=18511`, `RefacCap=443`) to the number, confirming the harness is faithful.

### Why exact DSE loses here (root cause)
The classic HiGHS DSE win is over **Dantzig** pricing. discopt already has **Devex**, a *good* DSE approximation — so exact DSE buys almost nothing on iteration count (-10% degenerate pivots) while paying: (a) per-warm-solve seeding of `m` BTRANs plus a reseed on every refactorization, and (b) different leaving-row choices that land on bases the FT-update/refactor machinery handles **worse** (RefacCap 3.4x, Phase-1 pivots 3.5x). Net: 2.5x slower. It is also **not node-count-neutral** (459->321 node LPs — the changed LP vertices change branching), so it could never be a silent default even if it were fast.

The #606/#598 per-node cost is **not a pricing gap**; it is degeneracy geometry + FT-update/refactor churn on lifted bases (DECOMP-1 §5.1a: iterative refinement / anti-degeneracy; the `RefacCap` is primal-side).

### What shipped
- `SimplexOptions::dual_exact_dse` (default-OFF; read from env `DISCOPT_DUAL_DSE`).
- Exact DSE weights `beta_i = ||B^-T e_i||^2` (`recompute_dse_weights`) + the **Forrest–Goldfarb 1992** rank-1 update recurrence in the dual loop, seeded at entry and reseeded on refactorization. When the flag is off, the path is **bit-identical** to before (the field defaults false).
- BFRT untouched (already present and correct).

### Verification
- `cargo test -p discopt-core`: **442 + 4 + 1 pass, 0 fail** — including 3 new tests:
  - `dse_weight_recurrence_matches_fresh_recompute` — the FG recurrence reproduces a from-scratch `||row of B^-1||^2` recompute to **1e-9** (this test caught and killed a wrong "floor" in the first implementation).
  - `dse_and_devex_agree_on_optimum_random_battery` — 80 random warm re-solves: DSE and Devex return the **same optimum to 1e-9** and same status (LP-level bound-neutrality).
  - `dse_vs_devex_degenerate_warm_pivot_counts` — degenerate-warm probe; asserts bound-neutrality, prints pivot aggregates (the entry-experiment evidence).
- **Bound-neutrality (default-OFF path):** all 10 pre-existing dual tests pass identically; the DSE-OFF repro reproduces the exact baseline counters. Flag-OFF is a no-op.
- `pytest -m smoke` (against this worktree): **643 passed, 1 skipped, 0 failed**.
- Adversarial suite (`pytest -m slow python/tests/test_adversarial_recent_fixes.py`): **10 passed, 0 failed**.

### Docs
- `gap-closing-execution-plan.md` §6: **F11** (binding falsification) + §7 progress-ledger row.
- `reference-solver-guide.md`: corrected the two inaccurate "MISSING" rows and Lever 1 (now RESOLVED/KILLED with evidence).

### Explicitly NOT done (per task + measurement)
- **Default not flipped.** `DISCOPT_DUAL_DSE` stays OFF. Do not merge-and-enable.
- BFRT not re-implemented (already present).

Refs verified present in `docs/references.bib`: `Huangfu2018`, `ForrestGoldfarb1992`.


jkitchin:
Superseded by a docs-only PR: the F11 falsification + guide correction are banked; the killed exact-DSE implementation is intentionally not shipped (no dead flags, CLAUDE.md §3).

Issue #632, opened 2026-07-11
Relaxation claim-boundary: replace _should_claim_composite defer-list with canonical AVM normal form (unblocks envelope graduation)


jkitchin:
## Problem: graduating any new relaxation "claimer" default-ON incurs a whack-a-mole tax

discopt's default global path is the auxiliary-variable method (AVM / lifted
McCormick), like BARON/Couenne/SCIP. But over time it has accreted a **federation
of overlapping specialized lift paths** — `relaxation-catalog.md` §4 lists
McCormick-NLP, McCormick-LP-lifted, TM2014 multivariate, piecewise-McCormick,
learned-ICNN, plus the monomial-secant lift, the square lift, the incremental
McCormick engine, the issue-267 univariate-product path, the finite-domain
trig-square table, repeated-factor lifting, and the H-UNI/H-LOG envelopes.

Several of these can relax the **same** expression. So each time a new envelope is
added as a claimer of "any composite of shape X," it overlaps the coverage of the
others, and an arbitration predicate (`_should_claim_composite`) must decide who
wins. That predicate is a **hand-maintained list of negative special-cases**
("defer to the square lift here, the monomial lift there, the trig table there"),
discovered empirically through test collisions rather than derived from an
invariant. The collisions are **order-masked under pytest-xdist** (a test can pass
under `-n4` and fail serially/in CI), so a green parallel suite is not proof of no
silent divert.

This blocked graduating H-UNI (`DISCOPT_UNIVARIATE_ENVELOPE`) default-ON in v0.6.0
(PR #631): three distinct collisions surfaced one test at a time, for a gain of a
single certified instance. H-UNI was reverted to a sound opt-in flag.

## How BARON and MAiNGO avoid this

Both apply **one uniform relaxation rule to a canonical decomposition** — there is
never a "which path claims this node" question:

- **BARON / Couenne / SCIP (AVM):** recursively decompose every nonlinear
  expression into a canonical DAG of elementary operations (one aux var per
  elementary op) until every node is an atom (`+`, `×`, `xⁿ`, `exp`, `log`, …), then
  relax each atom with its *one* known envelope. One decomposition, one envelope per
  atom type — nothing to arbitrate.
- **MAiNGO (reduced-space):** propagate a `(cv, cc, interval)` tuple bottom-up
  through the DAG via McCormick composition, no aux vars. One rule, every node.

[...]

PR #636, opened 2026-07-12
feat(#632): uniform factorable relaxation engine (default) replaces the federation + P3 per-node probing

jkitchin:
## Summary

Completes #632: replaces the federated claim system (~17 overlapping specialized relaxation collectors arbitrated by a hand-maintained defer-list) with a **uniform factorable relaxation engine** — the canonical AVM normal form SOTA global solvers (BARON/Couenne/SCIP) use: one decomposition, one sound envelope per atom. The engine (`uniform_relax.py`) is now the **default** per-node relaxation; `build_milp_relaxation` delegates to it unconditionally and the federation build body is bypassed (orphaned/dead — see correction below). Also lands the **P3 per-node probing** branch-and-reduce workstream.

> **Scope correction (this is not the original PR):** the branch began as a *byte-identical* R0+R1.1+R1.2-entry foundation. It is now a **bound-changing engine replacement** — the certificate path is new, not byte-identical to `main`. Correctness is preserved **by construction and verified**, not by byte-identity.

**Correctness — verified on the local host (maintainer, HEAD `0a8a7885`):** `global50` / `cert0` gate is **green — `incorrect_count = 0` across 116 instances**; no false optimal/infeasible, `bound ≤ incumbent` everywhere. +2 proved vs the prior HEAD, with much better trees on hard cases (`nvs09` optimal in **19 nodes / 18.5 s**, was 215 nodes; `st_miqp2` `obj=2.0` in 9 nodes). Wall time roughly flat.

Full plan + per-stage state ledger: [`docs/dev/avm-canonicalization-plan.md`](https://github.com/jkitchin/discopt/blob/claude/issue-632-opus-plan-ffxld4/docs/dev/avm-canonicalization-plan.md) §10.

## What's in this PR

**Uniform engine (the default relaxation)** — `python/discopt/_jax/uniform_relax.py` (`build_uniform_relaxation`): relaxes every canonical-DAG atom class (`atomize` taxonomy: univariate / power / product / ratio / multivar / opaque) soundly via the auxiliary-variable method, composing per-atom envelopes with aux-bound propagation. Every builder falls to a sound interval floor when it cannot tighten, so there is no "could not linearize" fallback. `build_milp_relaxation` delegates unconditionally; the ~3700-line federation build body + the collector/defer-list machinery are **orphaned** (unreachable behind the unconditional delegate — dead, not yet deleted; `milp_relaxation.py` is still ~4292 lines with the `_collect_*` family orphaned).

> **Reviewer corrections (2026-07-14):** two claims in the original description were inaccurate and are corrected here.
> - **Federation body: dead code now deleted.** ~1,466 lines of orphaned federation code (the `_collect_*` collectors, piecewise/trig relaxation dataclasses, `LogMonomialRelaxation`, `_convex_claimer_enabled`, …) were removed by an AST reachability analysis and verified **bound-neutral** (identical node_count/objective/bound on an 11-instance corpus). `milp_relaxation.py`: 4292 → 2719 lines. (Commit `c83dcb1c`.)
> - **This PR *does* add `DISCOPT_*` flags** (9, all default-off/functional): `DISCOPT_NODE_PROBING`, `DISCOPT_NODE_PROBE_MAX_VARS`, `DISCOPT_ENTROPY_ATOM`, `DISCOPT_LOGSUMEXP_ATOM`, `DISCOPT_XEXP_ATOM`, `DISCOPT_NORM_ATOM`, `DISCOPT_RELENT_ATOM`, `DISCOPT_ADAPTIVE_NLP`, `DISCOPT_ANALYTIC_SEPGRAD`. The earlier "No new `DISCOPT_*` flags" statement was wrong. (The dead `DISCOPT_LOG_MONOMIAL`/H-LOG flag has since been **deprecated and removed** — its log-space envelope lives in the uniform engine.)

**Relaxation-layer tightness (P1 / P2 / P4):**
- **Composite-convex OA (P2)** — certified convex/concave multivariate nodes lifted *additively* (decomposition kept, `w == dec`, OA on top → at-least-as-tight by construction), with `_separate_convex` adding exact supporting tangents lazily at the LP point (Kelley).
- **Log-space signomial envelope (P1/T3)** — sound additive `z=ln x / s=Σaz / w=exp(s)` band for positive products; recovers the `nvs09` `(∏x)^0.2` blocker and the positive ratio / wide-multilinear cases.
- **Two-piece odd-power hull (P1)** — exact facets for sign-straddling `t^p` (p odd ≥ 3, lo<0<hi).
- **P4 curvature routing** — each atom dispatches to its tightest sound envelope.

**P3 branch-and-reduce** — per-node **probing** (`crates/discopt-core/src/presolve/probing.rs`, wired into the in-tree reducer + `solver.py` with node fathoming, default-off / byte-neutral); FBBT and OBBT were already at nodes. Public `probe_box()` API in `tightening.py`. Contracts only on proven infeasibility.

## Verification

- **Correctness (local host, maintainer):** `global50` `cert0` **green — `incorrect_count = 0` / 116 instances**.
- **In-container CI:** fast suite + serial `-m claim_boundary -n0` green; `cargo test -p discopt-core` 445 passed; ruff + ruff-format + mypy clean.
- **At-least-as-tight** (assembled build vs pre-sweep `claim-baseline.jsonl`, in-house simplex): **0 looser, 0 unsound**, 2 static-tighter, 45 equal; `claim-baseline.jsonl` regenerated for the new shapes.
- **Soundness by construction:** every envelope row is a proven valid inequality at the lifted point; probing only contracts on proven-infeasible fixings; feasible-point sampling worst residual ≤ 2.7e-14, 0 cuts.

## Deferred / follow-ons (tracked in §10)

- Product-side tightness parity and residual heavy-math classes: sign-mixed high-arity multilinear (exponential hull), general linear-fractional `A(x)/B(x)` ratios (heatexch), `cvxnonsep_nsig30`.
- **`contvar` coverage regression** — its tightened LP (+248 rows) exceeds the Rust simplex root iteration budget → `iteration_limit` → None (a *sound* trivial bound, never a false certificate; breaks no test). Follow-on: simplex scaling/warm-start or a lift-count cap on very large instances (not band-aided).
- Local host: `≤ minlplib.solu` oracle detail, BARON node-count side-by-side, and the per-node-probing default-on decision + budget tuning.


jkitchin:
## R1.2 acceptance headline — nvs09 **certifies** locally ✅

Confirming the open item from "What's next": **nvs09 certifies with the current code in a local (non-CI) environment.** All three `test_lr2_nvs09_cert.py` cases pass, and the certificate is sound.

### Test run

```
tests/test_lr2_nvs09_cert.py::test_nvs09_certifies_with_univariate_envelope_flag PASSED
tests/test_lr2_nvs09_cert.py::test_nvs09_certifies_with_both_flags              PASSED
tests/test_lr2_nvs09_cert.py::test_nvs09_does_not_certify_without_flags         PASSED
========================= 3 passed in 96.20s =========================
```

### Concrete numbers (nvs09, optimum −43.134, root-cert budget 4.4e-3)

| Setting | status | bound | objective | nodes |
|---|---|---|---|---|
| `DISCOPT_UNIVARIATE_ENVELOPE=1` | **optimal** | −43.13433704 | −43.13433689 | 5 |
| both flags (`+DISCOPT_LOG_MONOMIAL=1`) | **optimal** | −43.13433704 | −43.13433689 | 5 |
| flags OFF (default) | feasible | −48.12954800 | −43.13433692 | 141 |

- **Sound cert**: bound −43.13433704 ≤ optimum, gap `|bound − opt| ≈ 1.5e-7` — well inside the 4.4e-3 budget. Certifies in **5 nodes**.
- **OFF baseline** confirms the guard: default leaves a materially looser dual bound (−48.13, ~5-unit gap) and does not certify (141 nodes, `feasible`). This is exactly the gap the univariate-envelope closes.
- `LOG_MONOMIAL` is inert here — the univariate envelope alone already tightens the root enough (identical bound/nodes with and without it).

### Environment

- Ran the PR branch (`2a86dba9`) Python source in an isolated git worktree, reusing the repo's prebuilt `_rust` extension (this PR touches no Rust — only Python).
- Python 3.12.11, `JAX_PLATFORMS=cpu`, `JAX_ENABLE_X64=1`, subprocess-per-solve (flags read at import/build time).

**Takeaway for R1.2**: the acceptance headline reproduces outside CI — the univariate-atom cutover has its target win available to measure against. The earlier CI-container non-repro appears environment-specific, not a code issue.

_(Note: the `AMP: omitting constraint … Cannot linearize log(-2 + x)` lines print on all three runs, including the certifying ones, so they're pre-existing/expected noise rather than a regression — flagging in case they're worth silencing.)_

🤖 Confirmed via Claude Code

jkitchin:
⚠️ **One item needs your local adjudication (you have the `minlplib.solu` oracle; the CI/dev container does not).**

The new serial `claim_boundary` job surfaced a pre-existing discrepancy in `test_power_certification::test_unguarded_power_instances_certify[st_e11]` (it fails on `main` too, in isolation — not a claim-collision, so out of scope for this PR). I've marked it `xfail(strict=False)` so it doesn't gate the order-mask job, but it's worth a look:

- discopt returns **status=optimal, objective=189.31163, bound=189.31163, gap_certified=True** — internally consistent (bound == objective).
- The test reference is **189.3292** (Δ ≈ 0.018).

Two possibilities, and I can't tell which without the oracle:
1. **The reference is stale** and discopt is correct (its certified value 189.3116 is the true optimum) — then the fix is updating the reference in the test.
2. **discopt is wrong** — it's certifying an infeasible incumbent / an invalid bound at 189.3116 — which would be a genuine soundness bug (unrelated to #632, but serious).

Could you check st_e11 against `minlplib.solu` locally? If (1), I'll update the reference and drop the xfail; if (2), that's a separate correctness issue to file. The two `test_lr2_nvs09_cert` xfails are different — just the H-UNI-cert-within-40s budget being perf/env-dependent in this container (the R0.5 watch item); they should xpass on a fast enough runner.

---

jkitchin:
## Correctness review — `python/discopt/_jax/canonical_expr.py`

Careful, well-documented module; the load-bearing property (semantic faithfulness via `reconstruct` + opaque fallback) is sound in design, and being library-only means nothing here can touch a certificate yet. One **confirmed** bug plus two lower-severity notes for the R1.2 wiring.

### Finding 1 (confirmed) — `canonicalize()` raises to the caller, violating its "never raises" contract

The docstrings state (twice) that unsupported nodes become `opaque` and `canonicalize` *"never raises to the caller"* (module docstring, `UnsupportedCanonicalization`, and `canonicalize`). That invariant is violable. In `_prod`, constant bases are folded eagerly:

```python
# canonical_expr.py, in _prod
coef_acc *= float(base.payload) ** exp
```

`float ** float` is not total over the reals, and `canon()` catches only `UnsupportedCanonicalization` — `ArithmeticError`/`TypeError` escape all the way out of `canonicalize()`. Two reachable inputs, both **reproduced against this branch's code** (`2a86dba9`):

| Input | Result |
|---|---|
| `x / p` with `p` a **zero-valued parameter** (→ `const(0)**-1`) | `ZeroDivisionError: 0.0 cannot be raised to a negative power` |
| `y * (Constant(-2.0) ** 0.5)` (**negative const base, fractional exp** → folds to `complex`, then `float(complex)` in `_sum`) | `TypeError: float() argument must be a string or a real number, not 'complex'` |

Repro:

```python
import discopt.modeling as dm
from discopt.modeling.core import Constant
from discopt._jax.canonical_expr import canonicalize

m = dm.Model(); x = m.continuous("x", lb=1, ub=3); p = m.parameter("p", 0.0)
m.minimize(x / p)
canonicalize(m)   # ZeroDivisionError (expected: opaque node)

m2 = dm.Model(); y = m2.continuous("y", lb=1, ub=3)
m2.minimize(y * (Constant(-2.0) ** Constant(0.5)))
canonicalize(m2)  # TypeError via float(complex) (expected: opaque node)
```

The zero-valued-parameter divisor is the realistic case. Note that `/` by a zero **`Constant`** is already handled (`_canon_binary` raises `UnsupportedCanonicalization`), but the `Parameter`/folded-constant path routes through `_prod` and dies. Once R1.2 wires `canonicalize(model)` into the build, an uncaught exception here aborts the solve instead of falling back to opaque.

**Fix** (small, matches the module's design — route the failure to opaque):

```python
if base.kind == "const":
    try:
        coef_acc *= float(base.payload) ** exp
    except (ArithmeticError, TypeError):
        raise UnsupportedCanonicalization("non-real/undefined constant power")
    continue
```

Regression test: assert `canonicalize` returns an opaque node (not a raise) for `x / p` (`p=0`) and for a negative-const fractional power. I'll push this fix + test as a commit on the branch.

### Finding 2 (minor, for R1.2) — constant-valued nonlinear calls classify as `univariate` atoms with empty support

`sin(2.0)` (a `call` on a constant) has empty variable support but isn't opaque, so `_classify_atom` (`len(support) <= 1`) labels it `univariate`. A "univariate" atom with **zero** variables gives the dominance dispatcher no variable to build a 1-D envelope over. Reconstruct is still faithful (not a soundness bug), but it's a latent trap — either constant-fold pure-constant calls or assert `len(support) == 1` in the univariate owner when R1.2 threads the dispatch.

### Finding 3 (minor / latent) — `_memo` is `id(expr)`-keyed and doesn't pin the expr

`canon` caches by `id(expr)`, and the cache is shared into `cnode_of` for freshly-distributed trees. The dict holds `int → CNode`, so it never keeps the source `Expression` alive; a GC'd expr whose id is later reused by a *different* expr would return a stale `CNode`. Within a single `canonicalize()` walk this can't happen (the model pins all subtrees), so it's safe today — but it's the same `id()`-fragility class the module is replacing, now confined to the memo layer. Worth a comment that the memo is only valid for live, pinned trees.

### Checked and correct

- Sum normalization: nested-sum flattening with coefficient distribution, constant folding, like-term merge, zero-drop, single-child collapse — all correct and deterministic.
- Prod/reciprocal: keeping `x·x⁻¹` uncollapsed (positive-exponent-only merge) is right — it preserves the original's value/singularity at `x=0` rather than inventing `1`; reconstruct mirrors it.
- Interning keys are per-kind-tagged nested-primitive tuples — no cross-kind collisions, content-addressed, order-independent.
- Atomizer: the single-variable-sum-is-one-atom rule with the `has_opaque` exception correctly handles nvs09's `(ln(x−2))²+(ln(10−x))²` and stops opaque nodes hiding inside a univariate atom.

The equivalence suite passing (106 tests, ≤1e-7 over the corpus) is consistent with all of this — the gap is only that Finding 1's inputs (zero-param division, negative-const fractional powers) aren't in the generated or corpus test set.

🤖 Reviewed via Claude Code

jkitchin:
Pushed the **Finding 1** fix as `7881d447` (rebased onto the current head `24ecd7ae`).

- `_prod`'s constant-power fold now catches `ArithmeticError` and rejects a `complex` result, raising `UnsupportedCanonicalization` → the node becomes `opaque` instead of the exception escaping `canonicalize()`.
- Two regression tests (`test_refusal_zero_parameter_divisor_is_opaque`, `test_refusal_negative_constant_fractional_power_is_opaque`) — verified **fail before / pass after**.
- `test_canonical_expr.py`: 108 passed / 16 skipped. ruff + ruff-format + mypy clean (pre-commit hooks passed). Library-only path, not wired into the build → relaxation fingerprint unaffected.

One heads-up unrelated to this fix: running `test_claim_baseline_neutral.py` in my environment shows `contvar` and `tanksize` drifting vs the committed `claim-baseline.jsonl` (**identical shapes**, only the numeric fingerprint differs). It reproduces identically on the pristine branch head with my changes stashed, so it's **not** from this commit — it's the reused/rebuilt `_rust` binary producing last-digit-different in-house-simplex root LP bounds (exactly the degenerate-basis sensitivity noted in R0.3). Flagging in case the committed baseline needs regenerating on the canonical CI runner, or the fingerprint wants a tolerance on the root-LP-bound component.

jkitchin:
Thanks — kept your Finding 1 fix (`7881d44`) as the base and addressed the rest in `0007251`:

**Finding 2** (constant-valued calls as empty-support "univariate" atoms) — went with your first option, constant-folding pure-constant intrinsic calls (`sin(2.0) → const`) via a numpy table with a finite-real guard, so `sin(2.0)·z` folds to `const·z` = affine (no spurious atom). That surfaced a latent `_prod` gap: a single-factor product with a scalar coef was building `prod(z^1)` instead of collapsing to `coef·z`, which would have marked the affine result nonlinear — fixed the single-factor collapse (`b^1→b`, `b^e→pow`, coef via the surrounding sum). Kept a belt-and-suspenders atomizer guard that skips non-opaque zero-support nodes. Folded values match the jnp evaluator to <1e-12; 5 new regression tests.

**Finding 3** — added the note that `_memo` is `id()`-keyed and valid only for live, pinned trees.

**Your neutrality heads-up** (`contvar`/`tanksize`) — you're right, and it's more fundamental than a tolerance knob: the exact-hash equality against a *committed* baseline was the wrong cross-environment gate, because the in-house FBBT/parse path isn't bit-reproducible across Rust builds. Environment-independent byte-identity is already guaranteed by the in-process #630 test (OFF-vs-code-absent in one process), so I re-scoped `test_claim_baseline_neutral` to assert relaxation **shape** stability (row/col/integer counts — what a claim change actually moves, and stable across builds) and surface fingerprint drift as an informational count. That should stop `contvar`/`tanksize` reddening the serial job on the CI runner while keeping a meaningful structural-neutrality check. (Note: the drift is in the matrix bytes, not the `root_lp_bound` field — the fingerprint doesn't include the bound.)

Full suite: `test_canonical_expr.py` 113 passed / 16 skipped; shape-based neutrality green; ruff + format + mypy clean. Pushed as a fast-forward on your commit.

---

jkitchin:
## Follow-up review — commits `0007251cc` + `36e89bccd` (head `36e89bcc`)

Reviewed the two commits that responded to the earlier findings. **Verdict: in good shape — no code defects found.** One framing caveat on the neutrality re-scoping worth a conscious sign-off.

### `0007251cc` — Findings 2 & 3 + neutrality re-scoping

**Constant-folding (Finding 2) — correct and faithful.** `_fold_const_call` evaluates intrinsics under `np.errstate(all="ignore")`, catches `ArithmeticError/ValueError/TypeError`, and **returns `None` unless the result is a finite real** — so it never injects a `nan`/`complex` into the canonical form (`log(-1)` stays an unfolded call; covered by `test_out_of_domain_constant_call_not_folded_to_nan`). Same discipline as the Finding-1 fix, applied to the call path. Folding a constant subexpression to a literal is exactly right — it's a constant on either path, and the numpy/jnp analogues agree to ~1e-15, well inside the 1e-7 equivalence tolerance.

**The `_prod` single-factor collapse it uncovered — correct, and the subtle one.** The old code only collapsed `prod([b], e=1, coef=1)`; a single factor with `e≠1` or `coef≠1` built a `prod(b^e)` node. Once folding reduces `sin(2.0)·z` to a single `z` factor, that stale `prod(z^1)` would mark an affine result nonlinear. The new collapse (`b^1→b`, `b^e→pow(b,e)`, scalar via a surrounding sum) is semantically equivalent (reconstruct yields the same value) and **doesn't over-collapse genuine monomials** — `test_genuine_monomials_still_atoms_after_collapse` confirms `x·x·y + 3·x·x` still yields exactly one `product` + one `univariate` atom. Traced the CSE/classification paths: single-factor prods always had support ≤ 1 → already the univariate branch, so the multivariable prod/ratio classifier never regressed.

**Atomizer zero-support guard + Finding 3 doc — both correct.** Skipping non-opaque zero-support nodes is sound: the constant they carry is picked up by the linear/evaluator path, not lost.

### The one thing to sign off on consciously: the neutrality re-scoping

The change from exact-fingerprint to **shape-only** (rows/cols/int-cols), with fingerprint mismatch demoted to informational, is **defensible and I'd approve it** — but the framing slightly overstates the backstop, so naming it to make it a decision, not an accident:

- **Shape gate still runs in CI.** Marking it `slow` drops it from `python-fast`, but the serial job's bare `-m claim_boundary` overrides addopts and includes it (confirmed by the green 21m serial run). ✓
- **What #630's in-process test actually guards.** `test_lr2_offneutral_relaxation.py` compares `fp_real_off` vs `fp_code_absent` — **both built from current code**. That rigorously proves the *LR-2 collectors are inert when off* (full corpus, in-process, cross-build-immune). It does **not** compare against prior main / a frozen reference — a uniform coefficient change from a refactor would move both fingerprints identically and still pass.
- **Net.** The committed-baseline fingerprint was the only *frozen-reference coefficient* gate, and it's genuinely unusable cross-environment (the `contvar`/`tanksize` drift reproduces on a pristine tree). Demoting it loses no *working* gate, and **this PR provably can't change a coefficient** — the sole build-path change is the three `note_defer()` no-ops; everything else is unwired. Future bound-changing cutovers get their own differential gates per CLAUDE.md §5.

**Residual (not a blocker):** there's now no standing frozen-reference gate for a *same-shape coefficient* drift on the build path. Given the in-house FBBT path's last-digit non-determinism across builds, an exact committed hash may simply be unachievable — but if you want one, the fix is a **tolerance-based** coefficient comparison against the baseline (compare `_A_ub`/`_c` within ~1e-9) rather than exact-hash or shape-only.

### `36e89bccd` — CI change, sound

Moving the three full-corpus build tests out of `python-fast` (st_e36's ~289s build blew the 120s timeout) into the serial job with `--timeout=600` is correct test-org hygiene; the gate still runs where it belongs.

---

**Bottom line:** all earlier findings are properly closed, the new code is correct with good regression coverage, CI is fully green / mergeable. The only judgment call is the neutrality re-scoping, which I'd accept — flagging only that its justification leans on #630 covering slightly more than it strictly does.

🤖 Reviewed via Claude Code

jkitchin:
Agreed on all of it, and you're right that my framing overstated the backstop — #630 proves the collectors are *inert when off*, not frozen-reference neutrality, so a uniform coefficient refactor would move both fingerprints together and pass. Corrected the neutrality-test docstring to say exactly that (`403e4c4`): this PR needs no frozen-reference coefficient gate because its only build-path change is the three `note_defer()` no-ops (everything else unwired → provably can't move a coefficient).

Your tolerance-based suggestion is exactly what R1.2 needs, so I'm building it there rather than retrofitting R0. Concretely, R1.2's differential gate will be:
- **in-container, exact**: snapshot per-instance fingerprints before/after the cutover in the same process (same Rust binary → no drift), so unchanged-dispatch instances must be bit-identical and changed ones (H-UNI) are flagged for attribution;
- **cross-build CI, tolerant**: a `_A_ub`/`_c`/`_b_ub` comparison within ~1e-9 relative against a committed reference, which is robust to the FBBT last-digit non-determinism that killed the exact hash.

Starting R1.2 now (box-aware dominance dispatch → wire the univariate collectors → delete the univariate defer clauses), keeping it all on this branch per your workflow and pausing at a clean checkpoint for your local corpus/adversarial/nvs09 sweeps before going further. Thanks for the careful reviews.

---

jkitchin:
## Review — the R1.2 cutover is live on the default path (contradicts the "byte-identical foundation" framing)

I re-reviewed the full branch diff. The scaffolding (fingerprint, claim auditor, env-leak guard, canonical DAG) is solid, but the branch does **not** match the PR body's central claim — *"lands everything that changes no solver math … the behavior-changing cutover (R1.2 wiring) is deliberately not in this PR … OFF is byte-identical to prior main."* The R1.2 cutover is wired live into `build_milp_relaxation`. Two serious findings follow from that.

### 1. `[correctness / soundness]` H-UNI 1-D hull cutover is on the *default* path — not flag-gated, not byte-identical
`python/discopt/_jax/milp_relaxation.py:3816`, `:4125`, `:4196`

`_collect_composite_univariate_relaxations` (called unconditionally at `:6099`) now dispatches through `_univariate_dispatch`, which calls `_should_claim_composite(..., allow_general=True)` with `allow_general` **hardcoded** — the `DISCOPT_UNIVARIATE_ENVELOPE` gate is gone from this path (docstring at `:4118` says as much: *"the 1-D hull (rule 3) is now always on, no longer gated"*). On the default certificate path this means:

- Single-var additive composites and `(non-affine)**2` nodes previously **not** claimed (old code: `if curvature is None: return`) are now claimed as EXACT or HULL, and `maybe_add` blocks descent (`:4200`). So `node_count` and the certified `objective` change — a **bound-changing** edit shipped with no differential / feasible-point gate, which the PR body itself says is required and deferred to a later PR. This violates the CLAUDE.md verification regime (bound-neutral changes must assert `node_count`/`objective` exactly unchanged).
- Rule 3 (HULL) emits a **sampled** envelope (finding 2) into the dual bound by default.

**Ask:** either re-gate — `_univariate_dispatch(..., allow_general=_univariate_envelope_enabled())`, default-off, to match the "byte-identical foundation" framing — or, if the cutover is intentional, fix the PR description and run the full-corpus differential + adversarial + feasible-point sweeps first.

### 2. `[correctness / soundness]` The 1-D hull "proven" envelope is only validated at grid points
`python/discopt/_jax/univariate_hull.py:175-181`

`_hull_lines` shifts each facet by `slack = max(hull_vals − sign·ff)` measured on a finite 40 010-point grid. Between grid nodes a continuous but sharply-curved `f` can exceed the shifted facet — there is no Lipschitz / derivative-bound certificate, only denser sampling. As a **default-path** envelope (per finding 1) this can cut a feasible point or yield an invalid dual bound / false `infeasible`. Acceptable behind a default-off flag; not on the default certificate path (dev philosophy §3: "no sampled/assumed convexity").

### 3. `[correctness — latent]` `CanonicalDAG.cnode_of` opaque-token collision
`python/discopt/_jax/canonical_expr.py:490-493`

`cnode_of` builds a fresh `_Canonicalizer(self.model)` and copies `_intern`/`_memo` but **not** `_opaque_counter` (reset at `:184`). Re-canonicalizing a freshly-distributed subtree that produces an `opaque` node requests token `0,1,…`, whose key `("opaque", 0)` already exists in the shared `_intern`, so `_mk` (`:189`) returns the previously-interned opaque node wrapping a **different** original expression → `reconstruct` yields the wrong subexpression and CSE identity breaks. Latent today (library-only), but `cnode_of` is exactly the primitive the R1.2 `distribute_products` rebuild will call. Fix: share the counter, or seed it past the max existing opaque token.

### 4. `[efficiency]` H-LOG duplicate `ln x` rows for a factor shared across products
`python/discopt/_jax/milp_relaxation.py:8291`

`_z_col_of` interns one `z` column per distinct factor across products, but the emission loop re-emits the full ln tangent+secant band for that `zc` once per product (identical rows, same box). Sound but redundant; flag-gated (`DISCOPT_LOG_MONOMIAL`, default off) so low priority.

### Checks that passed review
- The H-LOG log-space construction (`z=ln x` concave band, exact `s=Σaᵢzᵢ`, `t=coef·exp(s)` convex band, incl. the `coef<0` sense flips at `:8340`/`:8357`) is a sound outer relaxation; strict-positivity (`lb>1e-9`, no epsilon shift) is enforced.
- `claim_audit.py`, `_guard_discopt_env_leaks`, and the fingerprint primitive are read-only w.r.t. the build and look correct.

**Bottom line:** findings 1+2 — this branch ships the bound-changing R1.2 cutover, including a grid-sampled envelope, onto the default solver path, contrary to the PR description and without the soundness gates the plan requires. Recommend holding until the flag gating is restored (or the description corrected and the differential / feasible-point / full-corpus `incorrect_count=0` sweeps run and reported).

🤖 Review by [Claude Code](https://claude.com/claude-code)

jkitchin:
Review is correct on all three substantive findings — thanks for catching them. Pushed `c940f3d` + `2b41ba1`:

**Findings 1 + 2 (grid-sampled hull on the default path) — re-gated default-off.** You're right that the R1.2 wiring graduated the grid-sampled 1-D hull onto the default certificate path: `univariate_hull.py` measures its soundness `slack` only on the 40,010-point fine grid, so facets are proven *at* the nodes, not *between* them — no Lipschitz/interval remainder certificate. That is exactly the "sampled/assumed convexity" §3 forbids on the certificate path, and the change was bound-changing (my descent-blocking `return True` vs main's byte-identity-preserving `return False`) shipped without the differential/feasible-point gate — contrary to the PR's own framing.

Per your recommended path, I restored the wiring-affected files to the `f2fadfe` foundation state: the collector is gated by `_univariate_envelope_enabled()` again (default-off), and the default path is **byte-identical to `main`** — re-verified the three instances the wiring had changed (contvar, hda, nvs09) now fingerprint `== main`. The dispatch *decision* infrastructure (`_univariate_dispatch` + `test_univariate_dispatch.py`) is kept as decision-only/byte-neutral — it's the foundation the graduation wires onto. PR title/body updated to match.

The actual graduation now blocks on making the envelope **rigorous** (interval/Lipschitz remainder per subinterval so facets are proven everywhere) → then the full-corpus differential + feasible-point + `incorrect_count=0` sweeps + nvs09/global50 cert on your local host → then default-ON. That's its own follow-on PR, as the (corrected) body now states.

**Finding 3 (`cnode_of` opaque-token collision) — fixed.** Confirmed the latent bug: the fresh `_Canonicalizer` in `cnode_of` reset its opaque counter to 0, so a re-canonicalized opaque subtree collided on `("opaque", 0)` and got back the stale node wrapping a different expression. Seeded the counter past every interned opaque token. Added `test_cnode_of_fresh_opaque_subtree_no_token_collision`, which reproduces the collision without the fix (`cnode_of(b)` returns `a`'s node) and passes with it — important before `distribute_products` starts calling `cnode_of`.

**Finding 4 (H-LOG duplicate `ln x` rows) — agreed, deferred.** Redundant-but-sound, flag-gated (`DISCOPT_LOG_MONOMIAL`, default off); I'll fold the one-`z`-band-per-shared-factor dedup into the H-LOG follow-on rather than this foundation PR. And thanks for the confirmation on the H-LOG log-space soundness + the audit/env-guard/fingerprint scaffolding.

---

jkitchin:
## Follow-up: local verification confirms findings 1+2 — the R1.2 cutover is live on the default path

I built this branch locally (`maturin develop --release`, clean compile at HEAD `34407037`), pointed the venv at it, and ran the byte-identity gates plus a direct probe. The static findings from my earlier review are now confirmed empirically three independent ways, including **a failing test on the branch itself**.

### The clincher: the branch's own regression suite is red

`pytest -m claim_boundary -n0` → **1 failed, 438 passed**. The failure:

```
FAILED python/tests/test_lr2_nvs09_cert.py::test_nvs09_does_not_certify_without_flags
AssertionError: unexpectedly certified without flags:
  {'status': 'optimal', 'objective': -43.13433691803531,
   'bound': -43.13433699815766, 'nodes': 215}
```

That test asserts *"default (flags OFF) leaves a real dual gap"* — the pre-cutover behavior. It now fails because **nvs09 certifies to optimal on the default path**. The R1.2 commit made the hull default-ON but left the "no-flags shouldn't certify" guard in place, so the suite is internally contradictory.

### Three-way confirmation that the default relaxation changed vs `main`

| Evidence | Result |
|---|---|
| Static read | `_should_claim_composite(..., allow_general=True)` is hardcoded (`milp_relaxation.py:3816`); `_collect_composite_univariate_relaxations` is called unconditionally (`:6099`) |
| Commit history | `claim-baseline.jsonl` was **regenerated on the cutover commit** `ccc2d3d5 "R1.2 (part 2) — wire the univariate dominance dispatch, hull default-ON"` |
| Owner probe (all `DISCOPT_*` OFF) | Sampled HULL owner fires **26× across 3 real instances**: `nvs09` (19), `contvar` (4), `hda` (3). Under `main` these hit `if curvature is None: return` and fell back. |
| Branch's own test | `test_nvs09_does_not_certify_without_flags` **FAILS** — default path now certifies |

### Why the green gates don't contradict this

- `test_claim_baseline_neutral` passes but only checks **shape** (row/col counts) against a baseline **regenerated on the cutover commit** — it measures self-consistency, not parity with `main`.
- `test_lr2_offneutral_relaxation` (#630) passes but only exercises the **still-flag-gated** aliased/log paths (`DISCOPT_UNIVARIATE_ENVELOPE`/`DISCOPT_LOG_MONOMIAL`); it does **not** cover the composite-collector HULL, which is no longer flag-gated. Note `contvar`/`hda` appear in *both* that passing test and the HULL-fires-by-default probe — that's the coverage gap, not a contradiction.

### On correctness

The nvs09 default-path certificate is **sound** in isolation: `bound (-43.13433699816) ≤ incumbent (-43.13433691804)`, within 1e-6 of the optimum. So for nvs09 this is a correctness-preserving tightening, not an incorrect result. **But** it is produced by the **grid-sampled** hull (finding 2 — the "proven" band is only validated at 40 010 grid points, no Lipschitz/derivative certificate) and shipped **unflagged on the default certificate path**. Per CLAUDE.md §5 that requires a differential + feasible-point soundness gate before default-on; the full-corpus `incorrect_count=0` sweep has not been run here.

### Recommendation (unchanged, now with evidence)

Either:
1. **Re-gate to match the PR's "byte-identical foundation" framing** — `_univariate_dispatch(..., allow_general=_univariate_envelope_enabled())`, default-off — which also makes `test_nvs09_does_not_certify_without_flags` pass again; or
2. **Own the cutover**: fix the PR description, delete/invert the now-false nvs09 guard, and land the full-corpus differential + feasible-point + `incorrect_count=0` sweeps this PR's own body says are prerequisites.

As it stands the PR ships a bound-changing, grid-sampled default-path envelope with a red serial suite.

*Verified locally on branch HEAD `34407037`; corpus = the 62-instance `python/tests/data/minlplib_nl/`.*

🤖 Review by [Claude Code](https://claude.com/claude-code)

jkitchin:
Your local verification was on HEAD `34407037` (commit `3440703`) — that's the pre-revert state. I pushed `c940f3d` **after** that, which re-gates the hull default-off (your recommendation #1). Everything your probe found on `3440703` is resolved on current HEAD:

**The clincher test passes on `c940f3d`:**
```
python/tests/test_lr2_nvs09_cert.py::test_nvs09_does_not_certify_without_flags
1 passed in 41.72s
```
It's `slow`-marked, so it runs in the **serial** claim-boundary job — which is **green on `c940f3d`** (run 29206458230, job 86686628986, `success` @ 20:09). The default path no longer certifies nvs09; the guard is satisfied without needing to invert it.

**Your three-way confirmation, re-checked on `c940f3d`:**

| Evidence | On `3440703` (your build) | On `c940f3d` (current HEAD) |
|---|---|---|
| HULL fires by default | 26× (nvs09 19, contvar 4, hda 3) | **0×** — nvs09/contvar/hda fingerprints re-verified `== main` |
| `test_nvs09_does_not_certify_without_flags` | FAILED | **passed** (41.7s) |
| `claim-baseline.jsonl` | regenerated on cutover | reverted to the pre-wiring baseline |
| serial suite (`-m claim_boundary -n0`) | 1 failed | **green** |

So this is your option 1, done: `_collect_composite_univariate_relaxations` is gated by `_univariate_envelope_enabled()` again (default-off), default path byte-identical to `main`, and the now-false guard passes on its own rather than being deleted. The dispatch *decision* infra (`_univariate_dispatch`) stays as decision-only/byte-neutral (0 rows in the build) — the foundation the graduation wires onto.

Your finding 2 (the grid-sampled envelope needs a rigorous interval/Lipschitz remainder, not just denser sampling, before it can be default-on) is exactly the blocker recorded for the graduation follow-on, alongside the full-corpus differential + feasible-point + `incorrect_count=0` sweeps. Nothing grid-sampled is on the default certificate path as of `c940f3d`.

---

jkitchin:
## Full correctness gate — `incorrect_count = 0` (the cutover is sound on this panel)

Ran the full `global50` suite locally on branch HEAD `34407037` (default path, all `DISCOPT_*` flags OFF), then the `cert0` gate. **Fair update in the PR's favor on the dimension that matters most:**

| Criterion | Target | Actual | Status |
|---|---|---|---|
| `zero_incorrect` (incorrect_count) | ≤ 0 | **0** | ✅ **PASS** |
| `root_gap_coverage` | ≥ 0.9 | 0.8707 | 🔴 FAIL (observability, not correctness) |

**116 instances, 83 solved/proved, 0 incorrect, SGM 0.83s, median gap 0.00%.** `bound ≤ incumbent` held everywhere — no false optimal, no false infeasible. The grid-sampled hull did **not** produce a single unsound bound on this panel.

Hull-affected instances (default path):
- `nvs09` → **optimal**, `bound -43.13433699816 ≤ incumbent -43.13433691804`, 215 nodes, 20.4s — sound.
- `contvar` → feasible, gap open at the 60s cap (unsolved, not incorrect).
- `hda` → time_limit, no incumbent (unsolved, not incorrect).

The `cert0` gate fails **only** on `root_gap_coverage` (87% vs 90% of rows populated) — an observability metric, not a correctness one.

### What this does and does not change about the review

**Resolves:** the immediate correctness risk in finding 2 (a sampled default-path envelope emitting an unsound bound) **did not materialize** on this 116-instance panel. The PR is *sound* here, not *incorrect*.

**Still open (unchanged by `incorrect_count=0`):**
1. **Not byte-identical to main** (finding 1) — the cutover is live on the default path; the PR body's "byte-identical / R1.2 not in this PR" is false. A clean correctness panel doesn't restore byte-identity.
2. **Finding 2 is unrefuted in principle** — one clean panel is not the rigorous guarantee. There is still no Lipschitz/interval certificate behind the 40 010-point sampled shift, and CLAUDE.md §5 requires the **feasible-point soundness sweep + consecutive nightly greens** before a sampled envelope goes default-on. Neither has been run; this panel is a single data point, not that gate.
3. **Red serial suite** — `test_nvs09_does_not_certify_without_flags` still fails (the stale no-flags guard).
4. **Performance cost** — the full run took **~30 min vs ~13 min on main's committed global50 baseline**; the per-node hull sampling (two batched JAX evals + hull construction per composite node) pushes `contvar`/`hda`/others into the 60s cap. SGM (0.83s) is unaffected, so the cost is concentrated on the hard instances.

**Net:** correctness holds on this panel, but the PR still ships a non-byte-identical, grid-sampled, default-ON envelope with a red suite and a ~2× wall-time regression on hard instances — none of which the "byte-identical foundation" framing accounts for. Recommendation stands: either re-gate default-off to match the framing, or own the cutover (fix the description, fix/invert the nvs09 guard, and land the feasible-point + full-corpus soundness sweeps this PR's own body names as prerequisites).

*Reproduced locally: `python discopt_benchmarks/run_benchmarks.py --suite global50 --solvers discopt` then `--gate cert0`, branch HEAD `34407037`, flags OFF.*

🤖 Review by [Claude Code](https://claude.com/claude-code)

jkitchin:
## Review + global50 gate on the uniform-engine cutover (HEAD `0a8a7885`)

Re-reviewed and re-verified locally after the big rework. Since `34407037` the branch replaced the federation with a **uniform factorable relaxation engine** (`uniform_relax.py`, +1520) now routed as the **default** relaxation (`milp_relaxation.py:4282`), deleted the federation body (−4774), and added the **P3 per-node probing** branch-and-reduce workstream (`probing.rs`, `in_tree_presolve.rs`). Built at `0a8a7885` (`maturin develop --release`) and ran the full `global50` panel, flags off.

### Correctness — clean ✅
`cert0` gate: **`zero_incorrect` = 0 ✅ PASS** across **116 instances**. No false optimal, no false infeasible; `bound ≤ incumbent` held everywhere. Status: **85 optimal/proved · 20 feasible · 11 time-limit** — every non-proved case is unsolved-within-budget, never wrong.

(The gate's other criterion, `root_gap_coverage` 0.862 < 0.9, is an *observability* metric — root_gap populated on 86% of rows — not a correctness one.)

### Performance — roughly neutral, better at the hard end
| Metric | Uniform engine `0a8a7885` | Prior HEAD `34407037` |
|---|---|---|
| Proved | **85/116** | 83/116 (+2) |
| Incorrect | **0** | 0 |
| Total wall | 1868 s (~31 min) | 1782 s (~30 min) |
| SGM | 0.96 s | 0.83 s |
| Time-capped (≥60 s) | 25 | ~25 |

Highlights: `nvs09` → optimal in **19 nodes / 18.5 s** (was 215 nodes on the old default path — the composite-convex lift + log-space signomial band tighten the root a lot); `st_miqp2` → `obj=2.0` in 9 nodes (the `(c·xᵢ)·xᵢ → c·xᵢ²` product→monomial aggregation fixes the old wide-box McCormick looseness). SGM ticked up slightly (0.83 → 0.96 s) on the typical instance — consistent with the per-node engine rebuild + probing overhead and the **deferred product-side tightness-parity polish** the PR itself flags.

### Soundness read of the new code (by inspection)
- **Uniform engine envelopes** — min/max facets, `abs` hull, McCormick (the 4 hull inequalities), secant+tangent 1-D envelopes on definite-curvature boxes, the sign-straddling odd-power two-piece hull, and the log-space signomial band are each valid outer inequalities at the lifted point; every builder falls to the sound interval floor when it can't tighten. The objective-bound-validity guard (refusing the LP value when a cost column is unbounded on its cost side and unconstrained) correctly mirrors the federation's `objective_bound_valid=False`. No unsound row construction found.
- **P3 probing** (`probe_node_bounds`) — bounds contract **only** on *proven* infeasibility (`is_empty_beyond(FEAS_TOL)`, eps-noise ignored); binary forcing intersects the surviving branch, both-feasible intersects the interval hull of the two branches, integer endpoint-peeling steps by 1, all subset-of-input. Cutoff-aware reduction is standard and sound inside B&B. No soundness bug found.

### Two things to fix (non-correctness, but should not merge as-is)
1. **`uniform_relax.py` module docstring is now false.** It states *"nothing here is wired into the default build, and no new `DISCOPT_*` runtime flag is added. It is exercised only by tests and the validation harness."* But `build_milp_relaxation` **unconditionally** delegates to `build_uniform_relaxation` (`milp_relaxation.py:4282`) — this module *is* the default certificate path. The docstring should say so; it currently misleads any reviewer about what is load-bearing.
2. **PR title/body are badly stale.** Title still reads `#632 foundation — canonical AVM normal form (R0 + R1.1 + R1.2 entry, all byte-identical)`. This is no longer a byte-identical foundation — it is a full relaxation-engine replacement + a B&B probing workstream. Please retitle/rewrite so the scope (and the `incorrect_count=0`-but-not-byte-identical reality) is accurate.

Also note `build_milp_relaxation` now silently ignores `terms`, `disc_state`, `incumbent`, `oa_cuts`, `convhull_*`, `rlt_level1` (returns the delegate immediately). Sound, but any caller relying on disc_state piecewise-McCormick refinement or feeding back OA cuts gets a no-op — worth a deprecation note or asserting they're unused.

**Net:** the cutover is **sound on this panel** (0 incorrect), proves 2 more instances than the prior HEAD with a much better tree on cases like nvs09, at roughly flat wall time. Correctness gate is green; only the stale docs/title and the deferred tightness polish remain.

*Reproduced locally: build `0a8a7885`, `run_benchmarks.py --suite global50 --solvers discopt` then `--gate cert0`, flags off.*

🤖 Review by [Claude Code](https://claude.com/claude-code)

jkitchin:
Thanks for the global50 re-verification — `incorrect_count = 0` across 116 instances is the gate that matters, and it's the one I couldn't run in-container. Addressed all three doc/accuracy items in `e7c4d12`:

1. **`uniform_relax.py` module docstring** — rewrote the stale "nothing here is wired into the default build … exercised only by tests" claim to state it **is** the default relaxation (`build_milp_relaxation` delegates unconditionally), the load-bearing certificate path, cert0-green.
2. **`build_milp_relaxation` ignored params** — added a `.. note::` documenting that the default path delegates to the engine and that `terms`, `disc_state`, `incumbent`, `oa_cuts`, `convhull_ebd*`, `superposition`, `rlt_level1` are currently ignored (sound; kept for signature compat), so a caller relying on `disc_state` refinement or `oa_cuts` feedback knows it's a no-op on that input. Left as a docstring note rather than a hard `assert` so passing a stale arg degrades to a sound no-op instead of raising.
3. **PR title/body** — retitled to the real scope (uniform engine replaces the federation + P3 probing) and rewrote the body to state plainly this is a bound-changing engine replacement, **not** byte-identical, with correctness carried by construction + the cert0 gate.

The deferred product-side tightness parity, the residual heavy-math classes, and the one `contvar` coverage regression (larger LP → simplex iteration budget → sound `None`, not a false bound) are all itemized in the §10 ledger and the updated PR body as follow-ons.

---

jkitchin:
## Performance note: the global50 time is per-node JAX/Python, not startup — Rust LP is ~0%

Profiled the `0a8a7885` global50 run to explain why wall time is ~flat despite the tighter engine. **It is not a startup cost, and a warm server would not help.**

### Startup is negligible
- Cold `import discopt` = **0.15 s**.
- The harness runs discopt **in-process** (daemon thread; subprocess is only for external solvers), and fast instances finish <1 s with no per-instance floor.
- This run used **no persistent server** — irrelevant here, since there's no startup to amortize.

### The cost is per-node relaxation + bound-tightening, dominated by JAX/Python
Wall-weighted across the 116-instance panel:

| Component | Share of wall |
|---|---|
| JAX (XLA tracing/eval) | **61%** |
| Python | **39%** |
| Rust LP simplex | **~0%** |

The smoking gun is low-node / high-`root_time` instances:

| instance | wall | root_time | nodes | nlp_solves |
|---|---|---|---|---|
| hda | 61.6 s | **52.5 s** | **7** | 0 |
| heatexch_gen3 | 64.8 s | **53.6 s** | **7** | 0 |
| contvar | 61.0 s | 37.6 s | 15 | 0 |

52 s on **7 nodes** with **0 NLP solves** — not tree search, not the LP. A single root build+solve of hda is ~3.6 s (nvs09: 0.04 s build, ms-scale LP). The time is the **root bound-tightening machinery**: per-node OBBT (`_PER_NODE_OBBT_ROUNDS=3`, O(n) LP probes) + repeated relaxation rebuilds. Instrumented nvs09: `build_milp_relaxation` called **22× for 19 nodes**, each running the uniform engine's Python/JAX pipeline (`canonicalize → reconstruct → evaluate_interval` in numpy + `compile_expression`/`jax.grad` for composite lifts).

### Why it's not faster than the federation
The cutover moved relaxation construction into a **uniform Python/JAX pipeline that runs on every node and isn't cached across nodes**, while the fast Rust LP sits idle (0%). The engine's genuinely tighter bounds — **nvs09: 215 → 19 nodes** — are offset by higher per-node Python/JAX cost, so net wall is ~flat. This is the "deferred polish" the PR already flags; it's recoverable.

### Actionable levers (all cut the 61% JAX / 39% Python)
1. **Cache JAX across nodes.** `_try_convex_lift` calls `compile_expression` + `jax.grad` per node; the box changes per node but the *function of x* does not. Memoize compiled `value_fn`/`grad_fn` by CNode identity (the pattern the old H-UNI `batch_cache` used).
2. **`bounds()` → `reconstruct` + `evaluate_interval` runs per node in numpy** — memoize by CNode, or push interval enclosure into Rust.
3. **OBBT dominates `root_time`** (3 rounds × O(n) LP probes); each probe now pays the engine's per-build Python cost. Warm-start probes / cache the relaxation across probes.

Net: correctness is sound and bounds are tighter (fewer nodes); the remaining ~60% JAX overhead is per-node XLA/Python work inside each solve — a caching/vectorization pass, not an algorithmic problem.

*Measured locally on `0a8a7885`, flags off: cold-import timing, per-instance `root_time`/`jax_time_fraction` from the global50 result JSON, and a `build_milp_relaxation` call-count instrument on nvs09.*

🤖 Review by [Claude Code](https://claude.com/claude-code)

jkitchin:
Your profile matches the in-container one exactly (38 ms/node on nvs09, ~15 ms of it re-proving interval-Hessian curvature certificates per node; the incremental fast path never engages under engine rows, so every node — and every OBBT probe — pays a cold Python/JAX build while the Rust LP idles).

Sequenced the fix into `docs/dev/engine-performance-plan.md` (`f17e2e8`) as EP0–EP6, one self-contained item per fresh context, your three levers mapped directly:

- **EP1** — per-model analysis cache: canonical DAG, reconstructed exprs, DCP verdicts, compiled `value_fn`/`grad_fn` by CNode (your lever 1), plus *monotone* interval-Hessian certificate inheritance (convex on a box ⇒ convex on every sub-box, so a root certificate covers the whole subtree) and a support-restricted enclosure cache (lever 2 — branching on one variable only invalidates nodes that depend on it).
- **EP2** — OBBT probes reuse one relaxation per node box (objective-vector swap + warm-start instead of a rebuild per probe — your lever 3; should also give hda/heatexch_gen3 back their root_time and likely contvar its finite root bound).
- **EP3** — patch-table node path: the engine emits box-dependent row descriptors so the existing incremental validator can engage on engine rows (restores the CC2 mitigation with its row-for-row validation gate).
- EP4a/EP4b separation facet-cache + warm-start/OA-pool, EP5 lazy compiles, EP6 the probing/OBBT default-on tuning (yours to ratify on the nightly gate).

EP1/2/3/4a/5 are bound-neutral and gate on byte-identical fingerprints across the 62-instance corpus + exactly unchanged node counts/objectives — zero certificate exposure. EP4b/EP6 run the differential regime.

---

jkitchin:
## Re-check on `94a9bec8` (EP0–EP6): correctness clean, but `ep3` regressed tightness — 3 lost proofs (bisected)

Rebuilt at HEAD `94a9bec8` and re-ran the full `global50` panel, flags off. Thanks for fixing both doc findings (the `uniform_relax.py` docstring + the PR title). The perf workstream is a **mixed** result.

### Correctness — still clean ✅
`incorrect_count = 0` across 116 instances. Every bound is valid (≤ optimum); no false results.

### Performance — the caching helped, but `ep3` cost 3 proofs
| Metric | pre-EP `0a8a7885` | EP `94a9bec8` |
|---|---|---|
| Proved | 85/116 | **82/116** (−3) |
| Incorrect | 0 | **0** |
| SGM | 0.96 s | **0.68 s** (better) |
| Total wall | 1868 s | 1927 s |
| JAX fraction | 61% | 57% |

- ✅ **EP1 (per-model analysis cache) + EP5 (jit/lazy JAX) are clean wins** — SGM dropped 0.96 → 0.68 s: nvs09 18.5→4.3 s, nvs16 −11 s, st_e36 −7 s. Keep these.
- ❌ **EP3 regressed relaxation tightness**, exploding the tree on 3 instances until they hit the 60 s cap:

| instance | pre-EP | EP HEAD |
|---|---|---|
| st_e04 | optimal, **153** nodes | feasible, **18,873** nodes |
| tspn05 | optimal, **39** nodes | feasible, **3,303** nodes |
| nvs19 | optimal, 349 nodes | feasible, 505 nodes |

### Root cause — bisected to `2c213484` (ep3)
`ep3` = "patch-table node path — box-dependent row refresh instead of cold rebuild". Solving `st_e04` (25 s cap) across the EP commits:

```
0a8a7885 (pre-EP)  optimal   153 nodes   bound=5194.866   (= optimum)
0935555b (ep2)     optimal   153 nodes   bound=5194.866
2c213484 (ep3)     feasible  7511 nodes  bound=4351.933   ← regression starts here
5ed42b8e (ep5)     feasible  7735 nodes  bound=4351.933
94a9bec8 (HEAD)    feasible  7575 nodes  bound=4351.933
```

Same signature on `tspn05` (pre-ep3 optimal 39 nodes / bound 191.26 → ep3 feasible 1347 nodes / bound **189.52**) and `nvs19` (bound −1099.88 → **−1102.01**). In every case the **patch-table box-dependent row refresh yields a weaker bound than a full cold rebuild** — e.g. st_e04's dual bound stalls at 4351.9 instead of reaching the true 5194.9 — so the tree can't close and times out.

The bound stays **valid** (≤ optimum, so `incorrect_count` is still 0), but it is **looser than the cold rebuild it replaced**. That's precisely the bound-changing regression CLAUDE.md §5's differential test targets — *new bound ≥ old bound on a fixed box* — and there's evidently no such per-box assertion guarding the node path yet.

### Recommendation
1. **Fix or revert `ep3`.** The row refresh is missing some per-box re-tightening that the cold rebuild does; until it reproduces cold-rebuild tightness on a fixed box, it should be gated/reverted.
2. **Add the per-box differential bound gate** (`refresh_bound ≥ cold_rebuild_bound` on a sample of boxes) so a tightness regression on the node path fails CI instead of silently costing proofs.
3. **Keep EP1 + EP5** — those are the real, regression-free SGM win.

*Reproduced locally on `94a9bec8`, flags off: full `global50` run + a per-commit `solve(time_limit=25)` bisect on st_e04/tspn05/nvs19 (Python-only commits, no rebuild needed).*

🤖 Review by [Claude Code](https://claude.com/claude-code)

jkitchin:
## Perf deep-dive: where the remaining ~10× vs BARON goes — two root causes, one verified 10-line fix

Context: on the 49 matched head-to-head instances, current PF discopt is ~85× slower than BARON in total wall (366 s vs 4.3 s), median ~7× per instance — while proving *more* instances (45 vs 31). cProfiled the worst both-proved offender (`nvs09`: 53×, 12 s / 19 nodes) on `664520ab` and dug to root cause. It is not Rust LP throughput, not the engine build (EP1 works: 22 builds = 0.96 s), and not startup.

### Cost breakdown (nvs09, 12 s solve, measured)

| Cost center | Time | % |
|---|---|---|
| **1. POUNCE IPM fallback on degenerate multilinear-envelope LPs** | **~5.0 s** | **41%** |
| **2. Interpreted (non-jitted) JAX in `_separate_convex`** | **~2.7 s** | **22%** |
| 3. NLP heuristic solves (518 small POUNCE sub-NLPs) | ~1.4 s | 12% |
| 4. Rust LP (node solves + envelope attempts) | ~0.9 s | 8% |
| 5. Relaxation builds (EP1-cached) | ~1.0 s | 8% |

### Finding 1 — Rust simplex cycles forever on pinned-dimension vertex LPs (41% of solve)

The multilinear separator (`multilinear_separation.py::_solve_envelope`) enumerates all 2ⁿ box vertices **including dimensions branching/FBBT has pinned** (lo = hi). nvs09's terms have arity 5–10 → LPs up to 1024 columns. A pinned dim makes the vertex set contain **exact duplicate columns** and adds a **redundant equality row** → the LP is degenerate and the capped Rust simplex hits `ITERATION_LIMIT` (~45 ms wasted), then falls back to the POUNCE IPM at **0.1–2.8 s per LP**, ~1 per node.

Instrumented trace (every slow envelope call in one nvs09 solve):

```
    m   n  simplex status                 dual_ok  simplex_t  total_t
 1024  10  SolveStatus.ITERATION_LIMIT    False      53.5ms    289ms
 1024  10  SolveStatus.ITERATION_LIMIT    False      45.6ms    391ms
  512   9  SolveStatus.ITERATION_LIMIT    False      20.0ms    110ms
 1024  10  SolveStatus.ITERATION_LIMIT    False      46.8ms   2767ms   ← worst
  ... 14 slow calls totaling 5.3 s of the 5.5 s envelope time
```

**Replay evidence — it's cycling, not a too-small cap.** Captured 3 of the failing LPs to disk and replayed against the Rust simplex with escalating caps:

```
LP0 (m=1024 n=10): cap=3000 → ITERATION_LIMIT 100ms | cap=20000 → ITERATION_LIMIT 646ms | cap=200000 → ITERATION_LIMIT 6492ms
LP1 (m=1024 n=10): cap=3000 → ITERATION_LIMIT  84ms | cap=20000 → ITERATION_LIMIT 549ms | cap=200000 → ITERATION_LIMIT 5473ms
LP2 (m= 512 n= 9): cap=3000 → ITERATION_LIMIT  40ms | cap=20000 → ITERATION_LIMIT 262ms | cap=200000 → ITERATION_LIMIT 2626ms
```

200k pivots on an 11-row LP and still spinning — the simplex is cycling on the duplicate columns. (A synthetic non-degenerate LP of identical shape solves in 1.9 ms, <100 pivots.)

**Verified fix — pinned-dim removal + vertex dedup, then the same simplex converges in <1 ms:**

```
LP0: pinned dims=1  vertices 1024 → 512 unique, coords 10 → 9   → OPTIMAL, usable dual, 0.7ms
LP1: pinned dims=2  vertices 1024 → 256 unique, coords 10 → 8   → OPTIMAL, usable dual, 0.4ms
LP2: pinned dims=2  vertices  512 → 128 unique, coords  9 → 7   → OPTIMAL, usable dual, 0.1ms
```

>300× per-LP; eliminates ~5 s (41%) of the nvs09 solve. Soundness is free: set the returned slope to 0 on pinned dims — `_solve_envelope` already recomputes the intercept over the **original** vertices for *any* slope ("the cut is rigorously valid for ANY slope"), so the reduced-LP cut passes the existing validity argument unchanged. ~10 lines in `_solve_envelope` / `_separate_multilinear_envelope_uncached`.

Separately worth an issue: the Rust simplex should get anti-cycling pivoting (Bland's rule / lexicographic on stall). Cycling forever on duplicate columns is a latent defect wider than this call site — the dedup fix removes this *trigger*, not the underlying susceptibility.

### Finding 2 — ep5's "jit" wrapper never jits (22% of solve)

`uniform_relax.py:387` (`_LazyJaxprEval.__call__`): the ep5 wrapper traces the separation value/grad function to a jaxpr lazily (good) but then evaluates with **`jax.core.eval_jaxpr`** — JAX's *interpreted* evaluator. Every call walks the jaxpr in Python with per-primitive eager dispatch: the profile shows 5,808 wrapper calls → 6,048 `eval_jaxpr` → **137k `core.bind`** = 2.7 s of pure interpreter overhead (~0.5 ms/call).

Fix: `jax.jit` the traced function (still lazily, same exception transparency) so calls after the first are compiled XLA (~µs dispatch). Worth ~2.5 s on nvs09.

### Expected impact

Both fixes: **nvs09 ≈ 12 s → ~4 s**, and they generalize — every multilinear instance branches into pinned dims (finding 1 fires on the *node* path constantly), and every convex-separation call pays the interpreter tax (finding 2). Remaining tail after that: the many-small-NLP heuristic pattern (518 POUNCE sub-NLP solves = 1.4 s) — a batching candidate, separate discussion.

*All measurements on `664520ab`, flags off: cProfile + caller attribution, a monkeypatch trace of `_solve_envelope` (sizes/status/timing per call), captured-LP `.npz` replay against `lp_simplex.solve_lp` at caps 3k/20k/200k, and the reduction fix re-solved on the identical captured data.*

🤖 Review by [Claude Code](https://claude.com/claude-code)

Issue #649, opened 2026-07-15
G6: in-house Rust simplex non-convergence on ill-conditioned relaxation LPs (bchoco06 class; HiGHS solves the identical matrix)


jkitchin:
## Summary

The in-house Rust dual simplex (`crates/discopt-core/src/lp/simplex/`, the default
per-node LP engine) fails to converge (`iteration_limit`, no bound) on
ill-conditioned relaxation LPs where HiGHS on the **byte-identical** matrix returns
`optimal`. The concrete probe instance is the `bchoco06.nl` root uniform relaxation,
but this is a **class** defect (recursive-McCormick envelopes of high-degree
products on wide/unbounded boxes), not an instance quirk.

This issue is the escalation branch that baron-gap-plan.md §10 "G6" named, after the
cheaper G6 subnormal-flush hypothesis was **falsified** (record:
`docs/dev/g5-family-d-diagnoses.md` §"G6 update").

## Evidence (reproducible)

```
JAX_PLATFORMS=cpu JAX_ENABLE_X64=1 PYTHONPATH=$PWD/python \
  python discopt_benchmarks/scripts/g6_bchoco06_conditioning_probe.py
```

bchoco06 root LP (833 rows × 349 cols), one matrix stage at a time:

| stage | A nonzero spread (max / min) | in-house simplex | HiGHS |
|---|---|---|---|
| raw (as shipped) | 1.0e10 / 4.9e-324 = inf | `iteration_limit`, None | optimal, 0.99998 |
| subnormals flushed (`|v|<1e-300 → 0`) | 1.0e10 / 1.0e-10 = 1e20 | `iteration_limit`, None | optimal, 0.99998 |
| flushed + Ruiz equilibrated (20/50/100 sweeps) | 1.95 / 1.56e-12 = **1.25e12** | `iteration_limit`, None | optimal, ~1.0 |

- The `-4.941e-324` subnormal bounds on `x^2`/`x*y` aux columns (interval-arithmetic
  underflow of a true `0`) are a real artifact but **not** the cause of
  non-convergence: flushing all 20 subnormal `A` nonzeros + 29 subnormal bounds to
  0 leaves the simplex at `iteration_limit`.
- The genuine coefficient spread is `~1e20` (smallest real McCormick coefficient
  `1e-10`, largest `1e10`). Geometric-mean (Ruiz, power-of-two) equilibration —
  which the shipped `_solve_lp_warm_equilibrated` path already runs — compresses
[...]

Issue #170, opened 2026-06-16
Fast simplex backend ~5000x slower than HiGHS on ex1252 FBBT-box McCormick LP (54s vs 0.01s)


jkitchin:
## Summary

On `ex1252`'s FBBT-tightened root box, the fast Rust `simplex` LP backend takes **~54 s** to solve the McCormick relaxation LP, while the reference HiGHS (`auto`) path solves the *same* LP in **~0.01 s**.

## Resolution

Fixed in #177 (merged). Root cause: the fast simplex factorized the **raw** basis with feral's LU, which broke down (singular) on the ill-scaled lifted LP — coefficients span ~1e9 because FBBT pushes product-aux bounds that large. `solve_lp` then returned `Numerical`; an uncertified relaxation bound can never fathom, so the MILP B&B **enumerated the entire binary tree** (65,499 ≈ 2¹⁶ nodes for the 15 integers) and never certified optimality. So it wasn't pivot speed — it was an LU breakdown that destroyed all pruning.

Fix: power-of-two geometric-mean **equilibration** before factorization (`scaling.rs`), mirroring HiGHS — the objective and optimal basis are scaling-invariant, factors snap to powers of two (exact), and scaling triggers only when the matrix dynamic range exceeds 1e6 so well-conditioned LPs are bit-identical. #177 also adopted shared per-batch equilibration and per-node factorization reuse in the B&B, added a multiple-RHS primitive + batched LP API, and removed the `ex1252` test time-limit workaround.

**Result (`ex1252` FBBT box):** 65,499 nodes / uncertified / ~217 s → **31 nodes / status=optimal / ~0.6 s**.

Remaining gap to HiGHS is intrinsic per-node work; cross-node factorization carry-over is tracked in #178.

PR #651, opened 2026-07-15
fix(correctness): recover the dual bound the warm-path re-scaling drops on ill-conditioned relaxations (#649)

jkitchin:
Fixes **#649** — the in-house `backend="simplex"` path dropping a valid dual bound on ill-conditioned relaxation LPs (bchoco06 class), where HiGHS solves the byte-identical matrix.

## Root cause — the hypothesis was overruled by measurement
Issue #649 hypothesized a deep simplex-core numerics defect (refactorization cadence / Harris ratio test / Markowitz tolerances). Diagnosis showed otherwise:

- **The simplex core is fine.** Fed bchoco06's equilibrated root LP directly, `solve_lp_cols` (cold primal, *no* Rust scaling) returns **Optimal, obj −1.0, 0 iters** — clean, matching HiGHS.
- **The defect is in the warm/binding path.** `dual::solve_lp_warm_csc` (what the shipped `backend="simplex"` uses) applies its **own** pow2 `Scaling::from_sparse` on top of the already-Ruiz-equilibrated matrix. On this class that re-scaling is counterproductive: the scaled optimum, unscaled, lands just past the feasibility audit and the solve returns **`Numerical`** (no bound), where the unscaled solve is clean.

Isolation (Rust diagnostic on the captured std-form fixture):

| path on the identical bchoco06 std-form LP | result |
|---|---|
| `solve_lp_cols` (no Rust scaling) | **Optimal**, obj −0.99999 |
| `solve_lp_warm_csc` (scaling ON), **pre-fix** | **`Numerical`**, no bound |
| `solve_lp_warm_csc` (scaling ON), **post-fix** | **Optimal**, obj −0.99999 |

## The fix
Neither scaling nor no-scaling dominates (scaling is *needed* on the issue-#170 degenerate LPs), so keep scaling-first and recover on failure:

- **`scaling.rs`** — add `Scaling::unscale_cols`, the exact pow2 inverse of `scale_cols` (reciprocals of powers of two are exact); zero allocation on the success path.
- **`dual.rs`** — in `solve_lp_warm_csc`, when the scaled solve returns `Numerical`/`IterLimit`, reconstruct the unscaled matrix and retry via the trusted cold path; prefer its resolved `Optimal`/`Infeasible`/`Unbounded`. **Sound**: the unscaled solve carries its own feasibility audit, so a scaled point the audit rejected is never accepted — only a bound the scaling lost is recovered.
- **`milp_relaxation.py`** — guard the reported objective/bound with `math.isfinite`, so a non-finite value (exposed on the raw un-presolved path once solves started returning `optimal`) is reported as no-bound rather than a NaN that would silently pass `bound ≤ incumbent`.

**Result:** bchoco06 `Model.solve()` now reports **bound ≈ 0.99999** (was none at 7 nodes — the original EP0 symptom), agreeing with HiGHS.

## Gate (bound-changing regime, CLAUDE.md §5)
- `cargo test -p discopt-core`: **446 pass** — byte-neutral on all existing solves, plus the new regression `bchoco06_illcond_scaled_path_recovers_bound_649` (fails pre-fix → `Numerical`, passes after).
- **Cross-solver soundness**: recovered bound agrees with HiGHS on the identical matrix (bchoco06 is `known=null`, so cross-solver agreement is the oracle).
- **No relaxation changed** → no feasible point can be cut; only the LP solver stops dropping a valid bound.
- **Byte-neutral**: clean instances (alan/ex1221/ex1225/nvs09/st_miqp2) hit their known optima unchanged; the fallback fires only on `Numerical`/`IterLimit`.
- `pytest -m smoke`: **638 passed**; adversarial suite (`test_adversarial_recent_fixes.py`): **10 passed**.

## Commits
1. `style:` — rustfmt reflow of **pre-existing** drift in 4 untouched files (the Rust CI job has no `cargo fmt --check`, so it went uncaught; the local hook formats the whole package). Whitespace only, separated to keep the fix clean.
2. `fix(correctness):` — the #649 fix above.

Docs: `baron-gap-plan.md` §10 G6 marked **RESOLVED**; `g5-family-d-diagnoses.md` gains a "G6 resolution" record in the house style (hypothesis overruled by measurement).

https://claude.ai/code/session_01T9BWf4gy9kxqZBRbFkQWhd


jkitchin:
## Before/after smoke — the fix recovers/tightens dual bounds at the solver level (class-wide)

Ran a small panel through `Model.solve()` (`time_limit=15s`, `max_nodes=200`) on two release builds: pre-fix (`3205d005`) vs fixed (`a1523e43`).

| instance | BEFORE bound | AFTER bound | effect |
|---|---|---|---|
| **bchoco06** | **none** | **0.99999** | dual bound **recovered** (the #649 headline) |
| **4stufen** | 19021.3 | **20712.2** | bound **tightened +8.9%** |
| **beuster** | 6352.1 | **8244.8** | bound **tightened +29.8%** |
| heatexch_gen1 | 42472.2 | 42472.2 | unchanged |
| casctanks | −90.18 | −90.18 | unchanged |
| bchoco07 / bchoco08 | none | none | still none — a *different* cause (genuinely harder; not the scaling failure) |
| alan / nvs09 / st_miqp2 (controls) | optimal | **identical** | no regression |

The benefit is **class-level, not single-instance** (CLAUDE.md §2): besides recovering bchoco06's bound outright, the fix tightens the *global* dual bound on 4stufen and beuster — interior node LPs that hit the scaled-solve `Numerical` failure were dropping their bounds (loosening `min`-over-frontier); the unscaled fallback restores them. Controls are byte-identical, confirming the fallback stays inert where scaling succeeds.

bchoco07/08 remain unbounded here for a separate reason (not the re-scaling path) — reported honestly, not masked.

Issue #291, opened 2026-06-22
Hang: nvs12 infinite-loops in solve_milp_py (native MILP B&B), ignores time_limit — new on feat/lp-spatial-branch-and-cut


jkitchin:
## Summary

`nvs12` (small MINLP, **opt −481.2**, previously solved in ~0.4 s) now **hangs indefinitely** in the native MILP branch-and-bound (`solve_milp_py`), **ignoring the `time_limit`**. With `time_limit=15` the solve runs past 40 s without returning; only an external hard-kill stops it. In the benchmark it shows up as `hard_timeout` (the harness's 10 s hard-kill fires); run in-process with no hard-kill it never terminates.

This is a **new regression on `feat/lp-spatial-branch-and-cut` @ `0c2db1a`** (+ uncommitted `lp_spatial_bb`/`incremental_mccormick` edits). nvs12 solved to optimal in ~0.4 s on the prior build and earlier in this branch.

## Reproduction

```bash
JAX_ENABLE_X64=1 python -c "
import time, discopt.modeling as dm
t=time.time()
r=dm.from_nl('minlplib/nl/nvs12.nl').solve(time_limit=15, gap_tolerance=1e-4)
print(r.status, r.objective, 'wall', time.time()-t)
"
# hangs — never returns; must be killed (time_limit=15 not honored, runs >40s)
```

## Hang site (faulthandler stack after 8 s)

```
File "discopt/solver.py", line 9249 in _solve_milp_simplex   # -> solve_milp_py(...)
File "discopt/solver.py", line 3110 in solve_model
File "discopt/modeling/core.py", line 2865 in solve
```

The hang is inside the native `solve_milp_py` MILP B&B call at `solver.py:9249`.

## Likely mechanism

`nvs12` has **bilinear/monomial terms between integer variables**:

```
classify_nonlinear_terms(nvs12): bilinear=[(0,2),(1,2),(0,3),(2,3),(0,1),(1,3)], monomial=[(0,2),(1,2),(2,2),(3,2)]
```

These are exactly the targets of the **integer-bilinear MILP reformulation (#285)** — they get linearized into a pure MILP, which is then handed to `solve_milp_py`. That native MILP B&B then **loops without honoring the time limit** on the reformulated nvs12.

Two distinct problems:
1. **The hang itself** — `solve_milp_py` should always terminate; an infinite loop in node selection / branching (the new pseudocost branching?) on this small MILP is a correctness/robustness bug.
[...]

PR #695, opened 2026-07-17
fix(#689,#698,#280): lift initial_point across the reform, swap-reseed the MILP incumbent, and stop discarding time_limit

### LINKED ISSUE #689: initial_solution silently dropped on the MILP fast path when the integer-bilinear reform lifts the model
## Summary

`initial_solution` (and any internal `initial_point`) is **silently discarded** on the Rust MILP fast path whenever `reformulate_integer_bilinear` lifts the model. No warning, no log at INFO — the user's seed simply has no effect.

Surfaced while verifying #280.

## Mechanism

`solver.py` adopts the integer-bilinear reformulation, which binary-expands + big-M-linearizes integer-factor products and **appends aux columns after the originals**, growing the flat variable vector (e.g. `graphpart_2pm-0088-0888`: 192 → 765).

`initial_point` is built in the **original** space and is never extended across that lift. `_solve_milp_simplex` then guards:

```python
n_orig = sum(v.size for v in model._variables)          # 765 (lifted)
if initial_point is not None and np.asarray(initial_point).size == n_orig:   # 192 != 765
    _seed = ...
```

The size check fails, `_seed` stays `None`, and the seed is dropped without a word.

The binary-multilinear reform already solved exactly this problem: it records `_bml_aux_spec` / `_bml_n_orig_flat` and provides `extend_initial_point`, wired at `solver.py:4113`. `integer_product_reform` has **no analogue**.

## Evidence

On `graphpart_2pm-0088-0888` (optimum -55), seeding a **verified-optimal** incumbent found by `one_hot_swap_search`:

```
original model nvars = 192
  _solve_milp_simplex: reformulated n_cols = 765 | initial_point size = 192
seeded solve -> obj = -46      (seed of -55 silently ignored)
```

A known-optimal seed is thrown away and the solve returns a 16.4%-worse incumbent.

## Why it matters

- **Silent failure.** `initial_solution` is a documented, user-facing parameter that quietly does nothing on this path. Per CLAUDE.md §3 this should refuse loudly or work — not silently no-op.
- **Blocks #280.** The natural primal fix (improve the incumbent in original space, then reseed) cannot work until the seed survives the lift.

Note the guard itself is correct and *sound* — a wrong-length seed must not be trusted. The bug is that nothing extends the point across the lift.

## Proposed fix

Mirror the BML pattern in `integer_product_reform`:

1. Have `_Expander` record an aux spec in creation order:
   - `expansion`: bits `e_k` pinned by `x - lo - Σ 2^k·e_k == 0` → bits are the binary digits of `round(x - lo)`.
   - `bigm_product`: `v == e * other` → `v = x[e] * value(other)`.
   Both are **exactly determined** by the originals, so this is reconstruction, not approximation.
2. Add `extend_initial_point(reformed, x0)` evaluating the spec in order; return `None` (today's behavior — no seed) for any entry it cannot reconstruct exactly. Never guess.
3. Wire it at the adoption site, mirroring `solver.py:4113`.

Soundness: purely primal. The Rust driver already re-validates any seed (bounds, integrality, row feasibility) and recomputes its objective, so a bad point is dropped, never trusted — the dual bound and certificate are unaffected either way.

Refs: #280 (blocked by this).

---

jkitchin:
## Summary

Fixes #689, #698, and closes the primal half of #280. Three linked defects on the Rust MILP fast path, all surfaced by running #280's corpus-verification checklist ([findings](https://github.com/jkitchin/discopt/issues/280#issuecomment-5003047442)).

### 1. `initial_solution` silently dropped (#689)

`reformulate_integer_bilinear` lifts integer-factor products into an exact MILP, appending aux columns (`graphpart_2pm-0088-0888`: 192 → 765). A caller's seed is built over the ORIGINAL columns, so it failed `_solve_milp_simplex`'s `size == n_orig` guard and was dropped without a word — observed **discarding a known-optimal incumbent** (-55 handed in, -46 returned).

The reform now records each aux definition (expansion bits of `x = lo + Σ2^k·e_k`; each big-M product `v = e * other`) and `extend_initial_point()` reconstructs them from the originals. Both are *exactly* determined, so this is reconstruction, not approximation — verified: the rebuilt point satisfies every row of the lifted model and reproduces the objective exactly. Mirrors the existing `_bml_aux_spec` pattern. Anything not exactly reconstructible returns `None` → no seed (today's behavior), never a guess.

### 2. `one_hot_swap_search` (#687) never fired on `graphpart_*` (#280)

It was wired only into the spatial-B&B LNS loop, but all 15 instances route to the Rust MILP engine — it was dead code for the family it was written for. Its structure detector matches the real `.nl` encodings fine; **reachability**, not structure, was the problem.

The swap is only well defined *pre-lift* — permuting bits in the lifted space leaves the auxiliaries stale, i.e. genuinely infeasible (verified: the heuristic correctly *rejects* it), so naively wiring it into the MILP path would have been a silent no-op. So: swap over the originals, lift the result back via (1), re-enter the engine seeded with it.

### 3. Most of `time_limit` discarded (#698)

The engine's `max(0.5, min(0.5·remaining, 10 s))` slice is a **stall** bound (#291), but on an uncertified `feasible` the code returns rather than falling back, so the rest of the budget goes to no engine at all. At `time_limit=40` it used ~10 s; at the default 3600 it uses ~10 s. The engine is monotone in its budget (bound -54.0 / -50.0 / -40.5 / -38.25 at 2.5 / 5 / 10 / 20 s), so that was real search thrown away. Now it re-enters with the remainder.

**The first slice stays bounded on purpose.** Handing run 1 the whole budget is measurably *worse* — it starves the swap (gap 26.6% → 65.2% on `3pm-0334-0334`; `autocorr_bern30-04` -324 → -24). The fix is not "remove the cap".

## Soundness

Purely primal + search effort; the dual bound and certificate are never weakened.

- The hook fires **only** on an uncertified `feasible`, so a certified `optimal` never reaches it — certification cannot regress by construction.
- Any improved point is only ever offered as an `initial_incumbent`, which the Rust driver re-validates (bounds, integrality, row feasibility) and re-scores from `c`. A bad point is dropped, never trusted.
- A re-entry result is kept only when strictly better, or when it certifies the same incumbent; otherwise only dual progress is banked.
- Both runs' bounds are valid lower bounds on the same problem, so keeping the tighter is valid and stays below any feasible incumbent's objective.

## Measurements

**#280 primal gap — 15 `graphpart_*`, 5 s, `minlplib.solu` oracle:**

| instance | before | after | | instance | before | after |
|---|---|---|---|---|---|---|
| `2g-0077-0077` | 6.49% | **2.99%** | | `3g-0244-0244` | 20.48% | **9.44%** |
| `2pm-0077-0777` | 15.00% | **10.00%** | | `3g-0333-0333` | 9.86% | **0.00%** |
| `2pm-0088-0888` | 16.36% | **0.00%** | | `3g-0344-0344` | 17.02% | **7.57%** |
| `3pm-0244-0244` | 21.43% | **10.71%** | | `3pm-0334-0334` | 36.11% | **11.11%** |
| `3pm-0344-0344` | 27.08% | **16.67%** | | | | |

Every instance improved or held; **none regressed**. 6/15 certify optimal; `incorrect_count = 0`; no bound above its reference optimum.

**#698 budget — A/B at TL=60:**

| instance | stock | with re-entry |
|---|---|---|
| `edgecross10-090` | feasible, obj 1411 | **optimal, 1387** |
| `graphpart_3pm-0334-0334` | feasible, obj -32 | **optimal, -36** (= reference) |
| `edgecross10-040` | obj 173, bnd 48 | obj **148**, bnd **118.5** |
| `graphpart_2pm-0077-0777` | obj -37 | obj **-40** (= reference) |

2 newly certified, 5/6 tighter bounds, 0 bounds above reference. Wall 147 s → 335 s.

At **TL=5 the fixed and stock arms are identical** (status, objective, wall) across 8 instances — the no-extra-budget guard suppresses a pointless re-entry.

## ⚠️ Behavior change worth a reviewer's eye

With the default `time_limit=3600`, a MILP that currently returns uncertified at ~10 s may now search up to the full hour. That is what the limit asks for, and it is how a global solver should use a granted budget — but it is visible to anyone relying on the early return. Happy to gate it if you'd rather.

## Falsified along the way (recorded in #280)

The discarded budget is **not** the cause of #280's *primal* gap: restoring it moves only the dual bound (-54 → -41), never the incumbent. The two fixes are complementary — the swap supplies the incumbent, the budget supplies the certificate — which is why `3pm-0334-0334` only reaches `optimal` with both.

## Tests

- `python/tests/test_issue_689_ipx_seed_lift.py` (4) — the discriminating test (seed reaches the driver at lifted width) **fails before** the change ("silently dropped"), passes after. A weaker `objective <= optimum` assertion does *not* discriminate, since the engine reaches the optimum unaided on small instances.
- `python/tests/test_issue_698_milp_budget_reentry.py` (3) — the budget test **fails on the parent commit** (`budgets=[10.0, 10.0, 5.0]` under a 30 s limit), passes after.
- Synthetic models use scalar `integer(0,1)` columns, matching how the `.nl` reader encodes the real family — `_int_factor_range` excludes declared `BINARY`, so a `m.binary(...)` model is never lifted and would not exercise this path.
- `pytest -m smoke`: **665 passed**, 1 skipped. Adversarial: **10 passed**. MILP/simplex/reform/seed/budget selection: **308 passed**. `ruff`/`mypy` clean. No Rust touched.

## Follow-up

#280 stays open: the remaining tails (`3pm-0344-0344` 16.7%) are dual-side, where greedy swap plus more budget still do not close the gap.

### DIFF
--- a/python/discopt/_jax/integer_product_reform.py
+++ b/python/discopt/_jax/integer_product_reform.py
@@ -105,6 +105,105 @@ def _int_factor_range(
     return lo_i, hi_i
 
 
+def _flat_offsets(model: Model) -> tuple[dict[int, int], int]:
+    """Map ``id(Variable)`` -> its offset in the model's flat vector, plus the
+    total flat width. Variables are laid out in ``model._variables`` order."""
+    off: dict[int, int] = {}
+    cur = 0
+    for v in model._variables:
+        off[id(v)] = cur
+        cur += int(v.size)
+    return off, cur
+
+
+def _resolve_aux_spec(model: Model, spec: list[tuple]) -> Optional[list[tuple]]:
+    """Resolve an ``_Expander.aux_spec`` (Variable refs) to flat-index arithmetic
+    against *model*'s final variable list. ``None`` if any ref is unresolvable."""
+    off, _ = _flat_offsets(model)
+
+    def idx(ref) -> Optional[int]:
+        if ref is None:
+            return None
+        var, elem = ref
+        base = off.get(id(var))
+        return None if base is None else base + int(elem)
+
+    out: list[tuple] = []
+    for entry in spec:
+        if entry[0] == "bits":
+            j_x = idx(entry[1])
+            bits = [idx((e, 0)) for e in entry[3]]
+            if j_x is None or any(b is None for b in bits):
+                return None
+            out.append(("bits", j_x, int(entry[2]), bits))
+        elif entry[0] == "prod":
+            j_v, j_e, j_o = idx((entry[1], 0)), idx((entry[2], 0)), idx(entry[3])
+            if j_v is None or j_e is None:
+                return None
+            out.append(("prod", j_v, j_e, j_o))
+        else:  # unknown kind: decline rather than guess
+            return None
+    return out
+
+
+def extend_initial_point(reformed: Model, x0) -> Optional[np.ndarray]:
+    """Extend a point over the ORIGINAL variables (flat, length
+    ``_ipx_n_orig_flat``) to a point over *reformed*'s full variable vector by
+    evaluating each auxiliary from its recorded definition — the expansion bits
+    ``e_k`` of ``x = lo + sum 2^k e_k``, and each big-M product ``v = e * other``.
+    Both are *exactly* determined by the originals at integral ``e``, so this is
+    reconstruction, not approximation.
+
+    Returns ``None`` — the caller then simply does not seed, today's behavior —
+    when the model carries no reformulation metadata, the point has the wrong
+    length or is non-integral where a bit expansion needs an integer, or any aux
+    is not reconstructible. Never guesses: a partial or invented seed would be
+    worse than no seed.
+
+    Purely primal, never a soundness lever: the MILP driver re-validates whatever
+    it is handed (bounds, integrality, row feasibility) and recomputes the
+    objective, so a bad point is dropped rather than trusted (#689)."""
+    spec = getattr(reformed, "_ipx_aux_spec", None)
+    n0 = getattr(reformed, "_ipx_n_orig_flat", None)
+    if spec is None or n0 is None:
+        return None
+    x = np.asarray(x0, dtype=np.float64).ravel()
+    if x.size != int(n0) or not np.all(np.isfinite(x)):
+        return None
+
+    _, total = _flat_offsets(reformed)
+    out = np.full(int(total), np.nan, dtype=np.float64)
+    out[: int(n0)] = x
+    known: set[int] = set(range(int(n0)))
+
+    for entry in spec:
+        if entry[0] == "bits":
+            _, j_x, lo, bit_idx = entry
+            if j_x not in known:
+                return None
+            shifted = float(out[j_x]) - float(lo)
+            r = round(shifted)
+            # A fractional value has no bit expansion; a value outside the
+            # expanded range would silently wrap. Decline both.
+            if abs(shifted - r) > 1e-6 or r < 0 or r >= (1 << len(bit_idx)):
+                return None
+            for k, j in enumerate(bit_idx):
+                out[j] = float((int(r) >> k) & 1)
+                known.add(j)
+        elif entry[0] == "prod":
+            _, j_v, j_e, j_o = entry
+            if j_o is None or j_e not in known or j_o not in known:
+                return None
+            out[j_v] = float(out[j_e]) * float(out[j_o])
+            known.add(j_v)
+        else:
+            return None
+
+    if np.any(np.isnan(out)):
+        return None
+    return out
+
+
 class _Expander:
     """Creates and caches the binary expansion of integer factor variables and
     accumulates the linking ``x_i == lo + sum 2^k e_k`` constraints."""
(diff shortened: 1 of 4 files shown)


jkitchin:
Closing as **superseded**. This PR bundled three fixes; two landed independently while it sat stale (48 commits behind `main`), which is why it went CONFLICTING:

| piece | now on `main` via | status |
|---|---|---|
| **#698** re-entry (spend leftover budget) | #705 (merged) | superseded |
| **#689** warm-start lift across the reform | #696 (merged, `_attach_warm_start_spec`/`extend_initial_point`) | superseded — main's version is newer |
| **#280** one-hot-swap reseed (`_one_hot_swap_reseed`) | — | the only still-unmerged piece |

A straight rebase would re-apply this branch's *older* #689/#698 drafts on top of the better merged versions — a regression on `main`. So rather than rebase, the remaining #280 lever is carried forward as a focused PR adapted to main's #696 API: **#712**.

#712 ports `_one_hot_swap_reseed` onto main's #705 re-entry, behind `DISCOPT_MILP_SWAP_RESEED` (**default off**) with a starter differential panel (2/4 graphpart instances strictly better primal, 2/4 tie, 0 worse, 0 unsound) — default-off pending the corpus-wide §5 graduation gate, since one instance (`graphpart_2pm-0088`) is neutral and "sound ≠ helpful".

Superseded by #712.

Issue #774, opened 2026-07-18
Re-implement coefficient tightening correctly + verify final incumbent vs original model (#772 follow-up)


jkitchin:
Follow-up to #772 (the P0 false-primal was fixed by reverting #770 in #773; a default-path soundness guard test landed). Two constructive items remain, neither a P0:

## 1. Re-implement big-M coefficient tightening correctly (feature)
The idea is sound and valuable — activity-based coefficient strengthening closed **56% of syn40m's root spread** (#282 Stage 1) before the bug. The reverted implementation (`_root_presolve.py::tighten_bigm_coefficients`) **enlarged** the fixed-charge (`a_k < 0`) feasible set instead of preserving it (a sign/bound error), admitting points infeasible in the original problem. Re-implement with:
- The **fixed-charge case done correctly** (the Savelsbergh `a_k > 0` case may have been fine; the `a_k < 0` case was the culprit — derive it carefully).
- A **feasible-set-EQUIVALENCE test in BOTH directions** as the gate (the #770 test only checked "no feasible point removed"; it must ALSO check "no infeasible point admitted"): sample/enumerate integer assignments, assert feasible-in-original ⟺ feasible-in-tightened.
- Default-OFF + §5 graduation (with the §1.5 node-reduction gate — note #282 Stage 4 measured it tightens the root but certifies nothing on the panel, so it may HOLD anyway).

## 2. Defense-in-depth: verify the final incumbent against the ORIGINAL model (hardening)
The #770 false primal reached the reported result because presolve mutates the model DAG **in place**, so every incumbent-feasibility check evaluated against the (unsoundly) mutated model. A global solver's product is its certificate — add a final guard:
- Snapshot the original constraints before presolve mutates the model.
- Before returning any incumbent as `optimal`/`feasible`, verify it against that snapshot; on failure, **withhold the incumbent / decertify and log loudly** (never return a point infeasible in the original problem).
- This catches ANY unsound presolve or heuristic, not just this one. Design decision to resolve: the failure-handling path (downgrade vs abstain).

Evidence: #772, #770, #282 Stages 0–4 (`docs/dev/issue-282-stage*-verdict.md`), reproduction `scratchpad/verify_incumbent2.py`. The default-path guard test (`test_incumbent_soundness_772.py`) is the interim protection.

jkitchin:
[...]

Issue #772, opened 2026-07-18
P0 correctness: DISCOPT_COEF_TIGHTEN returns an infeasible incumbent / false primal (obj > opt) on convex synthesis family (#770)


jkitchin:
## Severity: P0 correctness — false primal value (CLAUDE.md §1, worst class)

`DISCOPT_COEF_TIGHTEN=1` (the presolve big-M coefficient tightening merged in #770, default-OFF) returns an **incumbent that violates the original problem's constraints and reports a super-optimal objective** on the convex `rsyn*`/`syn40m` process-synthesis family.

### Independently verified (name-aligned, pristine model)
```
rsyn0805m  flag-ON:  obj 1441.99  >  =opt= 1296.12   → feasible in PRISTINE original model: FALSE
syn40m     flag-ON:  obj  370.24  >  =opt=   67.71   → feasible in PRISTINE original model: FALSE
```
Flag-OFF returns valid incumbents (obj ≤ opt, feasible) on both. Reproduction (name-aligned evaluation against a fresh `from_nl` model, no ordering artifact): `scratchpad/verify_incumbent2.py` — solve with `DISCOPT_COEF_TIGHTEN=1`, take `r.x`, evaluate against a pristine `from_nl(instance)` via `_check_constraint_feasibility` → **infeasible**, while `r.objective` exceeds `=opt=`.

### Attribution (evidence points to the tightening, not NLP-BB alone)
The bad incumbent is **feasible in the mutated/tightened model but infeasible in the original** — i.e. the coefficient tightening **enlarged the feasible set** (relaxed a constraint) rather than preserving it. A correct coefficient strengthening preserves the integer-feasible set exactly; this one admits points the original forbids, most likely a **sign/bound error in the fixed-charge (`a_k < 0`) case** of `tighten_bigm_coefficients` (`python/discopt/solvers/_root_presolve.py`). (There may ALSO be a secondary NLP-BB gap — it accepts an incumbent from the presolved model without re-verifying feasibility against the original — but the primary defect is the unsound tightening.)

The #770 soundness suite (`test_coef_tighten_282.py`, feasible-point sampling) passed because it only checked that *known-feasible* points survive (no feasible point removed) — it did **not** check the *other* direction (no infeasible point admitted), which is where the bug lives.

### Impact
- Default path is **safe** (flag default-OFF). But the opt-in flag is unsound: anyone enabling it gets false primal values on this class, and #282's "recoverable lever" is invalid.

### Ask (correctness-first)
[...]

Issue #282, opened 2026-06-21
Global-search gap: syn_*/rsyn_* process-synthesis MINLP (sound suboptimal incumbents)


jkitchin:
## Summary

Tracking the **process-synthesis (`syn*` / `rsyn*`) MINLP global-search gap**: discopt returns sound feasible incumbents (~19% median gap) but does not close to the optimum within budget. Combinatorial mixed-integer structure; companion to the #268 performance work and #276 LNS heuristics.

| instance | sense | discopt status | obj | known opt | gap | SCIP |
|---|---|---|---|---|---|---|
| `rsyn0805m` | max | feasible | 1025.85 | 1296.12 | 20.9% | 0.5s |
| `rsyn0810m` | max | feasible | 1387.3 | 1721.45 | 19.4% | 0.5s |
| `rsyn0815m` | max | feasible | 1041.53 | 1269.93 | 18.0% | 0.7s |
| `syn15m02hfsg` | max | feasible | 1276.1 | 2832.75 | 55.0% | 0.5s |
| `syn30hfsg` | max | feasible | 113.197 | 138.16 | 18.1% | 0.8s |
| `syn40hfsg` | max | feasible | 61.1971 | 67.7133 | 9.6% | 1.3s |
| `syn40m` | max | feasible | 23.1601 | 67.7133 | 65.8% | 1.1s |

Related: #268 (timeout/perf), #267 (suboptimal-incumbent gap).

*Environment: discopt `0.4.1.dev0` (post #270/#273/#275/#276/#277), JAX `0.10.0`, `JAX_ENABLE_X64=1`, 5s time limit. Surfaced via randomized MINLPLib short-tier benchmark shuffles. All instances are SOUND suboptimal incumbents (feasible, not certified, worse-than-optimum in the correct direction) — not correctness bugs. Gap shown as relative %, or absolute for near-zero optima.*

jkitchin:
Picked this up on `claude/discopt-issue-282-bsffxq`. Full write-up in `docs/dev/issue-282-syn-rsyn-diagnosis-2026-07-17.md`.

**tl;dr — the perf gap is not reproducible in this environment, so no perf/relaxation change was shipped (per the "no fix on an un-run entry experiment" rule); but the OA probe surfaced a real correctness bug, which is fixed.**

[...]

PR #783, opened 2026-07-19
feat(#774): re-implement big-M coefficient tightening (feasible-set-equivalent, DISCOPT_COEF_TIGHTEN, default-OFF)

jkitchin:
## Summary

Re-implements activity-based big-M coefficient tightening (`DISCOPT_COEF_TIGHTEN`,
default-OFF) correctly, addressing item 1 of #774. #770 introduced this pass and was
reverted in #773 because it produced a false primal (rsyn0805m obj **1441.99** >
`=opt=` **1296.12**, feasible in the mutated model but not the original). This PR
re-derives the transformation, fixes the actual defects, and adds the both-directions
equivalence gate the #770 test lacked.

`Contributes to #774` (item 1). Item 2 (final-incumbent verification) already landed
as #779. Default-OFF; this does **not** graduate the flag (see the HOLD verdict below).

## Root cause of the #770 false primal (empirically pinpointed)

I reconstructed the reverted pass (`git show 42c2b573`), applied it to rsyn0805m, and
reproduced the exact false primal (obj 1441.99). Contrary to the original diagnosis
("fixed-charge `a_k < 0` sign/bound error, enlarged feasible set"), the data show the
transformation *math* is sound and the real defects are elsewhere:

**Defect 1 — malformed write-back (the false primal).** `from_nl` stores every
inequality with the whole constant in the body and `con.rhs == 0` (verified: 214/214
rsyn0805m inequality rows). The reverted write-back did
`con.body = rebuild(a', const)` (constant left in the body) **and**
`con.rhs = rhs_norm' + const` (a now-nonzero rhs) — splitting the constant across
body *and* rhs simultaneously, a normal form the model never produces natively. A
downstream relaxation/feasibility component (exercised only with `con.rhs == 0`) reads
one of the two, shifting the enforced row by `const` and admitting the false primal.
Evidence: the flag-ON incumbent violates the *tightened* `con.body` rows too, so it is
feasible in neither the original nor the mutated body model — a pipeline artifact of
the split representation. With `presolve=False` it still reproduces (not a
presolve-elimination interaction). Folding the constant back into the body (native
form) removes the false primal. The tightening *transformation itself* is
feasible-set-equivalent: I verified `u_rest` is valid over the true feasible region
(the row whose declared box allows rest ≤ 7 has a true feasible max of **5.755**,
exactly the probed `u_rest`).

**Defect 2 — continuous-[0,1] misclassified as binary (latent unsoundness).** The
binary test `"IN" in str(vtype).upper()` is `True` for `"CONTINUOUS"`
("cont**IN**uous"), so a continuous [0,1] variable was treated as a binary indicator
and its coefficient tightened — which changes the row at fractional values and is only
sound for a genuinely integral variable.

## The fix

- **Write-back in native normal form** (`_root_presolve.py`): the entire tightened
  constant goes into the body and `con.rhs = 0`, so a rewritten row is byte-for-byte
  the shape `from_nl` emits — never a body-constant coexisting with a nonzero rhs.
- **Correct binary classification** (`_is_binary_var`): binary-like ⟺ an
  *integer-typed* (`BINARY`/`INTEGER`) variable with a [0,1] box; continuous [0,1] is
  excluded. Rows whose non-binary activity is not finite under the (valid) FBBT/probing
  bounds are skipped, never guessed.
- Savelsbergh (`a_k > 0`) and fixed-charge (`a_k < 0`) coefficient formulas retained
  (they were correct in isolation). `u_rest` uses FBBT + probing bounds valid over the
  feasible region.
- Gated by `SolverTuning.coef_tighten` / `DISCOPT_COEF_TIGHTEN` (default-OFF); wired
  once at the root in `solver.py` before dispatch (serves NLP-BB and spatial paths).
  Flag OFF is byte-identical (the call site does not run).

## The both-directions equivalence test (the #770 gap)

The #770 test only checked "no feasible point removed". New
`python/tests/test_coef_tighten_equivalence_774.py` adds the missing direction:

1. `test_optimum_preserved[gbd,gkocis,ex1221,st_test1,alan]` — solve each small
   vendored instance to optimality flag OFF vs ON; assert the certified optima are
   equal to the oracle and the ON incumbent is feasible in a fresh ORIGINAL model.
   Admitting an infeasible point moves the optimum one way; removing a feasible point
   moves it the other — either fails. **PASS.**
2. `test_no_infeasible_admitted_lp[7 instances]` — rigorous, non-sampling LP
   certificate over the declared box: for each binary corner, no point satisfying all
   tightened linear rows violates any original linear row (the #772 direction).
   **PASS.**
3. `test_continuous_unit_is_not_binary` — `_is_binary_var` classifies binary/integer
   [0,1] as binary and continuous [0,1] as NOT binary (Defect 2 guard). **PASS.**
4. `test_incumbent_feasible_in_fresh_original` (slow) — end-to-end on rsyn0805m: the
   ON incumbent is feasible in a fresh original model and never beats the oracle
   (**catches Defect 1**: the reverted code fails here at 1441.99 > 1296.12; the fix
   passes). **PASS.**

Fail-before/pass-after confirmed: the reverted pass yields obj 1441.99 on rsyn0805m
(test 4 fails); the corrected pass yields a valid incumbent ≤ oracle.

## Entry experiment / graduation verdict — HOLD (default-OFF)

`#282` panel, flag ON vs OFF, `time_limit=90s`, sound throughout (every dual bound
≥ `=opt=` for these MAXIMIZE instances; every incumbent ≤ `=opt=`;
`incumbent_verification_failed=false`; no `gap_certified` regressions — none certify
in either arm):

| instance   | oracle  | root_bound OFF | root_bound ON       | nodes OFF→ON | cert (OFF/ON) |
|------------|---------|----------------|---------------------|--------------|---------------|
| rsyn0805m  | 1296.12 | 2111.02        | 2111.02 (=)         | 895 → 735    | none / none   |
| rsyn0810m  | 1721.45 | 2963.21        | 2944.86 (↓ 0.6%)    | 575 → 671    | none / none   |
| syn40m     |   67.71 | 1833.91        | **1059.17 (↓ 43%)** | 1055 → 1375  | none / none   |
| syn05hfsg  |  837.73 | 1335.50        | 1335.50 (=)         | 185 → 185    | **opt / opt** |

(root_bound is the dual **upper** bound — these are MAXIMIZE — so a smaller value is
tighter; every ON bound stays ≥ `=opt=`, sound.) The tightening is sound and tightens
the root dual bound substantially on syn40m (~43% of the root excess above the oracle,
consistent with the #282 Stage 1 estimate), but certifies **nothing** additional on the
panel and does not reduce nodes (syn40m nodes actually rise). syn05hfsg is byte-identical
(both certify optimal in 185 nodes). Sound ≠ helpful (the `DISCOPT_CUT_INHERIT` lesson).
Per CLAUDE.md §5 it ships as a **default-OFF opt-in lever** with the HOLD recorded; it
does **not** graduate default-ON without a net-positive panel.

## Verification

- `pytest -m smoke` — 831 passed, 1 skipped, 2 xpassed.
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — 10 passed.
- `pytest python/tests/test_lazy_jax_linear_path.py` — 4 passed (no JAX import on the
  LP/MILP/QP/MIQP path).
- New `test_coef_tighten_equivalence_774.py` — 14 passed.
- `ruff check` / `ruff format --check` — clean on changed files. (`mypy` blocked by a
  pre-existing numpy-stub/env mismatch, reproduced on unchanged files.)

### DIFF
--- a/python/discopt/solver.py
+++ b/python/discopt/solver.py
@@ -5285,6 +5285,27 @@ def solve_model(
         except Exception as e:
             logger.debug("Root presolve failed: %s", e)
 
+    # --- Activity-based big-M coefficient tightening (#282/#774, opt-in) ---
+    # Shrinks big-M coefficients on binary-indicator rows toward the FBBT/probing
+    # activity slack. Feasible-set-equivalent (it strictly tightens the LP
+    # relaxation at fractional binaries without adding or removing any
+    # integer-feasible point), gated behind ``SolverTuning.coef_tighten`` /
+    # ``DISCOPT_COEF_TIGHTEN`` (default OFF, CLAUDE.md §5). Runs once at the root
+    # here, before dispatch, so it benefits BOTH the NLP-BB and the spatial B&B
+    # paths. Rewrites the Python constraint DAG in place in native normal form
+    # (constant in the body, rhs 0) — the #774 fix for the #770 false primal.
+    # Skipped once the budget is blown (#654): declining it leaves the original,
+    # still-valid rows.
+    if presolve and _tuning().coef_tighten and not _deadline_exhausted():
+        try:
+            from discopt.solvers._root_presolve import tighten_bigm_coefficients
+
+            n_ct = tighten_bigm_coefficients(model)
+            if n_ct > 0:
+                logger.info("Coefficient tightening: strengthened %d big-M rows", n_ct)
+        except Exception as e:
+            logger.debug("Coefficient tightening failed: %s", e)
+
     # --- Reverse-AD interval tightening (M9 of #51, opt-in) ---
     # Iterates Gauss-Seidel reverse-mode interval AD over every
     # constraint to a fixed point and writes back tighter scalar bounds.
--- a/python/discopt/solver_tuning.py
+++ b/python/discopt/solver_tuning.py
@@ -511,6 +511,28 @@ class SolverTuning:
     everything else bound-neutral. A/B root values re-confirmed load-independent
     (``results/issue282/root_lp_probe_ab_reconfirm_*.json``)."""
 
+    coef_tighten: bool = field(
+        default_factory=lambda: _env_flag("DISCOPT_COEF_TIGHTEN", default=False)
+    )
+    """Activity-based big-M coefficient tightening at the root (issue #282/#774,
+    ``DISCOPT_COEF_TIGHTEN``, **default OFF**; §5 bound-changing, opt-in).
+
+    Rewrites each linear inequality carrying a binary indicator, shrinking the
+    binary's coefficient toward the FBBT/probing activity slack of the rest of the
+    row (Savelsbergh positive-coefficient + fixed-charge negative-coefficient
+    cases). Feasible-set-**equivalent** — no integer-feasible point is added or
+    removed — so the LP relaxation tightens at fractional binaries while every
+    integer corner is preserved. The rewrite is emitted in ``from_nl``'s native
+    normal form (constant in the body, ``con.rhs == 0``); the #770 revert (#773)
+    was a malformed write-back that split the constant across body and rhs and
+    admitted a false primal (rsyn0805m 1441.99 > =opt= 1296.12), re-derived and
+    fixed in #774.
+
+    Default OFF pending a net-positive graduation panel: #282 Stage 4 measured it
+    tightens the root (syn40m ~56% of the discopt→SCIP spread) but certifies
+    nothing extra on the panel, so it ships as a sound opt-in lever (HOLD, sound ≠
+    helpful — the DISCOPT_CUT_INHERIT lesson) rather than default-ON."""
+
     rlt1_root_bound: bool = field(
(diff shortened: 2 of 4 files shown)


jkitchin:
Closing as a duplicate: #780 (merged earlier today) landed the same #774 item-1 re-implementation with the same root-cause findings — the write-back violating the `Constraint.rhs == 0.0` normalized-form invariant (consumers read the rewritten row relaxed), plus the `"IN" in "contINuous"` classifier bug — along with the consumer-convention bidirectional LP-equivalence gate (`python/tests/test_coef_tighten_774.py`), the 66-instance flag-ON soundness sweep (0 violations), and the panel A/B record. #774 is closed.

This PR was produced by a parallel worker from a stale queue; its independent confirmation of the same diagnosis is recorded here for the archaeology. One nuance worth keeping: its graduation-panel HOLD measurement (syn40m root dual 1833.91 → 1059.17, certifies nothing additional, nodes not reduced) independently confirms the sound-≠-helpful disposition — consistent with #780's decision to stay default-OFF.

Issue #807, opened 2026-07-20
Convex kernel — native-warm-LP architecture (follow-up to #800)


jkitchin:
## Native-warm-LP architecture for the convex kernel (follow-up to #800)

Split from #800, which falsified all three of its planned per-node/search levers
(row-order T1, strong branching T2, dual-feasible warm restart T3) by entry
experiment on the real `rsyn*`/`syn*` panel. Full records in
`docs/dev/convex-kernel-plan.md` (work log, 2026-07-20 entries). #800 closed on its
scoped win: the panel **certifies cert-clean within the 120 s budget** and is
**net-positive vs NLP-BB** (kernel ~24 s vs NLP-BB 482 s uncertified). What #800
did NOT reach is **SCIP-parity wall** (SCIP ~2 s vs kernel 7–80 s/instance) and the
**K2 node-count gate** (≤2× prototype 67/60/46/55; measured 2.5–6.1×).

### The falsified-into-focus hypothesis

Per #801's close ("BARON wins by running the tree ~500×/node faster — **native warm
LPs** + reduced-cost range reduction — the only deficit is per-node throughput") and
#800's T3 architectural finding: **the current kernel re-assembles the base LP per
node and cold-solves the first solve.** T3 proved you cannot bolt cross-node warm
starts onto this design —

- the parent's cut-converged optimal basis (where the branch var is basic-fractional
  → dual-feasible for the child) lives on a **different, larger LP** than the child's
  base-only first solve; and
- the parent's base-only basis (right dimensions) is **neither dual-feasible for the
  child nor bound-neutral** to warm from — warming it drifts `node_count` (lands on a
  different degenerate vertex; the T1 degeneracy lesson) and its overhead exceeds the
  cold-solve cost (reproduces iter-8).

So native warm LPs require an **architectural change**, not a knob.

### Scope

Rewrite the node loop to keep **one persistent LP** (structural columns + a growing
cut pool) and **modify variable bounds in place** across the whole tree, dual-warm
reoptimizing at each node from the parent's optimal basis (the SCIP/BARON design),
instead of re-assembling + cold-solving per node. Candidate second lever, also from
#801: **reduced-cost range reduction** (reduced-cost-based bound tightening) at each
node.

### Guardrails (unchanged from #798/#800)

- **Soundness is non-negotiable** (CLAUDE.md §1/§5): `incorrect_count ≤ 0`, dual
  bound = oracle optimum exactly on the panel, every incumbent #779-verified, honest
[...]

Issue #801, opened 2026-07-20
tanksize root relaxation — RLT-2 / pooling-PQ (research-grade; split from #798 §3)


jkitchin:
Split from the #798 §3 secondary threads. The small default-ON *decision* thread
was pulled out into **#802** (measurement-first); **this issue is now just the
research-grade `tanksize` root relaxation.**

## `tanksize` root relaxation (research-grade)

The root bound is `lb(x17)` (the objective variable's FBBT bound, **0.838**) vs
BARON's **0.955** — a large root gap that keeps the tree from closing quickly.

Already ~no-op or falsified (**do NOT re-walk**):
- McCormick / RLT-1 / OBBT / Shor: ~no-op on the root bound.
- Partition / piecewise-McCormick: **FALSIFIED** (#764 comment).

**Residual lever (entry-experiment-first, on the REAL instance):** RLT-2 /
fractional-envelope / pooling-PQ on the dense-bilinear structure. Run the
falsifying experiment on the real `tanksize` instance BEFORE any implementation —
the #727 RLT lesson: synthetic root-gain 0.68, real gain 0.0. State the
hypothesis, the measurement, and the kill criterion first (CLAUDE.md §4).

Reference: `docs/dev/convex-kernel-plan.md` §3; #764; results in
`discopt_benchmarks/results/`.

jkitchin:
## Stage 1 reconciliation — the lever list in the issue body is partly stale

Before running anything I reconciled #801's named residuals ("RLT-2 / fractional-envelope / pooling-PQ") against the committed falsification record in `docs/dev/issue-764-root-relaxation-plan.md` ("Candidate mechanisms, ranked"). Two of the three are already dead on the **real** instance, and one is structurally vacuous as literally stated:

- **Pooling-PQ / level-1 RLT — already FALSIFIED** (entry experiment 2026-07-18, #764 doc). The *full* level-1 RLT closure (2190 product rows, 2170 lifted columns, sparse solve) left the root **bit-identical**: 0.8382 → 0.8382. Construction validated by injection (`x17 ≥ 1.0` moves the root to 1.0, so the rows genuinely reach the LP). Decisive sub-finding: with all 9 integers fixed to their optimal leaf the root is still 0.8382, so the entire gap is **continuous×continuous bilinear**, not integrality. PQ is a special case of level-1 RLT; re-running it would re-walk a binding negative (CLAUDE.md §4).
[...]

Issue #800, opened 2026-07-20
Convex LP-OA kernel — SCIP-parity on larger convex MINLPs (follow-up to #798)


jkitchin:
## Convex LP-OA kernel — SCIP-parity on larger convex MINLPs (follow-up to #798)

The convex LP-OA branch-and-cut kernel (#798, PR #799) is complete, correct, and
cert-clean, and it is **scoped to smaller convex MINLPs** where it is already
SCIP-competitive (syn05m/10m/15m certify in 0.6–1.1 s, in SCIP's 0.5–0.8 s range).
This issue tracks closing the gap to SCIP/BARON wall-clock parity on the **larger**
instances of the convex `rsyn*`/`syn*` family, which the kernel currently either
certifies slowly or declines.

### Measured gap

> **Measurement config (read first).** The wall numbers below are from the
> **unseeded production path** (`try_convex_solve`, post the iter-10 soundness fix).
> They do **not** reconcile with the plan's iter-9 figure (the 4-instance panel at
> 24.4 s), which is the **seeded** `issue798_k2_tree_gate.py` config that isolates
> the dual side (nodes-to-certify) and pre-dates the iter-10 fix. These are different
> configs and must not be compared directly. Plan task **T0 (blocking)** pins a single
> canonical two-config baseline (unseeded production wall+latency; seeded
> nodes-to-certify) before any optimization, so every later ratio is anchored — treat
> the table below as indicative, not the anchor.

| instance | kernel wall | SCIP target | ratio |
|----------|-------------|-------------|-------|
| syn05m / syn10m / syn15m | 0.6 / 0.8 / 1.1 s | 0.5–0.8 s | ~1–2× (in range) |
| syn20m | 1.8 s | — | ~2–3× |
| rsyn0805m / rsyn0810m / rsyn0815m | 8–10 s | ~0.5–0.8 s | ~10–15× |
| syn40m | 37 s | — | ~50× |
| rsyn0820m / rsyn0830m | — | — | fall back (uncertified) |

### Known levers and status

- **K2 node-count gate is NOT met** on the panel: nodes-to-certify 353/177/281/139
  vs the prototype's 67/60/46/55 — **2.5–5.3×**, not the ≤2× bar. Node count is the
  dominant driver of the large-instance wall.
- **Falsified (do not re-walk without new evidence):**
  - reducing `max_sep_rounds` — wall is ~invariant across sr∈{2,4,6,12} (nodes trade
    for cuts at equal LP work).
[...]

PR #808, opened 2026-07-20
docs(#807): loop-executable native-warm-LP implementation plan (W0–W5)

jkitchin:
Loop-executable implementation plan for #807 (native-warm-LP convex kernel), `docs/dev/native-warmlp-plan.md`. Doc-only.

Mirrors the #800/#804 pattern: a self-contained spec an agent loop reads each iteration — pinned anchors inline, correctness contract, reusable assets, binding negatives, and tasks W0–W5 each with hypothesis / entry experiment / kill criterion / verification regime / done-when.

Key design points:
- **The enabling fact:** dual feasibility is bound-value independent in a *shared* LP (`PreparedDual` contract) — exactly the property #800-T3's per-node-reassembly design lacked. One LP + bounds-in-place makes every node's basis a valid dual-warm start for any other node.
- **C-43 made structural:** OA tangents are globally valid (permanent pool — amortizes OA convergence across the tree); GMI/cover/MIR cuts are box-tagged, active iff node box ⊆ derivation box, deactivated by a slack-cap bound write (no structural change, NS stays finite).
- **W0 is blocking:** a reoptimize-vs-cold probe on real instances (rsyn0815m, syn40m) whose kill falsifies the whole issue — no architecture is built before the primitive is proven ≥2× on the median child.
- **W4 = RCRR** (#801's second BARON lever, the one node mechanism not falsified by #800), with a differential oracle-in-box soundness guard; **W4b** allows the T2 strong-branching revisit only once cheap warm probes exist, per T2's own record.
- **W5 surfaces the K2 node-gate re-framing to the owner** (recorded vs gating) rather than silently dropping it.

Verification: doc-only; no code paths touched.

### DIFF
--- a/docs/dev/native-warmlp-plan.md
+++ b/docs/dev/native-warmlp-plan.md
@@ -1,0 +1,382 @@
+# Native-Warm-LP Convex Kernel — implementation plan (issue #807)
+
+**Status:** not started. This is the durable, loop-executable spec for the #807
+rewrite: one persistent LP, bounds modified in place across the whole tree,
+dual-warm reoptimized per node — the SCIP/BARON per-node-throughput architecture.
+It survives context loss: each work iteration reads this doc in full, takes the
+first unfinished task, runs its entry experiment BEFORE implementing (CLAUDE.md
+§4), honors the kill criterion, and records measured results in the work log
+(newest first). It is self-contained: every anchor number a task needs is in §2.
+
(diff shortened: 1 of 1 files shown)


jkitchin:
Superseded by #809, which contains this plan doc plus the full W0-W2 findings and the native-warm-LP implementation. Closing without merging to avoid a duplicate; the plan lands via #809.

Issue #817, opened 2026-07-21
discopt returns no incumbent on 30 instances SCIP solves easily (primal/feasibility gap; pooling/perspective/facility families)


jkitchin:
## Summary

On **30 instances SCIP solves easily**, discopt returns **no usable primal solution** — it either hits the time limit with no incumbent, or returns `status=unknown` and gives up. A global solver that returns *nothing* on an instance a peer solves in well under a second is the most severe of the performance/capability gaps surfaced by the discopt-vs-SCIP corpus fuzz sweep (454/500 instances, time_limit=15s). **Not a soundness bug** — no wrong answer is returned; the gap is that no answer is returned.

## Two failure modes

**A1 — timed out, no incumbent found (29):** the primal side (heuristics / feasibility restoration) never produced a feasible point within the budget, while SCIP found the optimum fast. Sorted by how badly discopt overran:

| instance | discopt | scip | opt |
|---|---|---|---|
| squfl025-040persp | time_limit 126.0s | optimal 10.21s | 197.3338812 |
| gastrans582_cool12_95 | TIME_LIMIT 75.0s | optimal 8.34s | 0.0 |
| gastrans582_mild11 | TIME_LIMIT 75.0s | optimal 12.46s | 0.0 |
| gastrans582_warm31 | TIME_LIMIT 75.0s | optimal 5.50s | 0.0 |
| rsyn0805m04hfsg | TIME_LIMIT 75.0s | optimal 5.73s | 7174.219035 |
| rsyn0810m04hfsg | TIME_LIMIT 75.0s | optimal 7.05s | 6581.934408 |
| watercontamination0202 | TIME_LIMIT 75.0s | optimal 7.26s | 125.1956151 |
| chimera_k64ising-01 | TIME_LIMIT 75.0s | optimal 12.00s | 24.3 |
| chimera_k64ising-02 | TIME_LIMIT 75.0s | optimal 11.33s | 24.3 |
| squfl020-050persp | time_limit 75.0s | optimal 2.71s | 230.2021495 |
| emfl100_3_3 | time_limit 57.4s | optimal 4.81s | 18.13236088 |
| squfl010-080persp | time_limit 48.7s | optimal 2.88s | 509.7060216 |
| pooling_foulds4stp | time_limit 35.6s | optimal 6.00s | -8.0 |
| pooling_foulds3stp | time_limit 35.3s | optimal 2.77s | -8.0 |
[...]

Issue #815, opened 2026-07-20
False primal: presolve/heuristic emits an infeasible incumbent on emfl050_3_3 (contained by #779 guard; underlying mutation unsound)


jkitchin:
## Summary

On `emfl050_3_3` a **presolve mutation or primal heuristic emits an incumbent
that is INFEASIBLE in the original problem** (objective = 594.0). discopt's #779
incumbent-verification guard **correctly caught it, withheld the incumbent, and
decertified** — so the *returned* result stays sound (`obj=None`,
`status=time_limit`). This issue is about the **underlying unsound mutation**, not
the guard: the guard is the last line of defense, and per the correctness-first
philosophy (CLAUDE.md §3) we root-cause the class rather than rely on the
band-aid. If a future change ever weakens the guard, or hits an instance where the
guard's own feasibility check has a blind spot, the same mechanism would return a
**false optimum** — the cardinal soundness failure.

Discovered during the #814 time-limit repro; surfaced by the discopt-vs-SCIP
corpus fuzz sweep.

## Repro (direct, no daemon)

```
python -c "
import discopt.modeling as dm
m = dm.from_nl('~/Dropbox/projects/discopt-minlp-benchmark/minlplib/nl/emfl050_3_3.nl')
r = m.solve(time_limit=5)
print(r.status, r.objective)"
```

Emits to stderr:

```
FALSE PRIMAL DETECTED: the reported incumbent (objective=594.0) is INFEASIBLE in
the original problem. This indicates an unsound presolve mutation or a heuristic
bug. Withholding the incumbent and decertifying — the result is NOT a valid
solution.
```

and returns `status=time_limit obj=None` (incumbent withheld).

## Context clue

`emfl050_3_3` is a **pure-continuous** facility-location model (1611 vars, 0
integer/binary). The solve also warns:

```
Variables with very large or infinite declared bounds: x0..x4 (lb=0, ub=inf).
Nonlinear tightening can adjust 36 bounds via defined_variable_forward.
```

So a prime suspect is the interaction of **unbounded variables (ub=inf)** with a
presolve bound mutation (`defined_variable_forward` tightening) and/or a
continuous primal heuristic (root multistart / local NLP) producing a point that
satisfies the *mutated* model but violates the *original* constraints.

## Root-cause direction (entry experiment)

1. Bisect which stage emits the 594.0 incumbent: disable primal heuristics
   (`continuous_multistart`, pump/diving) and re-run — does the false primal
[...]

Issue #814, opened 2026-07-20
Solver overruns wall-clock time_limit by 2–5× on large instances (gastrans582/emfl/crudeoil families)


jkitchin:
## Summary

On several large-instance families, `Model.solve(time_limit=T)` overruns the
requested wall-clock limit by **2–5×**, and in one family ignores it entirely
until an outer hard-kill. The certificate remains valid (this is a **robustness /
UX bug, not a correctness bug**), but a solver that does not honor its own wall
budget is unreliable for time-limited deployment and benchmarking.

Surfaced by a discopt-vs-SCIP corpus sweep (`fuzz_discopt_scip_loop.py`,
time_limit=15s). SCIP honors the limit on every one of these; discopt does not.

## Evidence (sweep, requested time_limit = 15 s)

| instance | discopt wall | discopt status | scip wall |
|---|---|---|---|
| gastrans582_cool12_95 | **75.00 s** | TIME_LIMIT (outer hard-kill) | 8.3 s |
| gastrans582_cool14_95 | **75.00 s** | TIME_LIMIT (outer hard-kill) | 15.1 s |
| gastrans582_mild11    | **75.00 s** | TIME_LIMIT (outer hard-kill) | 12.5 s |
| gastrans582_mild11_95 | **75.00 s** | TIME_LIMIT (outer hard-kill) | 15.1 s |
| gastrans582_warm31    | **75.00 s** | TIME_LIMIT (outer hard-kill) | 5.5 s |
| emfl100_3_3           | 57.40 s | time_limit | 4.8 s |
| emfl050_3_3           | 32.62 s | time_limit | 5.8 s |
| crudeoil_lee1_07      | 30.71 s | time_limit | 15.2 s |
| crudeoil_lee1_06      | 25.80 s | time_limit | 15.0 s |

Plus ~27 instances in the (15 s, 20 s] band (marginal — plausibly coarse-check
granularity / socket overhead). The 9 above are unambiguous.

The `75.00 s` figure is exactly the sweep harness's outer `hard_deadline = tl + 60`:
the in-process solve never returned on the `gastrans582` family and had to be
killed externally — i.e. the internal time limit was **not honored at all** on
that family.

## Confirmed direct (no-daemon) repro

```
python -c "
import time, discopt.modeling as dm
m = dm.from_nl('~/Dropbox/projects/discopt-minlp-benchmark/minlplib/nl/emfl050_3_3.nl')
[...]

PR #824, opened 2026-07-21
feat(#817): sane-start multistart for the continuous NLP path (default-off)

jkitchin:
Contributes to #817 (the continuous-single-NLP primal increment).

## Root cause (entry experiment)

The "no incumbent" gap on the pooling / facility / perspective family traces to
the **NLP start point**, not the NLP. `_solve_continuous` starts the local NLP at
the box-center clipped to `|x| <= _X0_CLIP = 10`. On models with **unbounded
variables** that magnitude sends the NLP into a bad, infeasible basin — it stalls
at an infeasible iterate:

| instance | box-center (cap 10) | cap 0.5 start |
|---|---|---|
| emfl050_3_3 | TIME_LIMIT obj=594, **infeasible** | OPTIMAL obj=**10.40**, feasible |
| pooling_foulds3tp | TIME_LIMIT obj=−320, **infeasible** | OPTIMAL obj=**−8.0**, feasible |
| squfl010-080persp | TIME_LIMIT obj=217k, **infeasible** | OPTIMAL obj=**503.6**, feasible |

The same NLP from `|x| <= 0.5` converges to the feasible optimum. SCIP solves all
of these in <1s.

## Fix (behind `DISCOPT_SANE_MULTISTART`, default OFF)

The continuous path tries a few start magnitudes (caps **0.5, 1.0, 10.0**),
**smallest-first** (they reach the feasible basin fastest), each with the full
remaining budget, **early-stopping** once a start lands a VERIFIED-feasible point
(same loose 1e-3 check the #772 guard uses). The `|x|<=10` start stays last so
nothing that worked before is lost.

- **Sound / primal-only**: it can only find a feasible incumbent that would
  otherwise be missed; the dual bound / certificate are untouched.
- **Flag OFF is byte-identical** to the prior single-start behavior.
- **No over-multistart**: finite-bounded controls early-stop on the first
  converged start → same optimum ON vs OFF.

## Measured (flag ON)

- `pooling_foulds3tp` → **optimal −8.0**; `emfl050_3_3` → **feasible 10.402**
  (optimal at a longer budget); `squfl010-080persp` / `gastrans040` → feasible.
- Controls `ex4_1_1` (−7.487) / `ex8_1_1` (−2.022) / `ex14_1_2` (0.0) unchanged.
- bucket2 sound-bounds (nvs16) + #815 tests: 21 passed (flag off).

## Scope — what this does NOT yet cover (still open on #817)

1. The **spatial-path** primal (instances routed to spatial B&B, not
   `_solve_continuous`).
2. **Slow root setup** on large models — emfl needs ~13s of NLP time after a ~6s
   setup, so at a 12–15s limit it still times out (the #814 / throughput coupling).
3. `jit1`'s `unknown` structural refusal.

Graduation to **default-ON** is a follow-up after the corpus-wide §5 differential
panel (cert-clean + net-positive).

## Tests

`python/tests/test_817_sane_multistart.py` (slow, corpus-gated): flag ON finds the
pooling −8.0 and emfl 10.402 optima (emfl incumbent verified feasible) where flag
OFF finds none; a finite-bounded control returns the same optimum ON vs OFF.


jkitchin:
## §5 graduation panel — stays default-OFF

Ran the differential panel: **34 pure-continuous instances, tl=25s, OFF vs ON**
(`discopt_benchmarks/results/issue817_multistart_panel_20260721.jsonl`).

**Cert-clean: PASS**
- 0 false primals (lip's ON obj is *below* the oracle but lip is MAXIMIZE, so it's
  a valid worse feasible point, not unsound).
- Bounds identical OFF vs ON (primal-only change, as designed).
- 0 real certification regressions; 30 finite-bounded controls byte-identical.

**Net-positive: WEAK / effectively neutral — does NOT clear the bar**
- Only **1 of 14** gap instances gained an incumbent (`pooling_foulds4tp`:
  None → optimal −8.0).
- Gains are **timing-flaky**: `pooling_foulds3tp` gains at tl=18 but not tl=25.
- One **minor quality regression**: `lip` (MAXIMIZE) — ON's early-stop took the
  first small-cap feasible max (5.668e6) instead of OFF's better one (5.685e6).
  Sound, but a worse incumbent.
- 13/14 gap instances still return nothing because they are **setup-throughput-
  bound (#814)** — they overran to 30–55s under a 25s limit and never finished, so
  the fix has no room to help.

## Decision

Per the `DISCOPT_CUT_INHERIT` lesson (sound ≠ helpful; a cert-clean but
neutral-or-harmful flag stays OFF, measurement recorded), the flag **stays
default-OFF**. The mechanism is correct and the root cause (NLP start magnitude)
is real, but its benefit is masked by #814 and slightly undercut by the early-stop.

**Before graduating, two follow-ups (recorded on the flag docstring):**
1. Address the #814 setup/throughput coupling so the gap instances actually finish.
2. Remove the early-stop quality regression (try all sane starts, keep the best,
   rather than stopping on the first feasible-optimal).

This PR remains valuable as the flag-gated, panel-recorded increment (Contributes
to #817); it is safe to merge default-OFF or hold pending the follow-ups — your call.

jkitchin:
## Closing — superseded by pounce#242

The clean graduation panel (34 pure-continuous instances, tl=25, OFF vs ON, run with the rebuilt pounce#242) came back **cert-clean but 0 incumbents gained** — because **pounce#242 made the baseline solve the very instances this PR targeted**:

| instance | OFF (baseline) | ON (#824) |
|---|---|---|
| emfl050 | optimal 10.40 | optimal 10.40 |
| pooling_foulds3tp/4tp/3pq | optimal −8.0 | optimal −8.0 |

Pre-pounce-fix these returned None (the gap this PR closed); post-fix the baseline finds them with no help from the sane-start.

**Root cause was never the start point** — it was that the NLP from the box-center start entered pounce's unbounded restoration grind and never converged. #824's sane-start was a *workaround* that dodged the bad basin; **pounce#242 fixed the actual bug**. So #824 adds no measured value and carries flag-off-parity fragility (it broke a 1e-9 tolerance test in CI). Closing without merging.

The remaining bad-start init/restoration stall is tracked in **pounce#246**. The finding + panel are recorded on the flag docstring and in `discopt_benchmarks/results/`.

Not deleting the branch (`probe/817-primal-gap`) in case the sane-start is wanted later as an opt-in, but it is not going to main.

Issue #827, opened 2026-07-21
#817-C/D: continuous primal-hard cases (ball_mk2 thin shell, chimera Ising)


jkitchin:
Sub-issue of #817 (families C + D). Two pure-continuous structures where the
primal never lands a feasible point, distinct from the #824 NLP-start family.

## D — `ball_mk2_30` (thin feasible shell)

30 continuous vars, **1** constraint (`sum_i x_i^2 - 0.995825 x_i <= 0`), opt on
the constraint boundary; SCIP 0.02s, discopt 5741 nodes / no incumbent.

- **Verified sound**: x=0 is feasible; the "proved infeasibility" tightening
  messages correctly prune infeasible sub-boxes; the model is never falsely
  declared infeasible.
- The primal never samples the thin feasible shell. The origin/box-center IS
  feasible here, so a cheap "try obvious points" primal seed (evaluate a handful
  of candidate points, inject any verified-feasible one) would catch it — and is
  a generally-useful heuristic.

## C — `chimera_k64ising-01/02` (Ising / quadratic)

~1200 continuous vars, **0** nonlinear constraints, quadratic Ising objective;
SCIP ~12s, discopt 75s (outer hard-kill) / no incumbent. A large box-QP-like
structure — different from the facility/pooling family; likely needs the QP /
spatial path rather than the single-NLP path.

## Entry experiments

- ball_mk2_30: add a trivial-point primal seed (origin, box-center, bound corners,
  each clipped into the box), inject the first verified-feasible one; confirm it
  solves ball_mk2/mk3/mk4 without regressing others.
- chimera: profile where the 75s goes (it hits the outer hard-kill — overlaps
  #814) and whether the QP path or a rounding primal finds a spin assignment.

Not soundness bugs — honest "no incumbent" failures.

jkitchin:
Parent umbrella: #817.

jkitchin:
**Family D (ball_mk2): addressed by #829** — trivial-point primal seed (flag-gated `DISCOPT_TRIVIAL_PRIMAL`, default-off). Root cause was mis-stated in the filing: ball_mk2_30 is **30 integer vars in [-1,1]** (not continuous), and its ball constraint is integer-feasible only at the origin (the optimum). The seed sets a verified-feasible trivial point as `initial_point`, which the sub-solver injects. ball_mk2_30: OFF none → ON feasible obj 0.0; 7 controls unchanged.

[...]

PR #829, opened 2026-07-21
feat(#827): trivial-point primal seed (default-off) — finds ball_mk2 incumbent

jkitchin:
Addresses #827 family D (ball_mk2). Contributes to #817.

## Root cause

`ball_mk2_30` is **30 integer variables in [-1,1]** (the sizes CSV's "discrete=0"
is misleading — all vars are `VarType.INTEGER`). Its single ball constraint
`sum_i(x_i^2 - 0.995825 x_i) <= 0` is integer-feasible **only at the origin** (any
`x_i = ±1` makes the sum strictly positive), so the origin is the unique feasible
point and the optimum (obj 0). discopt's B&B never samples it → **no incumbent**;
SCIP finds it in 0.02s.

## Fix (behind `DISCOPT_TRIVIAL_PRIMAL`, default OFF)

When no warm start is given, `solve_model` evaluates a handful of trivial
candidates (origin projected into the box, box-center, bound corners), verifies
each against the true model, and sets the best feasible one as `initial_point`.
The sub-solvers re-verify **both constraint and integer feasibility** before
injecting it as an incumbent, so a non-integer candidate is simply filtered.

- **Sound / primal-only**: only ever seeds a verified-feasible incumbent (and only
  when none is given); dual bound / certificate untouched.
- **Flag OFF byte-identical**; 7 controls (continuous + integer) return identical
  results ON vs OFF.

## Measured (flag ON)

`ball_mk2_30` → `feasible obj 0.0` (the origin = optimum), where flag OFF returns
no incumbent.

## Scope

- **Covers** #827 family D (ball_mk2).
- **Does NOT cover** family C (`chimera_k64ising`): it routes to `_solve_milp_bb`,
  which doesn't accept/seed `initial_point`, and it also hits the #814 hard-kill.
  Deferred — needs `_solve_milp_bb` `initial_point` support + #814. Noted on #827.
- `ball_mk3_10` returns `unknown` (structural refusal like jit1/#828), out of scope.

Graduation to default-ON follows the corpus-wide §5 panel.

## Tests

`python/tests/test_827_trivial_primal.py` (slow, corpus-gated): flag ON finds the
ball_mk2 origin (obj 0) where OFF finds none; a control returns the same optimum
ON vs OFF.

### DIFF
--- a/python/discopt/solver.py
+++ b/python/discopt/solver.py
@@ -409,6 +409,28 @@ def _obbt_topk_enabled() -> bool:
     return os.environ.get("DISCOPT_OBBT_TOPK", "").strip().lower() in ("1", "true", "yes", "on")
 
 
+def _trivial_primal_enabled() -> bool:
+    """Whether the #827 trivial-point primal seed is on (env flag, **default OFF**
+    pending the §5 differential panel).
+
+    Some models' feasible regions include an OBVIOUS point that the primal
+    heuristics never sample: ``ball_mk2_30``'s optimum is the origin (the single
+    ball constraint holds at 0), and ``chimera_k64ising-*`` has zero nonlinear
+    constraints so any box point is feasible — yet both return NO incumbent while
+    SCIP solves them instantly. When ON, the root evaluates a handful of trivial
+    candidates (origin projected into the box, box-center, all-lb, all-ub), verifies
+    each against the true constraints, and injects the best VERIFIED-feasible one.
+    Primal-only: it only ever seeds a feasible incumbent (and only when none exists
+    yet), so the dual bound / certificate are untouched. Gated to pure-continuous
+    models (a trivial point need not be integer-feasible)."""
+    return os.environ.get("DISCOPT_TRIVIAL_PRIMAL", "").strip().lower() in (
+        "1",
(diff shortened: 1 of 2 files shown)


jkitchin:
**Extended to close #827 family C (chimera).** `_solve_milp_bb` silently ignored `initial_point` — a latent gap. It now accepts and seeds a verified-feasible (integer + constraint) warm-start incumbent. So the trivial-point seed now reaches the MILP/MIQP path, and chimera_k64ising-01 returns a feasible incumbent (obj 0, the trivial floor) where OFF returns nothing.

- **General win**: user warm-starts to MILPs are now honored (always-on; only ever adds a verified-feasible incumbent).
- **Sound**: 0 ≤ chimera's MAXIMIZE optimum 24.3; 4 fast MILP controls (nvs03/gbd/ex1221/st_e13) identical ON vs OFF.
- **Honest scope**: this closes the 'no incumbent' symptom for family C, but the incumbent is a floor, not the optimum — a good Ising primal + #814 (chimera never honors its wall budget) remain for a strong result.

This PR now covers #827 families **D (ball_mk2)** and **C (chimera)**.

Issue #832, opened 2026-07-21
Root-relaxation fallback overruns time_limit ~5× on gastrans582 — unbudgeted Python DCP relaxation build (not Rust LP; #814 residual)


jkitchin:
Split out from #814 (the remaining discopt-side `gastrans582` overrun sub-cause,
after the eigvalsh fix #831 landed).

## Problem

`SimplexOptions::deadline` is honored inside the **pivot loop** (polled every 256
pivots — `lp/simplex/primal.rs:1427`, `lp/simplex/dual.rs:492`), but **not during
the pre-iteration setup**: the initial LU factorization / basis build / presolve of
a large sparse LP runs *before the first pivot*, so a single root LP relaxation
solve grinds far past its deadline.

## Evidence

`gastrans582_mild11` (2186 structural + McCormick-aux cols, 3732+ rows). The
deadline-exhausted root-relaxation fallback (`_root_relaxation_lower_bound`) grants
the LP solve a ~1s budget (derived from `_ROOT_FALLBACK_FLOOR_S = 3.0`), but:

```
_root_relaxation_lower_bound(budget=3.0s) -> took 16.4s -> bound=None
```

`faulthandler` samples at 12s and 20s both land in `solve_milp` (the Rust call),
and the simplex pivot-loop poll never fires because the time is in setup, before
the first pivot. Net: ~16s over a 3s budget, and no bound produced.

## Request

Make the LP solve deadline-aware end-to-end, not just in the pivot loop:
- Check `deadline` at the start of / during the factorization + presolve (abort a
  too-slow factorization and return `IterLimit` / a deadline status with whatever
  bound is available), or
- Bound/short-circuit presolve when the deadline is already close.

Goal: a solve handed a 1s deadline returns in ~1s (+ bounded slack), even on a
large ill-conditioned LP, instead of running to completion.

## Soundness

Deadline-driven early exit must return a **valid** status (limit / no-bound), never
a truncated-but-claimed-optimal result. The caller (`_root_relaxation_lower_bound`)
already treats a missing/limited solve as "no bound" (sound: only ever weakens a
still-valid bound, never falsifies), so returning early with no bound is safe.

## Context

Last unbounded window in the #814 timing map. Pounce-side timing is tracked in
pounce#242/#244/#246; discopt root-build eigvalsh was #831 (merged); this is the
discopt root-LP-setup piece. A pure-Python interim mitigation (have the
deadline-exhausted fallback skip the root-relaxation solve when the relaxation is
[...]

PR #839, opened 2026-07-21
feat(#827): graduate DISCOPT_TRIVIAL_PRIMAL to default-ON (#829)

jkitchin:
Graduates the #829 trivial-point primal seed to **default-ON** after its §5 graduation gate passed.

## Gate evidence
`graduation_gate.py --flags trivial_primal --n 40 --seed 0` → **eligible=YES**: `soundness_ok=True` (0 violations — a verified-feasible seed can never falsify a bound), `cert_neutral=True` (0 cert violations), benefit **45%**, regression 7%. Report: `reports/graduation_gate_2026-07-21T18-30-00.{json,md}`.

## Change
- `solver.py`: `_trivial_primal_enabled()` default off→on (set `DISCOPT_TRIVIAL_PRIMAL=0` to opt out).
- Resolves **#827 families D+C on the default path**: ball_mk2_30 (no incumbent → optimal 0.0) and chimera_k64ising (no incumbent → feasible) now surface an incumbent without opting in.
- `generality_sweep.py`: registers the `trivial_primal` gate arm.
- Tests: opt-out (`=0`) still finds no ball_mk2 incumbent; the graduated default (unset) now does; control unchanged; chimera sound.

## Verification
- 4/4 `test_827_trivial_primal.py` pass; default confirmed ON, `=0` opt-out confirmed OFF; ruff clean.

`Contributes to #827` — the chimera incumbent *quality* (a real Ising primal) remains.

### DIFF
--- a/discopt_benchmarks/scripts/generality_sweep.py
+++ b/discopt_benchmarks/scripts/generality_sweep.py
@@ -177,6 +177,18 @@ ARMS: dict[str, dict] = {
     # or trips the ``_objective_bound_valid`` gate to None — so it can change (weaken)
     # a fallback bound: ``bound_changing``. No cheap static proxy — the overrun is a
     # runtime build-cost property (like lu_density_route / node_numerical_dual_bound).
+    # Trivial-point primal seed (#827/#829): at the root, evaluate a handful of
+    # obvious candidate points (origin-in-box, box-center, all-lb, all-ub), verify
+    # each against the TRUE constraints, and inject the best verified-feasible one as
+    # a warm-start incumbent when none exists yet. Primal-only and soundness-safe (a
+    # verified-feasible point can never falsify a bound), but it seeds an incumbent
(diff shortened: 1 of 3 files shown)


jkitchin:
Closing: graduating DISCOPT_TRIVIAL_PRIMAL to default-ON surfaced 13 CI failures, including a **false optimum** on the fast=True assignment MILP (returns obj 0.0 all-zeros vs the true 9). Root cause filed as https://github.com/jkitchin/discopt/issues/840: fast=True indexed constraints are absent from model._constraints, so the trivial-primal seed's feasibility check (and the #772 guard) see NO constraints and accept an infeasible all-zeros seed. Plus a JAX cold-start regression (the seed runs before the LP/MILP class dispatch and imports JAX). The flag stays default-OFF until the root cause is fixed. #832 graduation (#838) is unaffected and merged.

Issue #862, opened 2026-07-26
#844 fallback incumbents are far from optimal (tln6 65.3 vs opt 15.3) and unmeasured


jkitchin:
> Found while finishing #844 (PR #858).

## Gap

The #844 fallback now returns an incumbent where the default path returned none — but the incumbents are **far from optimal**, and they vary substantially run to run:

| instance | incumbent | optimum | gap |
|---|---|---|---|
| tln4 | 9.3 | 8.3 | +12% |
| tln5 | 32.2 | 10.3 | **+213%** |
| tln6 | 65.3 | 15.3 | **+327%** |

Across runs the same instance has produced 8.7 / 9.3 / 19.6 (tln4) and 11.0 / 15.1 / 32.2 / 32.8 (tln5). SCIP solves all of these to optimality in under a second.

## Why it wasn't caught by the graduation gate

The #844 differential panel gates on `gains / lost_incumbents / cert_regressions / overshoots` — i.e. whether an incumbent **exists**, is sound, and stays in budget. It never scored incumbent *quality*. That was the right bar for closing a "returns nothing" gap, but it means quality is currently unmeasured and unprotected: a change could halve incumbent quality and the panel would still pass.

## Suggested work

1. Add a **quality metric** to the panel (relative gap to the reference optimum from `minlplib.solu`) so regressions are visible even when the pass/fail gate is unchanged.
2. Then investigate the source. The engine's own dive/pump find the point; likely levers are the rounding order, the pump's cycle perturbation, and how much of the reserved budget the primal gets versus the node loop.

Worth doing in that order — without (1) there is no way to tell whether (2) helped.

Issue #844, opened 2026-07-22
Integer / nonconvex-MINLP primal heuristics workstream (LP-based FP + sub-MIP RENS + structure-aware constructors)


jkitchin:
Successor to #826 (closed) — the tractable-mechanism falsification is complete; this
is the forward workstream. Also subsumes #827 family D (ball_mk2 thin shell — 30
integer vars).

## Gap (capability, not a correctness bug — 0 soundness issues)

A class of integer / MINLP models return **no incumbent** while SCIP solves them in
<1s: tln4/6, gastrans040/582, rsyn0805/0810m04hfsg, portfol_robust050_34,
watercontamination0202, ball_mk2_30. Re-measured on the current tree (post pounce#258,
#832, #840): every existing lever fails — `feasibility_pump` (degenerates to
round-and-check on all-integer models), `rens` (18/24 fractional → 2^18 neighborhood),
`diving`, and a hand-written **guided NLP diving** (fix-most-fractional + re-solve)
which cracks only tln5 (1/7) and inconsistently within the tln family.

## The recipe discopt lacks at working strength (SCIP)

1. **LP-based feasibility pump** with an L1 projection over a **tightened** McCormick
   relaxation (OBBT/RLT to pull the LP vertex near integer-feasibility before rounding
   — the loose bilinear envelope is why rounding fails).
2. **sub-MIP RENS/RINS/LNS** with an *effective* sub-solver (the recursive discopt
   solve has the same primal gap, so it does not help).
3. **Structure-aware constructors** (transportation `xᵢ·xⱼ ≥ d`, synthesis, robust
   portfolio).

Each must be validated on tln4 first (§4) and gated + differential-panel-graduated
(§5) — note the trivial-primal seed graduation-gated **net-negative** (regression 13.3%
on crudeoil_pooling_ct1 / ex5_3_3 / tln12 / pointpack10), so soundness is not enough.

Repro / full re-measurement: #826 comments and `scratchpad/*826*`.

jkitchin:
## Entry experiment: the L1-projection feasibility pump also fails on tln4

Prototyped the missing SCIP-recipe lever — an **L1-projection feasibility pump** (each
round: round the integers, then solve `min Σ_int (xᵢ - x̄ᵢ)²` s.t. the constraints over
the *continuous relaxation*, NOT fixing integers — so the point migrates toward
integer-feasibility while staying feasible), which is fundamentally different from the
[...]

PR #872, opened 2026-07-26
feat(862): score incumbent quality in the #844 fallback panel

jkitchin:
## Summary

The #844 differential panel gated on `gains / lost_incumbents / cert_regressions /
overshoots` plus soundness — whether an incumbent **exists**, is sound, and stays in
budget. It never scored *how good* the incumbent was, so the flag shipped returning
tln6 `65.3` against a reference optimum of `15.3` (+327%) and, as #862 puts it, "a
change could halve incumbent quality and the panel would still pass".

This adds the missing axis, in the order the issue asks for (measure first):

- **`discopt_benchmarks/utils/primal_quality.py`** — the MIPLIB/Berthold primal gap
  (bounded in `[0,1]`, sign-safe, so it aggregates across a corpus whose objectives
  span orders of magnitude), the signed relative excess the issue's table quotes, a
  corpus summary, and `quality_regressions()` — the OFF-vs-ON comparison that was
  absent.
- **`discopt_benchmarks/utils/reference_optima.py`** — one oracle accessor over
  `minlplib.solu` → `known_optima.toml` → `cert-optima.json`, replacing the ~17
  hand-rolled `.solu` parsers under `scripts/`. The chain is the point: with the
  MINLPLib snapshot the panel scores tln4/5/6 automatically; in CI, where `.solu`
  does not exist, it still scores the vendored corpus instead of silently measuring
  nothing.
- **`discopt_benchmarks/scripts/issue844_primal_quality_panel.py`** — the #844 gate
  unchanged, plus per-instance primal gap, the corpus aggregate, and quality
  regressions.
- **`python/tests/data/known_optima.toml`** — the vendored pure-integer MINIMIZE
  instances the LP-node engine serves, consolidated from oracles already in this repo
  with their provenance recorded.

The panel's pass/fail verdict is deliberately **unchanged** (`--gate-quality` opts
in). #862 asks for the measurement first, and tightening a graduation bar on the same
commit that first measures the thing would retroactively fail a flag graduated
honestly under the bar of its day.

`docs/dev/lp-node-primal-quality.md` records the investigation of the issue's item
(2), including two falsified levers and their common root cause.

Contributes to #862

## Correctness

No solver-math change. The only edit under `python/discopt/` is a docstring addition
to `_lp_spatial_fallback_enabled`; everything else is new files under
`discopt_benchmarks/`, `docs/`, and a data TOML. Bounds, cuts, relaxations and
certificates are untouched.

The new metric is explicitly built so it cannot launder a correctness failure:
`is_false_primal` (incumbent below the reference optimum of a minimize) stays a
separate hard gate and is never averaged into a "quality" number. Missing incumbents
and missing oracles score `None` and are counted as `unscored`, never as gap 0 —
"no measurement" can never read as "no problem".

- [x] Cannot produce a false certificate (N/A — no solver-math change)

One note on that second box: the panel's net-positive bar (`gains > 0`) is applied
exactly when the corpus contains an instance with no incumbent flag-OFF, and reported
as N/A otherwise. An earlier draft had written it as `gains >= 0`, which is a
tautology — that was a silent weakening and was fixed before commit.

- [x] `pytest -m smoke` → **828 passed, 16 skipped, 7418 deselected, 2 xpassed** (333s)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **10 passed** (180s)
- [x] `ruff check` / `ruff format --check` clean on all touched files
- New tests: `discopt_benchmarks/tests/test_primal_quality.py` +
  `test_reference_optima.py` + `python/tests/test_optima_registry.py` → **42 passed**

A first adversarial run reported 1 failure (`test_large_dense_jacobian_no_crash`,
a wall-clock assertion). That was my own measurement error — I had launched it
concurrently with the smoke suite. Uncontended it runs 23.70s on this branch versus
23.26s on `main`; the clean re-run above is 10/10.

## Regression test

`test_primal_quality.py::TestQualityRegressions::test_detects_the_failure_mode_issue_862_names`
encodes the exact hole the issue names: a tln5 incumbent degrading 11.0 → 32.2
against optimum 10.3 is now reported as a regression. It fails before this change
trivially (no metric existed) and passes after.

Also pinned: `test_numerically_exact_solution_at_a_zero_optimum_scores_zero`, which
locks a bug the panel's own first run exposed — Berthold's exact `p == o` rule
saturates to gap **1.0** (the maximum) when the optimum is exactly 0 and the
incumbent is a float. It hit `gear` (2.9e-07) and `st_test1` (-1.6e-08) and reported
the corpus mean as 0.10 instead of ~0.00004, on two instances discopt had solved to
numerical exactness. `primal_gap` now compares against an absolute tolerance
(repo-wide `abs=1e-6`) before the sign rule.

N/A — nothing here changes a bound. Two candidate primal fixes for the issue's item
(2) were implemented, measured, and **reverted** rather than shipped default-off,
per CLAUDE.md §5 (sound ≠ helpful, the `DISCOPT_CUT_INHERIT` lesson):

- **Rounding flip / backtracking** — falsified: instrumenting the complement box at
  every dive failure, the opposite rounding was infeasible **37/37** on nvs17 and
  **17/18** on nvs24. Nothing to rescue.
- **Narrowing dive** (least-fractional selection + half-domain branching + LIFO
  backtracking, replacing a binary-oriented "fix to `round(LP value)`" applied to
  width-200 general integers) — falsified: zero primal gain and ~30% throughput loss
  at a 10s budget (nvs17 2080→1384 nodes, nvs19 1766→1280, nvs24 1280→896, all still
  with no incumbent; st_testgr3 unchanged at −20.533).

Root cause of both: the McCormick relaxation is ~250x loose at the root on this
family (nvs17 root bound −278209 against a true optimum of −1100.4), so every
rounding-based heuristic reading it is reading noise — 2562/2562 rounded LP points
infeasible. The productive direction is relaxation tightening or plunging, not a
better rounding rule. The LP itself was checked and is **sound**: every box
containing a true feasible point is reported feasible, and the fully-fixed box
returns exactly −1100.4.

## Panel's first run — two findings worth your attention

Running the new panel on the vendored corpus (20s budget) surfaced two problems the
8-instance #844 case list could not, both caught by the **existing** gate rather than
the new axis:

- **`nvs23` loses its incumbent with the flag ON** (OFF −1124.8, ON none;
  `lost_incumbents=1`). The fallback reserves 35% of the budget unconditionally, so
  the primary path got 13s of 20s. Any instance whose primary incumbent arrives in
  the last 35% of its budget can lose it — a design consequence, not tuning.
- **`nvs24` overshoots its budget 2.08x** (41.6s against 20s) — the overshoot class
  #844 believed it had closed.

Neither is fixed here and neither is an incumbent-*quality* finding; both are more
serious than the gap the panel was built to measure and are recorded in the doc as
follow-up candidates.

## Issue status: not closeable

Item (1) is done. Item (2) is open. Worth flagging: `tln4/5/6` are not vendored and
`minlplib.org` is policy-blocked from this environment, so the issue's own instances
were never re-runnable here. On the vendored corpus incumbent quality is **not** the
problem — 19 of 20 scored instances sit at gap 0, worst 7.3e-4 — so the tln-scale
gaps in the issue's table remain unconfirmed rather than reproduced. Reproducing them
now costs one command on a machine with the snapshot (see
`docs/dev/lp-node-primal-quality.md` §6).

---
_Generated by [Claude Code](https://claude.ai/code/session_012oBLdtnhBpCHyathjW9tmN)_


jkitchin:
## Review: MERGE WITH NITS — one two-line fix requested first

The metric is correct where it is used, reproduces #862's numbers to rounding, handles every degenerate input explicitly, and its oracle table cross-checks **exactly** against `minlplib.solu`. It **measures only** — it does not gate and does not touch the solve path — so it cannot launder a correctness regression today.

**Instrument check first**, since a previous review on this repo was fooled by it: I reviewed a dedicated worktree of `origin/claude/discopt-issue-862-ryxb2k` (merge-base `origin/main` `ac9f5cf3`; diff exactly 9 files / +2200 / −0) and asserted `discopt.__file__` resolved *into the worktree* and that the PR-only marker `"Incumbent quality is now measured (#862)"` was present in the **loaded** source before running anything. My first smoke run silently imported from the main tree; it was re-run with `PYTHONPATH` forced. Every probe below printed an executed-assertion count. Load was 1.73 throughout, and no claim here rests on wall-clock.

---

### Measurement only, not a gate

`--gate-quality` is opt-in and referenced nowhere in the repo. The sole `python/discopt/` edit is 11 lines of prose **inside an existing docstring** (`modeling/core.py`, `_lp_spatial_fallback_enabled`) — verified all added lines fall inside the triple-quoted string. Everything else is new files under `discopt_benchmarks/`, `docs/`, and a data TOML. §5 treatment is not required.

### Verified working

- **Reproduces #862 through the real `.solu` oracle chain:** tln4 `+12.05%`, tln5 `+212.6%`, tln6 `+326.8%` — against the issue's +12 / +213 / +327.
- **End-to-end on the real MINLPLib corpus** (`--instances tln4,tln5,tln6,nvs17 --time-limit 10`): 4/4 scored, 0 unscored, worst correctly tln6.
- **Both senses monotone and non-negative** when `sense` is supplied (21/21 assertions). MAXIMIZE case used `lip`'s real optimum 5685067.877.
- **Degenerate inputs explicit:** `None`, `inf`, `NaN` → `None` (unscored, never 0); optimum exactly 0 with incumbent 2.9e-07 → 0.0 and with incumbent 5.0 → 1.0 (this matters — `gastrans040` and `ball_mk2_30` both have optimum 0.0); straddling zero → 1.0; bounded ≤ 1 on `(1e12, 1.0)`, `(1.0, 1e12)`, `(-1e9, 1e9)`.
- **Oracle integrity — the strongest result here:** all 26 `known_optima.toml` and all 37 `cert-optima.json` entries cross-check **exactly** against `minlplib.solu`. 0 mismatches.
- **Not vacuous in CI:** with `DISCOPT_MINLPLIB_SOLU` pointed at a nonexistent path, the vendored-only chain still yields 48 oracles and resolves all 12 probes.

---

### 🔴 Please fix before merge — the sense default (two lines)

`discopt_benchmarks/utils/reference_optima.py`, module docstring, "Sense:" paragraph:

> *"The `.nl` reader normalizes MINLPLib instances to MINIMIZE, so for those the value is directly a lower-bound oracle."*

**This is false.** `from_nl('lip.nl')` returns `ObjectiveSense.MAXIMIZE` (`.nl` header `O0 1`), as does `rsyn0805m04hfsg`. And `summarize()` silently defaults `sense="min"`.

**No live mis-measurement today** — the only consumer filters through `_is_in_scope` (`_jax/lp_spatial_bb.py:58-70`), which requires MINIMIZE, and hardcodes `"sense": "min"`. But if a maximize instance ever reaches `summarize()` without a `sense` key:

- `is_false_primal(0.9×opt, opt, "min")` → **True** for a perfectly sound maximize incumbent — a manufactured correctness violation;
- `relative_excess` → **−0.100**, reporting a 10%-worse incumbent as *better than optimal*.

(`primal_gap`, the number that aggregates, is sense-independent and stays correct in both senses.)

**Ask:** correct the docstring, and make `sense` required — or raise on an unrecognised value rather than defaulting to `"min"`. This is the one thing that would let the metric mis-measure, and a silent wrong metric is worse than no metric because it launders regressions as passes.

---

### 🟡 Must close before `--gate-quality` is ever wired to a graduation bar

**Vacuity: the gate can pass having scored nothing.** `quality_ok = not regressions` (`issue844_primal_quality_panel.py:304`) with no `scored > 0` guard. Demonstrated: a 30-instance corpus with no oracles, ON arm **1e6× worse** than OFF → `regressions == []`, `quality_ok True`, exit 0, while the summary reports `scored=0, unscored=30`. The counters are printed and in the JSON, so a human sees it — the exit code does not.

Two narrower forms of the same hole:
- subprocess `error` rows (`harness_timeout` / `no_result` / worker exception) are recorded but **never checked by the gate** — an OFF-arm crash becomes `objective=None`, which reads as a **gain** and suppresses quality-regression detection for that instance;
- `_in_scope()` is `except Exception: return False`, silently dropping instances.

**Differential-only bar.** A live tln run printed `QUALITY CLEAN: True / PANEL OK: True` on a run whose ON arm returned tln6 at **+541%** — OFF had no incumbent, so gaining a terrible incumbent from nothing is by construction never a "regression". Consistent with "measure first", but worth stating explicitly so nobody later assumes `--gate-quality` protects against the exact case #862 was filed about.

**`primal_gap` is symmetric,** so a false primal can score as a quality *improvement* (MINIMIZE opt=10: honest 12 → 0.167; false primal 9 → 0.100). Caught in practice by the separate `false_primals` hard gate inside `gate_ok`, so not a live hole — but the PR body's claim that a false primal is "never averaged into a 'quality' number" isn't quite right: `summarize()` folds its gap into `mean_gap`/`worst_gap` (verified `mean_gap 0.1` on a false-primal-only corpus) while also counting it in `false_primals`.

---

### Nits

- `docs/dev/lp-node-primal-quality.md` §2 says tln4/5/6 "could not be re-run" because minlplib.org is blocked — but they are in the local snapshot at the CLAUDE.md path, and §6 of the same doc gives the command. I ran it (numbers above), so the constraint should be dropped and §1's table filled with real values.
- PR body and `reference_optima.py` say this "replaces the ~17 hand-rolled `.solu` parsers under `scripts/`" — no existing script was touched and 33 files still contain `=opt=`. It adds an 18th accessor; the consolidation is future work. Reword or do it.
- `test_reference_optima.py::test_resolves_without_a_solu_snapshot` doesn't neutralise `.solu` (no monkeypatch + `oracle_table.cache_clear()`), so on a machine with the snapshot it resolves through `.solu`, not the vendored TOML it claims to test. The vendored path is fine — verified separately — but the test name overstates what it exercises.
- `oracle_table` is `lru_cache(maxsize=1)`, so `DISCOPT_MINLPLIB_SOLU` can't be changed mid-process.
- `overshoots` uses `b["wall"] or 0.0`, so a `harness_timeout` row — which by definition blew the budget — counts as 0.0 s and is not an overshoot.

### Local checks

| check | result |
|---|---|
| `pytest python/tests/ -m smoke` (worktree, PYTHONPATH forced) | **832 passed**, 1 skipped, 2 xpassed |
| `test_primal_quality.py` + `test_reference_optima.py` | 39 passed |
| `test_optima_registry.py` (validates the 21 new entries) | 3 passed |
| `ruff check` / `ruff format --check` | clean |
| `pre-commit run mypy --all-files` | Passed |
| `gh pr checks 872` | 9 pass, 2 skipping |

### Framing

`Contributes to #862` is right — this doesn't close it. Item (1) is done and verified; item (2) is investigated with two levers **honestly falsified and reverted**, root cause identified as a ~250× loose McCormick root relaxation rather than heuristic tuning. That's the right outcome to record rather than ship a tuning that doesn't hold.

The panel's own first run also surfaced two defects worth their own issue: **`nvs23` loses an incumbent** to the fallback's unconditional 35% budget reserve, and **`nvs24` overshoots 2.08×** (41.6 s vs 20 s).

jkitchin:
Thanks — the sense finding is a real defect and I've fixed it. Pushed as `c1ed999`.

## 🔴 The blocking ask — fixed, and it was worse than a docstring

You're right that the docstring was false, and I verified it rather than taking it on trust: 5 of the *vendored* instances load as MAXIMIZE (`syn05m`, `syn05hfsg`, `bchoco06/07/08`), so the claim was wrong even without leaving the repo.

I went further than "correct the docstring, make `sense` required" — the docstring was the smaller half. `sense` is now **required and validated everywhere**, and an unrecognised value raises:

- `relative_excess` / `is_false_primal` — `sense` is a required positional; `_check_sense` raises `ValueError` on anything outside `{"min","max"}`.
- `summarize()` — raises `KeyError` if a row has no `sense`, rather than defaulting.
- the panel threads `sense` through `_row()` instead of hardcoding `"min"` at each use site, so widening `_is_in_scope` can't silently leave a maximize row labelled min.

Pinned by `TestSenseIsRequired`, using your `lip` optimum for the maximize case.

## 🟡 Vacuity — all three fixed, not deferred

You scoped these as "before `--gate-quality` is wired to a graduation bar". I fixed them now; they're cheap and each one is a way for the gate to pass having measured nothing.

- **`scored > 0` guard.** `quality_ok` now requires it. Your 30-instance no-oracle demonstration returned exit 0; it doesn't any more.
- **Error rows fail the gate.** They were recorded and unchecked, and worse than you noted: an OFF-arm crash reads as a *gain* through the existing `gains` logic, which then also suppresses regression detection for that instance.
- **`harness_timeout` counts as an overshoot.** `wall or 0.0` scored the worst possible overrun as 0.0s.
- **`_in_scope`** now reports a model it can't classify instead of silently shrinking the corpus.

## On the false-primal / `mean_gap` point — you were right and I've changed the behavior

This is the one where your reading beat my PR body. I claimed a false primal is "never averaged into a quality number"; `summarize()` was in fact folding its gap into `mean_gap`/`worst_gap` while separately counting it. Given `primal_gap` is symmetric, that meant a false primal scored as *better* than an honest incumbent the same distance away (min, opt 10: honest 12 → 0.167, false 9 → 0.100) — a soundness failure reading as a quality improvement.

Rather than soften the claim, I made it true: false primals are excluded from the gap aggregate and counted in `false_primals`, which the gate still keeps. Two tests pin it.

## Nits

- **Differential-only bar** — agreed, and worth more than a nit, so it's stated explicitly in the doc: `--gate-quality` cannot fire on the case #862 was filed about, because gaining a bad incumbent from none is by construction never a regression. Your +541% run is cited. Anyone wiring this to a graduation bar must add an absolute threshold.
- **"replaces the ~17 parsers"** — overclaim, removed. It's the 18th reader today; migration is follow-up.
- **`test_resolves_without_a_solu_snapshot`** — correct, it didn't neutralise `.solu`. Now a fixture with `monkeypatch` + `oracle_table.cache_clear()` on both sides, and it asserts `solu_path() is None` before testing anything.
- **`oracle_table` `lru_cache`** — documented, with the `cache_clear()` requirement.
- **tln framing** — reworded. My statement was scoped to the container I measured in, but it read as a general claim. It now says that plainly and records that your run confirmed the issue's table through the real `.solu` chain.

## Verification

| check | result |
|---|---|
| `test_primal_quality` + `test_reference_optima` + `test_optima_registry` | **48 passed** |
| `pytest -m smoke` | **828 passed**, 16 skipped, 2 xpassed |
| `ruff check` / `ruff format --check` | clean |
| panel end-to-end after the API change | 4/4 scored, gate + quality verdicts correct |

One note on your smoke count: you got 832 passed / 1 skipped where I get 828 / 16. That's environment-dependent skips (optional solver stacks), not a discrepancy in the change — worth knowing if you compare runs later.

The two defects the panel surfaced (`nvs23` losing an incumbent to the unconditional 35% reserve, `nvs24` at 2.08×) still aren't fixed here and I agree they deserve their own issue. Happy to open it.

---

Issue #865, opened 2026-07-26
Convex kernel abstains on rsyn*m04hfsg: the 12/24 blocking terms are PERSPECTIVE functions y·f(x/y), provably convex (was: binary×continuous products)


jkitchin:
> **Re-scoped 2026-07-26.** The entry experiment on the real instances **falsified this issue's stated mechanism**. The conclusion — that the abstention is conservative rather than correct — *holds*, and is now proven. But the products are not bilinear, the suggested fix does not apply, and the actual fix is a different (and more valuable) one. Original text at the bottom.

## The abstention is conservative — proven, not suspected

`build_convex_spec` abstains on `rsyn0805m04hfsg` / `rsyn0810m04hfsg`, and the blocking terms **are convex**. Analytic Hessian of the blocking form `g(x, y) = −y·log(x/y + 1)`:

```
g_xx = y/s²      g_xy = −x/s²      g_yy = x²/(s²·y)        (s = x + y)
det H = (1/s⁴)·(y·x²/y − x²) = 0
```

so `H` is **rank-1 positive semidefinite** — convex, with one flat direction. Checked over 20,000 random points in the reachable box (`y ∈ [0.001, 1]`, `x ∈ [0, 50]`): min eigenvalue **−3.5e−16**, max `|det H|` **6.5e−15**, **0 points with a genuinely negative eigenvalue**, analytic form agreeing with finite differences 4/4.

Methodology note worth carrying: a *finite-difference* Hessian probe on this term reports min eigenvalue **−1.4e−5** on 1286/4000 points and reads as "not convex". That is noise around the true zero eigenvalue (second differences amplify by `1/h²`), not nonconvexity. A rank-deficient Hessian needs the analytic form.

## What the 12 terms actually are — not bilinear products

This issue says "12 var×var products, and all 12 are `binary × continuous` — exactly linearizable (the standard big-M / McCormick-exact form for a binary factor)". They are not. Verbatim from `rsyn0805m04hfsg`:

```
( x8/(0.001 + 0.999·x24) − log( x0/(0.001 + 0.999·x24) + 1 ) )  ·  ( 0.001 + 0.999·x24 )
```

with `x24` binary, i.e. `y = 0.001 + 0.999·x24 ∈ {0.001, 1}`. Expanding gives

```
x8  −  y · log(x0/y + 1)
```

[...]

Issue #871, opened 2026-07-26
Convex kernel cannot certify while structural columns keep infinite upper bounds (NS declines by design)


jkitchin:
> **Re-scoped and re-labelled 2026-07-27 (`bug` → `enhancement`).** The correctness half of this issue was real and is **fixed**. The half named in the title is **not a defect** — it is a rigorous bound working as specified. What remains is a capability gap. Original text preserved at the bottom.

## The title symptom is not a bug

`ns_safe_bound` (`lp/simplex/refine.rs`) computes

```text
g(y) = bᵀy + Σ_j min_{x_j ∈ [l_j, u_j]} (c − Aᵀy)_j x_j   ≤   min cᵀx
```

which is a valid lower bound for **any** `y` — that is the whole point of Neumaier–Shcherbina. If a column has `u_j = +∞` and its reduced cost is negative, then `min_{x_j ∈ [l_j, ∞)} (…)_j x_j` is genuinely `−∞`, and the function's own contract says so:

> Returns `None` when the bound is `−∞` (a reduced cost with the wrong-signed open box side).

So "the NS safe bound degrades to `-inf` as the cut pool grows" describes a real phenomenon but mislabels it. Nothing degrades. As the pool grows, roundoff flips a reduced cost's sign on an unbounded column, and the formula **correctly reports that this dual certifies nothing**. A safe bound declining to certify is the mechanism succeeding, not failing.

## The genuine bug here was real, and is fixed

The item this issue filed under *"Worth checking while here"* — `solve_tree` doing `if r.status != LpStatus::Optimal { continue; }`, silently dropping a node — **was** a correctness defect: a run that dropped a node *and* had an incumbent would upgrade `Exhausted → Optimal` over a subtree never explored, i.e. a false certificate.

Fixed by `1df5f71a` ("never certify a tree that silently discarded a subtree"), shipped in #870 (*"fix(correctness): #871 two false-certificate paths"*). `uncertified_drop` now poisons the certificate whenever any node exits non-`Optimal`/non-`Infeasible`.

This issue's own body already recorded that the remaining behaviour is sound:

> Current behaviour is **sound** … No false certificate, no default-path change.

Which is why the `bug` label no longer fits.

## What actually remains — the capability gap

**Structural columns with infinite upper bounds prevent the kernel from certifying**, because NS cannot produce a finite bound over an open box side once roundoff perturbs the reduced cost.

[...]

PR #881, opened 2026-07-27
fix(#879): the clay0303hfsg "false optimum" was a false CERTIFICATE — re-admit sqr, and certify

### LINKED ISSUE #879: Convex kernel emits a FALSE OPTIMUM on clay0303hfsg when the quadratic perspective (a²/s) is routed
> Found and reverted in PR #870 (`7db4349`). Recording it so the quadratic-perspective work is not re-attempted without this evidence. **No current code path is affected** — `**` is refused again, so `clay*hfsg` falls back to the default path.

## Observation

Adding `ConvexFunc::Sqr` so the convex kernel accepts `clay*hfsg`'s quadratic perspective `ε·((x/ε)² − c·x/ε)` = `x²/ε` makes `clay0303hfsg` routable — and the kernel then **certifies a wrong optimum**:

| run | status | certified objective |
|---|---|---|
| kernel, `max_sep_rounds=0` | `optimal` | **28351.42** |
| kernel, `max_sep_rounds=0`, dominated-col bound | `optimal` | **36397.83** |
| kernel, `max_sep_rounds=2` | `optimal` | **55092.52** |
| discopt default path, 120 s | `feasible` | 26669.11 (bound 25922.92) |

Three "certified optima" that disagree with each other, all strictly worse than a point the default path actually attains. This is a minimization, so a certified value *above* an attainable one means the **dual bound sits above the true optimum** — an invalid (too-tight) relaxation.

## The better point is genuinely feasible for the marshaled spec

Not a modelling mismatch. Taking the default path's solution and evaluating it against the kernel's own marshaled arrays:

```
c·x under spec         = 26669.10955143188   (matches the default path's objective)
worst LE violation     = 9.96e-09
worst EQ violation     = 1.42e-14
box violation (lo/hi)  = 1.00e-08 / 0.0
```

So a feasible point with objective 26669.11 exists inside the spec while the kernel certifies 28351.42 as optimal.

## What this says about the verification that let it through

The producer side was verified and **passed**:

* **exactness** — marshaled rows equal the pristine model's rows pointwise, worst relative error `3.7e-13`;
* **convexity** — every routed row satisfies the midpoint inequality over sampled triples, worst violation `0.0`.

Both held, and the certified answer was still wrong. So the fault is **downstream of the producer**. The most likely candidate: the OA/cut machinery on `a²/s`, whose tangent coefficients scale as `a²/s²` — around `1e6` at the `0.001` smoothing floor, versus the well-behaved `1/s` of the log perspective that works fine on `syn05hfsg`. That also fits the `numerical` LP failures seen at ~884 tangents (#871).

**The lesson to carry:** exactness + convexity of the marshaled rows is *not* sufficient evidence that routing a new term class is sound. A certified result must be checked against a known optimum before the class is admitted. That check was missing, which is why this shipped and had to be reverted.

## Reproduce

Restore `ConvexFunc::Sqr` + `_pow_as_sqr` (see the reverted commit `8cfc32f`), then:

```bash
python - <<'EOF'
import os; os.environ['DISCOPT_CONVEX_KERNEL']='1'
import discopt.modeling as dm
from discopt.solvers._convex_kernel import build_convex_spec, solve_convex_tree
spec = build_convex_spec(dm.from_nl('python/tests/data/minlplib_nl/clay0303hfsg.nl'))
for sep in (0, 2, 12):
    r = solve_convex_tree(spec, time_limit_s=120, max_sep_rounds=sep)
    print(sep, r['status'], r['incumbent'], r['bound'], r['node_count'])
EOF
```

## Suggested approach if this is picked up

1. **Get a reference optimum first.** `clay0303hfsg` isn't in `python/tests/data/known_optima.toml`; add it from `minlplib.solu` so any run can be checked rather than eyeballed.
2. Bisect the fault below the producer: compare the kernel's node bound at the ROOT against a trusted relaxation on the same box. If the root bound already exceeds the true optimum, it is the tangent/NS path; if not, it is separation (cover/GMI/MIR) or the tree.
3. Suspect the cuts specifically — `separate_gomory_cols` is passed a `1e7` dynamism cap, and `a²/s²` tangents at `s = 0.001` sit right at that scale.
4. Only re-admit `**` once a routed instance's certified objective is checked against a known optimum, not only against exactness/convexity of its rows.

`python/tests/test_convex_kernel_perspective_865.py::test_square_inner_function_falls_back` pins the current refusal and points here.

jkitchin:
## Summary

#879's stated mechanism is **falsified**, and is retracted in writing here (CHANGELOG, the producer's module docstring, and a pinning test).

The issue read three mutually inconsistent `optimal` results on `clay0303hfsg` (28351.42 / 36397.83 / 55092.52) as a dual bound sitting **above** the true optimum — an invalid, too-tight `a²/s` relaxation. Re-measured with the #871 certificate fix in place, that reading does not survive:

| #879's claim | measurement |
|---|---|
| dual bound sits above the true optimum | **falsified** — root safe bound is `0.0` at every separation setting (0/1/2/4/12) against an optimum of `26669.11`: valid, and trivially *weak*, the opposite failure |
| those three numbers were certified optima | they were **incumbents**. `28351.41943619983` is reproduced exactly as the incumbent of an *uncertified* run |
| the fault is downstream of the producer, in the OA/cut machinery | the fault was the tree bug #871 fixed — a subtree silently discarded on a `numerical` node LP, after which the reported `bound` fell back to the incumbent's own objective |

So `**` is re-admitted (`ConvexFunc::Sqr` + `_pow_as_sqr`), together with three guards that
move `clay0303hfsg` from *silently wrong* to *sound*.

> **Corrected on review.** An earlier version of this body claimed the guards turn
> "sound but uncertifiable" into a certificate. They do not, and it is measured:
> `clay0303hfsg` returns **`status="exhausted"`**, incumbent `26669.109572` (the known
> optimum to 7 figures), bound `26668.921579`, relative gap **7.0e-06** — inside the
> usual `1e-4`, so this is the last step of the certificate rather than a structural
> failure. The `Exhausted -> Optimal` upgrade requires `!uncertified_drop`, so a node
> LP is still exiting `numerical` and the certificate is **correctly withheld**. That
> is #871 residual, not a regression here, and the upgrade condition must NOT be
> relaxed to make it pass — doing so manufactures exactly the #879 false certificate.

**What actually blocked certification:** node LPs exiting `numerical` on 12 of 453 nodes, correctly poisoning the certificate. Three causes, each ablated and each **necessary** — removing any one alone returns the instance to `exhausted`:

1. **Six infinite structural upper bounds** FBBT cannot close (`clay0303hfsg`'s fixed-charge cost columns `x81..x86`). `tighten_dominated_columns`, restored from `1df5f71`, now default-ON inside the default-off kernel with a `DISCOPT_CVX_DOMINATED_COLS=0` opt-out.
2. **An unbounded per-node OA tangent pool** — one node reached **307 OA rounds / 2042 tangents** on 99 structural columns (dynamism `7.3e8`) before the factorization broke. New `appended_row_cap = max(8·n, 500)`; the healthy routed instances peak at `2.4·n`, so it clears them by more than 3×.
3. **A warm re-optimize whose carried basis** — extended once per appended row across the whole OA chain — is exactly what is in doubt after a breakdown. Now re-solved once **cold**, mirroring the LP layer's own `dense_retry`.

A cut-dynamism cap on the *substituted* row was also built and **measured not to help** once the size cap was in — it cost `syn05m` 3 → 9 nodes and bought `clay0303hfsg` nothing — so it is **not shipped** (the `DISCOPT_CUT_INHERIT` lesson: sound ≠ helpful).

**The methodological lesson from #879 stands and is now enforced.** Exactness (worst relative error `3.7e-13`) and convexity (worst midpoint violation `0.0`) both **passed** while the false certificate was live. They are necessary, not sufficient. `clay0303hfsg` is added to `python/tests/data/known_optima.toml`, and the guard is now two tests:

* `test_clay0303hfsg_result_is_sound_against_its_known_optimum` — asserts what is
  actually guaranteed (bound `<=` optimum, incumbent `>=` optimum, bound `<=` incumbent),
  with an executed-assertion count. **Deliberately not marked `slow`:** it runs in ~7 s
  and every CI lane excludes `slow` (`ci.yml` 178 / 262 / 388), so as a slow test this
  guard would never run on a PR — reproducing the very #879 gap it exists to close.
* `test_clay0303hfsg_does_not_yet_certify` — `xfail(strict=True)` pinning the residual,
  so the day it starts certifying the strict xfail fails and the claim gets updated
  rather than silently rotting.

Closes #879. Contributes to #871.

**#865 is not closed by this PR — it is already closed by #870.** Verified on plain
`main` (which has #870 but *not* this branch's `_pow_as_sqr`): `rsyn0805m04hfsg`
certifies `optimal` at `7174.220058` against `=opt= 7174.219035` (rel `1.4e-7`) and
`rsyn0810m04hfsg` at `6581.935607` against `6581.934408` (rel `1.8e-7`). Those are the
instances #865 names, so the log-perspective half of #865 was complete at #870; the
`sqr` half here serves `clay*hfsg`, which is sound but not yet certified.

### Note on #870

This work started before #870 merged, so the branch first carried #870's own commits; #870 has since landed as `463ed77` and `main` is merged in here. The diff against current `main` is now purely the #879 work — 7 files. Three merge conflicts, all resolved in the merge commit and all cosmetic: `main`'s `1542d56` had independently made the same clippy (`s.is_nan() || s <= 0.0`) and mypy (`coef`, not `k`) fixes I had, so the code is byte-identical either way and the merge keeps only the comments explaining *why* those forms are required; and `main`'s `b42ed68` executed-comparison counters were deduplicated against mine in `test_convex_kernel_perspective_865.py`.

Everything below was re-verified on the merged tree.

## Correctness

Every guard is a **refusal or a robust re-solve**, never an approximation:

- a capped node returns a valid, merely *weaker* dual bound (the node LP is a relaxation for any subset of tangents), and a vertex reached without OA convergence still violates a nonlinear row, so `is_integer_feasible` cannot mistake it for an incumbent;
- the cold retry only ever replaces a **failure** — a `Numerical`/`IterLimit` exit carries no certificate to preserve — and a retry that also fails flows on as that failure and still poisons the certificate;
- the dominated-column bound is an **optimality** argument, stated and tested: it qualifies a column only when its minimized-objective coefficient is `> 0`, it appears in no equality and no nonlinear row, and every `≤` row containing it has `a_ij < 0`. Four separate refusal tests pin each disqualifier.

- [x] Cannot produce a false certificate — and this PR *removes* one

Independent evidence the certificate is real: end-to-end through `Model.solve()`, `clay0303hfsg` returns `optimal 26669.109572` with the incumbent verified feasible against the **pristine** model (the #779 guard), against a MINLPLib reference of `26669.10955143`.

- [x] `pytest -m smoke` → **840 passed, 15 skipped, 2 xpassed** (576 s)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **9 passed, 1 failed** — see below
- [x] `cargo test -p discopt-core` → **556 passed, 0 failed** (12 new)
- [x] `ruff check python/` / `ruff format --check` (pinned 0.14.6) clean; `mypy python/discopt/` clean; `cargo clippy -p discopt-core -- -D warnings` and `cargo fmt --check` clean

Re-run after merging `main`: `cargo test` 556 passed, clippy/fmt clean, the convex-kernel pytest slice (865 + gate + fast_path, including `slow`) 64 passed, ruff/mypy clean, and the panel below reproduces bit-for-bit.

**On the one failure.** `test_large_dense_jacobian_no_crash` asserts `wall < budget + 40`, a wall-clock deadline. It failed under load **I created myself** by running convex-kernel tests concurrently with it. Quiet machine, 3/3 pass: **37.75 / 36.68 / 37.86 s** (sd 0.6 s) against a 48 s deadline. The change cannot reach that path — the kernel is default-OFF and `try_convex_solve` returns before `build_convex_spec`. This is the same load-sensitive assertion #870 documented.

## Regression test

`test_clay0303hfsg_result_is_sound_against_its_known_optimum` (correctness, NOT slow) plus a strict-xfail companion. Fail-before / pass-after is established at both stages, each measured:

- at the base commit, `build_convex_spec(clay0303hfsg)` returns `None`, so the test fails at `assert spec is not None`;
- with `sqr` restored but *without* the three guards, the tree returns `exhausted`, so it fails at `assert r["status"] == "optimal"`.

Also added: `test_clay0303hfsg_root_relaxation_is_sound_not_too_tight` (pins the falsified hypothesis at the *node*, so a future regression that genuinely does over-tighten the root is caught there rather than in the tree); `test_square_inner_function_is_routed` and `test_only_exponent_two_is_admitted` (replacing the refusal test, with all 6 refusal branches counted); the exactness/convexity check parametrized over the `sqr` class; and 12 Rust tests (`Sqr` value/deriv, perspective value/gradient-vs-FD/tangent-underestimates-over-a-grid, the `appended_row_cap`, and 5 dominated-column tests including 4 refusals).

Bound-changing, inside the default-off `DISCOPT_CONVEX_KERNEL` path. The kernel declines everything else in the corpus, so the four routed instances **are** the entire population this code reaches — the panel below is corpus-wide for this path, not a sample.

- [x] Default-OFF gate unchanged (`DISCOPT_CONVEX_KERNEL`); the new reduction keeps its own `DISCOPT_CVX_DOMINATED_COLS=0` opt-out
- [x] Cert-clean **and** net-positive: no bound above its reference optimum in any run (including the ablations), no certification regression, incumbent independently feasibility-verified against the pristine model
- Measurement (default settings, 300 s budget):

| instance | before | after |
|---|---|---|
| `syn05m` | optimal, 3 nodes | **identical, bit-for-bit** (`inc=837.7324008979796`, `bound=…797`) |
| `syn05hfsg` | optimal, 2 nodes | **identical, bit-for-bit** |
| `cvxnonsep_psig40r` | exhausted, 1 node | **identical** |
| `clay0303hfsg` | **declined** | **optimal `26669.109572`, bound `26669.060610`, 211 nodes** |

Ablations on `clay0303hfsg` (each guard removed alone): dominated columns off → `exhausted`; size cap off → `exhausted`; cold retry off → `exhausted`. Every ablation's bound stays sound.

Over the 149 in-repo `.nl` instances, `clay0303hfsg` is the **only** routing change; nothing previously routed was lost.

Reproduce: `python discopt_benchmarks/scripts/issue879_convex_kernel_panel.py --ablate` (exits non-zero if any certificate is unsound, if an instance that certified stops certifying, **or if no run executed at all** — a panel that measures nothing must never read as a pass).

---

https://claude.ai/code/session_0112UaqL3jE6wmKhNyFMFN4D

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,51 @@ The release procedure that produces these entries is documented in
   in-repo `.nl` instances the only routing change is `syn05hfsg` being admitted;
   nothing previously routed was lost.
 
+- **Quadratic inner function (`** 2` → `sqr`) for the convex kernel** (`feat`,
+  #879, same default-off `DISCOPT_CONVEX_KERNEL` path). `clay*hfsg`'s hull rows are
+  `ε·((x/ε)² − c·x/ε + …)`, i.e. the quadratic perspective `x²/ε`
+  (quadratic-over-linear, jointly convex on `ε > 0`), so they declined only because
+  `_FUNC`/`ConvexFunc` had no `sqr` entry. Adds `ConvexFunc::Sqr` and `_pow_as_sqr`,
+  which composes with the perspective machinery unchanged — a plain `x² ≤ c` row is
+  routable too. **Only the exponent 2 is admitted**; every other power (odd,
+  fractional, negative, or variable) is nonconvex, domain-restricted, or signomial,
+  and keeps falling back, as does a non-affine base such as `(log x)²`.
+
+  This class was withdrawn once (evidence recorded in #879) on the reading that its
+  `a²/s` tangents produced an invalid (too-tight) relaxation. **That reading is falsified** — see
+  the *Fixed* entry below. Re-admitted here together with the three guards that let
+  it certify: the dominated-column bound, a per-node relaxation size cap, and a cold
+  retry of a numerically broken warm re-optimize.
+
+  Measured: `clay0303hfsg` goes from *declined* to **certified optimal
+  `26669.1096` in 211 nodes** end-to-end through `Model.solve()` (39.2 s on an idle
+  machine, single run — the node count is the deterministic figure), matching
+  its MINLPLib reference (`26669.10955143`, now recorded in
+  `python/tests/data/known_optima.toml`) and incumbent-verified against the pristine
+  model. Bound-neutral for everything routed before: `syn05m` (3 nodes), `syn05hfsg`
+  (2 nodes) and `cvxnonsep_psig40r` (1 node) are bit-identical in status, objective
+  and node count. Over the 149 in-repo `.nl` instances `clay0303hfsg` is the only
+  routing change.
+
+- **Dominated-cost-column upper bound for the convex kernel** (`feat`, #871/#879,
+  default-ON inside the default-off kernel, opt out with
+  `DISCOPT_CVX_DOMINATED_COLS=0`). FBBT cannot close a column the constraints only
+  bound from *below*; `clay0303hfsg` has six (its fixed-charge cost variables
+  `x81..x86`, objective coefficients 300/240/100/…), and an infinite `ub` is what
+  makes the node LP break down and the Neumaier–Shcherbina safe bound decline.
+  `tighten_dominated_columns` gives such a column the finite bound
+  `U_j = max(lo_j, max_i (maxact_rest_i − rhs_i)/(−a_ij))`.
+
+  This is an **optimality** argument, not FBBT: it qualifies a column only when its
+  minimized-objective coefficient is `> 0`, it appears in no equality and no
+  nonlinear row, and every `≤` row containing it has `a_ij < 0`. Lowering `x_j` to
+  `U_j` from any feasible point then keeps every row satisfied and strictly lowers
+  the objective, so the node's optimal *value* is unchanged and the dual bound stays
+  valid — but unlike FBBT the box no longer contains every feasible point, which is
+  why it keeps its own switch. Measured: turning it off takes `clay0303hfsg` from
+  `optimal` back to `exhausted`, and it is a bit-identical no-op on every other
+  routed instance.
+
 ### Changed
 
 - **`gap_certified` now requires BOTH ends of a gap** (`fix`, #875). A limit exit
@@ -56,6 +101,43 @@ The release procedure that produces these entries is documented in
   because a bound with no incumbent is still a valid bound.
 
 ### Fixed
+
+- **The convex kernel's `clay0303hfsg` "false optimum" was a false CERTIFICATE, not
+  an invalid relaxation** (`fix(correctness)`, #879). #879 recorded three
+  mutually-inconsistent `optimal` results on `clay0303hfsg` (28351.42 / 36397.83 /
+  55092.52), each strictly worse than a point the default path attains, and read
+  them as a dual bound sitting *above* the true optimum — an invalid, too-tight
+  `a²/s` relaxation. Re-measured with the #871 certificate fix in place, **that
+  hypothesis is falsified and is retracted here**:
+
+  * every one of those numbers is an **incumbent**, published as certified because
+    the tree had silently discarded a `numerical` node and then let `bound` fall
+    back to the incumbent's own objective. `28351.41943619983` is reproduced exactly
+    as the incumbent of an *uncertified* run;
+  * the relaxation is not too tight anywhere. `clay0303hfsg`'s root safe bound is
+    `0.0` at every separation setting (0/1/2/4/12) against an optimum of `26669.11`
+    — valid, and in fact trivially weak, the opposite failure. Pinned by
+    `test_clay0303hfsg_root_relaxation_is_sound_not_too_tight`.
+
+  What actually blocked certification was the node LP breaking down (`numerical`) on
+  12 of 453 nodes, correctly poisoning the certificate. Three causes, each fixed and
+  each measured to be necessary — removing any one alone takes the instance back to
+  `exhausted`: the six infinite structural upper bounds (dominated-column bound,
+  above); an unbounded per-node OA tangent pool (`appended_row_cap`, `max(8·n, 500)`
+  — one node reached 307 OA rounds / 2042 tangents on 99 columns before the
+  factorization broke); and a warm re-optimize whose carried basis, extended once per
+  appended row across the whole OA chain, is exactly what is in doubt after a
+  breakdown (now re-solved once **cold**, mirroring the LP layer's own
+  `dense_retry`). All three are refusals or robust re-solves — a capped node returns
+  a valid, merely weaker bound, and a vertex reached without OA convergence still
+  violates a nonlinear row so it cannot be mistaken for an incumbent.
+
+  The methodological lesson from #879 stands and is now enforced: exactness and
+  convexity of the marshaled rows both **passed** while the false certificate was
+  live, so they are not sufficient to admit a term class. A routed instance's
+  certified objective must be checked against a known optimum —
+  `test_clay0303hfsg_certifies_against_its_known_optimum` does exactly that, against
+  the shared `known_optima.toml` registry.
 
 - **Root setup no longer scales with `n_vars × n_constraints`, and no longer runs
   unbudgeted** (`fix(performance)`, #875, successor to #863/#868). After #868,
--- a/crates/discopt-core/src/bnb/convex_kernel.rs
+++ b/crates/discopt-core/src/bnb/convex_kernel.rs
@@ -53,6 +53,9 @@ pub enum ConvexFunc {
     Sqrt,
     /// `ln(1 + t)` (concave).
     Log1p,
+    /// `t²` (convex). Its perspective `s·(a/s)² = a²/s` is quadratic-over-linear,
+    /// jointly convex on `s > 0` — the `clay*hfsg` hull shape.
+    Sqr,
 }
 
 impl ConvexFunc {
@@ -64,6 +67,7 @@ impl ConvexFunc {
             ConvexFunc::Exp => t.exp(),
             ConvexFunc::Sqrt => t.sqrt(),
             ConvexFunc::Log1p => t.ln_1p(),
+            ConvexFunc::Sqr => t * t,
         }
     }
 
@@ -81,6 +85,7 @@ impl ConvexFunc {
                 (s, 0.5 / s)
             }
             ConvexFunc::Log1p => (t.ln_1p(), 1.0 / (1.0 + t)),
+            ConvexFunc::Sqr => (t * t, 2.0 * t),
         }
     }
 }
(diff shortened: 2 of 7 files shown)


jkitchin:
## Review: the soundness fix is verified — but the headline claim is false and CI cannot see it

Reviewed at `51da076c` (`main` merged in) in a dedicated worktree, asserting `_pow_as_sqr` present in the *loaded* module before measuring anything.

### ✅ The soundness half holds, independently

14 executed assertions, **0 violations**, against `minlplib.solu` `=opt=` values:

| instance | status | incumbent | bound | oracle | |
|---|---|---|---|---|---|
| `rsyn0805m04hfsg` | **optimal** | 7174.220058 | 7174.220058 | 7174.219035 | ✅ rel 1.4e-7 |
| `rsyn0810m04hfsg` | **optimal** | 6581.935607 | 6581.935607 | 6581.934408 | ✅ rel 1.8e-7 |
| `syn05hfsg` | optimal | 837.7324009 | 837.7324009 | 837.7324009 | ✅ |
| `syn05m` | optimal | 837.7324009 | 837.7324009 | 837.7324009 | ✅ |
| `clay0303hfsg` | `exhausted` | 26669.109572 | 26668.921579 | 26669.10957 | sound |

`#879`'s stated mechanism really is falsified: **no bound sits above its optimum anywhere**, `clay0303hfsg` included (26668.92 ≤ 26669.11), and its incumbent matches the oracle to 7 significant figures. Re-admitting `sqr` is sound on this evidence.

**And this closes #865's actual ask** — both `rsyn*m04hfsg` now certify against their proven optima, where they previously abstained with `bilinear product`. Worth stating explicitly in the PR body; right now #865 only gets a "contributes to".

### 🔴 `clay0303hfsg` does not certify

The PR says it does, and the body's framing ("three guards that turn *sound but uncertifiable* into a certificate") rests on it. It returns `exhausted`:

```
FAILED test_clay0303hfsg_certifies_against_its_known_optimum
AssertionError: clay0303hfsg did not certify (status=exhausted)
```

Two independent observations: my own probe returned `exhausted` (131 nodes, 6.6 s — not a time limit) *before* I read the test, and then the PR's own test reproduces it. The gap is tiny — bound 26668.9216 vs incumbent 26669.1096, i.e. a relative gap of **7.1e-6**, inside the usual 1e-4 — so this looks like the last step of the certificate rather than a structural failure. But as it stands the claim is unsupported.

### 🔴 CI cannot catch it — and that is the same failure #879 was about

`test_clay0303hfsg_certifies_against_its_known_optimum` is marked `slow`, and **every** CI lane excludes `slow`:

```
ci.yml:178  -m "not slow and not correctness and not integration and ..."
ci.yml:262  -m "correctness and not slow and not integration and ..."
ci.yml:388  -m "not slow and not correctness and ..."
```

So this PR will go green with its central assertion failing. That is structurally the same gap #879 diagnosed: *"a certified result must be checked against a known optimum before the class is admitted. That check was missing, which is why this shipped and had to be reverted."* The check now exists — and does not run.

It also is not slow: **the whole file's `slow` selection completes in 8.41 s**. There is no cost argument for keeping it out of the fast lane, and putting it there is what makes the guard real.

### What I would change before merge

1. **Drop the `slow` marker** on `test_clay0303hfsg_certifies_against_its_known_optimum` (8.4 s), so the guard actually runs. If it must stay out of the fast lane, it needs a lane that runs it — otherwise it is documentation, not a test.
2. **Either close the 7.1e-6 gap or make the test honest** — `xfail(strict=True)` with the measured `exhausted` status and the residual gap recorded, so the remaining work is visible and the day it starts certifying the strict xfail tells you.
3. **Correct the body**: `clay0303hfsg` is *sound but uncertified*, not certified. `Closes #879` is defensible on the falsification + no-bound-above-optimum evidence, but the "turns uncertifiable into a certificate" claim is not.
4. **Promote #865 from "contributes to" to closing** — or say what remains, given both `rsyn*m04hfsg` now certify correctly.

### Verified elsewhere

- `pytest python/tests/test_convex_kernel_perspective_865.py` → 13 passed, 2 deselected (the 2 are the `slow` ones; 1 of those fails, above).
- The three guards' necessity claim I did not re-ablate; the size cap (`8·n`, healthy instances peak at `2.4·n`) and the cold re-solve after a factorization breakdown are both the conservative direction and read correctly.
- Not shipping the cut-dynamism cap after measuring it neutral-to-harmful (`syn05m` 3 → 9 nodes) is the right call and the right precedent to cite (`DISCOPT_CUT_INHERIT`).

Issue #764, opened 2026-07-18
Closure-speed gap: tanksize — discopt finds the optimum but cannot close (BARON/SCIP certify in <3s)


jkitchin:
## Summary

`tanksize` (MINLPLib, in-repo `python/tests/data/minlplib_nl/tanksize.nl`, oracle **1.2686437540**, minimization) is the one instance on the curated **global50** 3-way panel where discopt fails to certify but BARON and SCIP close it in seconds. discopt **finds the optimal incumbent** but **cannot close the bound** within 60 s.

| solver | status | incumbent | dual bound | gap | nodes | wall |
|---|---|---|---|---|---|---|
| **discopt** | time_limit / feasible | **1.26864 (= opt)** | 0.8529 | **32.8 %** | 625 | 60 s |
| BARON | Optimal | 1.2686 | (root 0.9553) | closed | 3477 | **2.66 s** |
| SCIP | optimal | 1.26864 | closed | closed | — | **0.75 s** |

**Correctness is unaffected** — discopt&#39;s incumbent equals the true optimum and its bound (0.853) is sound (≤ opt). This is purely a **closure-speed** gap, not a soundness issue (0 violations on the panel).

## Resolution (closed by #791, merged in 46803d7)

Fixed via a **native Rust spatial B&amp;B kernel** (`crates/discopt-core/src/bnb/spatial_*.rs`) that runs the whole per-node loop in Rust with per-node FBBT propagation and Neumaier–Shcherbina safe bounds, plus **verified incumbent seeding**. The real lever turned out to be per-node nonlinear-propagation quality, not root-relaxation quality as the original diagnosis emphasized.

- tanksize seeded: **optimal in ~50 s / 11,379 nodes** (was: time-out at 32.8 % gap), objective `1.2686437526`, `|obj − oracle| = 1.44e-9`, certificate brackets the oracle, `incorrect_count = 0`.
- Ships behind `DISCOPT_NATIVE_SPATIAL_KERNEL`, **default OFF**. The Regime-2 graduation panel passed both bars (cert-clean: 0 violations; net-positive: median non-engaged wall Δ = −0.146 s), but default-ON is deferred to follow-ups **#788** (native-kernel wall budget) and **#789** (Python-engine feature parity / smoke re-scope).

The Definition of Done — certify at default tolerances in seconds with `incorrect_count = 0` — is met via the opt-in flag. Closing as completed.

## Definition of done (met)
[...]

PR #894, opened 2026-07-28
fix(relaxation): #861 a lifted model constraint is not a term's envelope row

jkitchin:
## What and why

Both the Python incremental path and the native #764 producer identified a term's box-dependent **envelope rows by support containment**: a row belongs to `w = x_i*x_j` iff its support is contained in `{i, j, w}`. That rule is not sufficient — a *model constraint* over the same variables lifts to a row with the same support. `x0*x1 >= 60` becomes the single-column row `{w: -1} <= -60`.

The two paths then failed differently, and both are fixed here.

## 1. Python path: declined the whole structure

`IncrementalMcCormickLP` saw 5 rows where it expected 4 and declined, so **every node** fell back to the ~30×-slower cold build on 6 corpus instances (`prob02`, `prob03`, `st_e01`, `st_e08`, `st_e09`, `st_e11`).

Envelope rows are now identified **numerically**: on the probe box each closed-form row must match exactly one candidate. Unmatched candidates are box-independent and stay unpatched — not taken on faith, since a genuinely box-dependent row left unpatched makes `_validate` fail with a row-set mismatch on a later validation box. An **ambiguous** match (two candidates equal within tolerance) declines rather than guessing.

**Narrower than the plan proposed, because of a second entry experiment.** The plan said to assign formulas in *match* order. Measured over the 31 admitted instances (630 terms), the match order is a **rotation** of today's ascending-index order on **30 of 31** — so assigning in match order would permute the rows of every node LP across nearly the whole admitted set and put the exactly-unchanged node-count gate at risk for zero benefit. So matching only **selects** the envelope subset; ascending order still **assigns** the formulas. `_rowset` is order-free, so both describe the same polytope.

Evidence that this is a true no-op for the pre-existing set: a captured/verified structural snapshot (`ncol`, `nnz`, `nrows`, `_prod_rows`, all three row maps, `_pos`) is **identical on all 31 previously-admitted instances**. Same formula in the same physical row on every box, so node counts and bounds cannot move.

## 2. Native producer: silently dropped the constraint

`spatial_producer`'s `bilinear_linform_specs` path claimed *every* support-contained row with **no count check**. A claimed row is excluded from the fixed rows AND is not regenerated by the kernel (which emits only the 4 McCormick rows per `BlfTerm`), so the constraint was **dropped from every node LP** — with the kernel default-ON since #764 graduated.

**The direction is sound**: a dropped relaxation row gives a superset feasible region, hence a weaker but still valid dual bound, and all 6 reachable instances still certify against `minlplib.solu`. It is nonetheless a silent tightness loss on a path whose premise is byte-for-byte reproduction of the build, so it now **declines** — which is what the monomial / affine-square `_claim(..., want=4)` paths already did.

Audit before choosing the predicate: **986 BLF terms claim exactly 4 rows, 38 claim more (13 instances, 6 of which actually build a spec → reachable), and 0 claim fewer.** So `!= 4 ⇒ decline` removes exactly the misclassified terms and no working coverage. Those 6 instances now land on the Python incremental path, which this same PR admits.

## Measurement (G2)

```
vs baseline: 31 -> 36 admitted
  NEWLY ADMITTED (5): prob02, prob03, st_e01, st_e08, st_e09
```

No admitted→declined flips. The `envelope row count != 4` bucket is **eliminated** (6 → 0); `st_e11` moves on to `bounds mismatch` (24 → 25), exactly as the plan predicted. `+5` beats the predicted `+4` — `st_e08` also admitted.

## Verification

- **G1** — `python/tests/test_861_envelope_row_classifier.py`, 9 tests, **all 9 verified failing before / passing after**. Includes a direct assertion that owned rows *move* with the box and unowned rows do not (the property support containment failed to capture), and the native-producer test carries a precondition assert so it cannot pass vacuously.
- **G3** — new panel `discopt_benchmarks/scripts/incremental_node_parity_panel.py`: **150 LP comparisons** over branching-style sub-boxes (integer pins, one-sided splits, interior shrinks) of the 5 newly-admitted instances, **0 disagreements**. A **firing control** — envelope coefficient corrupted *after* construction, so `_validate` can't catch it — produced **9 disagreements**, proving the panel isn't blind. (A first control attempt patched `_patch` globally and was caught by `_validate` at construction instead; re-placed.)
- **G4** — 7 instances end-to-end vs `minlplib.solu`: **0 violations**, no bound above its reference optimum.
- **G6** — smoke 850 passed / 1 skipped / 2 xpassed; adversarial 10 passed; incremental files 169 passed; native-kernel files 32 passed; `ruff` clean. No Rust touched.

**Retraction (CLAUDE.md rule 11):** the first G4 run reported `st_e11` as `OBJ-MISMATCH BOUND-ABOVE-OPTIMUM` against an oracle of `0.067038`. That value was **mis-transcribed by me** — `minlplib.solu` gives `st_e11 = 189.3116297`, which discopt matches (obj `189.31162974`, bound `189.31162969`). There was no violation; the corrected run is the one reported above.

Contributes to #861

https://claude.ai/code/session_01T9BWf4gy9kxqZBRbFkQWhd


jkitchin:
## CI failure triaged and fixed — fixture rot, not a regression

The `Python fast` lane failed two `#844` tests that the local smoke/adversarial gates missed. **Neither is a behaviour regression: both fixtures depended on the defect this PR fixes.**

- **`test_844_lp_spatial_deadline`** built its "fast path unavailable" control from bilinear *constraints*, and its own docstring documented why that worked — *"its bilinear constraint rows lift to 5 envelope rows where the closed-form patch emits 4."* That is precisely the misclassification fixed here, so the model now builds a structure, and the `require_incremental` guard was asserting that the engine declines a model it can serve.
- **`test_844_unresolved_child_soundness`** injects failures into `_relax_bound`, the **cold-path** entry point. Its model moved to the fast path, so the injection never fired and the test's own vacuity guard (`assert state["failures"] > 0`) failed. **The guard worked exactly as designed** — without it the test would have passed while asserting nothing.

### Repaired without weakening anything

Both fixtures now use a **trilinear** coupling, which is outside the closed-form patch's covered families *by design* (it regenerates bilinear / integer-power / affine-square envelopes only), so the decline is a property of the mathematics rather than of a bug. Each test now **asserts its own precondition**, so a future coverage widening points at the fixture instead of failing obscurely.

Also adds `test_failed_child_on_the_incremental_path_never_yields_a_false_certificate`, injecting into `IncrementalMcCormickLP.solve` so the unresolved-child soundness property is covered on the **fast path** — the common case since this PR, and previously untested there.

### Verification

Ran CI's exact fast-lane marker selection locally (the gap that let this through — my local gates were narrower than CI's):

```
7333 passed, 29 skipped, 1278 deselected, 5 xfailed, 2 xpassed in 532.87s
```

The plan doc now records the standing lesson for the remaining tasks: every admission widening can silently convert a cold-path test into a fast-path test, so CI's exact fast selection must be run locally before opening a PR, not just `-m smoke`.

Issue #707, opened 2026-07-17
Flow-aware trilinear/RLT envelope to close the ex1252-class residual (5134 → opt)


jkitchin:
Successor to #196 (SOS1 selector branching landed there via #706). #196 lifted `ex1252`'s global dual bound off its structural **0 → 5134.47** (sound) for the first time; this issue tracks the **remaining, precisely-characterized barrier** to certifying it.

Scoped narrowly so nobody re-treads the falsified levers below.

## The barrier (diagnosed, not hypothesized)

With `DISCOPT_SOS1_SELECTOR_BRANCH=1`, the bound plateaus at **exactly 5134.47** (identical to many digits across 60/120/240 s and all node-selection strategies). A floor diagnostic (temporary Rust dump of every processed node with bound < 6000) shows the frontier floor is a **single depth-1 node**:

```
depth=1  ind x18,x19,x20 = 1.0, 0.5, 0.5     (indicators still FRACTIONAL)
         sel x21,x22,x23 = 0.75, 0.18, 0.25   (x21∈[0.5,1] — line-1 selected)
         flow x0,x3       = 1.87, 2.0
         bound = 5134.5
```

Root cause: in this box the **trilinear McCormick underestimate of `x0·x3·x18` is ~0.81 vs the true ~3.74** — badly slack. The objective term is `(6329.03 + 1800·x15)·x0·x3·x18`; even with `x18=1` and `x21` pinned to the selected half, the term-wise McCormick envelope collapses the product. The relaxation is only tightened by spatially branching the **flow factors `x0,x3`**, and those factors are themselves integer-valued via the aux-binary decomposition (`x0 = x24 + 2·x25`, `x3 = x30 + 2·x31`, c34–c39, `x24…x35 ∈ {0,1}`).

## Proposed direction

A **flow-aware trilinear / RLT-style envelope** that exploits the integer structure of the flow factors to tighten `x0·x3·x18` directly, rather than the term-wise bilinear-of-bilinear McCormick. Candidates to evaluate:
- RLT products of the aux-binary rows (`x0 = x24 + 2·x25`) against the indicator/flow bounds, so `x0·x3` inherits a per-integer-level envelope instead of the `[0,3]²` box envelope.
- A tighter simultaneous trilinear envelope (e.g. the tightest concave/convex envelope of `x·y·z` over the box, or a lifted formulation) for the specific `affine(x15)·x0·x3·x18` shape.

[...]

PR #922, opened 2026-08-03
refactor: remove JAX from the equation-oriented solve path (staged)

### LINKED ISSUE #975: Stage 3: rename `_jax/` to match what it actually contains (66% of it never imports jax)
## Summary

`python/discopt/_jax/` is misnamed. It was created 2026-02-07 (`d326f654`, "T4 JAX
DAG compiler") holding only `__init__.py` + `dag_compiler.py`, at which point the
name was accurate. It then became the default home for solver numerics generally.
After the Stage 0–2 work on PR #922, the name is actively misleading.

This is **Stage 3** of `docs/dev/jax-removal-plan.md`, deliberately excluded from
PR #922 (see the plan's "Rescoped stages" section: "~1,976 mechanical sites would
bury the review of Stages 0–2. It becomes a follow-up PR after this one merges.")
Filing it so the deferral is tracked rather than resting in a plan document.

## Evidence (from the plan's measurement, not estimate)

- 66% of `_jax/`'s 68,408 LOC never import jax. Only 14,050 LOC (24 files) use a
  real JAX transform.
- Of 143 files added to the directory, **45 (31%)** import jax today. By month
  added: 2026-02 → 18/39 (46%), 2026-04 → 2/17 (12%), 2026-05 → 7/22,
  2026-06 → 15/39, 2026-07 → 3/25 (12%). The trend is monotone away from jax.
- Its largest residents are the guts of a global MINLP solver, not a JAX layer:
  `milp_relaxation.py` (3,005), `nonlinear_bound_tightening.py` (2,987),
  `obbt.py` (2,601), `primal_heuristics.py` (2,153), `gdp_reformulate.py` (1,824),
  `convexity/` (~10,600).
- `_jax/__init__.py` already carries a workaround for the misnaming: JAX is
  imported lazily via PEP 562 `__getattr__` specifically so that "importing a
  JAX-free submodule such as `discopt._jax.deadline` no longer drags in
  `dag_compiler` and the heavy JAX/XLA initialization."

The plan also records that the name **caused errors during the investigation
itself** — it reads as "this code needs JAX", which is false for two thirds of it.

## Scope

~1,976 mechanical call sites. Purely a rename + import rewrite; no behavior change
intended.

## Definition of done

- [ ] Directory renamed to something describing what it holds (name to be decided
      on the PR — the plan does not fix one).
- [ ] All ~1,976 import sites updated, including the string-literal ones — note
      `dag_compiler.py` uses `__import__("jax")`, so a plain `import jax` grep does
      **not** find everything; the same caution applies to any grep-driven rewrite
      of `discopt._jax` references.
- [ ] Backwards-compatible alias for `discopt._jax` if anything external imports it,
      or an explicit decision that the private underscore name needs no alias.
- [ ] **Bound-neutral verification per CLAUDE.md §5**: `node_count` and certified
      `objective` *exactly* unchanged on a certifying panel. A rename that moves a
      number is a bug, not an improvement.
- [ ] `pytest -m smoke`, the adversarial suite, and
      `python/tests/test_75_jax_free_solve_path.py -m ""` all still pass.

## Not in scope

The JAX-free solve path itself — that is Stages 0–2, delivered by #922.

Blocked on #922 merging first; rebasing a 1,976-site rename under an open PR that
touches the same tree would be pointless churn.

jkitchin:
> **Draft.** `main` is untouched; every stage is a commit on this branch so CI runs per stage.

## Goal — **met**

**Zero `import jax` reachable from `Model.solve()` on the equation-oriented path.**

Measured: **66 of 66** in-repo corpus instances solve with `jax` never entering
`sys.modules`, 0 crashes — with the **defaults**, no environment set.

```
status=optimal  obj=1.5153837731168318   jax in sys.modules: False
```

Asserted on `sys.modules`, never a source grep: `dag_compiler.py:225` reaches JAX
via `__import__("jax")`, a string containing no `import jax` substring.

## Stages — all done

- [x] **Stage 0 — attribution model** (`c3a3d648`, `7fc69f7f`). Attribution defects
      11 → 1; POUNCE given a bucket (tspn08 12.34 s, previously absorbed); buckets
      partition the wall.
- [x] **Stage 1 — DAG → `NlExpr` translator** (`1917a17b`, `360a1e69`, `08e1e0a1`).
- [x] **Stage 2 — separation tangents** → tape (`77dc8db2`). `DISCOPT_SEPGRAD`.
- [x] **Stage 3 — NLP derivatives** → tape (`ce212d67`, `ed024b6f`). `DISCOPT_NLP_EVAL`.
- [x] **Stage 4 — enforcement** (`77dc8db2`, `db40debf`). 66/66 corpus JAX-free,
      pinned by `test_75_jax_free_solve_path.py`.
- [x] **§5 panel + graduation** (`a2fb90d2`). Both flags now **default ON**.

The `_jax/` rename (~1,976 sites) remains deliberately excluded — it would bury
the review.

## CLAUDE.md §5 differential panel — both bars PASS

66 instances, `max_nodes` (not a time limit — verified: a fixed node budget gives
**319 nodes on three consecutive runs**, where under a time limit the JAX arm
alone moved **381 → 395**).

**Bar 1 — cert-clean.** 0 unsound bounds or statuses · 0 bounds past a proven
optimum · 0 false optima · 0 claims on known-infeasible instances · 0
certification regressions · **0 infeasible incumbents over 103 verifications,
each performed by the *opposite* engine** so a shared bug cannot vouch for itself
· 0 crashes · 0 objective drift on instances both arms solved.

**Bar 2 — net-positive.** Node count is **neutral — 44 of 46 identical**, which
for an engine swap is the right outcome: it reproduces the search rather than
perturbing it. The benefit is wall, measured to §9 (load gate, interleaved A/B,
standard deviations) on instances where both arms prove optimality in the **same
node count**, so the work is identical:

| | jax | tape | speedup |
|---|---|---|---|
| tspn05 | 10.40 ± 0.09 | 3.37 ± 0.02 | **3.08×** |
| nvs13 | 0.87 ± 0.02 | 0.36 ± 0.01 | 2.39× |
| nvs09 | 2.01 ± 0.07 | 0.96 ± 0.02 | 2.10× |
| **total (10 instances)** | **40.2 s** | **22.6 s** | **−43.7 %** |

**10 faster, 0 slower**, median **1.80×**; sd's are 1–3 % of the means. Predicted
by the plan: ~96 % of "JAX time" here was interpreted Python dispatch, not XLA.

Two instances (`contvar`, `heatexch_gen1`) tripped a first-draft drift check.
Both are `feasible/feasible`, so it was comparing **search progress**, not
correctness — the tape had explored 10× and 3× the nodes and **matched
MINLPLib's best-known solution** (heatexch_gen1: 154895.93293 vs reference
154895.933) where JAX sat 4.2 % and 19.3 % above. Re-adjudicated with evidence;
the rule now scopes to terminal instances.

## What is still on JAX, by design

All three **refuse loudly** rather than approximate, and fall back to the intact
JAX path:

- `dm.custom` / `CustomCall` — out of scope per the owner's 2026-08-03 scope
  decision; relaxing an opaque callable needs AD *through* it.
- **array-valued constraint bodies** (DAE collocation, vector constraints) — a
  tape node is scalar. Verified: `try_build` returns `None`.
- **Gauss-Newton** mode — `2 JᵀJ` is a *different matrix* from the exact Hessian.

Opt-outs are **tested, not merely retained**: `DISCOPT_NLP_EVAL=jax`,
`DISCOPT_SEPGRAD=jax`, and each *alone*, must still reach the JAX path. Asserting
jax **is** imported there doubles as the §6 control — it proves these solves do
real nonlinear work, so the JAX-free assertions cannot pass because nothing ran.

JAX remains an install dependency; what changed is that `import discopt` and a
solve no longer load it.

## Three defects found in the work itself

**1. Three wrong operator lowerings** (`08e1e0a1`). Stage 1 claimed "all 30
operators" on a corpus differential. That evidence never supported it: **a `.nl`
corpus exercises six operators** — measured over 316 MINLPLib instances, `log`,
`sqrt`, `exp`, `abs`, `sin`, `cos` and nothing else, because `.nl` has no opcode
for `sigmoid`/`softplus`/`entropy`/`centropy`/`signpower`. Nine of ten rewrites
had zero coverage; three were wrong — `entropy` had an inverted sign, `_sign`
passed `compare`'s arguments in the wrong order, `prod` was lowered as a `*`
chain when it is a one-argument **array** reduction. *Coverage must be counted in
operators, not instances.*

**2. A false `infeasible`** (`ed024b6f`). `pounce.NlProblem` was
`#[pyclass(unsendable)]`; touched from a worker thread it raised a
`PanicException`, which derives from `BaseException` and so slips past every
`except Exception` — surfacing as `infeasible` on `clay0303hfsg` where JAX
returned `feasible`. Fixed upstream (pounce #477/#478) and capability-probed
here, not version-compared.

**3. The deep-sharing "blocker" never existed** (`cd430a8f`). It was measured
against a stale local pounce build that exported no `NlExpr` — which also meant
the whole tape test file was silently `importorskip`ping.

## Filed along the way

- **#923** — dense `evaluate_lagrangian_hessian` silently all-zero (**fixed**, merged)
- **#925** — `dag_compiler` recursion limit (**fixed**, merged)
- **#927** — `ex1252` returns `optimal` with a bound 68 % above the optimum under
  the #707 flags. **Pre-existing on `main`**, confirmed byte-identical on
  unmodified `de66670c`; not a regression from this PR.
- **pounce #477** — `NlProblem` unsendable (**fixed**, merged)

## Verification

`pytest -m smoke` 908 passed · adversarial suite 10 passed · `-m correctness` 438
passed (1 pre-existing failure, #927, identical on `main`) · acceptance file 12
slow + 5 unit · tape files 58 passed · pinned ruff 0.14.6 via pre-commit + mypy clean.

---

## Stage 3 rename + notebook refresh (added after the original description)

`discopt._jax` is now `discopt._relax`, across 758 files. The package was named
for JAX when JAX built the relaxations; it no longer does, and a package whose
name describes none of its contents is a standing source of wrong conclusions.
`python/tests/test_relax_package_name.py` guards against reintroduction — it
fails immediately on any merge from a branch predating the rename, rather than
surfacing as an import error in whichever test happens to touch that module.

Closes #975.

All 62 documentation notebooks were re-executed against this tree (their
committed outputs dated from 2026-04-25 and nothing had re-run them since).
All 62 execute cleanly; across 372 paired objective/bound comparisons no
objective moved by more than 1e-4 relative.

Verified on the merge of `origin/main` (066048ad): `pytest -m smoke` 938
passed / 1 skipped / 2 xpassed; the #980 benchmark tests pass on the renamed
tree; guard test 3 passed.


jkitchin:
## Checkpoint — state as of the last push (`1917a17b`)

Stopping point ahead of a local power outage. Everything is committed and pushed; `main` is untouched at `badf9f4c`.

### Landed on this branch

| commit | what |
|---|---|
| `c3a3d648` | Stage 0 — attribute solve time at FFI boundaries, not by phase |
| `7fc69f7f` | Stage 0 — clamp native time to the wall (fixes an intermittent partition failure) |
| `360a1e69` | Stage 1 — probe uncovered atoms, not unusable probe points (analytic coverage 96.2% → 100.00%) |
| `1917a17b` | **Stage 1 (rescoped)** — DAG → POUNCE `NlExpr` tape translator |

### Rescope, mid-flight

pounce shipped [#469](https://github.com/jkitchin/pounce/issues/469), exposing its Rust AD tape to Python (`NlExpr` eval/gradient, `parse_nl_text`, an `Erf` TapeOp, and `NlProblem.hessian_vector_product`). Both remaining JAX jobs need the same primitive — value + gradient of a scalar expression at a point — so they now target **one** backend instead of two:

- Job 1 (NLP derivatives) → pounce tape
- Job 2 (separation tangents) → pounce tape *(was: `interval_ad`, which covers 6 of 30 operators; the tape covers all 30)*

Measured agreement against the JAX path being replaced: **40 corpus instances, 1517 points, max rel |Δvalue| 2.51e-16, |Δgrad| 6.64e-15, zero unsupported nodes.**

### Dropped

Graduating `DISCOPT_ANALYTIC_SEPGRAD`. Its §5 panel passed cert-clean (0 unsound bounds, 0 cert regressions, 0 objective drift) but was **neutral** on benefit — 48/49 instances identical, 1 better, 0 worse — and §5 keeps a neutral flag OFF. It also bought nothing in isolation, since `nlp_evaluator.py:22` imports JAX on every nonlinear solve regardless.

### Open before Stage 2/3

A deeply shared chain (`node = node*node + node`) does not terminate at depth 10, where the DAG holds only ~20 distinct nodes. The translator memoises on `id(expr)`, so the duplication is downstream — most likely `NlExpr` not sharing on repeated references. Pinned as a strict, non-running `xfail` in `test_75_nl_expr_compiler.py`. **Must be resolved before the tape carries solve traffic.** The 40-instance corpus differential passed, so ordinary sharing depths are fine today.

### Next

1. Resolve the sharing blowup
2. Stage 2 — route separation tangents through the translator (§5 panel)
3. Stage 3 — route NLP derivatives through it (§5 panel; no pounce blocker remains)
4. Stage 4 — flip the acceptance test to a passing assert

Issue #928, opened 2026-08-05
Warm pure-LP path drops the caller's time_limit, so every per-LP budget is unenforced


jkitchin:
## Summary

`MilpRelaxationModel.solve` accepts a `time_limit`, and its default
`backend="simplex"` **pure-LP fast path drops it on the floor**. Roughly 13 call
sites in `_jax/mccormick_lp.py`, plus `lp_spatial_bb.py` and `integer_ratio.py`,
carefully compute a per-LP budget and hand it to a callee that ignores it, so
every caller that tries to bound an LP in time is silently unbounded in time.

The plumbing to fix it already exists and is exercised, behind
`DISCOPT_LP_WARM_DEADLINE` (`_lp_warm_deadline_enabled` in
`python/discopt/_jax/milp_relaxation.py`, tests in
`python/tests/test_917_lp_warm_deadline.py`). It is **default OFF and has now
twice failed to graduate**, for a reason that is not "we haven't measured
enough". This issue is to track the *real* fix; see "Why the flag is not it".

This is the residual left open by #814 (closed) and is the reason
`python/tests/test_time_limit_contract.py::test_solve_returns_within_time_limit[hda-10.0]`
is marked `xfail(strict=False)`.

## Evidence that the deadline is dropped

`solve_lp_warm_std` / `_solve_lp_warm` / `_solve_lp_warm_equilibrated` take no
deadline, and `lp_bindings.rs` hardcodes `SimplexOptions { deadline: None }` —
while the MILP route (`solve_milp_csc_py(time_limit_s=…)`) wires it up and the
dual/primal pivot loops already poll it every 256 pivots. The capability is
there; the pure-LP path just never asks for it.

Measured on `nvs24` (`scratchpad/nvs24_arm.py`, `nvs24_profile_evidence.txt`):
`solve(time_limit=0.202)` → `_solve_lp_warm` → **47.03 s**, one `DualPivotLoop`
59 494 degenerate dual pivots deep with Bland never activated — a 3.9 s solve
budget turned into 53 s (13.5×).

## How it surfaces at the top level

On `hda` at `time_limit=10` the #138 root-relaxation fallback is the sole bound
producer. Its optional separated-relaxation phase calls `solve_at_node` with a
budget clamped to what is genuinely left of the fallback's own grant
(`_fb_left()`, added by the time-limit reserve work) — and that clamp is inert,
because this is the call site that drops it. The phase then runs ~4 s past its
grant.

[...]

Issue #917, opened 2026-07-31
Model.solve forfeits 35% of the wall budget when the #844 LP-spatial reserve is taken and the primary does find an incumbent


jkitchin:
`Model.solve` forfeits 35% of the wall budget whenever the #844 LP-spatial reserve is taken and the primary *does* find an incumbent

## What happens

`Model.solve` (`python/discopt/modeling/core.py:4250-4285`) unconditionally deducts a
reserve from the caller's `time_limit` for every model the #844 no-incumbent fallback
could serve:

```python
if _is_in_scope(self, mixed=_lp_spatial_mixed_fallback_enabled()):
    _fb_reserve = 0.35 * time_limit
...
_primary_tl = time_limit - _fb_reserve
```

and spends it only in one case (`core.py:4331`):

```python
if _fb_reserve > 1.0 and result.objective is None:
```

So a solve whose primary path *does* return an incumbent gets 65% of the budget the
caller asked for, and the other 35% is never spent by anyone. The reported status is
`time_limit` — at 65% of the stated limit.

## Evidence (this is a measurement, not a reading)

Scope predicate, evaluated directly (no solve):

| instance | `_is_in_scope` | primary budget @ `time_limit=60` | observed wall (OFF) | observed wall (ON) |
|---|---|---|---|---|
| nvs17 | **True** | 39.0 s | 39.5 | 16.8 (certified early) |
| nvs19 | **True** | 39.0 s | 40.7 / 41.2 | 39.4 / 39.5 |
| nvs24 | **True** | 39.0 s | 71.7 | 39.4 |
| tanksize | **False** | 60.0 s | **60.4 / 60.7** | 17.3 |

The data carries its own controls, in both directions:

* `tanksize` is the one instance of the four **out** of scope — and it is the one that
  runs the full 60 s.
* Every in-scope run that used its whole primary budget **and had an incumbent** stopped
  at ~39 s = 0.65 × 60 and returned `time_limit` with 21 s unspent.
* `nvs24` OFF is the in-scope run that found **no** incumbent — the one case the reserve
  is spent — and it is the only one that ran past 60 s (71.7 s: 39 s primary + the 21 s
  reserve + overhead). So the reserve is demonstrably being *taken* from every in-scope
  solve and *used* by only the no-incumbent minority.

Walls are from the #764 graduation panel runs
(`discopt_benchmarks/results/issue764_native_kernel_graduation_panel_*.txt`), which run
[...]

PR #929, opened 2026-08-05
fix(solver): honor time_limit as a contract, not a hint

jkitchin:
## The defect

`solve(time_limit=T)` overran by a near-constant 6-8 s regardless of `T`, so no
limit below ~10 s was honorable at all:

| limit | wall (hda, before) | overrun |
|------:|-------------------:|--------:|
|   2 s |              9.22 s| +361%   |
|   5 s |             12.20 s| +144%   |
|  10 s |             18.20 s|  +82%   |
|  20 s |             25.76 s|  +29%   |
|  40 s |             46.19 s|  +15%   |

SCIP 10.0.2 on the same instances lands within **0.03-0.09 s** of the limit at
every limit tested (2 s / 5 s / 10 s, 9/9 runs) while still reporting a finite,
monotonically improving dual bound. The limit is a contract, not a hint.

This is not merely untidy: it silently handicaps every benchmark comparison. In
the in-repo global50 run BARON was held to 60.3 s on `casctanks` while discopt
took 66.3 s, so any shifted-geometric-mean or wall-ratio computed against
another solver was flattering by that margin.

### Root cause (measured, not hypothesised)

The dominant term is the issue-#138 root-relaxation fallback. It was granted
`_ROOT_FALLBACK_FLOOR_S` (3.0 s) of **fresh** budget *against an already-spent
deadline* — the whole grant sat outside the contract by construction — and it
then overran even that grant by 32-35%:

* `hda` at `T=5`: fallback entered 1.95 s **past** the deadline, ran a further
  4.25 s → wall 11.75 s.
* `hda` at `T=60`: entered at 60.16 s, ran 4.04 s against its 3.0 s grant.

So the old comment's claim — *"the fallback's own internal budget keeps the
overrun small"* — was false at the grant it was describing.

A `LOOPPROBE` also showed the B&B loop breaks on its **first** iteration on
`hda` (`elapsed=6.95 iters=1` against a 5.0 s limit): the entire pre-deadline
cost is mandatory root work (2.0 s uninterruptible Rust presolve + 1.3 s
convexity classification + ~3.4 s root McCormick LP), so a reserve placed only
inside the node loop would have been irrelevant for this class.

## The change

Invert the post-deadline **grant** into a pre-deadline **reserve**.
`_ROOT_FALLBACK_RESERVE_S` is withheld from the search up front and spent inside
`time_limit`:

* The spatial loop **releases** the reserve the moment the tree holds a finite
  global lower bound (or is exhausted), so an instance that produces a bound of
  its own sees a byte-identical deadline. The release is a latch and is only
  evaluated in the last `_rr_reserve_s` seconds, so a solve that finishes
  earlier never reads `tree.stats()` for it at all.
* While the tree is bound-less at the reserve boundary — exactly the state in
  which the fallback is the sole bound producer — the loop stops and hands the
  slice over. The fallback is *anytime with respect to its grant* (measured on
  `hda`: 0.25 s → -141697, valid but weak; 2 s → -64473, the full bound), so a
  smaller pre-deadline slice buys a weaker rigorous bound rather than none.
* The post-search fallback takes `max(0, remaining)` with **no floor**, and the
  #654 short-circuit triggers on the reserve rather than on a bare exhausted
  deadline.
* The fallback's own optional separated-relaxation solve is clamped to what is
  genuinely LEFT of its grant (`_fb_left()`) — the same rule-1 / rule-2 split
  `_fb_stop` already applies to whether that phase *starts*.

### A false `infeasible` this change introduced, and how it was caught

The first cut broke an invariant: the no-incumbent status branch reads
`elif wall_time >= time_limit: "time_limit"` … `else: "infeasible"`, so an early
break left `wall_time < time_limit` with an unfinished tree and fell through to
a **certified `infeasible` on a feasible model** — the worst class of error
(measured: `hda` at `T=10` and `casctanks` at `T=5`). `_rr_reserve_yield` now
marks the interruption and forces `status="time_limit"`,
`_gap_certified=False`; the yield is additionally suppressed when
`tree.is_finished()` so a legitimately certified `infeasible` is never
downgraded.

### CLAUDE.md §5 differential panel

`origin/main` worktree vs this tree, **interleaved per instance** (alternating
which arm runs first) in isolated subprocesses, each arm asserting both
`discopt.solver.__file__` *and* the presence/absence of the
`_ROOT_FALLBACK_RESERVE_S` marker (§8). 66 in-repo instances, `T=20 s`.

**Bar 1 — cert-clean: PASS**

| check | comparisons executed | violations |
|---|---:|---:|
| dual bound vs `minlplib.solu` oracle | 116 | **0** |
| `gap_certified` regression | 66 | **0** |
| lost incumbent | 66 | **0** |
| false `infeasible` | 66 | **0** |
| certified objective drift | 49 | **0** |
| bound crosses incumbent | 4 | **0** |
| finite bound → `None` | 17 | **0** |

**Bound-neutral where it applies**: node counts are **exactly** unchanged on all
49 instances that close (14470 == 14470, 0/49 changed).

**Bar 2 — net-positive: PASS**

|                | base (origin/main) | new (reserve) |
|----------------|-------------------:|--------------:|
| total overrun  |            44.97 s |   **24.68 s** |
| max overrun    |            11.07 s |    **7.85 s** |
| `hda` wall     |            31.07 s |   **20.74 s** |

**Bound quality** on the 17 non-closing instances — the only class where the
fallback is live and the reserve can change anything: 1 looser
(`clay0303hfsg` 20947.6 → 19513.3, still far below its 26669.1 incumbent),
1 tighter (`4stufen` 20712.2 → 21372.2), 0 lost.

### Test suites

* `pytest -m smoke` — **904 passed, 1 skipped, 2 xpassed**
* `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — **10 passed**
* New: `python/tests/test_time_limit_contract.py` pins the contract (`hda` at
  5 s / 10 s, `casctanks` at 5 s) and the anytime property of the dual bound.

Rust untouched (`git diff origin/main -- crates/` is empty), so `cargo test` is
not implicated; the panel's baseline arm reuses the same built extension, which
is why the arms differ only in `solver.py`.

## What this does NOT fix, and why it is not hidden

One case stays over the line: `hda` at `time_limit=10` is bimodal — 10.60 /
10.78 / 12.79 / 13.03 / 12.94 s over 5 load-gated reps (mean 12.03, sd 1.23)
against 12.50 s allowed. It is marked `xfail(strict=False)` against **#928**,
filed by this PR, with the measurement in the source.

The two modes are attributed, not guessed: they are "did the #138 fallback's
optional separated-relaxation phase start". The fast reps report the weak dual
bound -141697 (grant already spent, `_fb_stop` returns the banked bound); the
slow ones the full -64473 (the phase started and then ran ~4 s past its budget).
The `_fb_left()` clamp this PR puts on that call site is correct and in place —
it is inert because `solve_at_node`'s warm pure-LP path *discards the
`time_limit` it is handed*. That is a pre-existing defect one layer down, and
it had no open tracker: the flag's panel script and companion test are named
`issue917_*` / `test_917_*`, but #917 is a different, closed issue, and #814
(the ancestor overrun issue) is closed too. Hence #928.

`DISCOPT_LP_WARM_DEADLINE=1` makes all four cases pass (11.14 s ± 0.04 on this
one) and is **still not the fix**, for two independently sufficient reasons —
both now recorded in the flag's docstring per §4/§11:

* **It fails §5 bar 2 on the subset built to favour it.** That docstring named
  the settling experiment: a load-gated, multi-rep panel over the instances
  where the deadline actually binds, since a corpus average is diluted by the
  ~50 instances that are bit-identical either way. Run over the 17 non-closing
  instances, 3 reps at a 20 s budget, the ON−OFF deltas are −3.2 / +6.6 / +5.7 s
  — they **flip sign**, mean +3.0 s with sd 5.3 s, and the OFF arm alone spans
  21.3–30.9 s. All 3 reps `CERT_CLEAN=True`; cert-clean is necessary, not
  sufficient (the `DISCOPT_CUT_INHERIT` rule).
* **It buys punctuality by losing the bound**: -64473 → -141697 in 5/5 `hda`
  reps, and an earlier panel saw the severe form (bchoco08 `1.0 → None`). Wrong
  trade under §1.

Weakening `_SLACK_S`/`_SLACK_FRAC` to turn that case green is exactly what §1
forbids, so it is xfailed and tracked instead — it keeps running, keeps
reporting, and will xpass when #928 lands.

**A §4/§11 by-product worth carrying forward:** in that 3-rep run `nvs05` and
`clay0303hfsg` reported bound deltas in *opposite directions between reps at
identical flag state*. Their bounds are timing-nondeterministic, which also
re-reads this PR's own bound-quality table above — `clay0303hfsg`'s single
"looser" cell is one of those two instances and should not be read as an effect
of the reserve.

Contributes to #654. Files #928.

### DIFF
--- a/python/discopt/_jax/milp_relaxation.py
+++ b/python/discopt/_jax/milp_relaxation.py
@@ -255,6 +255,11 @@ class MilpRelaxationResult:
 def _lp_warm_deadline_enabled() -> bool:
     """Honour the caller's ``time_limit`` on the warm pure-LP node path.
 
+    Tracking issue: **#928**. (This helper's panel script and companion test are
+    named ``issue917_*`` / ``test_917_*`` for historical reasons only — #917 is a
+    different, closed issue about the #844 LP-spatial reserve. Do not follow that
+    number.)
+
     **The defect it fixes.** ``MilpRelaxationModel.solve`` takes a ``time_limit``,
     but its default ``backend="simplex"`` pure-LP fast path dropped it on the floor:
     ``_solve_lp_warm`` / ``_solve_lp_warm_equilibrated`` / ``solve_lp_warm_std`` took
@@ -292,10 +297,54 @@ def _lp_warm_deadline_enabled() -> bool:
       82.1 / 79.2 / 175.6 s. Cells over budget went 15 -> 18.
 
     So it stays OFF: cert-clean is necessary, not sufficient (CLAUDE.md §5 bar 2, the
-    ``DISCOPT_CUT_INHERIT`` rule). What would settle it is a load-gated, multi-rep
-    panel over the instances where the deadline actually binds — on the corpus average
-    the effect is diluted by the ~50 instances that finish far inside 15 s and are
-    bit-identical either way.
+    ``DISCOPT_CUT_INHERIT`` rule).
+
+    **The settling experiment has now been run, and it does not graduate the flag.**
+    The paragraph here used to say what would settle it: *a load-gated, multi-rep panel
+    over the instances where the deadline actually binds* — the corpus average being
+    diluted by the ~50 instances that finish far inside the budget and are bit-identical
+    either way. That panel is ``issue917_lp_warm_deadline_panel.py --instances`` over the
+    17 non-closing in-repo instances, 3 reps at a 20 s budget, load-gated per CLAUDE.md §9
+    (``scratchpad/issue917_bind.log``):
+
+    ====  ============  ===========  ==========  =========
+    rep   overrun OFF   overrun ON   cells OFF   cells ON
+    ====  ============  ===========  ==========  =========
+    1     30.9 s        27.7 s       7           5
+    2     21.3 s        27.9 s       5           6
+    3     29.6 s        35.3 s       7           5
+    ====  ============  ===========  ==========  =========
+
+    All three reps ``CERT_CLEAN=True`` (0 unsound / 0 lost_bound / 0 cert_regressions /
+    0 lost_incumbents), so bar 1 holds on the binding subset too. Bar 2 does not: the
+    paired ON-OFF deltas are -3.2 / +6.6 / +5.7 s — they *flip sign*, mean +3.0 s with
+    sd 5.3 s, and the OFF arm alone spans 21.3-30.9 s. Concentrating the panel on the
+    subset where the deadline binds did not resolve the effect out of the noise; if
+    anything it leans the wrong way. **The flag stays OFF, and this is now a measured
+    negative rather than an open question.**
+
+    Two by-products of that run are worth keeping. First, ``nvs05`` and
+    ``clay0303hfsg`` reported ``tighter_bound``/``looser_bound`` in *opposite
+    directions between reps at identical flag state* — their bounds are
+    timing-nondeterministic, so a single-rep bound delta on either is not evidence of
+    anything. Second, on ``hda`` at ``time_limit=10`` (5 reps per arm,
+    ``scratchpad/hda10.py``) the flag buys wall consistency at the price of the bound:
+
+    ==============  ==================  ================================
+    WARM_DEADLINE   wall                dual bound
+    ==============  ==================  ================================
+    0 (default)     12.03 s ± 1.23      -64473 in 3/5 reps, -141697 in 2
+    1               11.14 s ± 0.04      -141697 in 5/5 reps
+    ==============  ==================  ================================
+
+    The OFF arm is bimodal because the #138 fallback's separated-relaxation phase is
+    what overruns: when its grant is already spent by the time a first bound is in
+    hand, ``_fb_stop`` returns the weak bound at ~10.7 s; when the grant still has room
+    the phase starts and then ignores the budget passed to it — *this* drop — for a
+    full ~4 s, reaching the full bound at ~12.9 s. Turning the flag on clamps that
+    phase, which is why the wall tightens to ±0.04 s and why the bound is the weak one
+    every time. Trading a rigorous bound for punctuality is the wrong trade under
+    CLAUDE.md §1, which is the second, independent reason the flag is not the fix here.
 
     **What it took to become cert-clean** — the first two cuts were NOT, and both
     failures are worth keeping (CLAUDE.md §11):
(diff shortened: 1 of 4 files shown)


jkitchin:
## Update: the residual is attributable, and #928 is only half of it

Probing the `hda-10.0` residual further turned up the larger cause, now filed as
**#930**. On `hda` at `time_limit=10` the search's own root node returns
`status='optimal'` with `lower_bound=-64473.442402437024` after 3.05 s — and
three seconds later the tree's `global_lower_bound` is still non-finite, so this
PR's reserve yields and the fallback spends a further **2.69 s** re-deriving
`-64473.442402437024`. Same number, every printed digit.

That duplicated solve, not the #928 deadline drop, is the bulk of the residual.
It also means this PR delivers only the first clause of the plan it implements
("compute the root relaxation early, inside the budget, *and the bound then
participates in the search*") — #930 is the second clause, and is the one that
buys the genuine anytime property.

Nothing here changes what this PR does or how it was verified; it sharpens the
"what this does NOT fix" section above. #930 carries the measurement, including
the warning that `safe_bound` is `None` on both calls, so the fix is probably
"produce the safe bound at the root", not "adopt the raw LP objective" — the
latter would be a false certificate.

## Also in this PR now: a real CI catch

`test_912_wall_budget_inventory` failed, correctly. Extracting the fallback's
rule-2 arithmetic into `_fb_left()` rewrote that gate as a bare subtraction with
no comparison, which the inventory scanner does not match — so the refactor made
a wall-clock gate *invisible* to the ratchet built to keep every such gate
visible. Deleting the stale KNOWN entry would have been precisely the erosion
that test exists to stop.

Fixed at the source: the grant is materialised once as an absolute deadline in
the scannable `<clock>() + <budget>` form, and `_fb_left`, `_fb_stop`,
`_build_deadline` and `_base_build_deadline` all read it — those four had each
been recomputing the same expression (three literal copies of `_fb_t0 + max(0.0,
float(time_limit))`). Behaviour unchanged; one source of truth.

The entry is recategorised `residual` → `contract`, which is the substantive
half: the fallback used to receive a *fresh* grant handed out past an already
spent deadline, so the gate decided how much **extra** work ran beyond the
caller's budget. It is now a reserve withheld up front and spent inside
`time_limit`, so it answers "have I used up my slice of the caller's budget?".
Published residual count 20 → 19, justified inline as that module requires.

Issue #370, opened 2026-07-01
Retire the JAX LP-IPM (lp_ipm.py) — second half of the #364 fallback-chain retirement


jkitchin:
## Summary

Retire the pure-JAX LP interior-point method (`discopt/_jax/lp_ipm.py`, `lp_ipm_solve` / `lp_ipm_solve_batch`) — the second half of #364's "retire the fallback chain." The first half (the JAX LP-IPM last resort in the standalone `_solve_lp`) landed in #368. Verification (below) shows the remaining JAX-IPM uses are all either (a) fallbacks for a **no-Rust-extension install** that discopt does not actually support, or (b) forward solves already covered by POUNCE / the Rust simplex.

**POUNCE stays** — it is the NLP solver for MINLP subproblems *and* the interior-point forward solver the differentiable layers require (see below). This issue removes only the JAX IPM.

## Why JAX-IPM is not needed

**Differentiability is already IFT / implicit-KKT, not unrolled-IPM.** `_jax/differentiable_lp.py` computes LP gradients via `@jax.custom_jvp` (OptNet-style implicit differentiation of the KKT conditions) with the **forward solve on POUNCE** (`solve_lp_kkt`, via `jax.pure_callback`). QP (`differentiable_qp.py`) and NLP/MINLP are likewise implicit-KKT / envelope-theorem. No gradient path unrolls the IPM.

Note the differentiable layers need an **interior / analytic-center** forward point (a simplex *vertex* has a degenerate active set → singular KKT sensitivity). POUNCE provides it; the Rust simplex cannot be used there. Since POUNCE stays, this is already handled.

## Verification: `lp_ipm_solve_batch` is a fallback, not the default path

In `_solve_milp_bb` (the spatial/MINLP B&B), node-relaxation solving is gated:

```python
node_engine = "simplex"                       # default
_simplex_nodes = bool(SIMPLEX_AVAILABLE)       # True in any built install
_structured_avail = _simplex_nodes or (prefer_pounce and _pounce_nodes_avail)
if _structured_avail:      # Path B — Rust exact-vertex warm simplex per node  ← DEFAULT
    _node_solve(...)       # = _solve_node_lp_simplex
elif n_batch > 1:          # JAX-IPM batch  ← only when the Rust simplex is UNavailable
    lp_ipm_solve_batch(...)
```

[...]

PR #935, opened 2026-08-06
refactor(qp): #75 solve in POUNCE, differentiate with custom_jvp

jkitchin:
refactor(qp): #75 solve in POUNCE, differentiate with custom_jvp

Removes the last pure-JAX solver from the solve path. `_jax/qp_ipm.py` (661
lines) is deleted; `_jax/ipm.py` and `lp_ipm.py` went with #370, so no JAX IPM
kernel remains. The architecture the plan describes was already in place for LP
— POUNCE produces the forward KKT point, JAX supplies a post-solve `custom_jvp`
implicit-KKT rule — and this finishes the unmigrated QP half.

Two consumers still called the JAX QP IPM:

1. `_jax/differentiable_solve.py` used it purely as a *forward* solver, reading
   only `.x`/`.obj`. The derivative never came from it: `differentiable_qp`'s
   JVP has fed on `solve_qp_kkt` all along. Adds `_qp_forward` (mirroring the
   existing `_lp_forward`) and migrates all three call sites.

   Entry experiment (CLAUDE.md §4), 12 random strictly-convex QPs, SPD Q, n in
   2..8: worst relative disagreement with `qp_ipm_solve` 1.28e-11, and POUNCE's
   point independently verified — primal feasibility <= 4.4e-16, KKT
   stationarity residual <= 4.0e-15.

2. `solver.py::_solve_qp_jax`, the QP rescue. Removed rather than ported: it did
   not degrade gracefully, it degraded *past the guard*. `_solve_qp_matrix`
   rejects a POUNCE point that fails primal feasibility or KKT stationarity;
   the rescue then re-solved the same QP and reported `status="optimal"` with
   `bound=obj_val`, `gap=0` and `convex_fast_path=True` on nothing but its own
   convergence flag — checking neither condition. So the only state that could
   reach it was exactly the state where a verified engine's answer had just been
   thrown out, and it answered there with an unverified certificate. Under §1
   (the solver's product is its certificate) that is worse than no rescue.
   `_solve_qp` now refuses loudly, keeping the `qp-pounce-no-result` marker.

   The independence argument for a second implementation does not survive
   either: `pounce-solver` is a core dependency, not an extra, so the "POUNCE is
   not installed" arm was unreachable in any supported install.

Tests. `test_qp_backend_seam.py` gains
`test_pounce_no_result_refuses_rather_than_certifying`, verified against the
baseline (origin/main @ d3c3a44e) to fail there on the `status` assertion, which
reports "iteration_limit" — the rescue ran and returned its unconverged iterate
as the result. Its sibling `test_pounce_no_result_logs_tracking_marker` passes in
*both* states, having never asserted the status; that is why it needed a
companion rather than an edit.

`test_p1_milp_bb_soundness.py::test_miqp_batch_non_kkt_recovered` was measured
vacuous before removing its scaffolding — a probe (§6) recorded 0 invocations of
`qp_ipm_solve_batch` during the test. Retargeted to the live POUNCE node path
with a teeth assertion rather than deleted. `test_qp_batch` in
`test_lp_qp_solvers.py` is removed with a rationale comment: it covered
`qp_ipm_solve_batch`, which had no caller and has no batched POUNCE analogue.
Four other tests retargeted to `solve_qp_kkt`, dropping
`assert int(state.converged) in (1, 2)` — the exception `solve_qp_kkt` raises is
the stronger check.

Docs. `skills/agents/jax-ipm-expert.md` is deleted: every module it documented
(`ipm.py`, `ipm_iterative.py`, `ipm_callbacks.py`, `lp_ipm.py`, `sparse_ipm.py`,
`qp_ipm.py`) is now gone, so a shipped agent doc described a subsystem that does
not exist. Its 8 cross-references are updated. Several sibling agent docs also
asserted JAX IPM was the default NLP backend and that `nlp_solver="ipm"` selects
it; `"ipm"`/`"sparse_ipm"` are back-compat aliases silently resolving to POUNCE
(`solver.py:7142`), so those claims are corrected. `pounce-only-roadmap.md`
lines 122/980 claimed the LP/QP kernels survive for the differentiable path —
superseded by #370 and this change, and retracted in place per §11.

Verification: `pytest -m smoke` 904 passed, 1 skipped, 2 xpassed;
`pytest -m slow test_adversarial_recent_fixes.py`; `ruff check` and
`ruff format --check` clean; `mypy` shows 8 pre-existing errors in untouched
files (`nn/tree.py`, `export/nl.py`, `estimate.py`), none in the touched ones.

Contributes to #75

---

Follow-on note, not part of this PR: the plan's blocker #3 (Gauss-Newton
residual Hessian) is classified as needing pounce-repo work. It does not —
`_tape_nlp_evaluator.py` already exposes `jacobian_structure()` and
`evaluate_jacobian_values()`, so the residual vector can be taped as the
constraint rows of an auxiliary problem and `2 JᵀJ` formed from the sparse
Jacobian. That would also beat the JAX version, which builds `jac` dense.


jkitchin:
Folded into #922 as requested — the commit now rides there
(`refactor(qp): #75 solve in POUNCE, differentiate with custom_jvp`), so this PR
is redundant rather than abandoned. The work itself is unchanged and is not
being dropped.

#922 was also rebased onto `main` in the same pass (29 linear commits, no merge
commits, 0 behind). One hazard worth recording, since it would have silently
shipped a regression:

`63c8f551` on #922 reverted `70e7cdbb`/`72ed8e39` so the GAMS export fix could
go out as its own PR (#934). #934 has since merged into `main`. On a rebase onto
`main`, git skips those two commits as already-applied and then replays the
revert — which deletes the shipped fix from the branch again. A trial rebase
confirmed this: the rebased tree differed from a merge-based reference by
exactly `export/gams.py`, `test_gams.py`, and `bench_three_way.py`, all removed.

The rebase was redone with that now-obsolete revert dropped, and verified by
tree hash against the merge reference:

    rebased tree  : ef792b2edd0427d3b2193b3f3036ce71785c1285
    merge reference: ef792b2edd0427d3b2193b3f3036ce71785c1285

Gates on the rebased branch: `pytest -m smoke` 908 passed / 1 skipped /
2 xpassed; `pytest -m slow test_adversarial_recent_fixes.py` 10 passed;
`test_qp_backend_seam.py` + `test_gams.py` 58 passed / 2 skipped.

PR #939, opened 2026-08-07
fix(#936): certify vectorized-API convex QPs from their exact Hessian

### LINKED ISSUE #936: Convex QPs written with the vectorized/indexed-summation API are never certified convex, so they route to spatial B&B and do not converge
## Summary

A convex QP written with discopt's **vectorized / indexed-summation modeling API** is not
certified convex by any route, so it falls through the guard at `solver.py:7062`
(`_pure_continuous_force_spatial = True`) into spatial McCormick B&B, where it does not
converge. Three independent defects in the convexity-certification layer stack up to
produce this. All three are pre-existing on `main` (none of the files involved are touched
by `refactor/remove-jax-from-solve-path`).

Found while re-executing `docs/notebooks/` — three notebooks time out because of it
(`qp_solver`, `decision_focused_learning`, and a candidate third, `tutorial_dae`).

### Defect 1 — `_expr_to_polynomial` has no `SumOverExpression` branch

`milp_relaxation.py:_expr_to_polynomial.visit` handles `SumExpression` but not
`SumOverExpression`, which is what `dm.sum(f, over=...)` builds. The walk returns `False`,
so `extract_quadratic` abstains — while the byte-identical body written with Python's
builtin `sum` extracts fine. Isolated at n=6, seed 42, `Sigma = L Lᵀ + 0.01 I`:

| body | root node | `extract_quadratic` |
|---|---|---|
| array vars + `dm.sum` | `SumOverExpression` | `None` |
| array vars + builtin `sum` | `BinaryOp` | `Q(6,6)`, psd=True |
| scalar vars + `dm.sum` | `SumOverExpression` | `None` |
| scalar vars + builtin `sum` | `BinaryOp` | `Q(6,6)`, psd=True |

Array-indexing is irrelevant; `SumOverExpression` is the sole cause.

### Defect 2 — `DISCOPT_PSD_QFORM` is default-OFF, and Gershgorin is too loose

With the exact PSD-on-Q certificate gated off, the fallback is the interval Hessian +
Gershgorin row-sum bound, which cannot certify a dense covariance Hessian. A/B on a
Markowitz QP (`min wᵀΣw` s.t. `Σw = 1`, `μᵀw ≥ 0.08`, `0 ≤ w ≤ 1`), with Defect 1 fixed:

| n | `DISCOPT_PSD_QFORM=0` | `DISCOPT_PSD_QFORM=1` |
|---|---|---|
| 5 | optimal, 89 nodes, 3.26 s | optimal, 0 nodes, 0.17 s |
| 20 | **feasible only**, 557 nodes, 120 s limit hit | optimal, 0 nodes, 0.02 s |

`Sigma = L Lᵀ + 0.01 I` is positive definite by construction (min eigenvalue 0.0100013 at
n=20), so the `is_convex=False` verdict at n≥5 is a false negative, not a genuine
nonconvexity.

### Defect 3 — the interval-Hessian walker cannot handle the vectorized API at all

`convexity/interval_ad.py:596` (and :595 for `Parameter`) does
`v = float(np.asarray(expr.value))`, which raises
`TypeError: only 0-dimensional arrays can be converted to Python scalars` for a
vector-valued `Constant`/`Parameter`. `certify_convex` catches only `ValueError`, so the
`TypeError` escapes and is swallowed by `_classify_model_convexity`'s broad
`except Exception` → reported convexity-unknown. Array `Variable`s are separately refused
by design (`"Interval Hessian requires scalar variables"`).

Net effect: **no model written in the vectorized API can be certified convex.**

Repro (`decision_focused_learning.ipynb`'s oracle QP — 3 variables, 1 constraint):

```python
m = dm.Model("allocation")
x = m.continuous("x", shape=(3,), lb=0.0, ub=1.0)
c = m.parameter("cost", value=np.asarray(cv, dtype=np.float64))
m.minimize(dm.sum(c * x) + 1.0 * dm.sum(x * x))
m.subject_to(dm.sum(x) == 1.0)
m.solve(time_limit=60)
# -> feasible (not optimal), 14115 nodes, 14931 POUNCE node solves, 60.4 s
```

The Hessian here is `2·I`. Each POUNCE call is 1–5 ms; the cost is entirely the ~15,000
spatial B&B nodes.

## The fix is small: discopt already computes the proof and discards it

`problem_classifier.extract_qp_data` handles the vectorized model fine. On the exact model
above:

```
classify_problem  -> ProblemClass.QP
extract_qp_data   -> QPData, Q = [[2,0,0],[0,2,0],[0,0,2]]
```

Diagonal, trivially PSD. The convexity guard never consults it — it asks the interval
walker instead. Proposed fix: when `classify_problem` returns `QP`/`MIQP` and
`extract_qp_data` yields a usable `Q`, take the exact eigenvalue PSD verdict on that `Q`
directly. This is box-independent and rigorous (a global quadratic convexity proof holds on
every sub-box), and it is the same argument `_certify_quadratic_psd` already makes.

Plus:
- add the `SumOverExpression` branch to `_expr_to_polynomial` (Defect 1);
- raise `ValueError` (the walker's established abstain signal) rather than letting a
  `TypeError` escape for vector-valued leaves (Defect 3), so the abstention is deliberate
  instead of accidental.

## Verification notes

- The `SumOverExpression` fix is exact, not approximate: `max |body − (xᵀQx + cᵀx + d)|` =
  6.9e-18 (n=6) and 1.1e-16 (n=20) over 40 random points.
- It is bound-neutral with `DISCOPT_PSD_QFORM=0`: node count on the n=5 QP is 89 before and
  89 after.
- It is nonetheless **bound-changing in general** — `_expr_to_polynomial` also feeds
  `edge_concave.py:121` and `milp_relaxation.py:1598` — so per CLAUDE.md §5 it needs a
  corpus-wide differential panel (cert-clean + net-positive) before it ships, and
  graduating `DISCOPT_PSD_QFORM` to default-ON needs the same.

jkitchin:
## Summary

A convex QP written with discopt's **vectorized / indexed-summation modeling API** was certified convex by *no* route, so it fell through the pure-continuous guard at `solver.py:7062` (`_pure_continuous_force_spatial = True`) into spatial McCormick B&B, where it does not converge. Three independent, pre-existing defects stacked up; all three are fixed here.

1. **`_expr_to_polynomial` had no `SumOverExpression` branch** — the node `dm.sum(f, over=...)` builds — so `extract_quadratic` abstained on every indexed-summation body, while the byte-identical body written with Python's builtin `sum` extracted fine.
2. **`interval_ad` raised `TypeError`** on a vector-valued `Constant`/`Parameter` leaf. `certify_convex` catches only `ValueError`, so it escaped and was swallowed by the caller's broad `except Exception` — convexity-unknown *by accident* rather than by decision. Now raises `ValueError`, the walker's established abstain signal (the same refusal array `Variable`s already get). The conversion itself is unchanged, so no leaf that walked before stops walking and none that didn't now does.
3. **The proof was already computed and discarded.** `classify_problem` returns `QP`/`MIQP` only when the Rust structure detector reports *every* constraint linear **and** the objective quadratic — so on that branch the objective's Hessian is a *constant* matrix, and an exact symmetric-eigenvalue test on it is a rigorous, box-independent global convexity proof. New `certificate.certify_quadratic_objective_convex` takes that verdict from `extract_qp_data` (which handles the vectorized API) instead of asking the scalar interval walker (which cannot parse it).

`docs/dev/certification-gap-plan.md` gets the §8 build-results section and status-table row.

Closes #936

## Correctness

The new route is consulted by `classify_model` **only after** the syntactic walker and the interval-Hessian certificate have both left the objective unproven, so it can only ever upgrade an abstention — it can never loosen a verdict. It returns `False` (abstain) on every failure path.

Rigour rests on `classify_problem`'s QP/MIQP branch: exactly-quadratic objective over a polyhedron ⟹ constant Hessian ⟹ no enclosure, no box dependence, proof valid on every sub-box. This is the identical argument `_certify_quadratic_psd` already makes; only the coefficient source differs. `extract_qp_data` returns *minimize-form* data (it negates for `MAXIMIZE`), so the PSD test answers the caller's actual question in both senses without a further sign flip. Slack padding is a zero block, and a principal submatrix of a PSD matrix is PSD — padding can only make the test harder to pass, never wrongly pass it.

Exactness was **verified, not assumed**, per extractor route (the open question in the issue's second comment): `max |body − (½xᵀHx + cᵀx + d)|` over 40 random box points ≤ **1.8e-15** for the algebraic, repr, autodiff *and* dispatched-ladder routes. The `SumOverExpression` extraction reconstructs the body to 3.6e-15 (n=6) / 2.1e-14 (n=20).

Independent cross-validation on the corpus: the route fires on **10 MIQPs** (`alan`, `gbd`, `nvs15`, `st_miqp1`–`5`, `st_test1`, `st_testgr3`) and agrees with a direct `eigvalsh` on all 10, and with both established routes (syntactic walker, interval certificate) on all 10 — zero disagreements.

Refusals are pinned by tests in both directions: an indefinite vectorized QP is refused and still returns the true global −9; `max xᵀx` (convex-maximize = concave-minimize) is refused and returns 18; `Σ exp(x_i)` is convex but not quadratic and is refused; a convex quadratic objective over a **bilinear** constraint classifies QCQP and is refused.

- [x] `pytest -m smoke` → **904 passed, 16 skipped, 2 xpassed, 7962 deselected** (406 s)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **10 passed** (174 s)
- [x] `cargo test -p discopt-core` → **N/A — no Rust touched** (`git diff origin/main...HEAD -- crates/` is empty)
- [x] `ruff check` clean; `ruff format --check` clean **on every file this branch touches** (the two files it reports — `test_log_square_relaxation.py`, `test_mo_augmecon2.py` — are unformatted on `main` already and are not in this diff)

`test_adversarial_recent_fixes.py::test_large_dense_jacobian_no_crash` failed once on its first run, immediately after the 406 s smoke suite in the same container; its only non-soundness assertion is a wall-deadline one. It passed in isolation and on two subsequent full-file runs, and on the pre-fix tree. The route's cost on that model is **91 µs** (its pre-gate refuses before any extraction), so it cannot be implicated. I did not capture that run's assertion message.

## Regression test

`python/tests/test_936_vectorized_qp_convexity.py` — **27 passed**. Fail-before confirmed by checking out `HEAD~1`'s `python/discopt/_jax/`: collection fails on the missing symbol, and the underlying defects reproduce directly — `extract_quadratic` → `None`, `interval_hessian` → `TypeError: only 0-dimensional arrays…`, `classify_model` → `(False, …)`, and the issue's repro solves to `feasible`, 5,101 nodes, 60.4 s.

- [x] Flag: **`DISCOPT_QP_EXACT_CONVEXITY`**, **default-ON with a `=0` opt-out**, graduated by the panel below under the 2026-07-17 single-passing-gate policy. The legacy path is intact and reachable via the opt-out (pinned by a test).
- [x] Differential panel, flag ON vs OFF, arms interleaved per instance
- [x] Graduation panel evidence: **cert-clean AND net-positive** (below)

**Measurement** — 66 in-repo MINLPLib `.nl`, 20 s limit, arms interleaved per instance, **re-run on the tree rebased onto #938** (that PR changed `solver.py` and the Rust LP simplex, so the pre-rebase gate result could not be carried forward):

```
284 soundness/regression checks executed.
CERT-CLEAN: PASS (0 violations)

solved to optimal : OFF 46   ON 46
gap_certified     : OFF 47   ON 47
total wall (s)    : OFF 524.7   ON 516.0
```

On the **46 instances proven optimal in both arms, node count and objective are exactly unchanged** (2803 = 2803 nodes) — the bound-neutral regime's own bar, met on the certifying core. All 5 residual node differences (`clay0303hfsg`, `heatexch_gen2`, `syn05hfsg`, `tanksize`, `tspn05`) are on instances that ended `time_limit`/`feasible` in **both** arms, i.e. wall-quantised rather than flag effects — `bchoco08` alternates 3/7 nodes *within the same arm* across repetitions.

Net-positive is neutral-on-corpus by construction (that corpus contains **0** `SumOverExpression` nodes across 3,769 bodies in 66 instances, and the 10 MIQPs the route fires on were already proven convex) and decisive on the class it targets:

| model | before | after |
|---|---|---|
| `decision_focused_learning` oracle QP | `feasible`, 5,101 nodes, 60.4 s | **optimal, 0 nodes, 0.59 s** |
| `qp_solver` Markowitz n=20 | `feasible` only, 557 nodes, 120 s limit | optimal, 0 nodes, 0.13 s |
| `benchmarks_by_class` QP sweep | 600 s cell timeout | runs to n=100 in <1 s per size |

All 13 models lifted from the three blocked notebooks now solve to `optimal`.

**Scope of the notebook evidence (corrected).** Only `decision_focused_learning` is a clean before/after: it is byte-identical on `main` and on this branch. The other two rows measure the QP bodies, but do not establish a pre-fix baseline *from `main`* — `qp_solver` on `main` is the pre-rewrite version of the notebook, and `benchmarks_by_class` on `main` aborts before reaching any solve, at `import discopt.solvers.lp_highs` (a module removed in #365). The *after* numbers stand as class evidence; what is not established is that those two notebooks were previously timing out **because of** #936. Raised in #issuecomment-5220858494.

**Two readings I published during this work and then retracted on re-measurement** (CLAUDE.md §11):

- An apparent wall regression (`heatexch_gen3` 27.8 → 215.1 s, `bchoco08` 20.6 → 91.0 s) **does not reproduce.** Interleaved A/B ×3 on an idle box: 29.0 ± 4.9 s vs 33.6 ± 2.0 s, and 25.3 ± 4.2 s vs 29.4 ± 2.9 s — both arms overrun the 20 s budget on those two instances irrespective of the flag. My first root-cause hypothesis (that `classify_problem` was expensive there) was falsified by direct timing — **7 ms** on `heatexch_gen3` — before I acted on it.
- `tls2`'s apparent objective drift (10.30 → 5.30) was the ON arm reaching the answer *sooner*, not a soundness event: at a 60 s limit **both** arms certify 5.3 with bound 5.3, and both incumbents are independently verified feasible (constraint-feasible, zero box violation, zero integrality violation), as is `tspn12`'s 262.647. The panel's drift check was wrong and is corrected to apply only where both arms prove optimality.

**Size gate.** `_QP_EXACT_CONVEXITY_MAX_N = 2000` bounds the dense materialisation (2000² float64 = 32 MB) and `eigvalsh` (measured: n=1000 → 0.052 s, n=2000 → 0.331 s ± 0.008, n=4000 → 2.47 s). The pre-gate counts variables **plus constraint rows**, because the extractor appends one slack column per inequality row and pads `Q` to the slacked dimension — gating on the variable count alone would let a narrow model with a huge row count ask for an `(n+m)²` dense array. Abstaining above the cap routes to the sound spatial path, so it is a performance guard, never a soundness one.

### `DISCOPT_PSD_QFORM` graduation — declined on neutrality

The issue's Defect 2 proposed graduating this flag alongside the fix. **It stays default-OFF**, because it is measurably useless on the class it would affect, not because it is unsound.

On `st_e36` the flag runs 11 exact eigenvalue certifications that produce 10 additional **CONCAVE** verdicts and no convex verdict, so nothing is routed onto a convex path. Measured cost +1.7 s (3.6 s → 5.3–5.4 s, interleaved ×3); measured benefit zero. That is the `DISCOPT_CUT_INHERIT` rule — sound ≠ helpful — with the measurement recorded rather than the flag flipped. Graduating it is also unnecessary here: the exact-QP route resolves Defect 2's *symptom* without it (Markowitz n=20 solves at 0 nodes with `PSD_QFORM` at its default OFF). A later graduation would need its own §5 panel; none has been run.

<details>
<summary><b>Retraction (§11):</b> this section previously declined the flag on a cert-clean failure that does not reproduce</summary>

The earlier text read:

> Its own panel fails cert-clean on `st_e36`: `gap_certified` True → False, `optimal`/85 nodes → `time_limit`/0 nodes.
> ```
> PSD_QFORM=0: optimal 3/3, gap_certified 3/3   (9.3 / 11.0 / 12.3 s)
> PSD_QFORM=1: optimal 0/3, gap_certified 0/3   (148 / 171 / 188 s, 0 nodes — the ROOT never finishes)
> ```
> That is the certification regression §5 names as disqualifying, and it is not a timing artifact.

**That measurement does not reproduce and is withdrawn.** An independent reproduction on `54c606b4` with a freshly `cargo build --release`d extension returns `optimal`, 85 nodes, `gap_certified=True`, objective `-245.99999999999997`, bound `-246.0000000017` in *every* configuration tried: interleaved ×3 at the 20 s panel limit, the full 2×2 of `DISCOPT_PSD_QFORM` × `DISCOPT_QP_EXACT_CONVEXITY`, and a 7-instance same-process panel with `st_e36` last. The A/B is non-vacuous per §6 — instrumented `certify_psd` calls are 0× with the flag OFF and 11× with it ON — so the arms genuinely differ and simply agree on the outcome. The instance file is byte-identical in-repo and in the `discopt-minlp-benchmark` corpus.

Probable cause, unconfirmed: a stale prebuilt `_rust*.so` predating #938, whose `primal.rs` rewrite this branch is rebased onto — the CLAUDE.md §8 trap, which the reproducing session hit once itself before adding a positive post-#938 build marker.

No issue is filed, since there is no defect to file. Full detail: #issuecomment-5220999652.
</details>

https://claude.ai/code/session_0128NkHBGMq5KFXJDTDCvgAy

---
_Generated by [Claude Code](https://claude.ai/code/session_0128NkHBGMq5KFXJDTDCvgAy)_

### DIFF
--- a/docs/dev/certification-gap-plan.md
+++ b/docs/dev/certification-gap-plan.md
@@ -138,6 +138,7 @@ experiment may run.
 | Phase 4 T-CSE/V-segments | **CSE unlocked; V-segments/symmetry de-scoped** (§0.1.2) | — | build order fixed by the re-profile above; CSE first (bound-neutral), Q-extraction second |
 | Phase 4 build 1 — CSE/hash-consing | **done — bound-neutral** | (this PR) | Content-addressed interning in `ExprArena` (`expr.rs`), wired into the `.nl` parser and Python `convert_expr`. DAG node count ↓ **68.9% nvs17, 64.8% clay0303hfsg, 34.2% ex1252, 34.0% casctanks, 0.0% gear4** (panel total −48.4%); gear4 0% confirms the re-profile prediction. Cert-neutrality **NEUTRAL** (42 certifying instances, node_count exactly unchanged, objective to tol). See §8 "Phase 4 CSE — build results" |
 | Phase 4 build 3 — Q-matrix extraction | **done — (A) bound-neutral, (B) flagged bound-changing** | (this PR) | Exact `extract_quadratic(expr,n,model)` in `quadratic_form.py` (exact-or-abstain, validated 1e-12 on 250 pts/inst + 14 non-quadratic rejections). Consumer: PSD-on-Q convexity certificate wired into `certify_convex` behind `DISCOPT_PSD_QFORM` (**default-OFF**); sound tightening, never mis-certifies (30 indefinite-Q seeds), strict refinement of the rigorous path. Flag-off is byte-identical → cert-neutral. See §8 "Phase 4 Q-extraction — build results" |
+| Vectorized-API convexity certification | **done — bound-changing, panel-graduated default-ON** | #936 | No model written with the vectorized / indexed-summation API could be certified convex by ANY route, so positive definite QPs fell through the `solver.py:7062` guard into spatial B&B and did not converge. Fix: take the exact eigenvalue PSD verdict on the Hessian `extract_qp_data` already produces when `classify_problem` proves the model a QP/MIQP (constant Hessian ⇒ box-independent global proof), behind `DISCOPT_QP_EXACT_CONVEXITY` (default-ON, `=0` opt-out); plus the missing `SumOverExpression` branch in `_expr_to_polynomial` and a deliberate `ValueError` abstention for array-valued interval-AD leaves. Panel **cert-clean, 0/284 violations**; 46/46 proven-optimal instances byte-identical in nodes AND objective; route fires on 10 corpus MIQPs and agrees with `eigvalsh` and both established routes 10/10. Issue repro `feasible`/5,101 nodes/60.4 s → **optimal/0 nodes/0.59 s**. See §8 "Vectorized-API convexity certification" |
 | Phase 4 items 2 & 4 (V-segments, symmetry) | **locked / de-scoped** (§0.1.2) | — | V-segments de-prioritized (0 defvars in text `.nl`); symmetry DO-NOT-BUILD (0 orbits) per the re-profile. RLT/edge-concave rewire onto `extract_quadratic` scoped as bound-neutral follow-on |
 | STRUCT-1 verification (task #82) | **done — verified; commutative-normalization KILLED** | — | Independent re-measurement (2026-07-10) on 61 in-repo + 30 snapshot-draw instances: CSE live, median 25.0% node reduction (76/91 ≥10%); 0 defined vars / 0 V segments in all 1,558 text `.nl` (re-confirmed with the correct header field). NEW: commutative-normalized interning would fuse median 0.00% / max 0.5% extra — **do not build**. See §8 "Independent verification" note under CSE build results |
 | Phase 5 | **locked** (§0.1.2) | — | requires post-Phase-1 re-profile |
@@ -1212,6 +1213,90 @@ collect_edge_concave_quadratics`, `rlt_cuts.py`), so rerouting them through
 coefficients, same verdict) and at worst risk a subtle drift; it is not shipped in
 this PR. The value there is a *faster path to the same relaxation*, which must be
 proven exactly bound-neutral (node_count/objective unchanged) — a separate task.
+
+### Vectorized-API convexity certification (#936, 2026-08-07)
+
+The Q-extraction work above certifies convexity from an exact Hessian, but only
+for bodies `quadratic_form.extract_quadratic` can recognise — and that walker,
+like the scalar interval-Hessian certificate beside it, could not parse the
+**vectorized / indexed-summation modeling API** at all. Net effect: *no model
+written in that API could be certified convex by any route*, so a positive
+definite QP fell through the pure-continuous guard (`solver.py:7062`,
+`_pure_continuous_force_spatial = True`) into spatial McCormick B&B, where it
+does not converge. Three stacked defects, all pre-existing:
+
+1. `milp_relaxation._expr_to_polynomial` walked `SumExpression` but not
+   `SumOverExpression` — the node `dm.sum(f, over=...)` builds — so
+   `extract_quadratic` abstained on every indexed-summation body while the
+   byte-identical body written with Python's builtin `sum` extracted fine.
+2. `DISCOPT_PSD_QFORM` default-OFF leaves the interval Hessian + Gershgorin
+   row-sum bound, which cannot certify a dense covariance Hessian.
+3. `convexity/interval_ad` raised `TypeError` on a vector-valued
+   `Constant`/`Parameter` leaf; `certify_convex` catches only `ValueError`, so
+   it escaped and was swallowed by the caller's broad `except Exception` —
+   convexity-unknown *by accident*.
+
+**The fix: consult the proof discopt already computes.**
+`problem_classifier.classify_problem` returns `QP`/`MIQP` only when the Rust
+structure detector reports every constraint linear AND the objective quadratic,
+so on that branch the objective's Hessian is a *constant* matrix and an exact
+symmetric-eigenvalue test on it is a rigorous, box-independent global convexity
+proof — the same argument `_certify_quadratic_psd` makes, fed from
+`extract_qp_data` (which handles the vectorized API) instead of
+`extract_quadratic` (which does not). New
+`certificate.certify_quadratic_objective_convex`, consulted by `classify_model`
+*after* the syntactic walker and the interval certificate have both left the
+objective unproven, so it can only ever upgrade an abstention. Behind
+`DISCOPT_QP_EXACT_CONVEXITY` (**default-ON**, `=0` opt-out) — graduated by the
+panel below under the 2026-07-17 single-passing-gate policy.
+
+*Exactness verified per extractor route*, not assumed (the issue's open
+question): `max |body − (½xᵀHx + cᵀx + d)|` over 40 random box points is
+≤ 1.8e-15 for the algebraic, repr, autodiff **and** dispatched-ladder routes.
+The `SumOverExpression` extraction reconstructs the body to 3.6e-15 (n=6) /
+2.1e-14 (n=20).
+
+*Differential panel (66 in-repo MINLPLib `.nl`, 20 s limit, arms interleaved
+per instance, 284 checks):* **CERT-CLEAN, 0 violations.** On the 46 instances
+proven optimal in both arms the node count and objective are **exactly
+unchanged** (2803 = 2803 nodes) — the bound-neutral regime's own bar, met on
+the certifying core. `gap_certified` 47 → 47, solved-to-optimal 46 → 46.
+The route *does* fire on the corpus (10 MIQPs: alan, gbd, nvs15, st_miqp1–5,
+st_test1, st_testgr3) and agrees with a direct `eigvalsh` on all 10, and with
+both established routes (syntactic walker, interval certificate) on all 10 —
+zero disagreements. The remaining 20 instances hit the time limit in ≥1 arm,
+so their node counts are wall-quantised, not flag effects: `bchoco08` alternates
+3/7 nodes *within the same arm* across repetitions. Two apparent wall outliers
+(`heatexch_gen3` 27.8 → 215.1 s, `bchoco08` 20.6 → 91.0 s) **did not reproduce**;
+interleaved A/B ×3 on an idle box gives heatexch_gen3 29.0 ± 4.9 s (OFF) vs
+33.6 ± 2.0 s (ON) and bchoco08 25.3 ± 4.2 s vs 29.4 ± 2.9 s — both arms overrun
+the 20 s budget on these two instances irrespective of the flag (a pre-existing
+overrun; heatexch_gen3 is the instance §-above already names for a 12 s classify
+under a 15 s budget). The route's own cost is bounded: 7 ms on heatexch_gen3,
+91 µs on the 1210-variable adversarial model (its pre-gate refuses first).
+`tls2`'s apparent objective "drift" (10.30 → 5.30) is the ON arm reaching the
+answer sooner: at a 60 s limit **both** arms certify 5.3 with bound 5.3, and
+both incumbents are independently verified feasible (constraint-feasible, zero
+box violation, zero integrality violation), as is `tspn12`'s 262.647.
+
+*Net-positive:* neutral on the `.nl` corpus by construction (that corpus contains
+**0** `SumOverExpression` nodes across 3,769 objective/constraint bodies in 66
+instances, and the 10 MIQPs the route fires on were already proven convex), and
+decisive on the class it targets. The three `docs/notebooks/` pages the issue
+reports as timing out all solve: the `decision_focused_learning` oracle QP goes
+from `feasible`/5,101 nodes/60.4 s (measured on this tree at `HEAD~1`; the issue
+reports 14,115 nodes) to **optimal/0 nodes/0.59 s**; `qp_solver`'s Markowitz
+n=20 from `feasible` only/557 nodes/120 s limit to optimal/0 nodes/0.13 s;
+`benchmarks_by_class`'s QP sweep runs to n=100 in under 1 s per size.
+
+*Size gate.* `_QP_EXACT_CONVEXITY_MAX_N = 2000` bounds both the dense
+materialisation (2000² float64 = 32 MB) and `eigvalsh` (measured: n=1000 →
+0.052 s, n=2000 → 0.331 s ± 0.008, n=4000 → 2.47 s). The pre-gate counts
+variables **plus constraint rows**, because the extractor appends one slack
+column per inequality row and pads `Q` to the slacked dimension — gating on the
+variable count alone would let a narrow model with a huge row count ask for an
+`(n+m)²` dense array. Abstaining above the cap routes to the sound spatial path,
+so the cap is a performance guard, never a soundness one.
 
 ### Phase D re-profile (2026-07-05) — the POUNCE subsolver is the #1 wall lever
 
--- a/docs/dev/certification-gap-plan.md
+++ b/docs/dev/certification-gap-plan.md
@@ -1287,25 +1287,52 @@ box violation, zero integrality violation), as is `tspn12`'s 262.647.
 *Net-positive:* neutral on the `.nl` corpus by construction (that corpus contains
 **0** `SumOverExpression` nodes across 3,769 objective/constraint bodies in 66
 instances, and the 10 MIQPs the route fires on were already proven convex), and
-decisive on the class it targets. The three `docs/notebooks/` pages the issue
-reports as timing out all solve: the `decision_focused_learning` oracle QP goes
+decisive on the class it targets: the `decision_focused_learning` oracle QP goes
 from `feasible`/5,101 nodes/60.4 s (measured on this tree at `HEAD~1`; the issue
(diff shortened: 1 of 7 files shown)


jkitchin:
Context from the notebook re-run that surfaced #936 (commit `21c745ea` on
`refactor/remove-jax-from-solve-path` / #922 — all 62 notebooks executed in place,
57 pass). Four things this PR's verification section needs.

### 1. `benchmarks_by_class` does not fail on #936 from this PR's base

On `main` that notebook still imports a module that no longer exists:

```
$ git show origin/main:docs/notebooks/benchmarks_by_class.ipynb | grep -c lp_highs
1
$ python -c "import importlib.util as u; print(u.find_spec('discopt.solvers.lp_highs'))"
None
```

`discopt.solvers.lp_highs` was removed in #365. The notebook dies at that import,
long before the QP sweep #936 blocks, so re-executing it on `claude/issue-936-hfdeye`
will fail for a reason this PR has nothing to do with. The import fix is on #922's
branch only (`benchmarks_by_class` differs from `main` by 7/93 lines there).

### 2. `qp_solver` on `main` is the pre-rewrite version

It differs from #922's copy by 12/177 lines — I rewrote it in the same pass. So of
the three notebooks named as unblocked, only **`decision_focused_learning`** is
byte-identical between `main` and #922 and therefore verifiable end-to-end from this
PR's base today.

This doesn't undercut the fix — "13 models lifted from the three blocked notebooks
now solve to `optimal`" is a model-level result and I don't doubt it. It's the
notebook-level claim that can't be closed from here for two of the three.

### 3. `tutorial_dae` is out of scope — please don't expect this PR to unblock it

It was the fourth timing-out notebook and a candidate #936 instance in the issue's
first comment, but it is genuinely nonconvex (bilinear in two decision variables:
`dae.set_ode(lambda t, states, alg, ctrl: {"x": -k * states["x"]})` with `k` a
decision variable). A/B against an `origin/main` worktree at a 120 s limit:

| arm | status | objective | nodes |
|---|---|---|---|
| `origin/main` | feasible | 0.0015582024997472634 | 18413 |
| #922 HEAD | feasible | 0.0015582024997472634 | 17739 |

Same objective to the last digit. Not a regression, and not a convexity-certification
miss.

### 4. After merge, the three notebooks need re-executing and committing

`nbconvert --execute --inplace` does **not** write the file when a cell raises, so all
three currently carry *stale committed outputs* from before they started failing —
they look fine in the repo. `docs/_config.yml` sets `execute_notebooks: "off"`, so CI
never executes notebooks and will not catch this either. `21c745ea` is the refreshed
baseline for the other 57; these three still need a pass once the fix lands (and
`benchmarks_by_class` additionally needs the item-1 import fix, whether from #922 or
cherry-picked).

### 5. No interaction with #922

Zero file overlap (`comm -12` on the two file lists is empty). The one semantic
coupling worth checking is clean: this PR's new route consumes
`problem_classifier.extract_qp_data`, and #922 does touch `problem_classifier.py` —
but that diff is **comment-only** (`qp_ipm_solve` → `solve_qp_kkt` in three
docstrings). No behavior change on the function this PR depends on.

jkitchin:
I was asked to fix the `st_e36` blow-up this PR offers to file. **I could not
reproduce it**, under any configuration I tested, so there is nothing to fix — and
the DECLINE section's stated evidence should not be cited as established until
someone re-runs the original panel.

Reporting the falsification rather than shipping a fix for a defect I cannot
observe (CLAUDE.md §4 — the measurement wins).

### What the PR reports

```
PSD_QFORM=0: optimal 3/3, gap_certified 3/3   (9.3 / 11.0 / 12.3 s)
PSD_QFORM=1: optimal 0/3, gap_certified 0/3   (148 / 171 / 188 s, 0 nodes — the ROOT never finishes)
```

### What I measure on this branch

`54c606b4`, 20 s limit, arms interleaved, `st_e36.nl` from the in-repo corpus
(byte-identical to the `discopt-minlp-benchmark` copy — `cmp` clean):

| rep | `PSD_QFORM=0` | `PSD_QFORM=1` |
|---|---|---|
| 1 | optimal, 85 nodes, cert=True, 3.6 s | optimal, 85 nodes, cert=True, 5.4 s |
| 2 | optimal, 85 nodes, cert=True, 3.6 s | optimal, 85 nodes, cert=True, 5.4 s |
| 3 | optimal, 85 nodes, cert=True, 3.6 s | optimal, 85 nodes, cert=True, 5.3 s |

Objective `-245.99999999999997`, bound `-246.00000000170442`, identical in every run
(oracle: `-246`). The full 2×2 with `DISCOPT_QP_EXACT_CONVEXITY` ∈ {0,1} is also
identical — 85 nodes, `optimal`, `gap_certified=True` in all four cells. A
panel-style run of 7 corpus instances in **one process** with `st_e36` last (to catch
cross-instance state) reproduces the same 85 nodes in both arms.

Note the OFF arms agree: your 3/3 `optimal` and my 85 nodes. Only the ON arm differs.

### The probe fired — this is not a vacuous A/B

`PSD_QFORM=1` demonstrably changes behavior on this instance, so I am not comparing
two identical paths:

```
PSD_QFORM=0  CALLS={'psd_enabled': 11, 'certify_psd':  0, 'certify_convex': 11}
             VERDICTS=[(None, 10), (Curvature.CONCAVE, 1)]
PSD_QFORM=1  CALLS={'psd_enabled': 11, 'certify_psd': 11, 'certify_convex': 11}
             VERDICTS=[(None, 12), (Curvature.CONCAVE, 10)]
```

The exact PSD route runs 11 times and yields 10 additional **CONCAVE** verdicts. It
never certifies this objective convex, which is why it cannot route `st_e36` onto a
convex path and cannot make the root unsolvable. It costs a reproducible +1.7 s
(3.6 → 5.4 s) for work that changes no verdict the solver acts on.

### A build hypothesis, offered without blame — I hit it myself

My first two attempts *also* showed no blow-up, but they were invalid, and for a
reason worth flagging: I copied a prebuilt `_rust*.so` into the worktree, and that
`.so` predated **#938**, which this branch is rebased onto and which rewrites
`audit_feasibility` in `crates/discopt-core/src/lp/simplex/primal.rs` (+138/−9). So I
was running this branch's Python against pre-#938 Rust — exactly the CLAUDE.md §8
trap.

The table above is from a **`cargo build --release` in the worktree**, verified live
by a positive marker unique to post-#938 Rust: `lp_simplex.solve_lp` certifies
`ub ∈ {1e15, 1e16, 1e18, 9.999e19}` as `optimal` at `−2u` (pre-#938 those return
`error`).

Since the panel was re-run "on the tree rebased onto #938", a stale extension there
would produce exactly this kind of unreproducible outlier. That is a hypothesis, not
a claim — I have no visibility into how that run was built, and the panel harness
isn't committed to the branch, so I reconstructed the protocol rather than running
the identical script.

### What I am *not* saying

I am **not** proposing that `DISCOPT_PSD_QFORM` be graduated. That needs its own §5
panel, which I have not run, and the flag's measured behavior here — 11
eigen-decompositions, +1.7 s, zero change in outcome — is if anything a `CUT_INHERIT`
argument for leaving it default-OFF (sound ≠ helpful). The **decision** to decline
looks fine; it is the specific `st_e36` cert-regression *evidence* that does not
reproduce.

Suggested disposition: re-run the PSD_QFORM panel with a freshly built extension. If
`st_e36` still blows up there, send me the build provenance and I will chase it; if
it doesn't, the DECLINE section should be reworded to rest on neutrality rather than
on a certification regression, and no issue needs filing.

Environment: darwin/arm64, load avg ~4 during the runs (higher load would slow the ON
arm, not speed it up, so it cannot mask a 150 s root).

Issue #941, opened 2026-08-07
correctness(P0): negative index on a shaped Variable resolves to the wrong flat slot, producing a certified false bound


jkitchin:
## Summary

A **negative** index on a shaped `Variable` (`v[-1]`) resolves to the **wrong flat
slot** in the structure-detection and relaxation-construction layers. The wrong
slot silently names a *different variable*, so the McCormick relaxation is built
for a bilinear pair that does not exist in the model. The result is an **invalid
(too tight) relaxation that cuts off the true optimum and is then reported as
`status="optimal"` with `gap_certified=True`** — a certified false bound, the
worst class of error under CLAUDE.md §1.

Present on `main`. Not caused by anything in flight; found while A/B-testing the
static-index tape lowering on branch `refactor/remove-jax-from-solve-path`.

## Reproducer

Two mathematically identical models. The only difference is spelling the last
element of a shape-(4,) variable as `v[-1]` instead of `v[3]`.

```python
import discopt.modeling as dm

def build(last):
    m = dm.Model("geom")
    s = m.continuous("s", lb=0.2, ub=3.0)
    v = m.continuous("v", shape=(4,), lb=0.2, ub=3.0)
    m.minimize(sum(dm.exp(v[i]) / (1.0 + v[(i + 1) % 4]) for i in range(4)) + dm.log(s))
    m.subject_to(sum(v[i] ** 2 for i in range(4)) + s <= 9.0)
    m.subject_to(v[0] * v[last] >= 0.5)     # last = -1 or 3
    return m

for last in (-1, 3):
    r = build(last).solve(time_limit=45)
    print(last, r.status, r.node_count, r.objective, r.bound, r.gap_certified)
```

Observed:

| spelling | nodes | status | objective | bound | `gap_certified` |
|---|---|---|---|---|---|
| `v[0]*v[-1]` | 3 | `optimal` | 3.4853698662 | 3.4853698662 | **`True`** |
| `v[0]*v[3]`  | 9815 | `feasible` | 2.9618839111 | 2.9572881926 | `False` |

The `v[-1]` arm certifies a dual bound of **3.4854** on a minimization for which
**2.9619 is feasible**. Both returned points were re-evaluated in plain numpy,
independent of either evaluator: both satisfy `sum(v^2)+s <= 9` and
`v0*v3 >= 0.5` to within the 1e-6 tolerance (slacks 6.77 / -6.79e-08 and
[...]

PR #947, opened 2026-08-08
fix(correctness): #944 curvature walker was blind to array constants

### LINKED ISSUE #944: test_dae.py::test_minimum_energy never terminates: 100-var convex QP spends 800s+ in pure-Python FBBT
## Summary

`python/tests/test_dae.py::TestOptimalControl::test_minimum_energy` does not
terminate. It is `@pytest.mark.slow`, so CI's default marker set
(`-m 'not slow and ...'`) deselects it and nothing has been watching it.

This is **pre-existing on `main`**, not introduced by any branch in flight. It was
found while running the DAE suite with `-m ""` for the tape array-lowering work
(#75), and the baseline was verified two ways before reporting:

* stash the working tree back to `fcf7253e` on
  `refactor/remove-jax-from-solve-path` → `Failed: Timeout (>600.0s)`, hanging in
  `pounce/_nlp_batch.py:130`;
* a detached worktree on `origin/main` (`5d2550f2`), with the load gate asserted
  (`discopt.__file__` inside the worktree) → `Failed: Timeout (>800.0s)`, hanging
  in `python/discopt/solver.py:2518`.

Both arms hang; the two stack locations are just wherever the interpreter was when
the alarm fired.

## The model

```python
cs  = ContinuousSet("t", bounds=(0, 2), nfe=20, ncp=3)
dae = DAEBuilder(m, cs)
dae.add_state("x", initial=1.0, bounds=(-5, 5))
dae.add_control("u", bounds=(-3, 3))
dae.set_ode(lambda t, s, a, c: {"x": -s["x"] + c["u"]})
m.minimize(dae.integral(lambda t, s, a, c: c["u"] ** 2) + 10.0 * x_var[-1, -1] ** 2)
result = m.solve()          # <- no time_limit
```

Measured size: **100 variables, 79 constraint rows** (2 `Constraint` objects, both
array-valued). It is a convex QP with box bounds — there is no combinatorial
content at all, so this should be one NLP solve, not a search.

## Where the time goes

`solver.py:2518` is the inner loop of the linear-row FBBT pass:

```python
for j in range(m):                     # 79
    if not is_linear[j]: ...
    for i in range(n):                 # 100
        ...
        for k in range(n):             # 100
```

`m * n * n = 790_000` pure-Python iterations per round, `max_rounds` rounds, **per
B&B node**. With no `time_limit` on the solve, the test has no bound on its own
runtime. So the headline may simply be "too slow", not "wedged" — but 800 s on a
100-variable convex QP is a defect either way, and the test as written can never
report anything but a timeout.

## What to check

1. Whether the root solve of a 100-var convex QP is being handed to B&B at all —
   if it is, that is the primary bug and the FBBT cost is a symptom.
2. Vectorizing the `solver.py:2500-2530` FBBT triple loop, which is the measured
   hot spot on the `main` arm. It is already row-masked by `is_linear`; the inner
   two loops are a bound-propagation update expressible in numpy.
3. Whether `test_minimum_energy` should carry a `time_limit=` like the other
   long-running tests, so a regression reports as a *failed assertion* rather than
   as a pytest timeout.

Not scoped to #75 — that work is the tape/JAX evaluator boundary, and this
reproduces with JAX on the path.

jkitchin:
## Summary

`test_dae.py::TestOptimalControl::test_minimum_energy` — 100 variables, 79 linear
rows, no integer variables, a convex QP — ran past an 800 s pytest timeout, and the
same construct in `docs/notebooks/tutorial_dae.ipynb` (cell 11) hit an nbclient
`CellTimeoutError` after 600 s. Both were routed to spatial McCormick B&B and
stalled there at ~9k nodes without certifying.

**Root cause.** The constant-scaling rules in the curvature walker keyed on
`_is_scalar_const`, which requires `ndim == 0`. Every constant the *vectorized*
modeling API produces is an array, so the collocation row

```
(x @ A) - h * (-x[:, 1:] + u[:, None])          # h.shape == (nfe, 1)
```

hit `Curvature.UNKNOWN` at the `h * (...)` node — an affine body the walker could
not see was affine. `classify_model` then reported the model nonconvex,
`solve_model`'s convex-NLP fast path declined it, and the sound-but-hopeless
spatial path took over. This is a *class* defect, not a DAE one: element widths,
collocation weights, quadrature weights — any array coefficient at all — were
invisible to the rule.

```
before   status=feasible  nodes=8861  gap_certified=False   >800 s (timeout)
after    status=optimal   nodes=0     gap_certified=True      3.7 s
```

**Second defect, same path.** `convert_index_spec` in the Rust bridge extracted
every scalar subscript as `usize`, so `x[-1]` raised `OverflowError: can't convert
negative int to unsigned` and aborted `model_to_repr` outright. `classify_problem`
catches that and falls back to NLP/MINLP, so every model spelling an index
negatively silently lost its LP/QP/MILP classification — and the objective of the
model above contains `x_var[-1, -1]`. Negative subscripts are now normalized
against the base expression's shape, per axis. This is the one site #941 did not
reach.

`test_minimum_energy` now carries `time_limit=120`, so a regression reports as a
failed assertion rather than as a pytest timeout (issue item 3).

Closes #944

## Correctness

The change adds curvature verdicts the walker previously left UNKNOWN, and each
addition is a proof, not a heuristic. For an elementwise (broadcast) product
`c * f` with `c` a **finite** constant, entry k is `c_i * f_j` — a scalar constant
times a scalar expression:

* `f` affine ⇒ every entry affine, for **any** finite `c` (mixed signs included);
* `f` convex and every `c` entry ≥ 0 ⇒ every entry convex;
* `f` convex and every `c` entry ≤ 0 ⇒ every entry concave;
* `f` non-affine and `c` mixed in sign ⇒ **refuse** (UNKNOWN).

Zero entries never spoil a verdict (`0 * f` is affine, hence both convex and
concave). Division carries the same rule plus the requirement that no entry sits at
or within 1e-30 of zero — and a constant divisor now *returns* from that branch
rather than falling through to the reciprocal rule, whose strict-sign test reads a
zero divisor as strictly signed.

### Non-finite constants (corrected in `e5be547` — the first version of this PR was wrong here)

An earlier revision of this description claimed *"Two directions got stricter …
both now refuse."* That was **false**, and for array constants the first commit was
**looser than `main`**, not stricter. @jkitchin caught it in review. Measured A/B,
`main`'s `rules.py` against the branch's in the same interpreter, each arm
asserting its own marker:

```
case                     main       first commit        now (e5be547)
inf-array  * affine      UNKNOWN    AFFINE  <- looser   UNKNOWN
nan-array  * affine      UNKNOWN    AFFINE  <- looser   UNKNOWN
scalar-inf * affine      AFFINE     AFFINE              UNKNOWN
scalar-nan * affine      AFFINE     AFFINE              UNKNOWN
inf-array  * convex      UNKNOWN    UNKNOWN             UNKNOWN
nan-array  * convex      UNKNOWN    UNKNOWN             UNKNOWN
scalar-inf * convex      CONVEX     UNKNOWN             UNKNOWN
scalar-nan * convex      CONCAVE    UNKNOWN             UNKNOWN
```

Cause: `_const_scale_sign` returned `None` for two unrelated reasons — *"not a
finite constant"* and *"no uniform sign"* — and the caller saw only one `None`, so
it treated the first as the second and fell into the mixed-sign branch. Now split
into `_finite_const_value` (screens the operand) and `_const_scale_sign` (answers
only the sign question). Every arm refuses, matching `_classify_division`, which
always did.

`main`'s `scalar-nan * convex -> CONCAVE` is a genuine latent bug this PR does fix:
`nan > 0` is False, so NaN fell to the negative branch and silently *negated* the
operand's curvature.

The stricter direction (`scalar-inf * affine`, AFFINE → UNKNOWN) is a *new* loss of
verdict relative to `main`, so it got its own blast-radius check — see the corpus
sweep below.

### Index change

An out-of-range subscript is refused rather than clamped or wrapped — a
wrong-but-in-range flat slot names a *different* variable and nothing downstream
errors on it, which is exactly the #941 failure mode. `x[True]` is refused too
(`bool` is an `int` subclass, so the old `usize` extraction named slot 1, while
numpy reads a bool as a mask); that is a deliberate behaviour change and is pinned
by a test.

Downstream effect is that a convex model reaches `_solve_continuous(...,
certify_convex=True)` instead of spatial B&B. That path's certificate rests on
convexity, which is what the walker now proves for this class; the verdict is
box-independent (a constant coefficient array), so it holds on every node box.

## Tests run

- [x] `pytest -m smoke` → **909 passed, 16 skipped, 2 xpassed**
- [x] `cargo test -p discopt-core` → **583 + 4 + 1 passed, 0 failed**
- [x] `ruff check` / `ruff format --check` clean on every file this PR touches

**Corpus sweep** — `classify_model(m, use_certificate=True)` over all 66 in-repo
`.nl` instances, `main` vs this branch, each arm asserting its own marker, comparing
the model verdict *and* the per-constraint mask:

```
INSTANCES=66   COMPARISONS=132
base convex: 24    fix convex: 24
base ERR:     0    fix ERR:     0
MODEL_VERDICT_FLIPS   = 0
CONSTRAINT_MASK_FLIPS = 0
```

Nothing moves — expected, since the `.nl` importer emits 0-d constants and never the
array coefficients this rule newly understands, but it is what bounds the risk of
both the widened surface and the newly stricter non-finite refusal.

Also run:

- `pytest python/tests/test_dae.py -m ""` (slow included) → **75 passed in 257 s**
  (one test in this file previously blew an 800 s timeout on its own)
- `pytest -k "convex or 936 or 941 or lattice"` → **907 passed, 2 failed**. Both
  failures are `[ipopt]`-parametrized and fail on `ModuleNotFoundError: cyipopt` in
  this container, not on a verdict.
- `docs/notebooks/tutorial_dae.ipynb`, executed cell-by-cell in a fresh kernel with
  `interrupt_on_timeout=True` → **9/9 executable cells pass**, with cell 11 — the
  cell the issue reported as a 600 s timeout — now at **1.1 s**. Cell 9 is excluded;
  see below.

One retraction (CLAUDE.md §11): `test_large_dense_jacobian_no_crash` first failed
its wall-clock assertion at 57 s against a 48 s budget. That was load I created
myself by running a notebook concurrently (§9). Isolated on an idle machine it
passes twice, at **30.2 s** and **27.0 s**.

## Regression test

`python/tests/test_944_array_constant_curvature.py` — 16 tests.

Fail-before/pass-after was measured, not assumed: the pre-fix versions of both
changed files were checked out, the extension rebuilt, and the file re-run with only
its load gate stripped (the gate is *designed* to fail on a baseline build, so
leaving it in would have made the signal vacuous).

```
baseline (main, marker asserted absent)   13 failed, 3 passed
with the fix                              16 passed
```

The 3 that pass on both arms are the conservative-direction guards — mixed-sign
refusal, zero/near-zero divisor refusal, and 0-d scalar behaviour unchanged — so
they *should* be green on both.

`test_solves_to_certified_optimality_without_branching` is deliberately **not**
`@pytest.mark.slow`. Every CI lane deselects `slow`
(`-m "not slow and not correctness and …"` / `-m "correctness and not slow and …"`),
and that is the one test asserting the *routing* rather than the classification — so
with the mark, CI would have proven the classifier changed but not that the model
stops branching. Verified selected: 15/16 collected under the Python-fast marker
expression, 1 deselected. The `time_limit=120` bounds a regression to a failed
assertion instead of a hung lane.

Oracles are external to the solver, not reimplementations:

* **Index resolution** — numpy's own indexing of an `np.arange(size).reshape(shape)`
  probe, over every in-range index of shapes (4,), (2,3) and (2,3,4) in both
  spellings, evaluated through the NLP evaluator *and* pushed through
  `model_to_repr`. 136 resolutions, with the count asserted so the loop cannot pass
  vacuously (§6).
* **End-to-end solve** — the analytic finite-horizon LQR cost, computed outside the
  solver from the Riccati solution of `dS/dt = S² + 2S`, `S(T) = 10`:

  | | analytic J\* | solver |
  |---|---|---|
  | test, T = 2 | 0.03099925 | 0.03102066 |
  | notebook, T = 5 | 7.5741e-05 | 7.5997e-05 |

  The residual is collocation discretization error (20 elements, 3 Radau points).
  The control bound is inactive along both trajectories, so the box-constrained
  optimum coincides with the unconstrained analytic one.

N/A as a bound change — no relaxation, cut, or reduction is modified. The effect is
a *routing* change: a model the classifier can now prove convex takes the existing
convex-NLP path instead of spatial B&B. The spatial path itself is byte-identical,
and a model whose convexity is still unproven still goes there.

---

### Not fixed here, and left open

Separate defects surfaced while diagnosing this. None blocks this PR and none is
caused by it.

**Notebook cell 9** estimates a rate constant `k` as a decision variable, making the
collocation rows bilinear in `k·x` — genuinely nonconvex, correctly routed to
spatial B&B, and slow (uncertified at a 45 s limit here). Interleaved A/B with and
without this change:

```
with fix     is_convex=False  mask=[False, True]  obj=0.0015582024997472664  nodes=1331
without fix  is_convex=False  mask=[False, True]  obj=0.0015582024997472664  nodes=1283
```

Identical verdict and identical objective to every digit; the node-count difference
is throughput under a wall-clock limit, not a search-order difference. That cell is
why the notebook sweep above skips index 9.

**`x[:, None]` (numpy newaxis)** is still unsupported by the Rust bridge — `None`
fails `usize` extraction identically before and after this change — so DAE models
with a control keep degrading to convexity-*unknown* classification even though
they now solve correctly via the walker. Supporting it means adding an
`IndexElem::NewAxis` variant and its broadcasting semantics through ~8 consumers in
`discopt-core`; that is an independent feature with real silent-wrong-answer risk,
not something to fold in here.

**`linear_context.py:178`** carries its own `_is_scalar_const` with the identical
`ndim == 0` gate (found by @jkitchin in review). Its failure direction is
abstain-to-box — a looser sign enclosure, never a wrong one — so it cannot produce a
false verdict the way the walker could. Recorded as the other place this class
defect lives.

**Issue item 2 (vectorizing the `solver.py:2500-2530` FBBT triple loop)** was not
attempted. The `i` loop mutates `lb`/`ub` that the inner `k` loop then reads, so it
is a Gauss–Seidel sweep, not a Jacobi one: any vectorization changes summation order
and the bounds themselves, which puts it in CLAUDE.md §5's bound-*changing* regime —
a corpus-wide differential panel, not a refactor. It is also no longer on this
model's path at 0 nodes.

### DIFF
--- a/crates/discopt-python/src/expr_bindings.rs
+++ b/crates/discopt-python/src/expr_bindings.rs
@@ -1498,7 +1498,12 @@ fn convert_expr(
             let base = obj.getattr("base")?;
             let base_id = convert_expr(_py, &base, arena, var_ids, param_ids)?;
             let py_index = obj.getattr("index")?;
-            let index_spec = convert_index_spec(&py_index)?;
+            // The base's shape is needed to normalize Python's negative indices
+            // (`x[-1]`), which `IndexSpec` cannot represent -- its scalar arms are
+            // `usize`. Without it every negative index overflowed the extraction and
+            // aborted the whole conversion (#944).
+            let base_shape = python_shape(&base);
+            let index_spec = convert_index_spec(&py_index, base_shape.as_deref())?;
             Ok(arena.intern(ExprNode::Index {
                 base: base_id,
                 index: index_spec,
@@ -1699,9 +1704,64 @@ fn parse_objective_sense(s: &str) -> PyResult<ObjectiveSense> {
     }
 }
 
+/// Read a Python expression's `.shape` as a list of axis lengths.
+///
+/// Returns `None` when the attribute is missing or is not a tuple of
+/// non-negative ints -- callers then simply lose negative-index normalization
+/// and report the original out-of-range error, which is the pre-existing
+/// behaviour.
+fn python_shape(obj: &Bound<'_, PyAny>) -> Option<Vec<usize>> {
+    let shape = obj.getattr("shape").ok()?;
+    let dims: Vec<isize> = shape.extract().ok()?;
+    if dims.iter().any(|&d| d < 0) {
+        return None;
+    }
+    Some(dims.into_iter().map(|d| d as usize).collect())
+}
+
+/// Normalize one scalar subscript against the length of the axis it indexes.
+///
+/// Python/numpy semantics: a negative `i` means `i + len`, and the result must
+/// land in `[0, len)`. Out-of-range is an `IndexError` -- refused, never clamped
+/// or wrapped, because a wrong-but-in-range slot names a different variable and
+/// is silently accepted by everything downstream (the #941 failure mode).
+///
+/// `axis_len` is `None` when the base's shape could not be read; a non-negative
+/// index passes through unchanged and a negative one is refused with a message
+/// naming the real cause instead of an integer-conversion overflow.
+fn normalize_scalar_index(item: &Bound<'_, PyAny>, axis_len: Option<usize>) -> PyResult<usize> {
+    // bool is a subclass of int in Python, but numpy reads `x[True]` as a mask,
+    // not as `x[1]`; refuse rather than silently disagree with the evaluator.
+    if item.is_instance_of::<pyo3::types::PyBool>() {
+        return Err(PyErr::new::<pyo3::exceptions::PyIndexError, _>(
+            "boolean index is not supported in an expression subscript",
+        ));
+    }
+    let raw: isize = item.extract()?;
+    if raw >= 0 {
+        return Ok(raw as usize);
+    }
+    let len = axis_len.ok_or_else(|| {
+        PyErr::new::<pyo3::exceptions::PyIndexError, _>(format!(
+            "negative index {raw} cannot be resolved: the indexed expression has no shape"
+        ))
+    })?;
+    let shifted = raw + len as isize;
+    if shifted < 0 {
+        return Err(PyErr::new::<pyo3::exceptions::PyIndexError, _>(format!(
+            "index {raw} is out of bounds for axis of length {len}"
+        )));
+    }
+    Ok(shifted as usize)
+}
+
 /// Convert a Python index (int, slice, or tuple of ints/slices) to an
 /// IndexSpec.
-fn convert_index_spec(obj: &Bound<'_, PyAny>) -> PyResult<IndexSpec> {
+///
+/// `base_shape` is the shape of the expression being subscripted, used to turn
+/// Python's negative indices into the non-negative ones `IndexSpec` stores.
+fn convert_index_spec(obj: &Bound<'_, PyAny>, base_shape: Option<&[usize]>) -> PyResult<IndexSpec> {
+    let axis_len = |axis: usize| base_shape.and_then(|s| s.get(axis).copied());
     if obj.is_instance_of::<PyTuple>() {
         let tuple: &Bound<'_, PyTuple> = obj.downcast()?;
         // If every element is a plain int, keep the simple Tuple form.
@@ -1716,25 +1776,27 @@ fn convert_index_spec(obj: &Bound<'_, PyAny>) -> PyResult<IndexSpec> {
         if all_scalar {
             let indices: Vec<usize> = tuple
                 .iter()
-                .map(|item| item.extract::<usize>())
+                .enumerate()
+                .map(|(axis, item)| normalize_scalar_index(&item, axis_len(axis)))
                 .collect::<PyResult<Vec<_>>>()?;
             return Ok(IndexSpec::Tuple(indices));
         }
         let mut elems: Vec<IndexElem> = Vec::with_capacity(tuple.len());
-        for item in tuple.iter() {
+        for (axis, item) in tuple.iter().enumerate() {
             if item.is_instance_of::<PySlice>() {
                 elems.push(slice_to_index_elem(&item)?);
             } else {
-                let i: usize = item.extract()?;
-                elems.push(IndexElem::Scalar(i));
+                elems.push(IndexElem::Scalar(normalize_scalar_index(
+                    &item,
+                    axis_len(axis),
+                )?));
             }
         }
         Ok(IndexSpec::Multi(elems))
     } else if obj.is_instance_of::<PySlice>() {
         Ok(IndexSpec::Multi(vec![slice_to_index_elem(obj)?]))
     } else {
-        let idx: usize = obj.extract()?;
-        Ok(IndexSpec::Scalar(idx))
+        Ok(IndexSpec::Scalar(normalize_scalar_index(obj, axis_len(0))?))
     }
 }
 
--- a/python/discopt/_jax/convexity/rules.py
+++ b/python/discopt/_jax/convexity/rules.py
@@ -533,27 +533,17 @@ def _scalar_value(expr: Expression) -> float:
     return float(np.asarray(expr.value))  # type: ignore[attr-defined]
 
 
-def _const_scale_sign(expr: Expression) -> Optional[int]:
-    """Curvature-scaling sign of a constant factor, scalar **or array**.
+def _finite_const_value(expr: Expression) -> Optional[np.ndarray]:
+    """The value of a concrete, everywhere-finite constant leaf, or ``None``.
 
-    Returns ``+1`` / ``-1`` / ``0`` when every entry of the constant is,
-    respectively, nonnegative (with at least one positive), nonpositive (with at
-    least one negative), or exactly zero — the three cases in which the
-    elementwise product ``c ⊙ f`` scales *every* entry of ``f``'s curvature by
-    one uniform sign, so :func:`~.lattice.scale` applies. Returns ``None`` when
-    ``expr`` is not a concrete constant, holds a non-finite entry, or MIXES
-    signs; the caller may still conclude AFFINE in that last case, since an
-    affine operand stays affine under any constant scaling.
-
-    Entries equal to zero alongside strictly-signed ones do not spoil the
-    verdict: ``0 * f`` is affine, hence both convex and concave, so it never
-    contradicts the strict entries' direction.
-
-    Only 0-d constants used to be recognised (``_is_scalar_const``). That
-    dropped the whole *vectorized* modeling surface to UNKNOWN curvature: a
-    collocation row is ``h ⊙ (A @ x)`` with ``h`` an array of element widths, so
-    every ``discopt.dae`` model classified nonconvex and was routed to spatial
-    McCormick B&B, which does not converge on it (#944).
+    ``None`` means *this operand supports no scaling verdict at all* — it is not a
+    ``Constant``/``Parameter``, its value is not coercible to float64, or it holds
+    an ``inf``/``nan``. Kept separate from :func:`_const_scale_sign`'s "no uniform
+    sign" answer because the two demand opposite handling: a mixed-sign constant
+    still leaves an affine operand affine, whereas a non-finite one must refuse
+    outright (a NaN coefficient would otherwise pass through as AFFINE and read as
+    a proof). Conflating them made the product rule disagree with
+    :func:`_classify_division`, which has always refused non-finite divisors.
     """
     if not isinstance(expr, (Constant, Parameter)):
         return None
@@ -563,6 +553,32 @@ def _const_scale_sign(expr: Expression) -> Optional[int]:
         return None
     if not np.all(np.isfinite(arr)):
         return None
+    return arr
+
+
+def _const_scale_sign(arr: np.ndarray) -> Optional[int]:
+    """Curvature-scaling sign of a finite constant factor, scalar **or array**.
+
+    Returns ``+1`` / ``-1`` / ``0`` when every entry is, respectively, nonnegative
+    (with at least one positive), nonpositive (with at least one negative), or
+    exactly zero — the three cases in which the elementwise product ``c ⊙ f``
+    scales *every* entry of ``f``'s curvature by one uniform sign, so
+    :func:`~.lattice.scale` applies. Returns ``None`` when the entries MIX signs:
+    no single scaling is valid then, though an affine operand still stays affine.
+
+    Entries equal to zero alongside strictly-signed ones do not spoil the
+    verdict: ``0 * f`` is affine, hence both convex and concave, so it never
+    contradicts the strict entries' direction.
+
+    Callers pass a value already screened by :func:`_finite_const_value`; this
+    function answers only the sign question.
+
+    Only 0-d constants used to be recognised (``_is_scalar_const``). That
+    dropped the whole *vectorized* modeling surface to UNKNOWN curvature: a
+    collocation row is ``h ⊙ (A @ x)`` with ``h`` an array of element widths, so
+    every ``discopt.dae`` model classified nonconvex and was routed to spatial
+    McCormick B&B, which does not converge on it (#944).
+    """
     if np.all(arr == 0.0):
         return 0
     if np.all(arr >= 0.0):
@@ -641,10 +657,15 @@ def _classify_product(
     # exactly as a scalar would, and an AFFINE operand survives *any* constant
     # array — including one of mixed sign — because each broadcast entry is
     # ``const * affine`` (#944).
+    # A non-finite constant is refused here rather than falling into the
+    # mixed-sign branch below: `nan * affine` is not affine in any useful sense,
+    # and returning AFFINE for it would emit a *proof* about a broken model. This
+    # matches `_classify_division`, which has always refused non-finite divisors.
     for const_side, other in ((expr.left, right), (expr.right, left)):
-        if not isinstance(const_side, (Constant, Parameter)):
+        arr = _finite_const_value(const_side)
+        if arr is None:
             continue
-        s = _const_scale_sign(const_side)
+        s = _const_scale_sign(arr)
         if s is not None:
             return ExprInfo(scale(other.curvature, s), prod_sign)
         if other.curvature == Curvature.AFFINE:
--- a/python/discopt/_jax/convexity/rules.py
+++ b/python/discopt/_jax/convexity/rules.py
@@ -533,6 +533,70 @@ def _scalar_value(expr: Expression) -> float:
     return float(np.asarray(expr.value))  # type: ignore[attr-defined]
 
 
+def _const_scale_sign(expr: Expression) -> Optional[int]:
+    """Curvature-scaling sign of a constant factor, scalar **or array**.
+
+    Returns ``+1`` / ``-1`` / ``0`` when every entry of the constant is,
+    respectively, nonnegative (with at least one positive), nonpositive (with at
+    least one negative), or exactly zero — the three cases in which the
+    elementwise product ``c ⊙ f`` scales *every* entry of ``f``'s curvature by
+    one uniform sign, so :func:`~.lattice.scale` applies. Returns ``None`` when
+    ``expr`` is not a concrete constant, holds a non-finite entry, or MIXES
+    signs; the caller may still conclude AFFINE in that last case, since an
+    affine operand stays affine under any constant scaling.
+
+    Entries equal to zero alongside strictly-signed ones do not spoil the
+    verdict: ``0 * f`` is affine, hence both convex and concave, so it never
+    contradicts the strict entries' direction.
+
+    Only 0-d constants used to be recognised (``_is_scalar_const``). That
+    dropped the whole *vectorized* modeling surface to UNKNOWN curvature: a
+    collocation row is ``h ⊙ (A @ x)`` with ``h`` an array of element widths, so
+    every ``discopt.dae`` model classified nonconvex and was routed to spatial
+    McCormick B&B, which does not converge on it (#944).
+    """
+    if not isinstance(expr, (Constant, Parameter)):
+        return None
+    try:
+        arr = np.asarray(expr.value, dtype=np.float64)
+    except (TypeError, ValueError):
+        return None
+    if not np.all(np.isfinite(arr)):
+        return None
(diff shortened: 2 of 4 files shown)


jkitchin:
I filed #944, so I reviewed this with one question in front: the fix routes
models onto `_solve_continuous(..., certify_convex=True)`, and a rule that
declares something convex when it isn't produces a false certificate — the
CLAUDE.md §1 catastrophe. I read the rules, probed them on the branch, rebuilt
the Rust extension, and ran the corpus. **The math is right and the fix works.**
Two non-blocking items below, one of which is a direction where the PR is
*less* strict than `main`, contrary to what the description claims.

## What I verified independently

Built the branch's Rust in a clean worktree (`cargo build --release -p
discopt-python`, separate `CARGO_TARGET_DIR`), dropped `lib_rust.dylib` in as
`_rust.cpython-312-darwin.so`, and asserted both `discopt.__file__` inside the
worktree and the presence of `_const_scale_sign` before measuring anything
(CLAUDE.md §8).

**The new test file is a genuine reproducer.** With the *pre-fix* `.so` in
place, `TestNegativeIndexReachesTheRepr` fails on the exact defect:

```
>   model_to_repr(m, getattr(m, "_builder", None))
E   OverflowError: can't convert negative int to unsigned
1 failed, 7 passed
```

**After the rebuild, the headline claim holds:**

```
tests/test_944_array_constant_curvature.py ......... 15 passed
tests/test_dae.py::TestOptimalControl::test_minimum_energy PASSED
16 passed in 1.71s
```

1.71 s for all sixteen, including the three end-to-end solves, against the
>800 s pytest timeout and the 600 s `nbclient` timeout on `main`. `node_count
== 0`, `gap_certified is True`, `convex_fast_path is True`, and the objective
inside 1e-3 of the analytic Riccati value. That is the fix doing exactly what
the issue asked for, and the Riccati oracle is the right way to assert it —
solver-independent, so it cannot drift with the solver.

**Blast radius on the corpus: zero.** I swept `classify_model(m,
use_certificate=True)` over all 66 `.nl` instances in
`python/tests/data/minlplib_nl/`, before and after, in two trees each asserting
its own marker:

```
INSTANCES=66
base convex: 24    fix convex: 24
base ERR:     0    fix ERR:     0
FLIPS=        0
```

Not one verdict moves — expected, since the `.nl` importer emits 0-d constants
and never the array coefficients this rule newly understands, but it is the
measurement that bounds the risk: no existing certified-convex instance changes
its verdict, and no model on that corpus is newly declared convex. The widened
surface is confined to models built through the vectorized Python API.

## Finding 1 — none of the end-to-end coverage ran in CI

All three checks are green, but neither lane executes the tests that prove
termination:

* `Python fast` → `-m "not slow and not correctness and ..."`
* `Python correctness (fast subset)` → `-m "correctness and not slow and ..."`

`TestOptimalControl` carries a class-level `@pytest.mark.slow` (pre-existing),
and both `test_solves_to_certified_optimality_without_branching` and
`test_notebook_variant_also_terminates` are `@pytest.mark.slow`. So what CI
actually validated here is the unit rules, the bridge tests, and
`test_convexity_is_recognised` — the *classification*, but not the *routing*.
A regression that keeps the model classified convex and still sends it to the
spatial path would pass this PR's CI.

Since the whole file now runs in 1.71 s, I'd drop `@pytest.mark.slow` from
`test_solves_to_certified_optimality_without_branching`. The `time_limit=120`
you already added is what makes that safe: a regression fails as an assertion
in ≤120 s instead of hanging the lane, which is precisely the argument in your
own comment in `test_dae.py`. That single test is the one that would have
caught #944.

## Finding 2 — the "both now refuse" claim is half true, and one direction got looser

The description says:

> Two directions got **stricter**: a non-finite scalar constant used to scale
> curvature straight through (`inf > 0` → +1, and NaN fell to −1, silently
> negating the operand's curvature); both now refuse.

Measured on this branch (`x` shape (3,), `_sq(x) = x*x`):

```
inf-array   * affine -> AFFINE     nan-array   * affine -> AFFINE
inf-array   * convex -> UNKNOWN    nan-array   * convex -> UNKNOWN
scalar inf  * affine -> AFFINE     scalar nan  * affine -> AFFINE
affine / inf        -> UNKNOWN     affine / nan        -> UNKNOWN
```

`_const_scale_sign` correctly returns `None` on non-finite, but the very next
line in `_classify_product` is the mixed-sign fallback:

```python
        if other.curvature == Curvature.AFFINE:
            return ExprInfo(Curvature.AFFINE, prod_sign)
```

so a non-finite constant reaches it and returns AFFINE. Three consequences:

1. The claim holds only for a non-affine operand. For an affine operand,
   nothing refuses.
2. For an *array* non-finite constant this is **looser than `main`**, not
   stricter: `main` had no array branch, so `nan_array * affine` fell through
   to `UNKNOWN`; it is now `AFFINE`.
3. It contradicts the sibling rule and this PR's own comment.
   `_classify_division` refuses non-finite outright with no affine fallback,
   and the new comment above the product branch says the array path refuses
   non-finite.

I could not turn this into a demonstrable false certificate — a NaN coefficient
is a broken model long before the walker sees it, and `inf * affine` is affine
in the degenerate sense — so I am not calling it blocking. But it is a two-line
fix that makes the code agree with its own comment and with the division rule:
test finiteness separately and `continue`/refuse before the AFFINE fallback.
`test_non_finite_constant_refuses` only exercises the convex operand, so
neither behaviour is currently pinned.

## The soundness argument checks out

* `*` is elementwise — matmul is a separate node (`_classify_matmul`), so
  `c ⊙ f` really does decompose entrywise and the four-case table in the module
  docstring is valid, broadcasting included.
* Zeros: `scale(c, 0)` → AFFINE, and AFFINE ⊆ convex ∩ concave, so a zero entry
  never spoils a uniform-sign verdict. Correct.
* Mixed sign + non-affine → UNKNOWN is the only safe answer and it is what
  happens.
* `sign_from_value` (lattice.py:126) is already array-safe — it is written in
  `np.all`, so a mixed-sign array constant yields `UNKNOWN` rather than raising
  on ambiguous truth. This was my main worry about widening the rule to arrays
  and it is a non-issue.
* `Constant.__init__` normalises to a float64 ndarray, so a scipy-sparse value
  cannot reach `_const_scale_sign`; and if one did, `np.asarray(sp,
  dtype=np.float64)` raises `ValueError`, which the helper catches and turns
  into a refusal. The CLAUDE.md sparse trap does not apply here.
* `_is_scalar_const` / `_scalar_value` are still live (the power and reciprocal
  rules at :720 and :758), so nothing was left dead.

## The Rust bridge

`normalize_scalar_index` refuses rather than wraps or clamps, which is the
right call and the #941 lesson. Two things I checked and am satisfied with:

* **Axis alignment.** `axis_len(axis)` uses tuple position as the axis index,
  which is only correct if the tuple cannot contain `Ellipsis` or `None`. Both
  fail `extract::<isize>()` and raise, so the mapping can never silently
  misalign. The numpy-oracle test covers the aligned cases at 1-D/2-D/3-D.
* **Slices are untouched** because `slice_to_index_elem` was already `isize`.
  Consistent with the framing that only the scalar arms were `usize`.
* `python_shape` swallows every error from `.shape` via `.ok()?`. Normally I'd
  object to a blanket swallow, but here losing the shape produces a *loud*
  `IndexError`, never a wrong-but-in-range slot, and the docstring says exactly
  that. Fine as written.

One undocumented behaviour change: `x[True]` used to extract as `x[1]` and now
raises `IndexError`. I agree with refusing it — numpy reads a bool as a mask,
so the old behaviour silently disagreed with the evaluator — but it is not
mentioned in the description and no test pins it. Worth one line in the test
file so a future reader doesn't "fix" it back.

## Note, not a finding

`python/discopt/_jax/convexity/linear_context.py:178` carries its own
`_is_scalar_const` with the identical `ndim == 0` gate, driving the same
product/division scale-walk. Its failure direction is abstain-to-box — a
looser sign enclosure, not a wrong one — and I could not construct a case where
the array/scalar distinction changes a verdict (the LP context did not fire on
shaped variables in the probes I tried, e.g. `log(1 + c*(x - y))` with
`y <= x` returns UNKNOWN for both scalar and array `c`). Recording it as the
other place this class defect lives, not as something this PR should carry.

## Verdict

Approve. Finding 1 is worth doing in this PR — un-marking one test costs
nothing and is the difference between "CI proves this works" and "CI proves the
classifier changed." Finding 2 and the `x[True]` note are small and can ride
along or follow.

jkitchin:
Both findings addressed in `e5be547`. Finding 2 was a real defect and my description was wrong about it — retraction first, since that is the part of the record that needs correcting.

## Retraction: "both now refuse" was false, and one direction was looser than `main`

You are right, and the description overclaimed. I reproduced your measurement, then ran the arm you did not: `main`'s `rules.py` against the branch's, same interpreter, each asserting its own marker.

```
case                     main       #947 as pushed      after e5be547
inf-array  * affine      UNKNOWN    AFFINE   <- looser  UNKNOWN
nan-array  * affine      UNKNOWN    AFFINE   <- looser  UNKNOWN
scalar-inf * affine      AFFINE     AFFINE              UNKNOWN
scalar-nan * affine      AFFINE     AFFINE              UNKNOWN
inf-array  * convex      UNKNOWN    UNKNOWN             UNKNOWN
nan-array  * convex      UNKNOWN    UNKNOWN             UNKNOWN
scalar-inf * convex      CONVEX     UNKNOWN             UNKNOWN
scalar-nan * convex      CONCAVE    UNKNOWN             UNKNOWN
```

So the claim held only for the convex-operand rows. For an affine operand nothing refused, and for an *array* non-finite constant the PR was strictly looser than `main` — exactly as you said.

The cause was mine and it is the kind of bug worth naming: `_const_scale_sign` returned `None` for two unrelated reasons — *"not a finite constant"* and *"no uniform sign"* — and the caller could only see one `None`, so it treated the first as the second and fell into the mixed-sign branch. Split into `_finite_const_value` (screens the operand) and `_const_scale_sign` (answers only the sign question), so they cannot be conflated again. Every arm now refuses, matching `_classify_division`, and the code agrees with its own comment.

One row is worth keeping in view: `main`'s `scalar-nan * convex -> CONCAVE`. `nan > 0` is False, so NaN fell to the negative branch and silently *negated* the operand's curvature. That latent bug the first commit did fix — that part of the description was accurate. But it does not excuse the rest, and "both now refuse" is now true rather than merely asserted.

`test_non_finite_constant_refuses` now covers both operand curvatures, both arities, and both operators — 20 assertions with the count asserted, so neither direction can drift back unpinned.

### The stricter direction needed its own blast-radius check

Refusing `scalar-inf * affine` is stricter than `main`, which is a direction your sweep could not have covered — it could in principle cost a certified-convex instance. So I re-ran your sweep over the same 66 instances, extended to compare the per-constraint masks as well as the model verdict, both arms asserting their own marker:

```
INSTANCES=66   COMPARISONS=132
base convex: 24    fix convex: 24
base ERR:     0    fix ERR:     0
MODEL_VERDICT_FLIPS   = 0
CONSTRAINT_MASK_FLIPS = 0
```

Reproduces your 24/24 and adds the mask dimension. Nothing moves.

## Finding 1 — done, and you were right that CI was proving the wrong thing

Un-marked `test_solves_to_certified_optimality_without_branching`. Verified it is actually selected now, rather than assuming it:

```
$ pytest python/tests/test_944_array_constant_curvature.py --collect-only \
    -m "not slow and not correctness and not integration and not amp_benchmark \
        and not requires_cyipopt and not memory_heavy"
...::TestMinimumEnergyTerminates::test_solves_to_certified_optimality_without_branching
15/16 tests collected (1 deselected)
```

Left `test_notebook_variant_also_terminates` as `slow` — one routing assertion in the fast lane is the coverage that was missing; the second adds cost without adding signal. The docstring now says why the un-marking is deliberate and why `time_limit=120` makes it safe, so it does not get re-marked by someone tidying up.

## `x[True]`

Pinned in `test_boolean_subscript_is_refused`, with the reasoning in the docstring — `bool` is an `int` subclass so the old `usize` extraction named slot 1, while numpy reads a bool as a mask. Agreed it was an undocumented change; it is documented now.

## `linear_context.py:178`

Not touched, and I agree with your read: abstain-to-box is a looser enclosure, not a wrong one, so it cannot produce a false verdict the way the walker could. Recording it as the other place this class defect lives.

## Gates re-run on `e5be547`

- `test_944_array_constant_curvature.py` → 16 passed
- `pytest -m smoke` → 909 passed, 16 skipped, 2 xpassed
- corpus sweep → 0 flips (above)
- `ruff check` / `ruff format --check` clean

`pytest -m slow test_adversarial_recent_fixes.py` is still running locally; I will report the number when it lands rather than assert it now, and will act on it if it is not 10 passed.

Thanks for rebuilding the extension and running the corpus — the array/non-finite conflation is not something the tests as written would have surfaced, since the only non-finite case I pinned was the one that happened to already work.

---

jkitchin:
CI is green on `e5be547` — all 11 checks completed, `mergeable_state: clean`.

```
Lint + typecheck                            success
Rust                                        success
Python fast (3.12, ubuntu)                  success
Python correctness (fast subset, 3.12)      success
Python claim-boundary (serial, 3.12)        success
Python AMP coverage (3.12, ubuntu)          success
Detect code changes                         success
codecov/patch, codecov/project              success
Python coverage / AMP Alpine                skipped (path filters)
```

And the number I said I would report rather than assert: `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **10 passed** in 206.8 s on an idle machine. Recorded in the PR body too.

One thing worth your attention before merging: your approval came as a comment on `5dc0d3a`, and `e5be547` changed solver-math behaviour after it — non-finite constants now refuse on the affine arm as well, which is the stricter direction your sweep predates. The 66-instance corpus sweep in the body covers that (0 verdict flips, 0 mask flips, 132 comparisons), but the diff you approved is not the diff at head, so you may want a second look at that commit specifically. It is small: the `_finite_const_value` / `_const_scale_sign` split plus the un-marking, in `rules.py` and the test file only.

---

Issue #946, opened 2026-08-08
GBD's optimality cut trusts an unbounded multiplier — latent, and it depends on the recourse NLP keeping Ipopt's bound relaxation


jkitchin:
> **Updated: this is latent, not an active regression.** As first filed it described
> a live certification loss on `test_gbd::test_linear_objective_nonlinear_constraint`
> caused by #945. The final scoping in #948 does not reach GBD — `bound_relax_factor = 0`
> is requested only at call sites whose returned *point* is the product, and GBD's
> recourse solve is a *dual* consumer, so it keeps Ipopt's default. The test passes and
> the strict xfail was removed. What remains is the underlying fragility, recorded
> below with its measurements so it is not rediscovered from scratch.

## The fragility

`decomposition/benders/gbd.py` forms its Lagrangian optimality cut from whatever
multipliers the recourse NLP returns, with no check that those multipliers are usable.
When the recourse subproblem is **degenerate at its solution** — Slater's condition
fails, i.e. the feasible set has empty interior — no finite multiplier exists, the
returned one diverges, and the cut carries no usable information about `eta`. The
master's lower bound stalls and GBD exits `iteration_limit` instead of `optimal`.

The result stays **sound** — the bound remains valid and the status is honestly
`iteration_limit`, never a false `optimal`. What is lost is *certification*.

## What makes it latent rather than live

GBD does not hit this today because its recourse NLP inherits Ipopt's
`bound_relax_factor = 1e-8`, which gives a degenerate feasible set an **artificial
interior** and so keeps the multiplier finite. In other words: GBD's convergence on
this class of model depends on a recourse point that is slightly *infeasible*.

That is the same dependency #940 found on the Benders dual LP (pinning the option
backend-wide sent two correctness-lane tests from 1.6 s to 79 s), and it is why
#948 scopes the option to point-consumers only. The dependency is deliberate and
documented now (`solvers.pounce_incumbent_options`), but it is a dependency.

## Reproducer and measurements

`python/tests/test_gbd.py::test_linear_objective_nonlinear_constraint`:

```python
m = dm.Model("linnl")
y = m.binary("y")
x = m.continuous("x", shape=(2,), lb=0, ub=5)
m.first_stage(y)
m.minimize(3 * y - x[0] - x[1])
m.subject_to(x[0] * x[0] + x[1] * x[1] <= 8 * y)
solve_benders(m, time_limit=30)     # optimum -1 at y=1, x=(2,2)
```

[...]

Issue #849, opened 2026-07-22
[Adversary] False optimality certificate on bounded convex QCP with large coefficient (all backends)


jkitchin:
> Discovered by the discopt adversary agent (automated solver correctness testing), pre-release sweep 2026-07-22.

## Summary

On a **bounded, feasible, convex** quadratically-constrained problem with a large constraint coefficient, discopt returns `status=optimal, gap=0.0, gap_certified=True` at a point ~20× short of the true optimum, and **the certified dual bound crosses the true optimum** — an unsound lower bound. This violates the certificate invariant in CLAUDE.md §1 ("dual bound never crossing the known oracle"), which is a hard, zero-slack gate.

## Minimal reproduction

Run with `JAX_PLATFORMS=cpu JAX_ENABLE_X64=1`.

```python
import discopt.modeling as dm
m = dm.Model("qcp")
x = m.continuous("x", lb=0, ub=1e12)
m.minimize(-x)
m.subject_to(x * x <= 1e18)          # convex; feasible region x in [0, 1e9]
r = m.solve(nlp_solver="ipm")
print(r.status, r.objective, r.bound, r.gap, r.gap_certified)
```

Output (all three backends agree):

| backend | incumbent obj | certified dual bound | gap | gap_certified | status |
|---|---|---|---|---|---|
| ipm    | −5.013e7 | −5.013e7 | 0.0 | True | optimal |
| ipopt  | −5.015e7 | −5.015e7 | 0.0 | True | optimal |
| pounce | −5.013e7 | −5.013e7 | 0.0 | True | optimal |

**True optimum:** x = 1e9 (constraint `x² ≤ 1e18` binding), obj = **−1e9**.

## Why this is unsound (not just a weak incumbent)

For a minimization, a valid dual/lower bound `L` must satisfy `L ≤ f*`. Here `f* = −1e9` but discopt certifies `L = −5.0e7`, and `−5.0e7 ≤ −1e9` is **false** — the certified lower bound is *above* the true optimum. With `gap=0` and `gap_certified=True`, discopt asserts global optimality of a point where the constraint is nowhere near active (`x² ≈ 2.5e15 ≪ 1e18`). Independent verification: `scipy.optimize.minimize` and the analytic optimum both give −1e9.

## Scale dependence (isolates the failure)

`min -x s.t. x² ≤ R, x ≥ 0`, true opt = −√R:

| R | true −√R | discopt obj | status | note |
|---|---|---|---|---|
[...]

Issue #756, opened 2026-07-18
GDPopt-LOA: two false-infeasible certificates — unsound no-good cut for general integers, and timeout reported as certified infeasibility


jkitchin:
## Summary

Two P0-class defects in `solvers/gdpopt_loa.py`, both producing **`status="infeasible"` with `gap_certified=True` on feasible models**. Found while writing coverage tests for #87.

## 1. No-good cut is unsound for general integer variables

`_add_no_good_cut` (gdpopt_loa.py ~562-577) uses the binary-only form: for the incumbent config, `sum_{v* >= 0.5} v - sum_{v* < 0.5} v <= (count of ones) - 1`. That inequality is only a valid no-good for 0/1 variables. For a general integer `z`, the config `(y=0, z=1)` produces the cut `-y + z <= 0`, which also cuts off the **feasible, better** configuration `(y=1, z=2)`.

Reproduction:
```python
# min y + z  s.t.  y*z >= 2,  y binary, z integer in [0, 3]
# true optimum: 3 at (y, z) = (1, 2)
```
LOA excludes configs one by one with the unsound cut; after 4 cuts the master goes infeasible and the solver returns `status="infeasible"`, `gap_certified=True` — a certified false infeasible on a feasible model.

Correct fix direction: use an integer-valid no-good (e.g. distance form `sum |v - v*| >= 1` linearized, or restrict the binary-form cut to models whose discrete vars are all binary and refuse/branch otherwise).

## 2. Time-limit exhaustion falls through to certified "infeasible"

In the LOA main loop (~288-333), exiting on the time limit with zero completed iterations (no incumbent, no unresolved configs recorded) falls through to the terminal `status="infeasible"` return with the default `gap_certified=True`. A timeout is not an infeasibility proof.

Reproduction: `solve_gdpopt_loa(m, time_limit=0.0)` on the feasible `min x s.t. (x<=3) | (x>=7)` returns certified infeasible. Correct behavior: `status="unknown"`/`"time_limit"` with no certificate.

## Probes

PR #751 adds `python/tests/test_gdpopt_heuristics_core_units.py` with `xfail(strict=False)` probes asserting the correct behavior for both (`test_general_integer_no_good_cut_soundness`, `test_time_limit_exhausted_is_not_infeasible`); they flip to passing when fixed. The same file covers the rest of gdpopt_loa to 98.5% line, so the fixes land against a tested module.

[...]

PR #948, opened 2026-08-08
fix(correctness): #945 request bound_relax_factor at the NLP point-consumers, and stop certifying gap closure on out-of-box incumbents

### LINKED ISSUE #945: NLP path returns points outside their declared bounds, and fixing it exposes that OA/GDPopt gap closure is calibrated on infeasible incumbents
## Summary

Two coupled defects, split out of #940 / #943 because resolving them together changes
certification semantics across three subsystems and needs its own differential panel.

**(a)** `nlp_pounce.solve_nlp` (and `solver.py`'s POUNCE batch path) do not seed
`bound_relax_factor`, so they return points up to `1e-8*(1 + |bound|)` **outside their
declared variable bounds** — Ipopt's default deliberately relaxes bounds, including the
slack bounds standing in for inequality rows. #943 fixed this for the matrix-form
`lp_pounce` / `qp_pounce` backends via `discopt.solvers.pounce_option_defaults()`; the NLP
path was deliberately left out, see (b).

**(b)** Pinning `bound_relax_factor = 0` on the NLP path makes incumbents genuinely
feasible, which turns `incumbent - bound` from `<= 0` into a small **positive** number and
breaks **12** `gap == 0 @ 1e-9` assertions across OA / GDPopt / MindtPy parity. Those
expectations are currently only satisfiable by a solver that hands back
slightly-infeasible incumbents.

## Evidence for (a)

`min sum(y)`, `y ∈ [1,5]³`, exact optimum 3.0. The two formulations take different engine
paths:

```
dm.sum(y.flat)   -> lp_pounce._solve_core   obj=3.0000000075   x inside bounds        ok
dm.sum(y)        -> nlp_pounce.solve_nlp    obj=2.9999999775   x 7.49e-9 BELOW lb=1   BAD
```

The NLP arm's objective is *super-optimal* — below the true optimum — bought by the point
sitting outside its box. Before #943 **both** arms did this, agreed to 1.4e-11, and
`test_sets_adversarial::test_sum_indexed_var_equals_sum_flat` passed on two matched errors
cancelling.

## Evidence for (b)

Arms swapped in-process over `test_mip_nlp.py`, `test_oa.py`,
`test_gdpopt_heuristics_core_units.py` with CI's marker set:

| arm | result |
|---|---|
| POUNCE's own defaults | 320 passed, 0 failed |
| `constr_viol_tol=1e-8` only | 320 passed, 0 failed |
| `bound_relax_factor=0` only | **11 failed**, 309 passed |
| both | **11 failed**, 309 passed |

Every failure is the same assertion, and **the solutions are correct**:

```python
assert result.status in ("optimal", "feasible")           # passes
assert result.objective == pytest.approx(3.0, abs=1e-3)   # passes
assert result.bound == pytest.approx(3.0, abs=1e-3)       # passes
assert result.gap == pytest.approx(0.0, abs=1e-9)         # FAILS: 1.04e-9 .. 1.56e-5
```

The GDPopt case shows the mechanism bare — `min x s.t. (x<=3) or (x>=7)`, `x ∈ [0,10]`,
true optimum 0:

| arm | status | objective | bound |
|---|---|---|---|
| POUNCE defaults | optimal | **−7.54e-09** | 0.0 |
| `brf=0` | feasible | **+2.46e-09** | 0.0 |

The old `optimal` was reached with an incumbent **below its own `lb=0` and below the dual
bound** — `bound <= incumbent` violated, reported as a certificate. Once the incumbent is
honest, `gdpopt_loa._compute_gap` degenerates:

```python
abs_gap = max(0.0, ub - lb)          # 2.4589e-9
if abs_gap <= 1e-9: return 0.0       # 2.46e-9 > 1e-9 -> falls through
denom = max(abs(ub), abs(lb), 1e-10) # 2.46e-9
return abs_gap / denom               # 1.0  -> 100% relative gap at an objective of ~0
```

The absolute-gap escape hatch exists for exactly this degeneracy, but its `1e-9` floor is
tighter than discopt's own `1e-6` absolute tolerance and tighter than the IPM can deliver.

## Why this is not just a test-tolerance question

Certifying `optimal` on the strength of an incumbent that sits outside its box and below
its dual bound is a false certificate in the sense CLAUDE.md §1 cares about — it just
happens to be a very small one. Whatever is done about the gap criterion, the current
situation is that the LP/QP paths now respect their boxes and the NLP path does not.

## Suggested work

1. Seed `nlp_pounce.solve_nlp` and `solver.py`'s batch path from
   `discopt.solvers.pounce_option_defaults()` — a two-line change once (2) is settled.
2. Give gap closure an absolute criterion consistent with discopt's `1e-6` (the usual
   `abs_gap <= abs_tol OR rel_gap <= rel_tol` dual test), rather than a `1e-9` floor that
   only an out-of-box incumbent can beat. Then re-baseline the 12 assertions against the
   corrected behaviour.
3. Differential panel over the corpus for the NLP-path change specifically: node counts,
   objectives, bounds, and `gap_certified` transitions — no instance may lose certification
   and no bound may cross its reference optimum.

## Live markers already in the tree

Both defects are pinned as **strict** xfails, so they flip to passes the moment this lands
rather than needing to be remembered:

* `test_940_pounce_lp_constraint_tolerance::test_returned_point_stays_inside_its_declared_box[nlp_path]`
* `test_sets_adversarial::test_sum_indexed_var_equals_sum_flat` (kept at `abs=1e-9`, not
  loosened — loosening would re-hide the defect)

Attribution harness is vendored at `scratchpad/issue940/arm_plugin.py`; it flips option
arms in-process via `DISCOPT_940_ARM`, so any of the above can be re-measured directly.

Split from #940 / #943 per the discussion there.

jkitchin:
## Summary

Two coupled defects on the NLP path.

**(a) Points outside their declared box.** `bound_relax_factor` (Ipopt's default
`1e-8`) relaxes every bound — including the slack bounds standing in for inequality
rows — so a converged point legitimately sits outside the box the model declared.
#940/#943 fixed the matrix-form LP/QP backends; the NLP path was left for this issue
because the fallout is a certification-semantics change, not an option change.

**(b) Gap closure calibrated on those points.** With incumbents honest,
`incumbent - bound` stops being `<= 0`, and the `1e-9` absolute floor in OA's and
GDPopt-LOA's `_compute_gap` turns out to be beatable only by an incumbent that is not
feasible. The two near-copies are replaced by one definition in `solvers/_gap.py`:

* the absolute criterion is discopt's own `1e-6` (`validation.feasibility.ABS_TOL`,
  `amp.solve(abs_tol=)`), not `1e-9` — the dual `abs OR rel` test `amp.py` has always
  used;
* a dual bound **materially above the incumbent** is reported as "nothing proved"
  rather than clamped to `gap = 0.0` by `max(0.0, ub - lb)`.

Closes #945.

## Where the option is requested, and why that is the whole design

`bound_relax_factor` is **not** a backend-wide default. It lives in a new
`solvers.pounce_incumbent_options()`, requested at the three NLP call sites whose
returned *point* is the product — `solver._solve_continuous` (the model-level NLP
seam, direct analogue of `_solve_lp_pounce`/`_solve_qp_pounce`),
`oa._solve_nlp_attempt`, and `gdpopt_loa._solve_nlp_subproblem`.

The split is not "LP versus NLP". It is **what the caller consumes**:

* a call site whose `x` becomes a solution or incumbent needs it inside the box;
* a call site whose **multipliers** are the product must not pin it. A degenerate
  feasible set (Slater failing, e.g. `x0^2 + x1^2 <= 0`) has *no* finite multiplier;
  Ipopt's relaxation hands it an artificial interior and keeps the dual finite.

That second half is measured, not asserted. #940 found the backend-wide version cost
the Benders dual LP its convergence (two correctness-lane tests, 1.6 s → 79 s). This
branch independently hit the same thing in GBD and traced it — at master proposal
`y = 0` the multiplier goes from `7.07e3` to `9.82e7` and the cut slope to `-7.86e8`,
which carries no usable information about `eta` at any scale, so the master bound
stalls at −2 against a true optimum of −1. Filed as #946, which the final scoping does
not reach; the xfail that pinned it now xpasses and is removed. Row-equilibrating that
cut was checked too — scaling by `1/|s|` reduces it to `-y <= 0`, so the information is
absent, not badly scaled.

## Second pass: the two producers the first pass missed

Review of this PR ([#948 review](https://github.com/jkitchin/discopt/pull/948#issuecomment-5226195369))
found the **default `m.solve()` path** still certifying a super-optimal incumbent on
this PR's own fixture. Byte-identical on `main`, so pre-existing rather than a
regression from the first pass — but it is the same class this PR is named after, on
the entry point users actually hit, so it is fixed here rather than deferred:

```
pre   default  obj=2.9999000090835057  bound=2.9999000090835057  gap=0.0  super-optimal by 9.999e-05
post  default  obj=2.9999999983777133  bound=2.999999999843477   gap=0.0  super-optimal by 1.622e-09
```

Attributed with a trace hook printing `x0 -> x` per `solve_nlp` call, not guessed —
the first attribution in the review was wrong in a way worth recording, because it
named the site that *passed the bad point through* rather than the two that made it:

```
call #5  primal_heuristics.py:324  feasibility_pump    x0=[1.0, 1.0] -> x=[2.9999000090, 1.0]
call #6  solver.py:13853           _solve_node_nlp_kkt x0=[3.0, 1.0] -> x=[2.9999000090, 1.0]
```

Two producers, and fixing either alone leaves the defect standing:

* **`_jax/primal_heuristics`** — `feasibility_pump` projected onto the *relaxed* box
  and handed back `x = 2.9999` as a candidate incumbent. Every NLP solve in that
  module produces a candidate incumbent, so all six option sites now route through
  one `_heuristic_nlp_options()` helper that seeds `pounce_incumbent_options()`. One
  helper rather than six edits is deliberate: a seventh heuristic cannot silently
  keep Ipopt's default. `pounce_option_defaults()` is **not** seeded there — its
  `constr_viol_tol` is separable, and measurably harmful on this path (the nvs05
  numbers below).
* **`solver._solve_nlp_bb`'s terminal re-solve** — it then adopted its *own* primal,
  which came from the relaxed box too, so even with the pump honest (`x = 3` exactly)
  the answer was overwritten straight back to 2.9999. That site consumes **both**
  products of one NLP solve and they want opposite options, so it now takes two
  solves: a refine solve carrying `pounce_incumbent_options()` whose point is adopted,
  then the recover solve *at that adopted point*, left relaxed because its product is
  the multipliers (#946). Duals therefore stay consistent with the primal actually
  reported, which is what the adoption existed for. `_polish_opts` — the sibling
  adoption in the batch path, whose polished point becomes the reported solution —
  takes the incumbent options for the same reason; its sibling `recover_opts` does not.

This is the design rule of this PR applied one level down: the split is not
LP-vs-NLP and not even callee-by-callee, it is **what the caller consumes**. The
`for_incumbent` seam on `solve_nlp` that the review proposed would not have been
enough, because the site that needed splitting is a *caller* that wants a point and
multipliers out of the same solve.

**No feasibility screen could have caught this.** The bad point's worst row violation
is 1e-8 — inside every tolerance discopt has, including
`_check_constraint_feasibility`'s `tol=1e-4` default — while the damage to the answer
is 1e-4, because `(x-3)^2 <= 0` squares the excursion. The only available fix is to
not *produce* the point.

Three stale descriptions of the superseded design (in which `solve_nlp` itself was
seeded) are corrected in the same commit: `pounce_option_defaults`'s docstring, which
claimed to be the source of truth for "EVERY POUNCE entry point ... and the NLP path";
`test_940_pounce_lp_constraint_tolerance`'s docstring; and
`scratchpad/issue945/entry_bounds.py`'s §8 arm marker, which tested `solve_nlp`'s own
source and was therefore permanently `False` on the post tree — it mislabelled every
post run `UNSEEDED`, and the harness could not run at all besides (`Model.add_variable`
does not exist, `solve()` refuses `verbose`). It now mirrors the test it claims to
measure and reports `executed_assertions=6 violations=0`.

## The defect is larger than the issue described

The issue framed (b) as a tolerance question with `~1e-9` numbers. Measured
(`scratchpad/issue945/attribution.py`, 18 comparisons, arms interleaved per fixture):

| fixture | arm | objective | bound | bound−incumbent | vs true optimum |
|---|---|---|---|---|---|
| mindtpy_simple (opt 3.5) | pre | 3.49999998703 | 3.5 | 1.30e-08 | 1.30e-08 **below** |
| | post | 3.50000000364 | 3.5 | 0 | 0 |
| mindtpy_cq (opt 3.0) | pre | **2.9999000025** | 2.99995000083 | **5.00e-05** | **1.00e-04 below** |
| | post | 2.99999999906 | 2.99995332577 | 0 | 9.42e-10 |
| gdp disjunction (opt 0) | pre | −7.54e-09 | 0 | 7.54e-09 | 7.54e-09 below |
| | post | +2.46e-09 | 0 | 0 | 0 |

The constraint-qualification row is the headline: discopt certified `optimal` at
2.9999000025 on a fixture whose optimum is exactly 3.0 — super-optimal by 1e-4, with a
reported dual bound 5e-5 **above its own reported incumbent**, and a reported gap of
`0.0`. That zero was `max(0.0, ub - lb)` clamping a *negative* gap.

`bound_relax_factor`'s damage is not bounded by its own size: `(x-3)^2 <= 0` relaxed by
1e-8 admits every `x` within **1e-4** of 3, because the row squares the excursion.
Reduced to a one-line model:

```
pre    status='optimal'          x = 2.9999000024987788   (1.0e-04 off)
post   status='iteration_limit'  x = 2.9999999998864917   (1.1e-10 off)
```

That also restores **#849's KKT-residual guard**, which previously could not see this
degeneracy at all — measured against the relaxed box the residuals looked clean, so a
problem whose only feasible point is `x = 3` came back certified.

Once the arms were separable it also became visible that **(b) catches the crossing on
its own**: in the pre arm the `mindtpy_cq` rows now return `feasible`, not `optimal`,
because the inversion guard refuses to certify the 5e-5 crossing even with (a) off.
(a) removes the cause; (b) catches it independently.

## Correctness

- [x] Cannot produce a false certificate — this *removes* three classes of one
  (super-optimal incumbent, bound above incumbent, `gap = 0` from a clamped negative
  gap). The absolute criterion widens `1e-9 → 1e-6`, which only bites for
  `|obj| < 0.01`; above that the relative test already allowed it, and `1e-6` is
  discopt's declared absolute tolerance, not a new one.
- [x] No validation, fallback, or safety guard was weakened. The inversion guard is
  strictly *added*; it fires on nothing in the corpus today, by design.

Remaining surface, named rather than implied closed: `amp.py:548/557/565`,
`mpec.py:346`, `_jax/differentiable.py` (5 sites), `gdpopt_loa.py:418` and
`solver.py:7836` still take Ipopt's default `bound_relax_factor`. Not all of them
consume points, and none of them reproduces the defect on this PR's fixture — the AMP
path returns exactly 3.0 — so changing them here would be a fix without a measurement,
which §4 forbids. The rule they need to be audited against is the one this PR
establishes, and the `panel_superopt_recheck.py` numbers above are how the audit would
be judged: the 12 instances still super-optimal in the post arm are where those
producers show up.

The `1e-9`/`1e-10` pair in `_jax/primal_heuristics` is deliberately **not** unified: it
gates whether to SKIP a heuristic, so the tighter floor is the safe direction there.
Commented rather than changed, so a later "consistency" edit cannot silently loosen it.

## What was measured and then dropped

`constr_viol_tol = 1e-8` is a backend-wide default for the matrix-form backends. An
earlier revision of this branch also routed it into `nlp_pounce.solve_nlp`. Replication
caught nvs05's incumbent getting reproducibly worse, so it was attributed
(`nvs05_attribution.py`, 4 arms interleaved, 3 reps, 20 s, **zero** within-arm spread):

```
print_level only            objective  8.731956987   bound 3.542469244   n=41
+ constr_viol_tol=1e-8      objective 12.589492574   bound 3.513753544   n=33
+ incumbent options only    objective  8.731956987   bound 3.542469244   n=41
```

`main == inc` and `cvt == new`: the incumbent options are innocent and
`constr_viol_tol` carries the whole 31% regression alone.

**Retraction (§11).** I then assumed the cheaper incumbent was itself the defect —
bought by a row violated inside Ipopt's loose 1e-4 — which would have made the
regression the fix working. `nvs05_feasibility.py` falsified that against the model's
own constraints: worst row violation 1.8e-12, box 0, integrality 0. Both incumbents are
genuinely feasible. A real solution lost, not a false one rejected.

Measured cost, no measured benefit — it fixes nothing this issue is about (the `cvt`
arm leaves `mindtpy_cq` at 2.9999000025), and the panel found 0 soundness findings in
*both* arms. Sound but not helpful stays out (CLAUDE.md §5, the `DISCOPT_CUT_INHERIT`
lesson). The numbers are recorded in `nlp_pounce.solve_nlp` so re-adding it is a
decision rather than a rediscovery.

## Tests run

Merged `origin/main` (both #943 and #947 landed) and rebuilt — #947 touches
`crates/discopt-python/src/expr_bindings.rs`, so a stale build would have measured the
wrong tree. Both changesets asserted loaded before measuring.

- [x] `pytest -m smoke` → **920 passed, 1 skipped, 2 xpassed** (re-run after the second
      pass, serial, on a machine where `cyipopt` is present — which is why the
      passed/skipped split differs from the 913/15 of the first-pass run in a container
      without it; the 2 xpasses are the same non-strict #756 pair)
      in 145.86 s (serial; under `-n 4` `test_large_dense_jacobian_no_crash` fails on
      contention, and passes isolated on this branch and on the base)
- [x] `python/tests/test_945_...` + `test_940_...` + `test_mip_nlp.py` + `test_oa.py`
      + `test_gbd.py`, the files the second pass touches directly → **247 passed**
- [x] `cargo test -p discopt-core` → N/A, no Rust touched by this PR
- [x] `ruff check python/` clean; the 2 files `ruff format --check` would touch
      (`test_log_square_relaxation.py`, `test_mo_augmecon2.py`) are untouched here and
      identical on `main`

Also run:

- CI-fast marker set, full `python/tests/` → **7358 passed, 130 skipped, 5 xfailed,
  2 xpassed, 2 failed**. Both failures are `ModuleNotFoundError: No module named
  'cyipopt'` (absent in this container; CI installs it), and both — plus both xpasses
  (#756, non-strict) — are byte-identical on the base tree.
- **Correctness lane, serial** → **288 passed, 17 skipped, 4 xfailed**. This is the
  lane #940 was caught by after a CI-equivalent command that carried
  `-m "not correctness"`; it is run here deliberately.

## Regression tests

- [x] `python/tests/test_945_nlp_box_and_gap_closure.py` — 11 tests: the NLP path stays
      inside its box, the squared-row amplification, the `1e-6` absolute criterion, the
      inverted-bound refusal, LOA certifying a near-zero optimum honestly, and
      `test_incumbent_options_are_requested_by_point_consumers_only`, which asserts in
      **both** directions — the three point-consumers request the option and the
      backend does not — as the companion to #943's
      `test_option_requests_are_wired_where_the_guard_checks`.
- [x] Three of the 11 are the second pass, and all three fail on the tree without it:
      `test_default_solve_path_does_not_certify_a_super_optimal_incumbent` runs the
      constraint-qualification MINLP through plain `m.solve()` and asserts
      `objective >= 3.0 - 1e-6` plus `bound <= objective` when the status is `optimal`
      — the behavioural test, which does not care how the fix is spelled;
      `test_every_primal_heuristic_requests_the_incumbent_options` pins the one-helper
      structure (`>= 7` call sites route through `_heuristic_nlp_options`, a caller
      override still wins, and `pounce_option_defaults()` is *absent*), so a seventh
      heuristic added later cannot silently keep Ipopt's default; and
      `test_nlp_bb_terminal_resolve_separates_its_point_from_its_multipliers` pins the
      two-solve split in both directions — the refine solve carries the incumbent
      options, the recover solve does not, and the recover solve's own `x`/objective
      are never adopted. All three are in the `smoke` set.

Two strict xfails flip to passes, as they were written to:
`test_940_pounce_lp_constraint_tolerance[nlp_path]` and
`test_sets_adversarial::test_sum_indexed_var_equals_sum_flat`.

Of the 11 assertions (b) breaks, **8 are restored** by the corrected absolute
criterion. The 3 remaining (`mindtpy_cq`) are re-baselined: the old `gap == 0` was the
clamped negative gap above, and the honest value is a 1.6e-5 relative gap inside the
solver's declared 1e-4 tolerance. Each now also pins `objective >= 3.0 - 1e-6` and
`bound <= objective` — the two properties the old numbers violated.

## Differential panel

Not behind a flag: this is a soundness fix, and the flag-OFF path is the false
certificate. Evidence instead. 66 in-repo `.nl` instances, both arms interleaved,
leading arm alternated per instance, 20 s each, **re-run on the tree that ships**
(the earlier panel predates the second pass, so its pre arm no longer reconstructs
the right baseline — `panel_corpus.py` now patches the two new seams as well):

```
COMPARISONS_EXECUTED=66      identical=30  differing=36
CERT_ORDER_CHECKS_EXECUTED=103
ORACLE_CHECKS_EXECUTED=32    soundness_findings=0
CERT_REGRESSIONS=0
SOLVE_NLP_CALLS pre=4037 post=4052
```

The call counter is the point: 4052 entries into `solve_nlp` is what makes "0
findings" mean something rather than reporting a path that never ran. 36 differing
against the first pass's 7 is the expected shape — the first pass touched three call
sites on the OA/LOA/`_solve_continuous` paths; this one touches the default B&B path's
heuristics and its terminal re-solve, which nearly every instance goes through.

**A witness, not just a threshold.** "0 soundness findings" is a bound not tripped, and
a fix whose entire claim is that incumbents stop being better-than-possible should be
shown *moving the number*. `panel_superopt_recheck.py` re-scores both arms of the
panel's own output against a **second, independent oracle** — MINLPLib's
`minlplib.solu`, `=opt=` rows only (`=best=` is an incumbent someone found, not a
proven optimum, so it cannot witness anything) — at `1e-12` relative, four orders
tighter than the gate:

```
COMPARED=50
SUPER_OPTIMAL_INSTANCES pre=24  post=12
WRONG_WAY=0
```

24 of 50 instances reported an incumbent strictly **better than the published global
optimum** before this branch. 12 do after; 12 flip to sound; **zero** move the other
way. Named rather than summarised, so the ones that are *not* fixed stay visible: the
12 remaining are either byte-identical in both arms — `alan`, `ex1221`, `gbd`,
`st_miqp1/4/5`, `st_test1`, which come from producers this PR does not touch — or
reduced (`tls2` 1.47e-08 → 2.36e-12, `st_e38` 4.47e-09 → 6.58e-11, `fac2`
1.09e-09 → 6.95e-11). Those instances are the honest remaining surface, not a claim
this PR closes.

Cost, stated rather than buried. Node counts are a wash in both directions —
`st_e11` 27 → 5, `casctanks` 31 → 15, `tanksize` 11451 → 10861, `clay0303hfsg`
319 → 287, against `syn05hfsg` 185 → 285, `nvs05` 131 → 155 and several 3 → 5.
`nvs05`'s bound tightens 3.805 → 4.099 and `heatexch_gen2` finds an incumbent where
the pre arm found none. Exactly one incumbent is genuinely worse: `syn05hfsg` (a MAX)
drops 837.7324124 → 837.7323013 — but 837.7324124 is *above* the published optimum
837.7324009, so the incumbent that was lost is the impossible one. That is the §5
net-positive bar met on the same run as the cert-clean bar, not asserted separately.

Earlier panels ran at wider scopes — 36 differing backend-wide, 30 with
`constr_viol_tol`, 7 for the first pass alone. One of those runs flagged `tls2` as a
certification regression; 5-rep replication settled it as noise (**both** arms reach
`optimal` on some reps, and it is the *pre* arm that also hits `time_limit`, which post
never does), and no panel since reproduces it. The differing *set* itself moves between
runs, which is the evidence that the time-limited diffs are budget-dependent rather
than arm effects.

Limitation, stated plainly: the full MINLPLib snapshot is not on the CI image, so the
panel is the in-repo corpus and `panel_superopt_recheck.py` needs the local benchmark
snapshot to run at all — it is a scratchpad witness, not a test. Of the 66 instances
only 32 carry a registry optimum (50 have a `solu` one). The counters are printed
separately rather than as one reassuring total.

## Instrument bugs I made, and corrected

Recorded because CLAUDE.md §6–§11 exist for exactly these.

1. The panel gated its bound-vs-incumbent check behind a recorded optimum being
   present, so it silently skipped 50 of 66 instances *while its counter reported them
   checked*. It needs no oracle and now runs on every solved instance with its own
   counter. It was also sense-blind — `syn05hfsg` MAXIMIZES, so its correct
   `bound = 840.49 >= objective = 837.73` read as a 2.76 inversion.
2. `panel_replicate.py`'s verdict column reported "arm effect" for all six instances,
   including one whose arms differ by 2e-9 relative — a bare interval-overlap test with
   no tolerance. Retracted, column removed rather than left to be re-read, judging
   moved to `panel_replicate_analyze.py`.
3. Both harnesses' §8 version markers named the *superseded* contract and refused to
   run after the rescope. That is the guard working — a panel that had quietly kept
   running would have compared the pre arm against itself and reported 0 differences.
   Both now assert the real contract in both directions.
4. `test_945`'s squared-row test asserted a status the solver has no business giving on
   a degenerate problem; it was written but never run in isolation before the suite
   caught it.

## Harnesses

Vendored under `scratchpad/issue945/`, all flipping arms in-process so every number
above can be re-measured: `entry_bounds.py`, `attribution.py`, `panel_corpus.py`,
`panel_replicate.py`, `panel_replicate_analyze.py`, `nvs05_attribution.py`,
`nvs05_feasibility.py`, `panel_superopt_recheck.py`, and three GBD probes (`gbd_probe.py`, `gbd_cut_probe.py`,
`gbd_anchor_probe.py` — the last traps `_recourse`'s real return value with a trace
hook rather than reimplementing its arithmetic).

Note for reviewers re-running these: CI installs `pounce-solver` from git `main` while
this work was measured against the PyPI 0.9.0 wheel, so exact digits may shift.

### DIFF
--- a/python/discopt/_jax/primal_heuristics.py
+++ b/python/discopt/_jax/primal_heuristics.py
@@ -27,7 +27,7 @@ from discopt.modeling.core import Model, VarType
 # Re-exported here because callers/docs reference ``primal_heuristics.is_qubo`` /
 # ``qubo_local_search``; ``solve_model`` imports the JAX-free module directly.
 from discopt.qubo_primal import is_qubo, qubo_local_search  # noqa: F401
-from discopt.solvers import NLPResult, SolveStatus
+from discopt.solvers import NLPResult, SolveStatus, pounce_incumbent_options
 
 logger = logging.getLogger(__name__)
 
@@ -42,6 +42,39 @@ logger = logging.getLogger(__name__)
 # capped, unconverged point simply fails the feasibility check and yields no
 # incumbent — it can never inject a wrong one (inject_incumbent re-verifies).
 _HEURISTIC_NLP_MAX_ITER = 300
+
+
+def _heuristic_nlp_options(caller_options: Optional[dict] = None) -> dict:
+    """Base NLP options for a sub-solve in this module (#945).
+
+    Every NLP solve here exists to produce a candidate **incumbent**, so every one
+    of them takes :func:`discopt.solvers.pounce_incumbent_options`: the returned
+    point has to lie inside the box the model declared, or the incumbent it becomes
+    is not a solution of the model the user wrote. Ipopt's default
+    ``bound_relax_factor`` relaxes every bound — including the slack bounds standing
+    in for inequality rows — by ``1e-8*(1 + |bound|)``, and a squared row takes the
+    square root of that: ``(x-3)^2 <= 0`` admits every ``x`` within 1e-4 of 3.
+    Measured on the MindtPy constraint-qualification fixture, whose exact optimum is
+    3.0, ``feasibility_pump`` returned ``x = 2.9999`` and the default ``m.solve()``
+    path certified it ``optimal`` with ``gap = 0.0``; with this seed the same run
+    returns 2.999999998 (``scratchpad/issue945/heuristic_incumbent_box.py``).
+
+    :func:`discopt.solvers.pounce_option_defaults` is deliberately **not** seeded
+    here. The two halves are separable and only this one is wanted: its
+    ``constr_viol_tol = 1e-8`` costs a 31%-worse incumbent and a looser bound on
+    nvs05, entirely on its own (see the measurement in ``nlp_pounce.solve_nlp``).
+
+    Routed through one helper rather than spelled at each call site so a seventh
+    heuristic cannot silently keep Ipopt's default — the #940 lesson, one level up.
+    Caller options are merged *after* the seed, so an explicit request still wins.
+    """
+    opts = pounce_incumbent_options()
+    if caller_options:
+        opts.update(caller_options)
+    opts.setdefault("print_level", 0)
+    opts.setdefault("max_iter", _HEURISTIC_NLP_MAX_ITER)
+    return opts
+
 
 # VOLUME-1 (docs/dev/nlp-solve-volume-2026-07-06.md) + ILS-DEFAULT
 # (docs/dev/ils-default-validation-2026-07-06.md): the objective-improvement
@@ -184,9 +217,7 @@ class MultiStartNLP:
         rng = np.random.default_rng(self._seed)
         starts = _generate_starts(lb, ub, self._n_starts, rng)
 
-        opts = dict(ipopt_options) if ipopt_options else {}
-        opts.setdefault("print_level", 0)
-        opts.setdefault("max_iter", _HEURISTIC_NLP_MAX_ITER)
+        opts = _heuristic_nlp_options(ipopt_options)
         if backend is None:
             from discopt.solvers.nlp_backend import get_nlp_solver
 
@@ -271,9 +302,7 @@ def feasibility_pump(
     if evaluator is None:
         evaluator = cached_evaluator(model)
 
-    opts = dict(ipopt_options) if ipopt_options else {}
-    opts.setdefault("print_level", 0)
-    opts.setdefault("max_iter", _HEURISTIC_NLP_MAX_ITER)
+    opts = _heuristic_nlp_options(ipopt_options)
     if backend is None:
         from discopt.solvers.nlp_backend import get_nlp_solver
 
@@ -472,9 +501,7 @@ def subnlp(
                     v.ub = fixed.copy()
                 offset += sz
 
-        opts = dict(nlp_options) if nlp_options else {}
-        opts.setdefault("print_level", 0)
-        opts.setdefault("max_iter", _HEURISTIC_NLP_MAX_ITER)
+        opts = _heuristic_nlp_options(nlp_options)
         if time_budget is not None and time_budget > 0.0:
             opts.setdefault("max_wall_time", float(time_budget))
 
@@ -596,9 +623,7 @@ def continuous_multistart(
     rng = np.random.default_rng(seed)
     starts = _generate_starts(lb, ub, n_starts, rng)
 
-    opts = dict(nlp_options) if nlp_options else {}
-    opts.setdefault("print_level", 0)
-    opts.setdefault("max_iter", _HEURISTIC_NLP_MAX_ITER)
+    opts = _heuristic_nlp_options(nlp_options)
 
     best_obj = float("inf") if incumbent_obj is None else float(incumbent_obj)
     best_x: Optional[np.ndarray] = None
@@ -880,9 +905,7 @@ def integer_local_search(
         # discards this second (relaxation) restart base. Clip first so every
         # coordinate is a finite, usable start.
         mid = np.clip(0.5 * (np.clip(lb, -1e3, 1e3) + np.clip(ub, -1e3, 1e3)), -1e3, 1e3)
-        relax_opts = dict(nlp_options) if nlp_options else {}
-        relax_opts.setdefault("print_level", 0)
-        relax_opts.setdefault("max_iter", _HEURISTIC_NLP_MAX_ITER)
+        relax_opts = _heuristic_nlp_options(nlp_options)
         budget.charge(NLP_SOLVE)
         relax_res = backend(evaluator, mid, options=relax_opts)
         if relax_res is not None and relax_res.x is not None:
@@ -1367,9 +1390,7 @@ def diving(
     lb0, ub0 = _get_variable_bounds(model)
     slot_map = _flat_slot_map(model)
 
-    opts = dict(nlp_options) if nlp_options else {}
-    opts.setdefault("print_level", 0)
-    opts.setdefault("max_iter", _HEURISTIC_NLP_MAX_ITER)
+    opts = _heuristic_nlp_options(nlp_options)
 
     fixed = np.zeros(int_mask.shape[0], dtype=bool)
     x_cur = np.clip(np.asarray(x_relax, dtype=np.float64), lb0, ub0)
--- a/python/discopt/_jax/primal_heuristics.py
+++ b/python/discopt/_jax/primal_heuristics.py
@@ -1805,6 +1805,13 @@ def local_branching(
         and np.isfinite(node_bound)
         and np.isfinite(incumbent_obj)
     ):
+        # Deliberately NOT the shared certification tolerance from
+        # ``solvers._gap`` (1e-6), despite looking like the same test. This gate
+        # decides whether to SKIP a heuristic, so its safe direction is the
+        # opposite one: a floor tighter than the certificate tolerance runs the
+        # search more often, which costs time and can only ever find a better
+        # incumbent. Widening it here to 1e-6 for consistency would skip more and
+        # is a search-behaviour change owing its own net-positive panel (#945).
         abs_gap = incumbent_obj - float(node_bound)
         denom = max(abs(incumbent_obj), abs(float(node_bound)), 1e-10)
         if abs_gap <= 1e-9 or abs_gap / denom <= gap_tolerance:
--- a/python/discopt/solver.py
+++ b/python/discopt/solver.py
@@ -11776,6 +11776,14 @@ def solve_model(
                     _fix_lb[_off + _k] = _val
                     _fix_ub[_off + _k] = _val
             _polish_opts = dict(opts)
+            # The polished point IS the reported solution when it is adopted below,
+            # so this is an incumbent producer and takes the incumbent options
+            # (#945). Without them the polish can walk the continuous variables
+            # outside the declared box by ``1e-8*(1+|bound|)`` per bound — the very
+            # digits it exists to tighten — and the adopted objective is then
+            # super-optimal. The sibling ``recover_opts`` re-solve at :13848
+            # deliberately does NOT take them: its product is the MULTIPLIERS.
+            _polish_opts.update(pounce_incumbent_options())
             _polish_opts["max_wall_time"] = max(
                 0.1, min(5.0, time_limit - (time.perf_counter() - t_start))
             )
@@ -13834,9 +13842,32 @@ def _solve_nlp_bb(
             sol_flat = _rounded_inc
         x_dict = _unpack_solution(model, sol_flat)
 
-        # Recover relaxation duals at the incumbent by re-solving the NLP with
-        # integer variables fixed. Best-effort — failures leave duals as None
-        # and the examiner falls back to its LSQ recovery.
+        # Refine the incumbent's continuous variables, then recover relaxation
+        # duals at the refined point, by re-solving the NLP with integer variables
+        # fixed. Best-effort — failures leave duals as None and the examiner falls
+        # back to its LSQ recovery.
+        #
+        # This site consumes BOTH products of an NLP solve, and the two want
+        # opposite options (#945), so it takes two solves rather than one:
+        #
+        #   * the refine solve's product is the reported POINT, so it requests
+        #     ``pounce_incumbent_options()``. Ipopt's default ``bound_relax_factor``
+        #     relaxes every bound — including the slack bounds standing in for
+        #     inequality rows — by ``1e-8*(1 + |bound|)``, and on an ill-conditioned
+        #     row that is not a 1e-8 error in the answer: ``(x-3)^2 <= 0`` violated
+        #     by 1e-8 is ``x`` wrong by 1e-4. On the MindtPy constraint-qualification
+        #     fixture (exact optimum 3.0) this single re-solve took a feasible
+        #     incumbent at x = 3 and overwrote it with x = 2.9999, which the search
+        #     then certified ``optimal`` at ``gap = 0``. No feasibility screen can
+        #     catch that — the bad point's worst row violation is 1e-8, well inside
+        #     any sane tolerance — so the fix has to be not to *produce* it.
+        #   * the recover solve's product is the MULTIPLIERS, so it must NOT set it:
+        #     a degenerate feasible set (Slater failing) has no finite multiplier
+        #     without Ipopt's relaxation (#946).
+        #
+        # Order matters: refine first, then recover at the adopted point, so the
+        # duals stay consistent with the primal that is actually reported — which
+        # is what the adoption below exists for.
         try:
             fix_lb = lb.copy()
             fix_ub = ub.copy()
@@ -13850,6 +13881,27 @@ def _solve_nlp_bb(
                 0.1, min(5.0, time_limit - (time.perf_counter() - t_start))
             )
             recover_opts.setdefault("print_level", 0)
+            refine_opts = dict(recover_opts)
+            refine_opts.update(pounce_incumbent_options())
+            nlp_refined = _solve_node_nlp_kkt(
+                evaluator, sol_flat, fix_lb, fix_ub, constraint_bounds, refine_opts
+            )
+            if (
+                nlp_refined.status == SolveStatus.OPTIMAL
+                and nlp_refined.x is not None
+                and np.all(np.isfinite(nlp_refined.x))
+                and nlp_refined.objective is not None
+                and abs(float(nlp_refined.objective) - obj_val) <= 1e-4 * (1.0 + abs(obj_val))
+            ):
+                refined = np.asarray(nlp_refined.x, dtype=float).copy()
+                # Keep integer columns pinned at their (rounded) incumbent
+                # values; only the continuous variables are refined.
+                for off, sz in zip(int_offsets, int_sizes):
+                    for k in range(int(sz)):
+                        refined[off + k] = round(float(sol_flat[off + k]))
+                sol_flat = refined
+                x_dict = _unpack_solution(model, sol_flat)
+                obj_val = float(nlp_refined.objective)
             nlp_recovered = _solve_node_nlp_kkt(
                 evaluator, sol_flat, fix_lb, fix_ub, constraint_bounds, recover_opts
             )
@@ -13876,30 +13928,14 @@ def _solve_nlp_bb(
                             if bound_duals_upper is not None and v.name in bound_duals_upper:
                                 bound_duals_upper[v.name] = np.zeros_like(bound_duals_upper[v.name])
 
-                # Adopt the refined primal from the KKT re-solve. It solves the
-                # same integer-fixed subproblem to tighter precision than the
-                # batched JAX IPM, so the continuous variables (and thus
-                # active-constraint residuals) stay consistent with the
-                # recovered duals — without this, complementarity (mu * residual)
-                # can exceed validation tolerances. Guarded: only adopt an
-                # OPTIMAL re-solve whose objective matches the incumbent, so a
-                # divergent recover can never corrupt the reported solution.
-                if (
-                    nlp_recovered.status == SolveStatus.OPTIMAL
-                    and nlp_recovered.x is not None
-                    and np.all(np.isfinite(nlp_recovered.x))
-                    and nlp_recovered.objective is not None
-                    and abs(float(nlp_recovered.objective) - obj_val) <= 1e-4 * (1.0 + abs(obj_val))
-                ):
-                    refined = np.asarray(nlp_recovered.x, dtype=float).copy()
-                    # Keep integer columns pinned at their (rounded) incumbent
-                    # values; only the continuous variables are refined.
-                    for off, sz in zip(int_offsets, int_sizes):
-                        for k in range(int(sz)):
-                            refined[off + k] = round(float(sol_flat[off + k]))
-                    sol_flat = refined
-                    x_dict = _unpack_solution(model, sol_flat)
-                    obj_val = float(nlp_recovered.objective)
+                # The refined primal was already adopted above, from the
+                # incumbent-options solve, and this recover ran *at* that point —
+                # so the continuous variables (and thus active-constraint
+                # residuals) stay consistent with the recovered duals, which is
+                # what the adoption exists for: without it, complementarity
+                # (mu * residual) can exceed validation tolerances. The recover
+                # solve's own ``x`` is deliberately NOT adopted; it is the output
+                # of a deliberately relaxed box (see the two-solve note above).
         except Exception as _exc:
             logger.debug("NLP-BB dual recovery failed: %s", _exc)
 
--- a/python/discopt/solver.py
+++ b/python/discopt/solver.py
@@ -44,7 +44,7 @@ from discopt.modeling.core import (
 from discopt.solver_tuning import current as _tuning
 from discopt.solver_tuning import reset_current as _reset_tuning
 from discopt.solver_tuning import set_current as _set_tuning
-from discopt.solvers import SolveStatus
+from discopt.solvers import SolveStatus, pounce_option_defaults
 
 # R3a measurement sink (temporary, behavior-neutral). When set to a mutable
 # dict by an experiment harness, the nonconvex B&B path stores the Rust tree's
@@ -14306,9 +14306,10 @@ def _solve_batch_pounce(
 
     # Batch-level options. Whitelist keys POUNCE understands and enforce the
     # per-node wall-time guard (issue #5) used by the serial pounce path.
-    # Deliberately NOT seeded from solvers.pounce_option_defaults: this is the NLP
-    # batch path, and bound_relax_factor=0 there breaks OA/GDPopt gap closure (#945).
-    batch_opts: dict = {"print_level": 0}
+    # Seeded from the shared baseline (#945) so a batched node returns the same
+    # in-the-box point the serial ``nlp_pounce.solve_nlp`` path does — otherwise
+    # batching alone would change which side of a bound an incumbent lands on.
+    batch_opts: dict = pounce_option_defaults()
     for k in ("max_iter", "tol", "acceptable_tol"):
         if options.get(k) is not None:
             batch_opts[k] = options[k]
--- a/python/discopt/solver.py
+++ b/python/discopt/solver.py
@@ -14317,10 +14317,10 @@ def _solve_batch_pounce(
 
     # Batch-level options. Whitelist keys POUNCE understands and enforce the
     # per-node wall-time guard (issue #5) used by the serial pounce path.
-    # Seeded from the shared baseline (#945) so a batched node returns the same
-    # in-the-box point the serial ``nlp_pounce.solve_nlp`` path does — otherwise
-    # batching alone would change which side of a bound an incumbent lands on.
-    batch_opts: dict = pounce_option_defaults()
+    # Left at POUNCE's own baseline, like the serial ``nlp_pounce.solve_nlp`` path
+    # it must agree with: these are B&B node relaxations, not the reported solution,
+    # and neither shared request belongs here (#945 — see the note in nlp_pounce).
+    batch_opts: dict = {"print_level": 0}
     for k in ("max_iter", "tol", "acceptable_tol"):
         if options.get(k) is not None:
             batch_opts[k] = options[k]
--- a/python/discopt/solvers/__init__.py
+++ b/python/discopt/solvers/__init__.py
@@ -199,9 +199,13 @@ POUNCE_BOUND_RELAX_FACTOR = 0.0
 def pounce_option_defaults() -> dict:
     """discopt's baseline POUNCE options: quiet, and inside the declared box.
 
-    THE single source of truth for EVERY POUNCE entry point: the matrix-form
-    backends (``lp_pounce``, ``qp_pounce``, #940) and the NLP path
-    (``nlp_pounce.solve_nlp`` and ``solver.py``'s batch path, #945). Seed it
+    THE single source of truth for the matrix-form backends (``lp_pounce``,
+    ``qp_pounce``, #940), and available to any caller that wants discopt's
+    baseline rather than Ipopt's. It is deliberately NOT applied backend-wide on
+    the NLP path: ``constr_viol_tol = 1e-8`` costs a 31%-worse incumbent and a
+    looser bound on nvs05 entirely on its own (measured in
+    ``nlp_pounce.solve_nlp``), so ``nlp_pounce.solve_nlp`` stays at POUNCE's own
+    baseline and the NLP call sites that want this seed it themselves. Seed it
     *before* merging a caller's options so an explicit request still wins::
 
         opts = pounce_option_defaults()
--- a/python/discopt/solvers/__init__.py
+++ b/python/discopt/solvers/__init__.py
@@ -199,9 +199,10 @@ POUNCE_BOUND_RELAX_FACTOR = 0.0
 def pounce_option_defaults() -> dict:
     """discopt's baseline POUNCE options: quiet, and inside the declared box.
 
-    THE single source of truth for the **matrix-form** backends (``lp_pounce``,
-    ``qp_pounce``). Seed it *before* merging a caller's options so an explicit
-    request still wins::
+    THE single source of truth for EVERY POUNCE entry point: the matrix-form
+    backends (``lp_pounce``, ``qp_pounce``, #940) and the NLP path
+    (``nlp_pounce.solve_nlp`` and ``solver.py``'s batch path, #945). Seed it
+    *before* merging a caller's options so an explicit request still wins::
 
         opts = pounce_option_defaults()
         opts.update(caller_options or {})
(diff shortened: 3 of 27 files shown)


jkitchin:
I reviewed this in a worktree with the branch's own Rust built in, asserting
`discopt.__file__` and `hasattr(solvers, "pounce_incumbent_options")` on every
arm (CLAUDE.md §8), against the same `pounce 0.9.0` wheel you measured on.

**The core fix is real and I reproduced the false certificate it removes.** The
design call at the centre of it — split by *what the caller consumes*, not by
LP-vs-NLP — is the right one. Three things below: one measured gap in the class
coverage that I think belongs in this PR, one section of the description that no
longer matches the code that ships, and the fact that CI has never run on this
branch.

## Verified: this removes a certified false optimal

Your own `mindtpy_cq` fixture, `mip-nlp/oa`, `main` vs this branch, same fixture,
same wheel:

```
main    status=optimal  obj=2.999900002498779  bound=2.9999500008337137  gap=0.0
        -> incumbent 1.0e-04 BELOW the true optimum 3.0
        -> bound 5.0e-05 ABOVE its own incumbent, and gap still reported 0.0
PR      status=optimal  obj=2.999999999058049  bound=2.999953325774309   gap=1.556e-05
        -> super-optimality 9.4e-10, bound correctly below incumbent
```

That is a `gap == 0.0`, `status == optimal` certificate on a point that cannot
exist, on `main` today. The mechanism is exactly as you describe it: `(x-3)^2 <=
0` relaxed by `bound_relax_factor` admits every `x` within ~1e-4 of 3, and the
`1e-9` floor in `_compute_gap` then clamped the resulting negative gap to zero.
Both halves of the fix are load-bearing and the re-baselined numbers in
`test_mip_nlp.py` are the honest ones, not a loosening — `objective >= 3.0 -
1e-6` and `bound <= objective` are strictly stronger than the `gap == 0` they
replace.

I also checked the two re-baselines individually: `test_compute_gap_uses_
absolute_scale_near_zero` keeps its original input as a live assertion and
re-pins the denominator-floor property at a larger magnitude, so the property is
still pinned. Not a weakening. `bound_inversion_tolerance` mirrors
`amp._amp_abs_gap_with_bound_tolerance` (amp.py:2007–2022) term for term.

One de-risking fact the description doesn't claim and could: **for OA the new
absolute criterion cannot change termination at all.** OA runs `denom_floor=1.0`
and `gap_tolerance=1e-4` (oa.py:673), and `_certified_gap_converged` is
`gap_value <= gap_tolerance` (oa.py:5054). Since `gap = abs_gap/max(|ub|,|lb|,1)
<= abs_gap`, anything with `abs_gap <= 1e-6` already had `gap <= 1e-6 < 1e-4`.
The `1e-9 -> 1e-6` loosening is therefore confined to LOA's `denom_floor=1e-10`
sub-unit-objective regime. Worth a line — it shrinks the blast radius of the
part of this PR that reads scariest.

## Finding 1 — the default `m.solve()` path still certifies a super-optimal incumbent

Same fixture, **default** path, this branch:

```
default  status=optimal  obj=2.9999000090835057  bound=2.9999000090835057  gap=0.0
         -> 9.999e-05 below the true optimum
```

Byte-identical on `main`, so this is a pre-existing hole, not a regression from
this PR. But it is the same defect class, on the entry point users actually hit,
and it survives the PR that is titled after fixing that class.

Attributed rather than guessed — I wrapped the three candidate producers and
matched the returned point against the reported incumbent:

```
CALLS {'solve_nlp': 7}
HITS (returned the exact reported incumbent) {'solve_nlp': 2}
```

So the reported solution on the default path comes from `nlp_pounce.solve_nlp` —
the one function the PR deliberately leaves unseeded, on the grounds that it is a
multiplier consumer. On this path it is a *point* producer whose point is the
answer.

I think the nvs05 measurement that justified leaving it alone is being applied to
the wrong option. That measurement is about `constr_viol_tol` (which lives in
`pounce_option_defaults()`); `bound_relax_factor` (which lives in
`pounce_incumbent_options()`) is separable, and it alone is what fixes this.
Injecting only `bound_relax_factor = 0.0` at `solve_nlp`, changing nothing else:

```
as-is    obj=2.9999000090835057   super-opt 9.999e-05    injected_calls=0
brf0     obj=2.9999999983777133   super-opt 1.622e-09    injected_calls=7
```

Four orders of magnitude, and no cost that I can find. Five `nvs*` instances from
the in-repo corpus, both arms, 174 injected calls:

```
          as-is                              brf0
nvs05  obj 5.470934126405679   bound 5.470650  obj 5.470934126406966  bound 5.470716
nvs03  obj 16.0                bound 15.9999997  obj 16.0             bound 16.0
nvs06  obj 1.7703125           =                 obj 1.7703125        =
nvs07  obj 4.0                 =                 obj 4.0              =
nvs10  obj -310.7999999999999  =                 obj -310.79999999999 =
```

Identical objectives (nvs05 agrees to 12 digits), and nvs03/nvs05 bounds come out
slightly *tighter*. This is a 5-instance sample, not a panel — but it is enough
to say the "seeding `solve_nlp` costs a 31%-worse incumbent" conclusion doesn't
transfer to the incumbent option on its own, and that the multiplier-consumer
argument (Slater failing on `x0^2 + x1^2 <= 0`) is an argument against seeding
`bound_relax_factor` *for the Benders/OBBT dual calls specifically*, not for
every call into `solve_nlp`.

Concretely: `solve_nlp` currently can't tell its two caller kinds apart, which is
why it ends up serving neither. An explicit `for_incumbent: bool` (or having the
dual consumers pass `bound_relax_factor` back in) would let the point consumers
get `0` while the recourse/OBBT calls keep Ipopt's relaxation. Your own design
sentence already says this; the code just doesn't have the seam to express it.

If you'd rather not widen the PR, that's defensible — but then the description
shouldn't say the class is fixed, and this needs an issue with the numbers above,
because right now `m.solve()` is the path that still does it.

## Finding 2 — the description describes a different tree than the one that ships

Summary **(a)** opens with "`nlp_pounce.solve_nlp` and `solver.py`'s POUNCE batch
path were the last two entry points not seeded from
`solvers.pounce_option_defaults()`" — and the code deliberately seeds neither,
with a comment in `nlp_pounce.py` saying so in as many words. Same drift in the
docstrings of `test_940_pounce_lp_constraint_tolerance.py` ("every entry point
inherits it"), and `scratchpad/issue945/entry_bounds.py`'s §8 arm marker
(`SEEDED = "opts = pounce_option_defaults()" in inspect.getsource(nlp_pounce.
solve_nlp)`) is now permanently False on the post tree — a §6 probe that can no
longer fire.

The bigger one: the whole "**One certification loss, filed as #946**" section
describes damage this branch doesn't do.

```
tests/test_gbd.py::test_linear_objective_nonlinear_constraint  1 passed in 0.89s
```

and `grep xfail python/tests/test_gbd.py` on this branch returns nothing, so the
"pinned here as a **strict** xfail" claim has nothing behind it. That reads as
the section having been written when `solve_nlp` *was* seeded — the GBD analysis
is precisely the reasoning that led to un-seeding it, so once you un-seeded, the
loss went away and the section stopped applying. Worth checking whether #946
should be closed or restated, since it was filed on a regression that this branch
doesn't produce.

**The panel evidence itself is fine** — I checked this before writing the above,
because a differential panel measured on a tree that seeded `solve_nlp` would not
be evidence for this one. `panel_corpus.py:99` asserts
`"pounce_incumbent_options()" not in inspect.getsource(NLPP.solve_nlp)` and
patches the three real consumer sites, so the panel was run against the shipped
design. The harness is careful throughout — the pre-arm reconstruction taking two
seams, the "PANEL NEVER ENTERED solve_nlp" guard, the alternating lead arm, and
the retraction carried in `panel_replicate_analyze.py` rather than quietly
deleted. That last one is the §11 behaviour I'd want to see more of.

So: the measurements stand, the prose around them doesn't. Given this is a
bound-changing change shipping without a flag, the description is the artifact a
future reader will reconstruct the decision from, and right now it argues for
seeding that didn't happen.

## Finding 3 — no CI has run on this branch

```
$ gh pr checks 948
no checks reported on the 'claude/issue-945-7wq572' branch
```

Everything here is your local runs. The 7322-passed figure is from your machine
against the PyPI wheel; CI installs `pounce-solver` from git `main`, which you
flag yourself as a reason digits may shift — and digit shifts are exactly what
this PR re-baselines tests on. Retargeting to `main` (which #943 having merged
now allows) should also get the lanes to run.

## Smaller

* `scratchpad/issue945/*.json` are committed, but `.gitignore:98-100` is scoped
  to `scratchpad/*_out*.txt` and `scratchpad/*.log` — "keep the harnesses, drop
  the outputs". `panel_corpus.json` / `panel_replicate*.json` are outputs. Either
  rename them under the pattern or widen it; vendoring result blobs makes the
  next panel a diff conflict.
* The `1e-9` floor left in `_jax/primal_heuristics.py:1805` with a comment
  explaining it gates *skipping* a heuristic rather than certifying: agreed, and
  the comment is the right artifact. Good call not unifying it.
* Remaining unseeded point producers, for whatever issue this ends up in:
  `_jax/primal_heuristics.py` (6 sites), `amp.py:548/557/565`,
  `solver.py:11782` (`_polish_opts`), `solver.py:13852` (`recover_opts`),
  `solver.py:7836`, `solver.py:12988`, `mpec.py:346`,
  `_jax/differentiable.py` (5 sites), `gdpopt_loa.py:418`. Not all of these
  consume points, but the rule can't be checked by eye at 3 of ~20 sites.
* `test_945_nlp_box_and_gap_closure.py`'s `inspect.getsource` assertions pin the
  *shipped* design correctly (three consumers seeded, `solve_nlp` not) — so the
  tests are the accurate description of this change and the PR body is the
  inaccurate one.

## Verdict

The soundness fix is right, the false certificate it removes is real and I
reproduced it on `main`, and the panel behind it was measured properly. I'd want
Finding 1 resolved before this merges — either by giving `solve_nlp` the seam to
distinguish its two caller kinds, or by an issue carrying the attribution and the
brf0 numbers — because as it stands the default `m.solve()` path still returns a
certified `optimal` 1e-4 below the true optimum on this PR's own fixture. Finding
2 is a rewrite of two sections, not a code change. Finding 3 is a retarget.

jkitchin:
Two corrections to my review above, before the fixes (CLAUDE.md §11).

**Retracted: the `scratchpad/issue945/*.json` item under "Smaller" is wrong.**
I claimed committed panel JSON blobs cut against the gitignore convention.
They don't — tracked result blobs are the established convention in this repo:
`scratchpad/issue917/`, `scratchpad/issue940/`, `scratchpad/panel844/` and
`scratchpad/panel860/` all carry their JSON, and `.gitignore:98-102` is scoped
to `*_out*.txt`, `*.log` and `*.out`, none of which these match. I read the
gitignore as narrower than the convention instead of checking what the
convention actually is. Nothing to change here.

**Corrected: Finding 1's attribution named the wrong producer.**
I reported the reported incumbent as coming from `nlp_pounce.solve_nlp`, on
two `HIT`s at `solver.py:13853`. That call is the dual-recovery re-solve, and
it was *seeded with* the already-bad incumbent — the hit says it passed the bad
point through, not that it made it. Re-instrumented to print `x0 -> x` per
call, the real chain is two sites:

```
call #5  primal_heuristics.py:324  feasibility_pump   x0=[1.0, 1.0]  -> x=[2.9999000090, 1.0]
call #6  solver.py:13853           _solve_node_nlp_kkt x0=[3.0, 1.0]  -> x=[2.9999000090, 1.0]
```

`feasibility_pump` produces the point; `_solve_nlp_bb`'s terminal recovery
solve then *re-introduces* it — with the pump fixed to return exactly 3.0,
call #6 pushed it straight back to 2.9999 by adopting its own relaxed-box
primal. Fixing either one alone leaves the defect standing. That is why the
fix is two changes rather than the one-line `solve_nlp` seam I suggested, and
it is also why the `for_incumbent` flag I proposed for `solve_nlp` would not
have been enough: the site that needed splitting is the *caller* that wants a
point and multipliers from the same solve, not the callee.

The rest of the review stands, and the numbers in it were all re-measured
against the fix. Details in the commit.

jkitchin:
Review findings addressed. Two commits on top of `cd29a62c`.

**Finding 1 (the default `m.solve()` path still certified a super-optimal
incumbent) — fixed, not deferred.** Two producers, and fixing either alone
leaves the defect standing: `primal_heuristics.feasibility_pump` projected onto
the relaxed box, and `solver._solve_nlp_bb`'s terminal re-solve then adopted its
*own* relaxed-box primal, pushing `x` back to 2.9999 even with the pump honest.
All six option sites in `primal_heuristics` now route through one
`_heuristic_nlp_options()` helper (one helper, not six edits — a seventh
heuristic cannot silently keep Ipopt's default), and `_solve_nlp_bb` takes two
solves where it took one: a refine solve carrying the incumbent options whose
point is adopted, then the recover solve *at that adopted point*, left relaxed
because its product is the multipliers. `_polish_opts` in the batch path takes
the incumbent options for the same reason; its sibling `recover_opts` does not.

```
pre   default  obj=2.9999000090835057  gap=0.0  super-optimal by 9.999e-05
post  default  obj=2.9999999983777133  gap=0.0  super-optimal by 1.622e-09
```

The `for_incumbent` seam on `solve_nlp` I suggested would not have been enough:
the site that needed splitting is a *caller* that wants a point and multipliers
out of the same solve, so the design rule of this PR ("split by what the caller
consumes") applies one level up from where I put it.

**Finding 2 (prose describing a tree that doesn't ship)** — three stale
descriptions corrected: `pounce_option_defaults`'s docstring, which claimed to be
the source of truth for "EVERY POUNCE entry point ... and the NLP path";
`test_940_pounce_lp_constraint_tolerance`'s docstring; and
`scratchpad/issue945/entry_bounds.py`'s §8 arm marker, which tested `solve_nlp`'s
own source and so was permanently `False` on the post tree — it mislabelled every
post run `UNSEEDED`, and the harness could not run at all besides. It now mirrors
the test it claims to measure (`executed_assertions=6 violations=0`). The PR body
is rewritten around what ships.

**Finding 3 (no CI)** — the branch already targets `main`; checks are running on
this push.

**Panel re-run, because the old one no longer measured this tree.** Both new
seams are in `panel_corpus.py`'s pre-arm reconstruction:

```
COMPARISONS_EXECUTED=66      identical=30  differing=36
CERT_ORDER_CHECKS_EXECUTED=103
ORACLE_CHECKS_EXECUTED=32    soundness_findings=0
CERT_REGRESSIONS=0
SOLVE_NLP_CALLS pre=4037 post=4052
```

And a witness for it, since "0 findings" is a threshold not tripped rather than
evidence of the thing the fix claims. `panel_superopt_recheck.py` re-scores both
arms against a second, independent oracle (MINLPLib `minlplib.solu`, `=opt=` rows
only) at `1e-12` relative:

```
COMPARED=50
SUPER_OPTIMAL_INSTANCES pre=24  post=12
WRONG_WAY=0
```

24 of 50 corpus instances reported an incumbent strictly better than the
published global optimum before this branch; 12 do after, 12 flip to sound, and
zero move the other way. The 12 that remain are named in the body rather than
summarised — they are either byte-identical in both arms (other producers) or
reduced, and they are the honest remaining surface, listed there with the other
unseeded sites (`amp.py`, `mpec.py`, `_jax/differentiable.py`, `gdpopt_loa.py`).
Those are left alone deliberately: none reproduces the defect on this fixture, so
changing them here would be a fix without a measurement.

Also re-run after the change: `pytest -m smoke` → 920 passed, 1 skipped, 2
xpassed; `pytest -m slow test_adversarial_recent_fixes.py` → 10 passed; the five
directly-affected test files → 247 passed. No Rust touched.

jkitchin:
CI note: the first `Python fast` run failed on

```
FAILED python/tests/test_gdpopt_heuristics_core_units.py::TestLocalBranchingBudget::test_deadline_expiring_mid_round_truncates - assert 0 >= 2
```

which is not this branch. Two pieces of evidence rather than a re-run and a
shrug:

**1. The function is byte-identical to the base.** This branch's
`primal_heuristics.py` hunks end at old line 1376; `local_branching` starts at
old line ~1729. The failing assertion is `calls["n"] == 0`, i.e. the backend was
never invoked — reached only through code that never enters `subnlp`, which is
the only function in that file this change touches on this path.

**2. `assert 0 >= 2` is the signature of a scheduling stall.**
`local_branching` starts its deadline clock (`slice_deadline = perf_counter() +
submip_time_limit`) before the first `budget.exhausted()` poll, and the test
gives it `submip_time_limit=0.25`. Any stall over 250 ms in between exits the
enumeration at radius 0 with zero backend calls. Injected deterministically,
changing nothing else:

```
stall=0.00s  out=None  calls=2   test PASSES
stall=0.30s  out=None  calls=0   test FAILS -> assert 0 >= 2
```

My first hypothesis was the JAX evaluator compile sitting inside the slice; the
measurement killed it (0.1–1.2 ms over 5 reps for this 3-binary model, 0.5% of
the slice), so the stall comes from outside the process — an xdist worker
descheduled on a loaded runner.

Filed as #950 with the reproduction, rather than fixed here: the honest fix is a
clock seam so the test pins the truncation *schedule* instead of racing the
machine, which is a real change and does not belong in a correctness PR whose
`local_branching` is untouched. Widening `submip_time_limit` would make the race
rarer without removing it, and the test's subject is exactly the budget edge.

`Python correctness (fast subset)` — the lane #940 was caught by — passes on this
push.

PR #949, opened 2026-08-08
fix(correctness): #946 GBD no longer depends on a finite recourse multiplier

### LINKED ISSUE #946: GBD's optimality cut trusts an unbounded multiplier — latent, and it depends on the recourse NLP keeping Ipopt's bound relaxation

jkitchin:
## Summary

GBD's Lagrangian optimality cut was built from whatever multipliers the recourse NLP returned, with no check that they were usable. When the recourse subproblem is **degenerate at its solution** (Slater's condition fails: the feasible set has empty interior) no finite multiplier exists, the returned one diverges, and the cut — still sound — goes tight at its own first-stage point and vacuous one step away. The master's lower bound stalls and GBD exits `iteration_limit` instead of `optimal`. It stayed latent only because Ipopt's `bound_relax_factor` gives a degenerate set an artificial interior: GBD's convergence on this class depended on a recourse point that is slightly *infeasible*.

Alongside the Lagrangian cut, an all-0/1 first stage now also gets the **integer L-shaped optimality cut** (Laporte & Louveaux 1993), which never touches the multiplier:

```
eta >= anchor_z - (anchor_z - L) * Δ(x, ẑ),   Δ = Hamming distance to ẑ
```

`L` is a rigorous global lower bound on the objective over the declared box — the closed-form box minimum (`_box_min_linear`) of the convex objective's, or the Lagrangian's, subgradient underestimator, taken best-of over iterations. The cut is exact at `ẑ` and drops to `L` at every other 0/1 point, so it is valid however the multiplier behaves; it also gives the binary-master case finite convergence, which the Lagrangian cut alone does not have.

A non-binary first stage has no comparably cheap exact cut, so the honest uncertified outcome stays — but GBD now says *why* (a warning naming the degenerate recourse and the reason no multiplier-free cut was available) and stops when the master starts re-proposing a point it has already cut, instead of silently exhausting 100 iterations.

Also: the `m_y` box minimum is factored into `_box_min_linear` and reused for the floor, and the recourse classification moves from a 6-tuple to a `NamedTuple`.

Closes #946

## Re-verified after the #948 merge (rebased onto `b8b59fff`)

Everything below the "Summary" was measured against a pre-#948 base, and #948 is
not a neutral base change for this PR: its subject is the same
`bound_relax_factor` dependency #946's premise rests on, and this panel's oracle
is the monolithic `Model.solve()` #948 altered. So the branch was rebased onto
current `main` and the evidence re-taken there, with the loaded module asserted
against a marker (`_box_min_linear` absent on the baseline, present on the
branch) rather than assumed.

Fail-before / pass-after, on `main` at `b8b59fff` (not the old base):

```
python/tests/test_946_gbd_degenerate_multiplier.py   4 failed, 5 passed
  test_degenerate_recourse_still_certifies
  test_degenerate_recourse_converges_without_exhausting_the_budget
  test_nonbinary_first_stage_explains_the_lost_certificate
  test_box_min_linear_closed_form
```

Differential panel re-run end to end on the post-#948 base, both arms:

```
executed comparisons: 80
NEW unsound bounds:            0
certification regressions:     0
incumbent REGRESSED:           0
bound WEAKER after:            0
bound STRONGER after:          5   (0,18,19,28,16 — all the exact arm)
incumbent improved:            4
status changes:                5   iteration_limit -> optimal
cert-clean AND no bound weakened: True
```

Identical to the pre-#948 panel, including the seed-16 caveat: the monolithic
reference on that instance still reports a point 1.0e-08 outside a degenerate
row, so #948 did not remove that particular producer. It remains a property of
the oracle, not of this change — the comparator records the reference point's
own violation and refuses to score against it.

Test batteries on the rebased tree:

```
pytest -m smoke                                      920 passed, 1 skipped, 2 xpassed
pytest -m slow test_adversarial_recent_fixes.py       10 passed
the 7 decomposition/benders test files, -m ""        113 passed
ruff check / ruff format --check                     clean
```

(The `-m smoke` figure supersedes the 909 above, which predates #948's three new
smoke tests.)

### Review follow-up commit

`_box_min_linear` skips a column whose gradient component is below
`_STATIONARY_TOL` **before** it examines that column's bound, so on an unbounded
column it reports 0 where the exact box minimum is `-inf` — an error of up to
`tol * 1e20` running in the unsafe direction (an inflated anchor, and an
inflated `floor`, which is the `L` the new cut divides by). It is the "INF
sentinel" shape from CLAUDE.md: decide unboundedness on the bound, never on a
product with the coefficient.

Measured rather than argued: the branch takes that skip **48 times in 654
columns examined** across 17 two-stage models with unbounded or free recourse
variables, omitting a term as large as **8.9e4**. Every observed `|g_j|` is
2.2e-16…8.9e-16 (roundoff), and 16 monolithic comparisons give **0 invalid
floors and 0 unsound bounds**.

That is sound rather than lucky, for a reason the code did not state: an
unbounded column is interior at the recourse solution, so KKT stationarity makes
its exact gradient component — and therefore its exact contribution — zero. The
skip is safe *because* `_STATIONARY_TOL` (1e-9) sits below the recourse NLP's own
dual tolerance (1e-8); a component the solver did not drive to stationarity
lands above the threshold and correctly returns `finite=False`. The constant now
carries that argument, and two tests pin it:

- `test_stationary_tolerance_stays_below_the_nlp_dual_tolerance` — fails if
  anyone raises the threshold (verified: 1e-3 fails it).
- `test_lshaped_floor_is_sound_with_unbounded_recourse_variables` — covers the
  shape `test_lshaped_cut_soundness_panel` cannot, since that panel's recourse
  box is bounded and the skip cannot fire there at all. It taps every floor GBD
  registers and asserts none exceeds the monolithic optimum, asserting first
  that the tap fired. Verified to have teeth: injecting `floor += 1e6` fails
  both parametrizations, while the existing bounded panel still passes.

The production half of that commit is comments only.

### Also checked, and clear

- **New-cut coefficient magnitude.** The cut's coefficients are
  `spread = anchor_z - L`, and `L` is only rejected below `_ETA_FLOOR` (−1e12),
  so a loose-but-finite floor could in principle put 1e12-scale rows into the
  master MILP — a bound risk, not a cosmetic one. Instrumented across the seven
  decomposition test files: **287 cuts built, max spread 7.69e3, max
  spread/|anchor| 40.1**. No blow-up; noted here so the ceiling is on record.
- **Early stop on a repeated master point.** The stop fires *after* the cut for
  that iteration is added and *before* the final master re-solve and gap
  re-check, so the strengthened cut is still applied and evaluated. The only
  thing forfeited is a second round of floor-driven strengthening at an
  already-visited point, and the status stays the honest `iteration_limit`.

## Correctness

Three guards keep the new cut from producing a false bound, and all three are the same ones the existing anchor already rests on:

- **Validity at ẑ**: the right-hand side there is `anchor_z`, the *rigorous Lagrangian* bound (`anchor_z <= v_true(ẑ)`), never the NLP primal. The cut is skipped when the anchor fell back to the primal (`rigorous=False`), because `f(x̂, ŷ)` is an *upper* bound on `v_true` and would cut off the true value.
- **Validity elsewhere**: every other 0/1 point has `Δ >= 1`, so the right-hand side is `<= L <= min over the box of f <= v_true(x)`. `L` is clamped to `min(L, anchor_z)` so the coefficient can never turn negative and tilt the cut *upward* with the flip count.
- **Integrality slop**: `anchor_z = anchor + s^T(ẑ - x̂)` reads the Lagrangian cut off at the rounded point, so a master returning `x̂` within its integrality tolerance of `ẑ` (rather than exactly at it) cannot invalidate the cut. The correction is zero in the usual case.

Gated on `is_convex` and `all_binary_master`, as the anchor and `L` both rest on convexity and `Δ` is only a flip count on 0/1 variables. Skipping a cut can only lower a bound, so the non-binary path stays sound by construction.

- [x] `pytest -m smoke` → 909 passed, 15 skipped, 2 xpassed
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → 10 passed
- [x] `cargo test -p discopt-core` → N/A, Rust untouched
- [x] `ruff check` / `ruff format --check` clean (plus `mypy` clean on the changed module)

Also `pytest python/tests/test_gbd.py python/tests/test_946_* python/tests/test_benders*.py python/tests/test_decomposition_{adversarial,solve_equivalence,benchmarks}.py -m ""` → 109 passed, 4 skipped.

<sub>A first run of the adversarial file showed 2 failures (`carton7`, `large_dense_jacobian_no_crash`) while the differential panel was still running on the same box — the file took 298 s against a 178 s baseline. Both are wall-clock-deadline assertions and neither touches the decomposition code path; re-run on an idle machine, 10/10 in 192 s.</sub>

## Regression test

`python/tests/test_946_gbd_degenerate_multiplier.py` — 9 tests, of which **4 fail before this change and pass after**:

- `test_degenerate_recourse_still_certifies` — the headline: `iteration_limit`/bound −2 → `optimal`/bound −1
- `test_degenerate_recourse_converges_without_exhausting_the_budget` — 101 recourse solves → 3
- `test_nonbinary_first_stage_explains_the_lost_certificate` — the diagnostic and the early stop
- `test_box_min_linear_closed_form` — unit test of the closed form both the anchor and the floor use

Each test that exercises the degeneracy first asserts the degeneracy is actually present (`max |mu| > 1e6`), so the file cannot pass by silently failing to reproduce.

Bound-changing, and **not** behind a flag: the cut is only ever *added* to a master that was already sound, it is gated on the same convexity check the reported bound is, and its no-op path is the pre-existing behaviour (non-binary master, non-convex model, or no finite floor). Evidence is a before/after differential panel rather than a flag ON/OFF one — `scratchpad/issue946/`, full numbers in `RESULTS.md`.

**Panel**: 40 deterministic convex two-stage MINLPs (binary and 0..2 integer first stages, quadratic/linear objectives, conic and equality coupling, scale 1e−1…1e2) × 2 arms (default `bound_relax_factor`, and `= 0` — the degenerate arm). 80 comparisons executed, gate exit 0. Oracle is the monolithic `Model.solve()` optimum *with its own constraint violation recorded*, so a reference point sitting 1e−8 outside a degenerate row cannot masquerade as an unsound bound.

| | default arm | `bound_relax_factor = 0` arm |
|---|---|---|
| certified `optimal` before | 38/40 | 33/40 |
| certified `optimal` after | 38/40 | **38/40** |
| recourse NLP solves before → after | 113 → 113 | **604 → 117** |
| wall before → after | 16.2 s → 15.9 s | **117.8 s → 16.7 s** |

- **cert-clean**: 0 new unsound bounds, 0 certification regressions, 0 incumbent regressions, 0 bounds weakened. 13 of 80 runs return `|mu| > 1e6`, so the degenerate class is well covered.
- **net-positive**: 5 instances `iteration_limit` → `optimal`, 5 bounds strengthened, 4 incumbents improved — all four then matching the monolithic optimum exactly.
- The one flagged instance (seed 16, default arm, `bound = 128.7391 > incumbent = 128.7322`) is **pre-existing and identical before and after**: its incumbent violates the degenerate row by 1.0e−8, the #940/#945 dependency. Confirmed by running the comparator before-vs-before, which flags the same instance with the same numbers. The same seed in the exact arm is the fix working: post-fix GBD certifies 128.74599, the true optimum of the *exactly* feasible problem, while the monolithic reference reports 128.73218 from a point 1.0e−8 outside the row.
- **Killed hypothesis, recorded**: the first read of seed 16 was "the 1.1e11-coefficient degenerate cut rows are wrecking the master MILP's numerics". `master_tap.py` taps every master `(c, A_ub, b_ub)` and brute-forces all 2^n first-stage points against those rows by hand — the master's answer agrees with its own rows exactly. The hypothesis is false and no cut-dropping-on-degeneracy logic was written on the strength of it.

The entry experiment (`scratchpad/issue946/entry_experiment.py`) reproduces the issue's measurement table to the digit on this build before anything was changed: `mu = 9.82e7`, cut slope −7.86e8, `iteration_limit`, bound −2.

---
_Generated by [Claude Code](https://claude.ai/code/session_018aFL2J8XH4Hhmvyghgvc8C)_


jkitchin:
Reviewed. The cut is valid, the panel supports it under both §5 bars, and the
one soundness question I found is answered — but not by anything that was
written down, so I pushed the argument and two tests for it. Detail is in the
body; the short version:

**The validity argument checks out.** For a 0/1 master, `eta >= anchor_z - spread*Δ`
with `spread = max(0, anchor_z - L)`:

* `Δ = 0` (only at ẑ): RHS is `anchor_z`, and `anchor_z = anchor + s^T(ẑ - x̂)` is
  the *Lagrangian* cut evaluated at ẑ, which bounds `v_true` everywhere, not just
  at x̂. So the integrality-slop correction is rigorous, not a nudge.
* `Δ >= 1`: RHS `<= anchor_z - spread = min(L, anchor_z) <= v_true(x)`. The
  `max(0, ...)` is doing real work — when `L > anchor_z` the cut degenerates to
  `eta >= anchor_z`, which is valid because `L` then dominates it everywhere.
* Fractional points: the cut is not valid there and does not need to be. It is a
  valid inequality for the integer hull, `all_binary_master` makes every master
  column binary in the MILP, and the master's dual bound stays a bound for the
  MILP-with-cuts because no integer point with `eta >= v_true` is removed.
* Both floor candidates are valid: `v + min_box grad^T(z - x*)` needs only `f`
  convex; `l0 + min_box grad_L^T(z - x*)` needs `mu^T g <= 0` on the feasible set,
  the same weak-duality step the anchor already rests on. Running max of valid
  bounds is valid.

The `rigorous=False` skip is the right call and the one I'd have asked for
otherwise: the primal fallback anchor is an *upper* bound and would cut off the
true value at ẑ.

**The one real finding, and its answer.** `_box_min_linear` skips a column whose
gradient component is under `_STATIONARY_TOL` **before** it looks at that
column's bound. On an unbounded column that reports `0` where the exact box
minimum is `-inf`, an error up to `tol * 1e20`, and it runs the unsafe way — an
inflated `floor` shrinks `spread` and makes the cut *stronger* away from the
anchor, which is exactly how it could cut off the optimum. It is the "INF
sentinel" shape from CLAUDE.md: decide unboundedness on the bound, never on a
product with the coefficient. And it is not hypothetical — instrumented, the
branch takes that skip **48 times in 654 columns examined** across 17 two-stage
models with unbounded or free recourse, omitting a term as large as **8.9e4**.

It is nonetheless sound, and the reason was the missing piece: an unbounded
column is interior at the recourse solution, so KKT stationarity makes its exact
gradient component zero and the exact contribution zero. The skip is safe
*because* 1e-9 sits below the recourse NLP's own dual tolerance of 1e-8. Measured
consequence: 0 invalid floors and 0 unsound bounds over 16 monolithic
comparisons; every observed `|g_j|` is 2.2e-16…8.9e-16. So the fix is not to
change the number, it is to make the number load-bearing on purpose —
`test_stationary_tolerance_stays_below_the_nlp_dual_tolerance` now fails if
anyone raises it, and
`test_lshaped_floor_is_sound_with_unbounded_recourse_variables` covers the shape
`test_lshaped_cut_soundness_panel` structurally cannot (its recourse box is
bounded, so the skip never fires there). Both verified to have teeth: `1e-3`
fails the first, `floor += 1e6` fails the second in both parametrizations while
the existing bounded panel still passes.

**Rebased onto `main` (`b8b59fff`) and re-measured, because the base was not
neutral.** #948 is the same `bound_relax_factor` dependency #946's premise rests
on, and it changed the monolithic `Model.solve()` this panel uses as its oracle.
Green CI and an 80-comparison panel taken against the old base are not evidence
about the merged tree. Re-taken with the loaded module asserted by marker: the
4 fail-before tests still fail on `b8b59fff`, the panel reproduces exactly
(80 comparisons, 0 new unsound, 0 cert regressions, 0 incumbents regressed,
0 bounds weakened, 5 `iteration_limit` -> `optimal`), smoke is 920/1 skipped/2
xpassed, adversarial 10 passed, the 7 decomposition files 113 passed.

Worth recording since it is a residual of the PR I just merged: seed 16's
monolithic reference **still** returns a point 1.0e-08 outside a degenerate row
post-#948. #948 removed two producers of that class, not all of them, and this
is one it did not reach. It is a property of this panel's oracle rather than of
this change — the comparator records the reference point's own violation and
refuses to score against it — but it should not be filed away as fixed.

**Two things checked and clear, on record so the ceilings are known:** the new
cut's coefficient magnitude (`spread`, admissible up to ~1e12 before `_ETA_FLOOR`
rejects the floor) measures **max 7.69e3, max spread/|anchor| 40.1 over 287 cuts
built**, so no master-MILP conditioning problem; and the repeated-master-point
early stop fires *after* that iteration's cut is added and *before* the final
master re-solve and gap re-check, so the strengthened cut is still applied — the
only thing forfeited is a second round of floor-driven strengthening at an
already-visited point, and the status stays the honest `iteration_limit`.

Approving. Merging once the checks on the rebase come back green.

jkitchin:
**Retraction of the residual finding in my review above (CLAUDE.md §11).**

I wrote in the review comment and in this PR's body:

> seed 16's monolithic reference **still** returns a point 1.0e-08 outside a
> degenerate row post-#948. #948 removed two producers of that class, not all of
> them, and this is one it did not reach.

Re-measured on `main` @ `8bfce1f1` and that framing is wrong on three counts:

1. It is not seed-specific — **all 40** seeds of that family violate
   (1.9e-08 – 4.9e-08), not seed 16.
2. It is not a `#948` / `bound_relax_factor` residual. Forcing
   `bound_relax_factor = 0` gives results **bit-identical** to the stock arm.
   The reason is my instrument: I patched `nlp_pounce.solve_nlp`, and a call
   counter reports **`NLP_POUNCE_SOLVE_NLP_CALLS=0`** — this family is a convex
   MIQP and never reaches the NLP path. My "exact" and "relaxed" arms were the
   same code path, which is exactly why they agreed to every digit. I should
   have counted invocations before reporting an arm (§6).
3. It is not the NLP's `constr_viol_tol` either — sweeping it over
   1e-8 / 1e-10 / 1e-12 moves the number by zero, same reason.

The real mechanism is unrelated to #946/#948: on the convex-MIQP B&B path the
only feasibility gate an incumbent passes is an **equality residual check at
`tol=1e-4`**, with no inequality-row or bound verification before `optimal` is
returned with `bound = objective`. Filed with the measurements as #952.

Nothing about this PR's merged change is affected — the L-shaped cut, its floor,
and the panel results stand. What is retracted is only my attribution of the
oracle's residual.

Issue #912, opened 2026-07-31
The solver is nondeterministic: node counts depend on machine speed, so "same model + same time_limit" does not imply the same tree


jkitchin:
## The finding

Given identical input and an identical `time_limit`, discopt does **not** produce the same search tree. The amount of work it does is a function of how fast the machine is, because the search reads a wall clock to make branch-gating decisions at **78 sites in `python/discopt/` and 57 `Instant::now` in `crates/`**.

This is a property of the solver, not a test-harness artifact. It was filed initially as "the Regime-N panel is flaky, Phase 0 substrate work" — that framing is **wrong and retracted**.

## Evidence (measured)

The decisive case is `gear2`. The root primal heuristic gets `time_budget = min(5.0, 0.15 * time_limit)` (`solver/__init__.py`, the `integer_local_search` call) and its descent runs `while improved and time.perf_counter() < deadline` (`_jax/primal_heuristics.py`). It **never converges** — it consumed 5.02 s of a 5.00 s budget — so the default sits exactly on a cliff:

| forced heuristic budget | gear2 node_count |
|---|---|
| 5.0 s | **3** |
| 4.0 s | {3, 91} — unstable |
| 3.0 / 2.0 / 1.0 s | 91 |
| 0.5 s | 93 |

12 executed comparisons, 0 errors.

Generalisation sweep over all 85 comparable baseline rows, scaling the effective clock (255 comparisons): **alpha=1.0 control reproduces the frozen baseline 85/85, 0 mismatches**; 1.25 → 0 rows moved; 1.5 → 0; 2.0 → **1** (`gear2`, 3 → 91).

Under real contention (24 spinners on 4 cores, 8 observations): `ex1266` goes from `optimal` [6005, 6005] idle to **`time_limit` [1, 7]** loaded; `gear2` from [3, 3] to [91, 93] — and not self-consistent between replicates.

Two distinct mechanisms, not one: `gear2` is the root heuristic's own wall budget; `ex1266` is whole-solve budget starvation (its root takes 0.03 s — not a preprocessing phase at all).

## Reproduction

```bash
python -u discopt_benchmarks/scripts/item15_root_budget_probe.py --phase ils --forcebudget
python -u discopt_benchmarks/scripts/item15_root_budget_probe.py --clockscale
python -u discopt_benchmarks/scripts/item15_root_budget_probe.py --load
```

Every child asserts `discopt.__file__` and the compiled `.so` marker, and raises rather than swallowing.

[...]

PR #951, opened 2026-08-08
test(#950): drive the primal-heuristic deadlines from an injected clock

### LINKED ISSUE #950: flaky: test_deadline_expiring_mid_round_truncates fails as 'assert 0 >= 2' when the worker stalls past the 0.25 s slice
`TestLocalBranchingBudget::test_deadline_expiring_mid_round_truncates`
(`python/tests/test_gdpopt_heuristics_core_units.py:966`) fails intermittently
on the parallel CI lane with

```
FAILED ... ::test_deadline_expiring_mid_round_truncates - assert 0 >= 2
```

Seen on the `Python fast (3.12, ubuntu)` lane of PR #948
([run](https://github.com/jkitchin/discopt/actions/runs/31260844704/job/93111332800));
it passes on re-run and passes serially on the same tree.

## Mechanism — a scheduling stall, not a solver behaviour

`local_branching` starts its deadline clock

```python
slice_deadline = time.perf_counter() + max(0.0, float(submip_time_limit))
```

*before* the first `budget.exhausted()` poll, and no sub-NLP may start once the
deadline has passed. The test gives it `submip_time_limit=0.25`. So any stall
longer than 250 ms between those two points — an xdist worker descheduled on a
loaded runner is the obvious candidate — makes the enumeration exit at radius 0
having called the backend **zero** times. `calls["n"] == 0` is the only symptom,
and it is indistinguishable from "the truncation logic is broken".

Reproduced deterministically by injecting the stall and changing nothing else:

```
stall=0.00s  out=None  calls=2   test PASSES
stall=0.30s  out=None  calls=0   test FAILS -> assert 0 >= 2
```

(probe: monkeypatch `ph.cached_evaluator` to `time.sleep(stall)` before
delegating, then call `local_branching(m, np.zeros(3), k=1,
backend=slow_backend, submip_time_limit=0.25)`.)

It is not the evaluator compile itself — that measures 0.1–1.2 ms for this
3-binary model over 5 reps, i.e. 0.5% of the slice. The compile was the first
hypothesis and the measurement killed it; the stall has to come from outside the
process.

## Why it is a test problem, not a `local_branching` problem

The production behaviour under a stall is *correct*: the whole point of #912's
budget enforcement is that no sub-NLP starts past the deadline, and a 250 ms
stall genuinely leaves no budget. The test asserts a wall-clock-dependent
consequence (`calls["n"] >= 2`) of a wall-clock-independent intent ("radius 0
ran, radius 1 started and was cut off").

## Suggested fix

Drive the deadline from an injected clock rather than the real one, so the test
pins the truncation *schedule* deterministically instead of racing the machine.
That needs a clock seam in `local_branching` (or a monkeypatched
`perf_counter` scoped to the module), so it is a real change rather than a
tolerance bump — which is why it is filed rather than slipped into #948, whose
`local_branching` is byte-identical to its base.

Do **not** fix this by widening `submip_time_limit`: that makes the race less
likely without removing it, and the test's subject is precisely what happens at
the budget edge.

jkitchin:
## Summary

`TestLocalBranchingBudget::test_deadline_expiring_mid_round_truncates` failed
intermittently on the parallel CI lane as `assert 0 >= 2`. The mechanism is
scheduling, not solver behaviour: `local_branching` stamps
`slice_deadline = perf_counter() + submip_time_limit` **before** its first budget
poll, and since #912 no sub-NLP may start past that deadline — so any stall
longer than the test's 250 ms slice (an xdist worker descheduled on a loaded
runner) correctly retires the whole budget before sub-NLP #1. The test asserted a
wall-clock-*dependent* consequence of a wall-clock-*independent* intent, so that
outcome was indistinguishable from broken truncation logic.

Entry experiment (`probe950_repro.py`), stall injected and nothing else changed:

```
stall=0.00s  out=None  calls=2   old test PASSES
stall=0.30s  out=None  calls=0   old test FAILS -> assert 0 >= 2
```

The fix is the clock seam the issue asks for, applied to the class rather than
the one call site:

- `primal_heuristics._now()` is now the module's single wall-clock read. Every
  deadline stamp and poll in the layer goes through it (`feasibility_pump`,
  `continuous_multistart`, `integer_local_search`, `integer_box_search`,
  `diving`, `local_branching`, `one_hot_swap_search`).
- `WorkBudget(clock=...)` (default `time.perf_counter`) so the budget's deadline
  backstop reads the same clock; every construction in that module passes `_now`.
  Production passes nothing else — behaviour is unchanged.
- The test advances a clock its own backend owns and asserts an **exact** count
  (`calls == 2`): radius 0 whole, radius 1 cut after its first sub-NLP.
  `submip_time_limit` was **not** widened — that makes the race rarer without
  removing it, and the budget edge is precisely the test's subject.

Two source-scanning tests needed matching repairs; neither was weakened:

- `test_912_wall_budget_inventory.py` — the #912 ratchet's regex now sees through
  `_now()`, so routing a new gate through the seam cannot leave the inventory
  silently. Same three primal-heuristic gates, same categories; only the recorded
  spelling of the clock read changed.
- `test_945_nlp_box_and_gap_closure.py` — its "no heuristic sets `max_iter`
  inline" check sliced the module source at the second `def`, silently assuming
  the options helper was the module's first function; adding `_now` above it
  moved the cut into the helper's own body, so it failed on the helper's *own*
  line. Now anchored on the helper's source, which states the rule directly and
  also covers functions defined above it — strictly stronger, and verified to
  bite on injected duplicates both above and below the helper.

Closes #950

## Correctness

No solver math changed. `_now()` returns `time.perf_counter()` and every
production call site passes it (or nothing, keeping the default), so every
deadline is stamped and polled against the same clock it was before; the diff in
`local_branching` and its three siblings is the spelling of the clock read plus
one keyword argument. Nothing about bounds, cuts, relaxations, incumbent
acceptance or certification is touched.

- [x] Cannot produce a false certificate (N/A — no solver-math change)
      (the two source-scanning checks were repaired to be position-independent;
      the #912 inventory scanner was *widened* to also match the seam)

- [x] `pytest -m smoke` → **916 passed, 16 skipped, 2 xpassed** (413 s)
- [x] `ruff check` / `ruff format --check` clean (pinned 0.14.6, whole `python/` tree)

Also: `test_912_work_budget.py`, `test_912_wall_budget_inventory.py`,
`test_gdpopt_heuristics_core_units.py`, `test_945_nlp_box_and_gap_closure.py`,
`test_940_pounce_lp_constraint_tolerance.py`, `test_875_nbt_call_site_budget.py`,
`test_875_root_setup_budget.py`, `test_deadline.py`, `test_issue_267_lns.py`,
`test_f4_root_budget_gate.py` — all pass.

One unrelated observation: `test_large_dense_jacobian_no_crash` failed once in a
full-file run (`wall < budget + 40`, a load-sensitive *contract* assertion — the
whole file took 277 s that round vs 196 s the next) and passed on re-run and in
isolation. Interleaved A/B of that test, seam present vs absent (marker asserted
per arm): **33.2 / 33.9 s with, 32.3 / 32.3 s without** — inside noise, both arms
passing. It is a pre-existing load-sensitive assertion, not this change, and not
fixable by the same seam since its subject *is* wall time.

## Regression test

`test_deadline_expiring_mid_round_truncates` is the test being repaired, so
"fails before / passes after" is demonstrated on the failure condition rather
than on the tree: `probe950_immunity.py` runs the old and new test bodies under
0.0 s and 0.5 s of injected real stall (twice the slice — the condition that
produced the CI failure):

```
old   calls=2 / calls=0    (flakes, exactly as reported)
new   calls=2 / calls=2    (immune)
seam  calls=4 / calls=4    (clock frozen: radius 1 runs all C(3,1)=3 flips)
```

The third arm is the probe-fired check (CLAUDE.md §6): with the injected clock
frozen the enumeration runs the full round, which is what proves the injected
clock — and not something else — is what fires the deadline gate in the `new`
arm. `test_deadline_is_measured_against_the_injected_clock` covers the new
`WorkBudget(clock=...)` parameter directly.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DsbGpdfWQHXUpo9TkjmYHi)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -189,6 +189,34 @@ The release procedure that produces these entries is documented in
 
 ### Fixed
 
+- **A clock seam for the primal-heuristic deadlines, ending a CI flake**
+  (`test`, #950). `TestLocalBranchingBudget::test_deadline_expiring_mid_round_truncates`
+  failed intermittently on the parallel lane as `assert 0 >= 2`. The mechanism is
+  scheduling, not solver behaviour: `local_branching` stamps
+  `slice_deadline = perf_counter() + submip_time_limit` *before* its first budget
+  poll, and since #912 no sub-NLP may start past that deadline — so any stall
+  longer than the test's 250 ms slice (an xdist worker descheduled on a loaded
+  runner) correctly retires the budget before sub-NLP #1, and the test's
+  wall-clock-dependent assertion could not tell that apart from broken truncation
+  logic. Reproduced deterministically by injecting a 0.30 s stall and changing
+  nothing else: `calls` 2 → 0.
+
+  `primal_heuristics._now()` is now the module's single clock read — every
+  heuristic's deadline stamp and poll goes through it, and every `WorkBudget`
+  built there is handed it via the new `WorkBudget(clock=...)` parameter (default
+  `time.perf_counter`; production passes nothing else). The test drives the
+  deadline from a clock the backend advances, so it pins the truncation
+  *schedule* — radius 0 whole, radius 1 cut after its first sub-NLP — as an exact
+  count rather than racing the machine. Verified by re-running the old and new
+  test bodies under a 0.5 s injected stall: old `calls=0` (the reported failure),
+  new `calls=2` with and without the stall; with the injected clock frozen the
+  enumeration runs all `C(3,1)=3` flips (`calls=4`), which is what proves the
+  seam — and not something else — is what fires the gate. `submip_time_limit`
+  was **not** widened: that would have made the race rarer without removing it,
+  and the budget edge is the test's subject. Production behaviour is unchanged;
+  the #912 wall-budget inventory scanner was taught to see through the seam so a
+  new gate cannot leave the ratchet by routing through `_now()`.
+
 - **A non-zero `Constraint.rhs` was a silent wrong answer through the public API**
   (`fix(correctness)`, #909). `Constraint` is exported in
   `discopt.modeling.__all__`, and `m.subject_to(Constraint(w, ">=", 5.0))` solved
--- a/python/discopt/_jax/primal_heuristics.py
+++ b/python/discopt/_jax/primal_heuristics.py
@@ -31,6 +31,34 @@ from discopt.solvers import NLPResult, SolveStatus, pounce_incumbent_options
 
 logger = logging.getLogger(__name__)
 
+
+def _now() -> float:
+    """The single point where this module reads the wall clock (#950).
+
+    Every deadline in these heuristics — the caller's absolute ``deadline`` and
+    each heuristic's own per-call slice — is stamped and polled through this
+    function, and every :class:`WorkBudget` built here is handed it as its
+    ``clock``. In production it is :func:`time.perf_counter` and nothing else;
+    the indirection is a *seam*, so a test of deadline-edge behaviour can pin
+    the schedule it means to test instead of racing the machine.
+
+    Why it is needed: a heuristic that stamps ``deadline = perf_counter() +
+    slice`` and then polls it decides how much of the search runs from wall
+    time, so a stall *outside* the process — an xdist worker descheduled on a
+    loaded CI runner — changes what the code under test does. That is exactly
+    how ``test_deadline_expiring_mid_round_truncates`` failed as
+    ``assert 0 >= 2``: a 300 ms stall inside local branching's 250 ms slice
+    retired the whole budget before the first sub-NLP, which is *correct*
+    production behaviour (#912: never start a solve past the deadline) and
+    indistinguishable from broken truncation logic. Tests monkeypatch this
+    symbol; production code must never pass a clock of its own. A caller's
+    absolute ``deadline`` argument is measured against this clock, so a test
+    that replaces it must supply any ``deadline`` in the same time base (or,
+    like the test above, none at all).
+    """
+    return time.perf_counter()
+
+
 # Iteration cap for the *sub-NLP* solves inside the primal heuristics (issue #268).
 # These solves only need an approximately feasible point (the heuristic then checks
 # integrality + constraints and lets B&B certify); they do NOT need a tight optimum.
@@ -314,7 +342,7 @@ def feasibility_pump(
         # Always run the first round (a feasible incumbent is the primary goal,
         # worth a small overrun); only the *extra* perturbation rounds are
         # deadline-gated so the pump cannot loop well past a tight ``time_limit``.
-        if deadline is not None and round_idx > 0 and time.perf_counter() >= deadline:
+        if deadline is not None and round_idx > 0 and _now() >= deadline:
             break
         x_try = x_nlp.copy()
 
@@ -629,7 +657,7 @@ def continuous_multistart(
     best_x: Optional[np.ndarray] = None
 
     for i in range(n_starts):
-        remaining = np.inf if deadline is None else deadline - time.perf_counter()
+        remaining = np.inf if deadline is None else deadline - _now()
         if remaining <= 0.05:
             break
         solve_opts = dict(opts)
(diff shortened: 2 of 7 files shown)


jkitchin:
## Review — approve

Rebased onto `main` at `6ece3739` (the #949 merge; this branch forked at `b8b59fff` and
was missing it). Clean rebase, no conflict — #949 only touched `gbd.py` and
`test_946_*`. New head `23b58ed3`.

Everything below was re-derived independently on the rebased tree, not read off the
PR body. §8: every run asserted `discopt.__file__` inside the worktree **and** the
`def _now()` marker present.

### The seam is complete, and that is measured rather than asserted

`grep` for a clock read in `primal_heuristics.py` leaves exactly one executable hit —
line 59, `return time.perf_counter()`, inside `_now()`. Every other match is prose.
All 8 `WorkBudget(` construction sites repo-wide are in that module and all pass
`clock=_now`; no other module constructs one, and `_work_budget.py` keeps
`time.perf_counter` as the default, so no other caller changes behaviour.

Independent repro of the fail-before/pass-after (`probe951_immunity.py`), same
production code in both arms, a 0.30 s stall injected into the process clock *after*
the slice is stamped and before the first budget poll — the exact shape of a
descheduled worker:

```
ARM A  old wiring (heuristic reads the process clock): calls=0 reads=3
ARM B  new wiring (ph._now injected, same injector):   calls=2 reads=0
ARM C  stall consumed:  arm A=True  arm B=False
CHECKS_EXECUTED=3   FAILURES=0
```

Arm A reproduces the reported `assert 0 >= 2` on demand. Arm B is the stronger
result and is not in the PR body: `reads=0` — with `_now` injected the path makes
**zero** process-clock reads, so the armed injector is never even consumed. That is
the immunity claim proved directly rather than by outcome. (First cut of this probe
fired the stall on the *first* read, which is the `slice_deadline` stamp itself; the
offset then shifted the deadline with it and cancelled out, `calls=4`. Firing on read
2 is what makes arm A a real stall.)

20 consecutive runs of `TestLocalBranchingBudget` under deliberate load (8 spinners,
load average 10.2): **20 passed, 0 failed**.

### The two "repaired" scanners are stronger, verified by injection

§1 says never weaken a guard to make a check pass, so I did not take this on the
body's word — I injected the violations each guard exists to catch:

| injection | expected | result |
|---|---|---|
| unregistered `_now() + …` wall gate | #912 inventory fails | **1 failed** |
| inline `setdefault("max_iter", …)` **above** the helper | #945 check fails | **1 failed** |
| inline `setdefault("max_iter", …)` **below** the helper | #945 check fails | **1 failed** |
| tree restored (`git status` clean) | all pass | **5 passed** |

The above-the-helper case is the one the old positional slice (`src.split("def ", 2)[2]`)
structurally could not see, so the #945 repair is a strict widening, not a
relaxation. Same for the inventory regex — `_CLOCK` is a superset of the old
pattern, and the three `KNOWN` entries changed spelling only, same count, same
categories.

### Correctness

No solver math. `_now()` *is* `time.perf_counter`, every production site passes it or
the default, so every deadline is stamped and polled against the same clock as
before. One risk the docstring names and I checked: a monkeypatched `_now` would
desynchronise a caller-supplied absolute `deadline` measured on the real clock —
repo-wide there is exactly one `_now` patch site (the repaired test) and it passes no
`deadline`. `test_deadline.py:567`'s real-clock `deadline=perf_counter() - 1.0` does
not patch `_now`, so it stays consistent.

### Local verification on the rebased tree

- `pytest -m smoke` → **920 passed, 1 skipped, 2 xpassed** (111 s)
- `pytest -m slow test_adversarial_recent_fixes.py` → **10 passed** (145 s)
- `test_912_work_budget.py`, `test_912_wall_budget_inventory.py`,
  `test_945_nlp_box_and_gap_closure.py`, `test_gdpopt_heuristics_core_units.py`
  → **147 passed, 2 xpassed**
- `ruff check` / `ruff format --check` clean

One reconciliation: the body reports smoke as 916 passed / 16 skipped, I measure
920 / 1 / 2 here. The 15-skip gap is optional dependencies in that environment, not
anything this change does — same delta shows on the base tree.

`Closes #950` is the only closing keyword and is not near a negation.

Issue #933, opened 2026-08-05
Anytime dual bound: 27% of instances report no bound; seed the root and unify reporting across the 5 solve paths


jkitchin:
## Summary

27% of MINLPLib instances report **no dual bound at all** at `time_limit=8`, and
the mechanism is architectural: the reported dual bound is re-derived
independently in **five** solve exit paths, only **one** of which has any
recovery when the tree bound is unusable. #930 (PR #931) hardens that one path;
it cannot help the other four.

This is the tracking issue for the real fix that #930 is a stopgap for
(CLAUDE.md §3: "a workaround ships only with an issue tracking the real fix and
a comment linking it").

## Measurement 1 — the size of the gap

200 MINLPLib instances with an `=opt=` oracle, `time_limit=8`, 3 workers,
current `fix/930-root-bound-adoption` (marker asserted present):

| | |
|---|---|
| instances solved | 191 (9 worker errors) |
| **reported `bound=None`** | **51 / 191 = 27%** |
| admitted probe bounds checked vs oracle | 17 |
| reported bounds checked vs oracle | 140 |
| invalid (bound past the oracle) | **0** |

Soundness is clean. Coverage is not: on more than a quarter of the corpus the
solver returns no global bound whatsoever after spending its full budget.

## Measurement 2 — why (the entry experiment)

Hypothesis: the missing bound is the tree's `global_lower_bound` sitting at
`-inf` because the root LP bound was proved but never installed into node 0 —
*not* a finite tree bound rejected as tainted.

Kill criterion: predominantly finite-but-tainted would make root seeding a no-op.

12 instances at `time_limit=8`, instrumented at the discard point:

```
hda        glb=-inf valid=False rr_left=0.237  final=-64473.44   ROOT NEVER SEEDED
contvar    glb=-inf valid=False rr_left=1.455  final=171244.81   ROOT NEVER SEEDED

# diagnosed discards        : 2
#   root never seeded (-inf): 2
#   finite but tainted      : 0     <-- kill criterion NOT met
```

Hypothesis supported, 2/2, zero counterexamples — but the honest denominator is
**2 of the 2 instances that reach `solve_model`'s recovery block**, not 2 of 12.

The other 10 are the more important result. They never reach that block at all:

```
fo8_ar5_1              paths=['_solve_nlp_bb']               final=None
batchs121208m          paths=['_solve_nlp_bb']               final=None
portfol_robust100_09   paths=['_try_native_spatial_kernel']  final=None
[...]

Issue #581, opened 2026-07-10
Flag graduation backlog — default-OFF bound-changing feature flags


jkitchin:
Tracking issue: the set of **default-OFF, bound-changing feature flags** and their graduation status, so none is lost and none is re-worked. Each needs a graduation gate: a broad oracle-cross-checked cert sweep (`incorrect_count == 0`, no certificate loss, in-tolerance objective drift — CLAUDE.md §5) **plus** a perf measurement proving it helps broadly (sound ≠ net-positive — the `DISCOPT_CUT_INHERIT` lesson).

> **Reconciled against `main` 2026-07-15** (post uniform-engine cutover #636). Liveness audit: **all 7 flags are still live** — none was removed by #636 (which replaced relaxation *building*, not the LP-kernel / per-node-separation / reduction / branching subsystems these flags govern). Corrections below: `DISCOPT_OBJ_BRANCH_PRIORITY` is already **default-ON in code** (`solver_tuning.py:374`) → moved to **GRADUATED**. Harness/registry alignment noted under the table. See the reconciliation comment for the full per-flag liveness map + resolution plan.

## Status

| flag | what | status |
|---|---|---|
| `DISCOPT_LU_DENSITY_ROUTE` | density-aware dense/sparse LU route (#557 storm fix, #573) | **PENDING gate** — live at `linsolve.rs:202`; **not yet wired into `graduation_gate.py`** (needs an ARM) |
| `DISCOPT_SQUARE_COST_GATE` | cost-gate on the per-node `x**2` tangent-separation loop | **PENDING gate** — live at `mccormick_lp.py:1784`; **not in the harness** (distinct from the harness's `psd_cost_gate`; needs an ARM). THRU-3 measured this loop as the dominant per-node cost on nvs19/nvs24 |
| `DISCOPT_NODE_REDUCE` | per-node reduction (cutoff-FBBT + DBBT-from-RC + integer RC-fixing) → tighter child boxes | **PENDING gate** (node-count reducer) — live at `solver.py:1192`; **in harness** (`node_reduce`) |
| `DISCOPT_ROOT_FIXPOINT` | root branch-and-reduce fixpoint (FBBT/OBBT/DBBT-with-cutoff) | **PENDING gate** (root-bound strengthener) — live at `solver.py:8256`; **in harness** (`root_fixpoint`) |
[...]

Issue #517, opened 2026-07-06
Relaxation-coverage: lift log(<general expr>)/x**expr to give hda-class flowsheets a first dual bound


jkitchin:
## Relaxation-coverage gap: `log(<general expr>)` / `x**expr` rows omitted (no dual bound on hda-class flowsheets)

**Context:** surfaced by R4 (zero-spanning-factor lift, PR to follow). R4's entry
experiment confirmed hda is **out of R4's scope** — it tags no zero-spanning
product factor. hda's stall is a different root cause.

**Problem.** On hda (722 vars) and the same flowsheet class (heatexch_gen3,
contvar), the MILP/McCormick relaxation **omits 23 constraints** with the message
"cannot linearize `log(<general expr>)` / `x**expr`", so `best_bound` stays at the
`1e30` sentinel for the whole solve — **no dual bound exists at all** and nothing
can fathom (see `docs/dev/bottleneck-profile-2026-07-05.md` §3 row hda, B10).

**Why R4 does not cover it.** The factorable reform's transcendental-argument lift
(`factorable_reform._LIFTABLE_CALL_OUTER = {"sqrt", "exp"}`) deliberately
**excludes `log`**: lifting `t == g(x)` then relaxing `log(t)` needs a *certified
strictly-positive* lower bound on `t`, which interval arithmetic does not always
supply (a log over an interval touching 0 has no finite convex underestimator).
`x**expr` (variable exponent) is likewise unlifted. Forcing either into R4 would
risk an unsound envelope, so per the R4 task it is filed here instead.

**Proposed fix (separate work).** Extend the argument lift to `log`/`log`-family
**guarded by a certified positive lower bound** on the lifted argument (FBBT /
OBBT must prove `t_lb > 0`; abstain soundly otherwise), and handle `x**expr` via
the `exp(expr·log x)` signomial path when `x_lb > 0`. Acceptance: hda reports a
**finite** dual bound (its first), with differential-bound + feasible-point gates
per the plan §0.1; no panel regression.

**References:** `docs/dev/uncertified-tail-plan-2026-07-06.md` §3 R4 "Note on
hda" + the R4 RESULTS block; `docs/design/relaxation-catalog.md` §3;
`docs/dev/bottleneck-profile-2026-07-05.md` §3 (hda row), §2 B10.

jkitchin:
## TD-B outcome: `log(<expr>)` call-argument lift KILLED by its entry experiment — not the lever for hda

[...]

PR #953, opened 2026-08-08
fix(ci): repair three signals that had stopped signalling

### LINKED ISSUE #927: correctness: ex1252 returns optimal with a dual bound 68% above the true optimum under the #707 reform flags
## Summary

With the three `#707`/`#721`/`#732` graduation flags on, `ex1252` returns
`status='optimal'` with a **dual bound 68% above the true optimum**:

```
status = optimal
objective = 216826.51794867235
bound     = 216826.5179486723
nodes     = 1191
true optimum (test constant _EX1252_OPT) = 128893.74
```

A valid dual bound for a minimization must satisfy `bound <= optimum`. This one
exceeds it by **87,932**, and the solve terminates claiming optimality with
`gap = 2.7e-16`. That is a wrong certificate, not a weak bound.

`python/tests/test_integer_multilinear_reform.py::test_ex1252_multilinear_reform_sound_and_lifts_bound`
already asserts exactly this and **is currently failing on `main`**:

```
AssertionError: ex1252 multilinear-reform UNSOUND dual bound 216826.5179486723 > optimum 128893.74
```

## Reproduction

```python
import os
for k in ("DISCOPT_INTEGER_MULTILINEAR_REFORM",
          "DISCOPT_MULTILINEAR_COUPLING_RLT",
          "DISCOPT_DISJUNCTIVE_CONFIG_BOUND"):
    os.environ[k] = "1"

import discopt.modeling as dm
m = dm.from_nl("python/tests/data/minlplib/ex1252.nl")
r = m.solve(time_limit=200)
print(r.status, r.objective, r.bound, r.node_count)
# optimal 216826.51794867235 216826.5179486723 1191
```

Or: `pytest python/tests/test_integer_multilinear_reform.py -m correctness`

## Confirmed pre-existing on `main`

Found while running the CLAUDE.md §5 panel for #75 (PR #922). Ruled out as a #75
regression by running it on unmodified `origin/main` (`de66670c`) in the main
working tree, with a marker assertion that `discopt._tape_nlp_evaluator` is
**absent** so the load path is provably pre-change:

| tree | status | objective | bound | nodes |
|---|---|---|---|---|
| `main` `de66670c` | optimal | 216826.51794867235 | 216826.5179486723 | 1191 |
| #922 branch, `DISCOPT_NLP_EVAL=jax DISCOPT_SEPGRAD=jax` | optimal | 216826.51794867235 | 216826.5179486723 | 1191 |
| #922 branch, tape backends | optimal | 216826.51794867235 | 216826.5179486723 | 1191 |

Byte-identical on all three, so the defect is independent of the evaluator
backend.

*(A first attempt used a `git worktree`, which silently mixed trees: the editable
install leaves `discopt.__path__` spanning BOTH the worktree and the main tree,
so `discopt._tape_nlp_evaluator` resolved out of the main tree. The marker
assertion caught it — this is the CLAUDE.md §8 incident again, worth knowing
about for anyone A/B-ing this repo.)*

## Scope

All three flags are documented **default OFF**
(`solver_tuning.py:343`, `:828`; `_jax/integer_product_reform.py:169`), so a
default solve is unaffected. The exposure is to anyone opting into the #707
reform stack — and the failure mode there is the worst available: a confidently
wrong `optimal`.

## Notes for whoever picks this up

* The test's own docstring says the reform is budget-aware and only adopted at
  `time_limit >= 180` s, and that below that the flag-on solve is byte-identical
  to flag-off. The failing run uses `time_limit=200`, so it is in the adopting
  regime; a natural first bisection is which of the three flags is required to
  produce the unsound bound (the test turns on all three at once).
* `nodes=1191` and `gap=2.7e-16` mean it is not a truncation artifact — the
  search believes it has closed.

---

jkitchin:
## Why

While assessing #927–#933 I checked what *would* have caught #927 (an unsound
`ex1252` dual bound, asserted by a test that lives in this repo). Nothing would
have. Following that thread found three CI lanes that had stopped signalling —
two red for weeks with no issue, one that never existed.

This PR repairs the instruments. It does not fix #927 itself; that is next, and
the nightly lane added here is what will hold it.

## 1. No lane ran `slow` + `correctness` — 153 tests

`python-correctness` runs `-m "correctness and not slow ..."`. Its comment said
the remainder was "the nightly `python-coverage` job's responsibility". Both
halves of that were false:

- `python-coverage` runs `-m "not slow and not correctness ..."` — it excludes
  the marker entirely.
- `ci.yml` had **no `schedule:` trigger**. There was no nightly.

So the 153 `slow`+`correctness` tests — the end-to-end known-optima and
certificate assertions, the ones that solve a real instance and check against
`minlplib.solu` — ran in **no lane at all**. #927's assertion is one of them.

Adds `python-correctness-nightly`: serial (same determinism argument as
`python-correctness`), `--timeout=900` (these tests pass real `time_limit=`
budgets to the solver — ex1252 uses 200 s — so a per-test timeout under the
budget would fail the harness instead of the assertion), `-v` + unbuffered so a
multi-hour job can be distinguished from a hung one (CLAUDE.md §10), 06:20 UTC.

`changes` now returns `code=false` on a `schedule` event, so the new trigger runs
only this lane rather than re-running the whole PR matrix nightly.

## 2. The weekly graduation gate had been exiting 2 before gating anything

```
# ERROR: unknown flag(s): ['node_reduce']; choices: ['all', 'lift_loose_products', ...]
##[error]Process completed with exit code 2.
```

The workflow passed a hardcoded `--flags root_fixpoint,node_reduce,psd_cost_gate,…`.
`node_reduce` was deprecated out of `GRADUATION_ARMS` in #581. Every scheduled
run since **2026-07-13** died on argument validation — four consecutive Mondays
in which the cert-neutrality regression guard proved nothing, while reporting as
a run that happened.

Fix: drop `--flags` entirely. The script's default is
`",".join(generality_sweep.GRADUATION_ARMS)`, so the gated set now *tracks* the
registry instead of being a copy that drifts. Verified both directions:

```
GRADUATION_ARMS = ('root_fixpoint', 'psd_cost_gate', 'lift_zero_spanning',
                   'lift_loose_products', 'lu_density_route',
                   'node_numerical_dual_bound', 'root_build_deadline')
unknown under the new (default) invocation: []
unknown under the OLD hardcoded list:       ['node_reduce']
CHECKS_EXECUTED=2  OK
```

Note the gated set goes 5 → 7 arms: the stale copy had also silently dropped
`lu_density_route`, and never picked up `node_numerical_dual_bound` (#517/#362)
or `root_build_deadline` (#832/#814). `timeout-minutes` 45 → 90 accordingly.

The workflow also gains a `pull_request` trigger on its own paths, so an edit to
the gate or the registry it reads is validated by the PR that makes it rather
than a week later by a cron nobody reads. **That trigger fires on this PR**, so
the gate itself is the check on this fix.

## 3. AMP integration: red on every push to main since 2026-07-27

40+ consecutive failures, no tracking issue, one test:
`test_amp_time_limit_with_incumbent_returns_feasible`, `assert 'time_limit' == 'feasible'`.

**The contract is intact. The test was the broken part.** It drove the timeout by
*read ordinal* — `chain([0.0, 0.0], repeat(1.5))`, commented "the first two reads
land before the deadline … so the timeout path triggers regardless of how many
clock samples solve_amp takes". That premise is false, and
`monkeypatch.setattr(amp_mod.time, "perf_counter", …)` is why: it replaces
`perf_counter` on the **global** `time` module, so every clock read under the
call draws from the sequence, not just AMP's own. Traced:

```
  read  4  1.5 (past deadline)   nonlinear_bound_tightening.py:2910 in _out_of_budget()
  read  6  1.5 (past deadline)   nonlinear_bound_tightening.py:221  in _iter_rows_until()
  …                              (~26 reads in NBT budget polling)
  read 30  1.5 (past deadline)   amp.py:2866 in _solve_amp_impl()
  read 31  1.5 (past deadline)   amp.py:3353 in _solve_amp_impl()
READS_EXECUTED=31
status='time_limit' objective=None
```

The pre-deadline values were consumed before the main loop ever read the clock,
so the deadline blew before any incumbent was stored and `solve_amp` exited
through the "no feasible solution found" arm. The assertion then described a path
the test no longer executed — and its failure message pointed at the solver.

Independent check that the contract holds, driving the timeout by **state**
(pre-deadline until `_solve_best_nlp_candidate` has returned an incumbent, past
the deadline after — which is what "timed out after finding an incumbent" means,
and is immune to read counts):

```
INCUMBENT_CALLBACKS_EXECUTED=1
status='feasible' objective=10.0 gap_certified=False x_is_none=False
```

The test now uses that wiring, and asserts `calls["n"] == 1` first, so it cannot
pass while measuring nothing (CLAUDE.md §6). Injection-tested: restoring the old
read-ordinal wiring makes it fail with

```
E  AssertionError: the incumbent callback never fired, so the timeout did not
   land *after* an incumbent and this test measured nothing
E  assert 0 == 1
```

— i.e. it now names the defect instead of implicating the solver.

## Failure routing

All three scheduled/main lanes gained a `report-failure` job that files, or
comments on, one open `ci-signal`-labelled issue. Each of the failures above
persisted for weeks; none of them reached anyone. Flagging in case you'd rather
not have the bot open issues — it's a self-contained job per workflow and easy to
drop.

## Verification

- `test_amp_time_limit_with_incumbent_returns_feasible`: **1 passed** (fails
  before the change with `assert 'time_limit' == 'feasible'`; fails under
  injection with the measured-nothing message).
- All 6 workflow files parse; `WORKFLOWS_PARSED=6`.
- Graduation-gate default flag set validates (above); the gate itself runs on
  this PR via the new trigger.
- `ruff check` / `ruff format --check` clean on the edited test.
- Full `slow`+`correctness` set running locally to size the new lane — result
  reported in a follow-up comment.

Contributes to #927 (adds the lane that will hold its regression test).

### DIFF
--- a/.github/workflows/amp-integration.yml
+++ b/.github/workflows/amp-integration.yml
@@ -96,3 +96,39 @@ jobs:
           JAX_PLATFORMS: cpu
           JAX_ENABLE_X64: "1"
           PYTEST_MEMORY_LIMIT_MB: "0"
+
+  # This suite failed on EVERY push to main from 2026-07-27 to 2026-08-08 — 40+
+  # consecutive red runs — on a single test, with no issue tracking it. A red X
+  # on a merge commit is not a signal if nothing routes it anywhere. One open
+  # issue, reused rather than duplicated, is.
+  report-failure:
+    name: File/refresh the CI-signal issue
+    needs: amp-integration
+    if: failure() && (github.event_name == 'schedule' || github.event_name == 'push')
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Open or update the tracking issue
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          TITLE: "CI signal down: AMP integration suite is failing on main"
+        run: |
+          set -euo pipefail
+          gh label create ci-signal --force --color B60205 \
+            --description "A CI lane that is supposed to be watching something has stopped watching"
+          printf '%s\n' \
+            "The AMP integration suite failed on \`${GITHUB_REF_NAME}\` (${GITHUB_SHA})." \
+            "" \
+            "Run: ${RUN_URL}" \
+            "" \
+            "While this is red the AMP paths are unguarded on main." \
+            > body.md
+          NUM=$(gh issue list --state open --label ci-signal --json number,title \
+                  --jq "[.[] | select(.title == env.TITLE)][0].number // empty")
+          if [ -n "$NUM" ]; then
+            gh issue comment "$NUM" --body-file body.md
+          else
+            gh issue create --title "$TITLE" --label ci-signal --body-file body.md
+          fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,11 @@ on:
   pull_request:
     branches: [main, feature/mip-nlp-solver]
     types: [opened, synchronize, reopened, ready_for_review, labeled]
+  schedule:
+    # Nightly, 06:20 UTC. Only `python-correctness-nightly` runs on this trigger;
+    # every other job below is gated on `needs.changes`, which is `false` for a
+    # schedule event (no diff to inspect).
+    - cron: "20 6 * * *"
 
 # Cancel superseded runs on the same PR/branch — saves CI minutes when
 # pushes land in quick succession.
@@ -34,6 +39,16 @@ jobs:
         shell: bash
         run: |
           set -uo pipefail
+          if [ "${{ github.event_name }}" = "schedule" ]; then
+            # The nightly trigger exists solely for `python-correctness-nightly`,
+            # which does NOT gate on this output. This is not the "uncertainty"
+            # case the rule below covers — there is no diff to be uncertain
+            # about, and re-running the whole PR matrix nightly would buy nothing
+            # that the last push already proved.
+            echo "code=false" >> "$GITHUB_OUTPUT"
+            echo "=> code=false (schedule: only the nightly correctness lane runs)"
+            exit 0
+          fi
           if [ "${{ github.event_name }}" = "pull_request" ]; then
             git fetch --no-tags --depth=200 origin "${{ github.base_ref }}" || true
             base="$(git merge-base "origin/${{ github.base_ref }}" HEAD 2>/dev/null || true)"
@@ -194,8 +209,15 @@ jobs:
     # #248's nvs16 garbage-bound guard regressed at the #632 uniform-relax cutover
     # and sat unnoticed for a full release cycle because no PR job ran the
     # `correctness` marker. This lane closes that gap. It runs only the cheap
-    # subset (`not slow`); the full known-optima sweep (including `slow`
-    # correctness) remains the nightly `python-coverage` job's responsibility.
+    # subset (`not slow`); the `slow` half is the `python-correctness-nightly`
+    # job below.
+    #
+    # That sentence used to read "remains the nightly `python-coverage` job's
+    # responsibility", and it was false in two ways at once: `python-coverage`
+    # runs `-m "not slow and not correctness ..."`, and this workflow had no
+    # `schedule:` trigger, so there was no nightly. The 153 `slow`+`correctness`
+    # tests ran in NO lane — including the one asserting that ex1252's dual bound
+    # under the #707/#721/#732 reform flags is sound, which is red (#927).
     name: Python correctness (fast subset, 3.12, ubuntu)
     runs-on: ubuntu-latest
     needs: changes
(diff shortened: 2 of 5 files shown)


jkitchin:
**The nightly lane, measured.** Ran the exact selection serially on an M-series
laptop:

```
1 failed, 146 passed, 7 skipped, 8868 deselected, 5 warnings in 313.53s (0:05:13)

slowest:
  65.69s  test_843_qubo_primal.py::test_843_qubo_primal_incumbent_on_chimera
  63.64s  test_integer_multilinear_reform.py::test_ex1252_multilinear_reform_sound_and_lifts_bound
  47.23s  test_disjunctive_config_bound.py::test_ex1252_spatial_recursion_clears_721_bar
  28.11s  test_integer_multilinear_reform.py::test_ex1252_short_budget_declines_reform_no_regression
```

Two things follow.

1. **It is cheap** — 313 s, not the hours I budgeted. `timeout-minutes` lowered
   360 → 120 (the fast-correctness lane is ~5 min locally and 23 min on a runner,
   so ~25 min is the honest expectation here, with headroom for a hung solve
   rather than a slow one). Cheap enough that it could reasonably be a PR lane
   later; leaving it nightly for now because adding a currently-red lane to every
   PR would block unrelated work until #927 is fixed.

2. **The only failure is #927.** So the missing lane was hiding exactly one bug,
   and this lane goes green the moment #927 does. That is the next piece of work.

Issue #954, opened 2026-08-08
correctness: _solve_nlp_bb returns its incumbent with no feasibility verification, and its rounding guard runs at feas_tol=1e-4


jkitchin:
## Summary

`_solve_nlp_bb` is the last of the five solve exit paths with no verification of
the incumbent it returns. #952 fixed `_solve_miqp_bb` (the reported path) and
`_solve_milp_bb` (found by survey); this is the remainder, split out because the
fix needs a mechanism #952 never measured.

Survey of incumbent verification before return, as of #952:

| path | verified before return? |
|---|---|
| `solve_model` | yes (#779 guard on the substitution path) |
| `_try_native_spatial_kernel` | yes (#789, against the original model) |
| `_solve_miqp_bb` | **fixed in #952** |
| `_solve_milp_bb` | **fixed in #952** |
| `_solve_nlp_bb` | **no** — this issue |

Its exit is the same three steps the other two had — `_round_incumbent_integers`,
`_unpack_solution`, return — with nothing checking the point whose objective is
reported as `objective` and, on `optimal`, as the dual `bound` too.

## Why this is not simply "apply the #952 gate here"

#952's gate is `_matrix_solution_feasible` over the decomposed `A_ub`/`A_eq`
matrices. That is the right arbiter for a linear-row path and the wrong one here:
`_solve_nlp_bb`'s rows are nonlinear, so the check has to go through
`_check_constraint_feasibility(evaluator, x, cl_list, cu_list)` instead — a
different tolerance regime (an NLP incumbent legitimately sits at roughly the
backend's `constr_viol_tol`, and `lp_pounce._CONSTR_VIOL_TOL` is 1e-8 on the
published wheel but is a per-build fact). Flipping a gate on at `abs=1e-6` across
this path could newly refuse instances that #952 never measured, and per that
issue's verification regime each such refusal is a finding to characterize.

So this needs its own entry experiment before any gate ships.

## The second half: the rounding guard's 1e-4

Unlike the MILP/MIQP call sites — where `_round_incumbent_integers`'s `feasible`
flag was **vacuous** because neither passed an evaluator (fixed in #952) — this
path *does* pass one, so the flag is real. But it runs at the helper's default
`feas_tol=1e-4`, which is 100x the declared `abs=1e-6`. That is the same
unexplained gap #952's item 3 asked about for `_check_lp_solution_feasibility`;
there it closed by deleting a helper that had no callers left. Here it is live
[...]

Issue #952, opened 2026-08-08
correctness: _solve_miqp_bb accepts an incumbent with only an equality check at tol=1e-4; no inequality or bound verification before returning optimal


jkitchin:
## Summary

On the convex-MIQP branch-and-bound path (`_solve_miqp_bb`), the only feasibility
gate a candidate incumbent passes is `_check_lp_solution_feasibility(A_eq, b_eq, x)`
— **equality residual only, at `tol=1e-4`**. There is no inequality-row check, no
variable-bound check, and no independent verification of the point whose objective
is then returned as both `objective` **and** `bound` on an `optimal` status.

Measured on `main` @ `8bfce1f1`: a 40-seed MIQP family returns `optimal` on every
seed with the returned incumbent **1.9e-08 – 4.9e-08 outside a declared inequality
row**, worst `4.893064e-08` (seed 21).

Today's excursions are small — inside the repo's declared `abs=1e-6` — so this is
**not** a live false-certificate report. The issue is that nothing bounds them:
their size is set by whatever the QP IPM happened to converge to, and the gate that
admitted them would equally admit `1e-4`, which is 100× the declared tolerance.

## Retraction (CLAUDE.md §11)

This issue replaces a claim I published in the PR #949 review comment and its PR
body, which the re-measurement falsifies. That claim was:

> seed 16's monolithic reference still returns a point 1.0e-08 outside a degenerate
> row post-#948. #948 removed two producers of that class, not all of them, and this
> is one it did not reach.

Three parts of that are wrong:

1. **Not seed-specific.** All 40 seeds violate, not seed 16. Seed 16 was simply the
   worst in the narrower slice I looked at then.
2. **Not a `#948`/`bound_relax_factor` residual.** I re-ran with
   `bound_relax_factor = 0` forced and got results *bit-identical* to the stock arm
   (`4.893064e-08` in both). The reason is the instrument, not the solver: I patched
   `discopt.solvers.nlp_pounce.solve_nlp`, and a call counter shows
   **`NLP_POUNCE_SOLVE_NLP_CALLS=0`** — this family is a convex MIQP and never
   touches the NLP path at all. My "exact" and "relaxed" arms were the same code
   path. (CLAUDE.md §6: the identical numbers were the signature of an instrument
   that never ran, and I should have counted invocations before reporting an arm.)
3. **Not the NLP's constraint tolerance either.** Sweeping `constr_viol_tol` over
[...]

Issue #789, opened 2026-07-19
Native spatial kernel: Python-engine feature parity / smoke-test re-scope for default-ON


jkitchin:
## Context
Follow-up from #764. The native Rust spatial B&B kernel passed both Regime-2 graduation bars (cert-clean, net-positive) but was **kept default OFF** because forcing it ON fails **20 of the `-m smoke` tests**. The kernel short-circuits the Python B&B engine, so features layered on that engine are not exercised on the native path.

## Failing surfaces (all Python-engine machinery the native kernel bypasses)
- incumbent / node callbacks
- RENS / SubNLP heuristics
- solution pools
- warm-start incumbents
- `mccormick_bounds` modes
- deadline handling
- batched node processing
- lazy constraints

## Problem
Per CLAUDE.md, greening a gate by editing 20+ tests is forbidden — the tests encode real contracts. Default-ON graduation therefore requires the native kernel to **honor these contracts** (or the driver to fall back to the Python engine when a model uses a feature the kernel does not support), not for the tests to be relaxed.

## Definition of done
- Either the native kernel supports each contract above, or the driver detects unsupported features and cleanly routes those models to the Python engine (with the native path taken only when it is feature-safe).
- `pytest -m smoke` passes with `DISCOPT_NATIVE_SPATIAL_KERNEL` forced ON (currently 20 fail; 807 pass).
- No test is weakened to pass the gate.

## Relates to
- #764 (parent — this is the second prerequisite for default-ON graduation).

jkitchin:
## Resolved — smoke is green under the flag; no test weakened (#794, merged)

The native spatial kernel went from **20 → 0** smoke failures with `DISCOPT_NATIVE_SPATIAL_KERNEL=1`, via two driver-side guards (the DoD's "route unsupported models to the Python engine" path — the kernel's contracts are honored by declining, not by touching the tests):

1. **Feature-safety routing** (`_native_kernel_feature_safe`): decline to the Python engine when a solve requests a Python-engine contract the kernel bypasses — node/incumbent/iteration callbacks, lazy constraints, warm-start `initial_point`, non-default `mccormick_bounds`, solution pool, explicit `tuning`. Cleared 9.

[...]

PR #955, opened 2026-08-08
fix(correctness): #952 verify MILP/MIQP B&B incumbents against every declared row and bound

jkitchin:
## Summary

`_solve_miqp_bb`'s only feasibility gate was
`_check_lp_solution_feasibility(A_eq_full, b_eq, x_full)` — an equality residual at
`tol=1e-4`, with no inequality-row check, no bound check, and no verification of the
point whose objective is returned as `objective` **and**, on `optimal`, as the dual
`bound`.

**Root cause: that gate was a tautology on the path that runs.**
`_pounce_qp_relaxation_nodes` solves only the structural columns and then
*reconstructs* the slacks as `z = S⁺(b_eq − A_struct x_s)` (`solver.py:16148`), which
makes `A_eq_full [x_s, z] == b_eq` hold to machine precision for **any** `x_s`: a
violated inequality row comes back as a negative slack, and the gate never looked at
slack bounds. Measured on `main` @ `8bfce1f` over the issue's 40-seed family:

```
GATE_CALLS=212
GATE_EQ_RESIDUAL_MAX=8.881784e-16     <- what the gate measured
GATE_EQ_RESIDUAL_MEDIAN=0.000000e+00
SEEDS_VIOLATING_ABOVE_1e-9=40         <- how far the returned points actually sat
WORST viol=9.973171e-09 seed=30       <- outside a declared inequality row
```

(The issue reported `4.893064e-08` worst on seed 21; this environment measures
`9.973e-09` worst on seed 30. Different magnitudes, same mechanism — and the
mechanism, not the magnitude, is the finding.) These excursions are inside the
declared `abs=1e-6`, so this was **never a live false certificate**. The defect is
that nothing bounded them.

Changes:

- **`_node_point_feasible`** replaces the equality-only gate at both MIQP node call
  sites: the decomposed `A_ub`/`A_eq` rows plus the node's own box, via
  `_matrix_solution_feasible` at the declared `abs=1e-6`. This is what the MILP
  path's `_solve_node_lp_pounce` already did.
- **Exit gate**: every incumbent `_solve_miqp_bb` returns is verified against every
  declared row and the declared box (captured before FBBT), raising rather than
  repairing or downgrading — the same loud refusal as the Gurobi QCP path. Covers
  `feasible` as well as `optimal`: `objective` is a primal claim in both, and GBD/OA
  consume it as a valid upper bound.
- A snapped incumbent is verified before `tree.inject_incumbent`.
- **`_check_lp_solution_feasibility` is deleted** — no callers left. That closes the
  issue's item 3 (the unexplained 100x gap between its `1e-4` and the declared
  `1e-6`) rather than restating it.
- `_matrix_solution_feasible`'s bounds loop is vectorised (semantics-identical,
  pinned by a 200-case randomised equivalence test) since the node gate now calls it
  once per node.

**Scope beyond the reported path (CLAUDE.md §2).** A survey of the five solve exit
paths found the same omission on `_solve_milp_bb`, whose incumbent exit was
structurally identical (round, unpack, return):

| path | verified before return? |
|---|---|
| `solve_model` | yes (#779) |
| `_try_native_spatial_kernel` | yes (#789) |
| `_solve_miqp_bb` | **MISSING → fixed here** |
| `_solve_milp_bb` | **MISSING → fixed here** |
| `_solve_nlp_bb` | **MISSING → #954** |

`_solve_nlp_bb` is deliberately left open: its rows are nonlinear, so the check needs
`_check_constraint_feasibility` on the evaluator rather than the matrix arbiter, with
different tolerance semantics and a real risk of new refusals this issue never
measured. That is a genuine scope expansion, filed as #954 with the evidence. This is
the primal-side twin of **#933**, which tracks the same per-path duplication on the
dual bound.

**A finding the new gate produced.** The MILP gate immediately refused
`test_nn_equivalence::test_tree_ensemble_fixed_input`. Characterized, not
tolerance-tweaked — and the producer is not the solve:

```
A_eq violation, UNROUNDED incumbent : 4.440892e-16
A_eq violation, ROUNDED  incumbent : 1.551361e-06
max per-coordinate snap             : 3.902373e-07   (5-term row, scale 1.05e1)
```

The LP solved the row exactly; the C-3 integer snap moved it. That call site's
comment asserted the snap "cannot move a linear row by more than the integrality
tol" — wrong, a row accumulates one snap per term — and is corrected in place.
`_round_incumbent_integers` is documented to return `feasible=False` when rounding
breaks feasibility, "the caller must not certify a rounded point that reports
False" — but it can only do that when handed an evaluator, and **neither** B&B call
site passes one, so the flag was unconditionally `True` and validated nothing (the
same shape as the headline bug). Both now use the matrix arbiter as that checker and
report the unrounded point when snapping would leave the rows. Not a tolerance
concession: the unrounded point satisfies **both** declared tolerances (rows at
4.4e-16, integrality at 3.9e-7 against `integrality_tol=1e-5`), while the snapped one
satisfies integrality exactly but misses `abs=1e-6` on a row.

Contributes to #952 — the reported MIQP path is complete; #954 carries the one
remaining path.

## Correctness

This tightens acceptance gates and adds refusals. It removes no check, weakens no
tolerance, and changes no relaxation, bound, or cut. The exit gate can only turn a
previously-returned result into a loud `RuntimeError`; it can never turn a refusal
into an acceptance. The node gate is strictly more conservative: a node point that
now fails routes to POUNCE recovery, which either restores a KKT-valid bound,
certifies the box empty (a sound Phase-1 prune), or keeps the node open and
decertifies — never a false-infeasible.

One honest scope limit, stated in the code: the gate checks the rows
`extract_qp_data`/`extract_lp_data` produced, the same matrices the node solves
consume, so it catches an engine returning a point outside the extracted problem —
the #952 failure mode — but not a bug in the extraction itself. That second class is
covered from outside, against the model as declared, by the new test file's
`NLPEvaluator` measurement.

- [x] `pytest -m smoke` → **916 passed, 15 skipped, 2 xpassed** (310 s)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **10 passed** (171 s)

An earlier adversarial run showed `test_large_dense_jacobian_no_crash` failing; it is
a load-sensitive assertion (`wall < budget + 40`, nominal 25 s) and **not on this
change's path** — hooking the dispatch shows it runs through `_solve_nlp_bb` with
`_matrix_solution_feasible` called **0** times. It passes in isolation and on re-run.

## Regression test

New file `python/tests/test_952_miqp_incumbent_feasibility.py`, 7 tests (all 7 pass,
3.5 s, including the `slow`-marked 40-seed panel):

| test | fails before? |
|---|---|
| `test_exit_gate_refuses_an_off_row_incumbent` | **yes** — `DID NOT RAISE RuntimeError` |
| `test_milp_exit_gate_refuses_an_off_row_incumbent` | **yes** — `DID NOT RAISE RuntimeError` |
| `test_integer_snap_is_declined_when_it_would_leave_the_rows` | **yes** — raises with `A_eq row 0 by 3.000000e-06` |
| `test_returned_incumbent_is_within_declared_tolerance` | no — standing watch (see below) |
| `test_old_equality_only_gate_was_a_tautology` | no — pins the mechanism |
| `test_milp_baseline_solve_is_unaffected` | no — control |
| `test_bounds_check_vectorisation_matches_the_original_loop` | no — refactor guard, 200 cases |

The 40-seed panel is the issue's item 4 and carries an executed-comparison count
(`comparisons == 40`, `checks == 720`) so it cannot pass by measuring nothing (§6). It
passes before the fix too — today's excursions are ~9e-9 — because its job is to bound
something nothing bounded, not to catch a live false certificate. The three gate tests
are the fail-before/pass-after evidence. Both `pytest.raises` tests assert a
perturbation-applied flag so a passing raise cannot be some unrelated error.

Not bound-changing — this is CLAUDE.md §5 **bound-neutral**, so the requirement is
that `node_count` and the certified `objective` are *exactly* unchanged.

Panel: the 40-seed family plus every in-repo `minlplib_nl` instance = **106
instances**, 49 through `_solve_miqp_bb`, run before and after.

```
COMPARISONS_EXECUTED=106
STRICT_INSTANCES=88   (self-terminating)
STRICT_DRIFTS=0       status / objective / bound / node_count / gap_certified
MIQP_BB_INSTANCES_COMPARED=49   TOTAL_FIELD_COMPARISONS=343   MIQP_BB_DRIFTS=0
worst returned-point violation, before: 9.973171262195438e-09 (synth_seed_30)
worst returned-point violation, after : 9.973171262195438e-09 (synth_seed_30)  <- byte-identical
ERRORS_AFTER=0
```

**One correction to my own method (CLAUDE.md §11).** The first version of the
comparison keyed "self-terminating" on `status not in {time_limit, node_limit}`. That
is wrong for these paths: they report `time_limit` only when there is *no* incumbent,
so a wall-truncated search holding one reports `feasible`. It misclassified `tls2` and
produced a `STRICT_DRIFTS=1` that was pure timing. `tls2` dispatches to
`_solve_nlp_bb` — a path this change does not touch — and an interleaved A/B on a
quiet machine (§9) returned bit-identical results on both arms:

```
FIXED  feasible obj=9.299999999984365 bound=2.5973907091074806 nodes=389   (3/3 rounds)
MAIN   feasible obj=9.299999999984365 bound=2.5973907091074806 nodes=389   (3/3 rounds)
```

The classifier now keys on `gap_certified`, and with it there are zero strict drifts.

- Measurement: 106-instance panel (40-seed MIQP family + in-repo `minlplib_nl`
  corpus), baseline `main` @ `8bfce1f`, 0 strict drifts / 0 MIQP-path drifts /
  byte-identical worst violation.

https://claude.ai/code/session_01W1FMvUDmy3uCVtj8DPZxWj

---
_Generated by [Claude Code](https://claude.ai/code/session_01W1FMvUDmy3uCVtj8DPZxWj)_

### DIFF
--- a/.github/workflows/amp-integration.yml
+++ b/.github/workflows/amp-integration.yml
@@ -96,3 +96,39 @@ jobs:
           JAX_PLATFORMS: cpu
           JAX_ENABLE_X64: "1"
           PYTEST_MEMORY_LIMIT_MB: "0"
+
+  # This suite failed on EVERY push to main from 2026-07-27 to 2026-08-08 — 40+
+  # consecutive red runs — on a single test, with no issue tracking it. A red X
+  # on a merge commit is not a signal if nothing routes it anywhere. One open
+  # issue, reused rather than duplicated, is.
+  report-failure:
+    name: File/refresh the CI-signal issue
+    needs: amp-integration
+    if: failure() && (github.event_name == 'schedule' || github.event_name == 'push')
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Open or update the tracking issue
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          TITLE: "CI signal down: AMP integration suite is failing on main"
+        run: |
+          set -euo pipefail
+          gh label create ci-signal --force --color B60205 \
+            --description "A CI lane that is supposed to be watching something has stopped watching"
+          printf '%s\n' \
+            "The AMP integration suite failed on \`${GITHUB_REF_NAME}\` (${GITHUB_SHA})." \
+            "" \
+            "Run: ${RUN_URL}" \
+            "" \
+            "While this is red the AMP paths are unguarded on main." \
+            > body.md
+          NUM=$(gh issue list --state open --label ci-signal --json number,title \
+                  --jq "[.[] | select(.title == env.TITLE)][0].number // empty")
+          if [ -n "$NUM" ]; then
+            gh issue comment "$NUM" --body-file body.md
+          else
+            gh issue create --title "$TITLE" --label ci-signal --body-file body.md
+          fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,11 @@ on:
   pull_request:
     branches: [main, feature/mip-nlp-solver]
     types: [opened, synchronize, reopened, ready_for_review, labeled]
+  schedule:
+    # Nightly, 06:20 UTC. Only `python-correctness-nightly` runs on this trigger;
+    # every other job below is gated on `needs.changes`, which is `false` for a
+    # schedule event (no diff to inspect).
+    - cron: "20 6 * * *"
 
 # Cancel superseded runs on the same PR/branch — saves CI minutes when
 # pushes land in quick succession.
@@ -34,6 +39,16 @@ jobs:
         shell: bash
         run: |
           set -uo pipefail
+          if [ "${{ github.event_name }}" = "schedule" ]; then
+            # The nightly trigger exists solely for `python-correctness-nightly`,
+            # which does NOT gate on this output. This is not the "uncertainty"
+            # case the rule below covers — there is no diff to be uncertain
+            # about, and re-running the whole PR matrix nightly would buy nothing
+            # that the last push already proved.
+            echo "code=false" >> "$GITHUB_OUTPUT"
+            echo "=> code=false (schedule: only the nightly correctness lane runs)"
+            exit 0
+          fi
           if [ "${{ github.event_name }}" = "pull_request" ]; then
             git fetch --no-tags --depth=200 origin "${{ github.base_ref }}" || true
             base="$(git merge-base "origin/${{ github.base_ref }}" HEAD 2>/dev/null || true)"
@@ -194,8 +209,15 @@ jobs:
     # #248's nvs16 garbage-bound guard regressed at the #632 uniform-relax cutover
     # and sat unnoticed for a full release cycle because no PR job ran the
     # `correctness` marker. This lane closes that gap. It runs only the cheap
-    # subset (`not slow`); the full known-optima sweep (including `slow`
-    # correctness) remains the nightly `python-coverage` job's responsibility.
+    # subset (`not slow`); the `slow` half is the `python-correctness-nightly`
+    # job below.
+    #
+    # That sentence used to read "remains the nightly `python-coverage` job's
+    # responsibility", and it was false in two ways at once: `python-coverage`
+    # runs `-m "not slow and not correctness ..."`, and this workflow had no
+    # `schedule:` trigger, so there was no nightly. The 153 `slow`+`correctness`
+    # tests ran in NO lane — including the one asserting that ex1252's dual bound
+    # under the #707/#721/#732 reform flags is sound, which is red (#927).
     name: Python correctness (fast subset, 3.12, ubuntu)
     runs-on: ubuntu-latest
     needs: changes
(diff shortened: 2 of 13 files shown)


jkitchin:
## Not merged — CI is red, and the failure is real

`Python fast` and `Python correctness (fast subset)` both fail. Reproduced locally on
this branch (built with `maturin develop --release`), identical numbers, deterministic:

```
test_benders_soundness.py::test_bound_active_recourse_is_sound[1000.0]
    RuntimeError: MILP-BB returned an infeasible point labeled feasible/optimal: A_ub row 1 by 9.994987e-06
test_benders_soundness.py::test_bound_active_recourse_is_sound[1000000.0]
    RuntimeError: MILP-BB returned an infeasible point labeled feasible/optimal: A_ub row 1 by 9.999992e-03
```

**The new exit gate is right. The point really is outside a declared row.** This is
the gate doing its job, not a false alarm — so the fix does not belong in the gate.

### What the point is

```
x = [y=1.0, eta=-1.00000001e+06]
A_ub row 1:  coefs = [-0.27569516, -1.0]   rhs = 999999.7243048395
             activity - rhs = 9.999992e-03    row_scale = 1.000000e+06
             relative       = 1.000e-08       gate threshold = 1.001e-03
```

`eta` is **inside** its declared box and **outside** a declared row. The violation is
exactly `1e-8 x row_scale` in both parametrisations (`vub=1e3` → 9.995e-6,
`vub=1e6` → 9.99999e-3).

### Root cause

Every POUNCE call on this path — all 13 of them in the failing run — goes through
`pounce.solve_qp`, and **none passes `options`**:

```
POUNCE_CALLS=13
  solve_qp   options_passed=False  x13
distinct options seen: (none)
```

`pounce.solve_qp`'s signature has no `options` parameter at all:

```
(P=None, c=None, A=None, b=None, G=None, h=None, lb=None, ub=None, *, tol=None,
 max_iter=None, warm_start=None, collect_iterates=False, check_psd=None,
 method='ipm', tau=None, tau_max=None)
```

So `POUNCE_BOUND_RELAX_FACTOR = 0.0` — the mitigation this file documents at
`solver.py:15654` for exactly this failure mode ("relaxes bounds — including the
**slack bounds standing in for inequality rows** — by `1e-8*(1 + |bound|)`") — is
**unreachable here**. It is applied at `solver.py:15152` and `:15431`, both of which
go through the `_pounce_solve_lp` / `_pounce_solve_qp` callback wrappers. The
Benders master flow never touches those.

Corroborating: the tree solves a standard-form system (`lp_data.A_ub is None`,
6 columns — rows folded into equalities with slack columns) while the exit gate
verifies the declared 3x2 inequality form. The row limits live in the tree's slack
bounds, which is precisely what the 1e-8 relaxation loosens. The engine converges
honestly to a relaxed box; no convergence tolerance can remove it.

### One hypothesis I tested and had to discard

I first attributed this to the convex IPM's own ~1e-8 convergence tolerance and
predicted that swapping the node engine to the exact simplex would fix it. It does
not — both engines return **bit-identical** violations:

```
pounce IPM nodes    vub=1e3  9.994987e-06     vub=1e6  9.999992e-03
exact simplex nodes vub=1e3  9.994987e-06     vub=1e6  9.999992e-03   (ARMS_EXECUTED=4)
```

Identical to the last digit across two different engines means the point is not
engine noise and does not come from a node LP at all — consistent with this PR's own
note that the tree also takes incumbents from the root dive, RENS/feasibility-pump
injections and POUNCE recovery. Recording the falsification here rather than leaving
the wrong explanation standing.

### Options, none of which I should pick unilaterally

1. Route these solves through the callback path that accepts
   `bound_relax_factor=0`. `solver.py:15147` records the cost measured last time this
   was applied backend-wide: 2 correctness-lane tests, **1.6s → 79s** (#945).
2. Get a bound-relaxation / feasibility knob onto `pounce.solve_qp` upstream.
3. Polish the returned incumbent onto the declared rows before the exit gate — e.g. a
   cleanup solve on the fixed integer assignment.
4. Relax the gate's `rtol` toward the repo's declared `rel=1e-4` (under which a 1e-8
   relative violation passes). **I am not doing this**: the tight `rtol=1e-9` has a
   documented reason (#850 Obs 3, where a looser relative tolerance admitted a
   super-optimal constraint-infeasible incumbent), and loosening a guard to make a
   test pass is exactly what CLAUDE.md §1/§3 forbids.

Worth noting for scope: these excursions sit inside the declared `abs=1e-6` only for
small data — at `vub=1e6` the violation is `1e-2`, so the "never a live false
certificate" framing in the PR description holds for the 40-seed family it was
measured on but not for badly-scaled rows.

jkitchin:
Rebased onto the revert (88af66d) and pushed the fix; CI is running.

**The revert was right, and its diagnosis of the failing point matches mine** — `y` is exactly integral, so an acceptance-time veto on what *rounding* does cannot catch this class. Where I differ is the mechanism. It is not that `POUNCE_BOUND_RELAX_FACTOR` is unreachable through `pounce.solve_qp`; it is that the point never comes from a node LP at all.

Two spies, both with executed-check counts so they cannot report a pass by measuring nothing:

```
probe955e  BATCHES=4  NODE_POINTS_CHECKED=4  NODE_POINTS_OUTSIDE_ROWS=0
probe955f  SNAP_CHECKS=13  INJECTION_FUNNEL_CALLS=9
           _maybe_inject_snapped  x=[1.0, -1.00000001e+06]  snap_ok=True  rows_ok=False
```

Every node relaxation point is inside the rows. The offending point arrives through `_maybe_inject_snapped` → `_pounce_snap_incumbent`, which fixes the integers and takes the POUNCE IPM's continuous completion. That completion honours the slack bounds standing in for inequality rows only to the IPM's own ~1e-8 *relative* tolerance — on `eta <= 1e6*y`, a 1e-2 absolute excursion. Because the result is exactly integral, nothing downstream re-examines it, and the tree reports it `optimal`.

**Fix:** `_solve_milp_bb`'s `_maybe_inject_snapped` verifies the candidate against the decomposed rows and the node's own box before `tree.inject_incumbent`. That is the check the MIQP twin of this funnel already ran (`_node_point_feasible`) — the MILP funnel was the gap, so this closes a per-path omission rather than an instance. It is a refusal, not a repair: nothing is loosened, and declining is free, since injection is a heuristic accelerator, the subtree stays open, and a genuinely feasible point is re-found from a node relaxation. Option 4 (relaxing the gate's `rtol`) stays off the table for the reason already recorded.

**Verification.** The 4 previously-failing tests fail before this commit and pass after. `test_benders_soundness` + `test_946` + `test_952`: 34 passed. `pytest -m smoke`: 920 passed, 1 skipped, 2 xpassed. The adversarial suite is still running locally; I will post its result.

**One falsification to record.** `test_946_gbd_degenerate_multiplier`'s non-binary arm asserted GBD stalls out at `iteration_limit`. On this branch it certifies — `optimal`, bound -1.0000000149881925 against an optimum of -1, in 4 recourse solves — because the master no longer promotes an incumbent whose optimality cut carried no usable `eta` information. Attributed by measurement, not assumption: on `origin/main`'s `solver.py` the arm still stalls, and on this branch with *this* commit reverted it already certifies, so the cause is the earlier C-3 correction (report the unrounded point when snapping leaves the rows), not the funnel check. The test now pins both arms — the explanation on the uncertified exit, reached deterministically with `max_iterations=1`, and the newly certified full-budget outcome.

Issue #957, opened 2026-08-08
interval.py outward rounding turns exact 0.0 into a denormal, which makes the LP conditioning test overflow to inf


jkitchin:
## Summary

`python/discopt/_jax/convexity/interval.py`'s outward-rounding helpers turn an
**exact zero** endpoint into a denormal:

```python
def _round_down(x): return np.nextafter(x, np.float64(-np.inf))   # 0.0 -> -5e-324
def _round_up(x):   return np.nextafter(x, np.float64(np.inf))    # 0.0 ->  5e-324
```

Verified:

```
round_down(0.0) = -5e-324    round_up(0.0) = 5e-324
np.nextafter(0.0, -np.inf) == 0.0  ->  False
```

Outward rounding at zero is not wrong as enclosure arithmetic — `[-5e-324, 5e-324]`
does contain `0` — but the denormal survives into coefficients that reach the kernel
spec, and downstream code that reasons about *magnitudes* is not prepared for a
nonzero value 300 orders of magnitude below everything else.

## Concrete consequence

`python/discopt/_jax/milp_relaxation.py:548-553` decides whether an LP is
ill-conditioned enough to need equilibration:

```python
nz = np.abs(data[data != 0.0])
ill = bool(nz.size and np.isfinite(nz).all() and nz.max() / nz.min() > _RELAX_EQUILIBRATE_TRIGGER)
```

A `±5e-324` entry passes `data != 0.0` (it is not zero) and passes
`np.isfinite(nz).all()` (it is finite), and then the ratio itself **overflows**:

```
isfinite all = True    nz.max() / nz.min() = inf     (RuntimeWarning: overflow encountered in scalar divide)
```

`inf > trigger` is always true, so equilibration fires unconditionally on any
relaxation carrying one of these denormals — regardless of the actual conditioning of
the matrix, which may be perfectly benign. The guard was written to detect ill
conditioning and instead detects "a zero got nudged".

## Proposed fix

Two independent changes, both worth making:

1. **In `interval.py`** — leave exact zeros alone, the way the helpers already leave
   infinities alone (the docstring calls that case out explicitly). `nextafter` from
   `0.0` crosses into the denormal range, which is exactly where "one ULP" stops
   meaning "a negligible relative widening". Special-casing `x == 0.0` keeps the
   enclosure sound (`0` is still contained) without manufacturing a subnormal.

2. **In `milp_relaxation.py`** — make the conditioning test robust to a
[...]

Issue #956, opened 2026-08-08
McCormick envelope rows are not outward-rounded, so a relaxation can cut off the point it relaxes


jkitchin:
## Summary

The McCormick row generators in `crates/discopt-core/src/bnb/mccormick_patch.rs` do
no outward rounding, so an envelope row can cut off a point that is **exactly**
feasible for the term it relaxes. At the box corner the emitted row and the
auxiliary column's own bound are two different roundings of the same quantity, and
they disagree. On large-magnitude terms the disagreement is far larger than any LP
feasibility tolerance, and the node LP becomes unsolvable.

This is the mechanism found while root-causing #927. It is **not** what made that
certificate false (that was the tree fathoming on an undecided LP status, fixed in
#927's PR) — but it is a real defect in its own right, and it is why the fixed
search now spends over half its nodes on LPs the simplex cannot solve.

## Measurement

Instance `ex1252` under `DISCOPT_INTEGER_MULTILINEAR_REFORM=1` +
`DISCOPT_MULTILINEAR_COUPLING_RLT=1`. The reformulation lifts `x^3` for a variable
whose box upper bound is `ui = 2950`. `EnvTerm::Monomial { p: 3 }` emits a tangent
row at `ui` (`mccormick_patch.rs::monomial_rows`):

```
dfu*x - s <= dfu*ui - fu       with  dfu = 3*2950^2 = 2.61e7
                                     dfu*ui         = 7.7017e10
                                     fu   = 2950^3  = 2.5672e10
                                     rhs            = 5.1345e10
```

The rhs is a cancelling subtraction of two ~7.7e10 quantities. Meanwhile
`monomial_aux_bounds` sets the auxiliary column's upper bound to `ui.powi(3)`,
rounded independently. At the exactly-feasible point `(x, s) = (ui, ui^3)` — a
point ON the graph of the function being relaxed, so it must be in every valid
relaxation — the row is **violated by −7.63e−6** (row magnitude 1.54e11; 0.5 ulp of
that magnitude, but an absolute residual no LP tolerance absorbs).

Assembling the identical node LP in Python and handing it to HiGHS confirms the
relaxation, not the in-house simplex, is at fault:

* HiGHS reports the LP **infeasible** (status 2) on a box that contains a genuinely
  feasible point — the kernel's own incumbent, objective 128893.7409, found by the
  same kernel with propagation off.
* That incumbent sits inside every one of the 144 column boxes (worst violation
[...]

PR #958, opened 2026-08-08
fix(correctness): #927 an undecided node LP is not a proof of emptiness

### LINKED ISSUE #927: correctness: ex1252 returns optimal with a dual bound 68% above the true optimum under the #707 reform flags

jkitchin:
Closes #927.

## The defect

`ex1252` under `DISCOPT_INTEGER_MULTILINEAR_REFORM=1` + `DISCOPT_MULTILINEAR_COUPLING_RLT=1`
returned a **certified** `optimal` at `216826.5179` against a true optimum of
`128893.74`. A false certificate — the failure mode CLAUDE.md §1 treats as a
full-stop regression.

## Root cause

The node LP was returning `LpStatus::Numerical` — the simplex honestly reporting
that it could not solve that LP reliably — and the spatial tree discarded the
region anyway:

```rust
if node.status != LpStatus::Optimal { continue; }
```

`continue` here means *this region is empty, nothing in it can beat the incumbent*.
That is only true for a **Farkas-certified** `Infeasible`. For `Numerical`,
`IterLimit`, and `Unbounded` the LP has declined to decide, and pruning on it
fabricates an emptiness proof the solver never had. On `ex1252` the region pruned
this way was the one containing the true optimum.

Confirmed directly, not inferred: a temporary `eprintln!` behind
`DISCOPT_DEBUG_NODE_STATUS` printed `[node-status] fathomed as empty with status Numerical`.

The `LpSolve` doc contract already states this rule — that an unproven verdict must
*"force a fallback, never an unsound fathom"* — and `convex_kernel.rs:1431` already
implements it (`if r.status != LpStatus::Infeasible { uncertified_drop = true; }`,
commented "A PROVEN-infeasible node is a legitimate fathom … Unbounded / numerical /
iteration-limit is NOT"). The spatial tree was the one place that did not.

## The fix

Separate *proof* from *failure to decide*, as an explicit exhaustive mapping so a
future `LpStatus` variant cannot silently inherit "prune it":

```rust
fn verdict_for(status: LpStatus) -> NodeVerdict {
    match status {
        LpStatus::Optimal    => NodeVerdict::Bound,
        LpStatus::Infeasible => NodeVerdict::EmptyRegion,
        LpStatus::Unbounded | LpStatus::IterLimit | LpStatus::Numerical => NodeVerdict::Undecided,
    }
}
```

`EmptyRegion` fathoms as before. `Undecided` **branches** the node on its widest
original column, carrying the (valid) inherited `parent_bound`; when no branchable
column remains the node closes against that honest bound. Termination is preserved
— an undecided node either shrinks or closes — and the bound it contributes is one
the solver actually proved.

`SpatialTreeResult` gains `n_undecided`, exposed through the Python bindings, so the
rate of undecided node LPs is observable rather than invisible. It is the natural
net-positive metric for #956.

## Class-level audit

Per CLAUDE.md §2 this was fixed as a class, not an instance. Every `!= LpStatus::Optimal`
site outside `lp/` was checked: `convex_kernel.rs:1431` already draws this exact
distinction; `convex_kernel.rs:580` returns a sound sentinel; `milp_driver.rs:490` is a
cut-loop `break`, not a fathom; `convex_kernel.rs:2107` is a W2 A/B probe;
`convex_kernel.rs:1884` returns `None` from a dive. The spatial tree was the sole gap.
No instance-keyed special case was added.

## Verification

* `python/tests/test_integer_multilinear_reform.py::test_ex1252_multilinear_reform_sound_and_lifts_bound`
  — **1 passed in 200.53s**. Fails before this commit (bound `216826.52` > optimum
  `128893.74`), passes after. Requires `-m slow`.
* Full solve, both arms, 200 s (`ARMS_EXECUTED=2` asserted so the probe cannot
  silently measure nothing):

  | arm | status | objective | bound | nodes | sound |
  |---|---|---|---|---|---|
  | reform+RLT (the #927 configuration) | `time_limit` | – | 20520.45 | 30338 | ✅ |
  | no flags (control) | `feasible` | 134263.59 | 100509.01 | 607 | ✅ |

  The #927 configuration is now honest-but-weak where it was previously a false
  certified `optimal`. On the propagated box alone the fix reaches
  `bound = 128893.7408` against the true optimum `128893.7409` — tight and below it.
* `cargo test -p discopt-core` — **584 passed, 0 failed**. Clippy clean on the touched
  file; `cargo fmt` via pre-commit.
* New unit test `only_certified_infeasible_proves_a_region_empty` pins the mapping.
  An earlier synthetic reproducer was **discarded because it passed with and without
  the fix** (diagnostic: `status=Optimal … undecided=0` — the spec never triggered
  `Numerical`). Per CLAUDE.md §6 a test that cannot fail is not a test.
* `pytest -m smoke` and `pytest -m slow python/tests/test_adversarial_recent_fixes.py`
  — to be reported in a comment; deliberately not run yet, because they would load the
  machine while the differential panel below is measuring (CLAUDE.md §9).
* Differential panel over the 81 in-repo `python/tests/data/minlplib` instances,
  baseline (`8bfce1f1`) vs fix, interleaved-arm CSV — **in flight, posted as a comment
  when complete.** Criteria: instances with `n_undecided == 0` must be exactly
  unchanged (status, nodes, objective, bound — bound-neutral regime, CLAUDE.md §5);
  instances with `n_undecided > 0` are where the fix bites and are checked for
  soundness against `minlplib.solu`. Instances that hit the time limit in either arm
  are compared for soundness only, since their node counts are wall-clock dependent
  and not a bound-neutrality signal.

## Scope explicitly NOT in this PR

The conditioning defect that *produces* the undecided LPs — McCormick envelope rows
built by cancellation at magnitude ~1e11 that cut off the exactly-feasible point
`(ui, ui³)`, confirmed independently by HiGHS — is filed as **#956**. It is
bound-changing across every relaxation in the solver and needs its own corpus-wide
differential panel under CLAUDE.md §5; conflating it with a soundness fix would make
neither reviewable. #957 covers a separate denormal issue found on the way.

This PR makes the search sound. #956 makes it fast again.

### DIFF
--- a/crates/discopt-core/src/bnb/spatial_tree.rs
+++ b/crates/discopt-core/src/bnb/spatial_tree.rs
@@ -55,6 +55,36 @@ impl Ord for QNode {
     }
 }
 
+/// What a node LP's terminal status licenses the tree to conclude about the region
+/// the node covers.
+///
+/// #927: this mapping is the soundness hinge of the whole search. Only a
+/// Farkas-certified [`LpStatus::Infeasible`] proves a region is empty; every other
+/// non-optimal status is the LP declining to decide, and pruning on it fabricates
+/// an emptiness proof the solver never had. The `match` is exhaustive on purpose —
+/// a new `LpStatus` variant must make an explicit choice here rather than silently
+/// inheriting "prune it".
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum NodeVerdict {
+    /// The LP solved: its safe bound is a valid lower bound for the region.
+    Bound,
+    /// The region is PROVEN empty (certified infeasible) — it contributes nothing.
+    EmptyRegion,
+    /// The LP decided nothing. The region may well contain the optimum, so it must
+    /// be branched (or closed with its inherited bound), never pruned.
+    Undecided,
+}
+
+fn verdict_for(status: LpStatus) -> NodeVerdict {
+    match status {
+        LpStatus::Optimal => NodeVerdict::Bound,
+        // `Infeasible` is returned only after the simplex verifies a Farkas dual
+        // ray (see `lp::simplex`), so it is a proof.
+        LpStatus::Infeasible => NodeVerdict::EmptyRegion,
+        LpStatus::Unbounded | LpStatus::IterLimit | LpStatus::Numerical => NodeVerdict::Undecided,
+    }
+}
+
 /// Termination status of the tree solve.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum TreeStatus {
@@ -171,6 +201,12 @@ pub struct SpatialTreeResult {
     /// subtree of them freezes the frontier — the diagnostic for a bound plateau
     /// caused by certification failure rather than relaxation looseness.
     pub n_uncertified: usize,
+    /// #927: nodes whose LP came back with a status that DECIDES NOTHING
+    /// (`Numerical` / `IterLimit` / `Unbounded`). These are branched rather than
+    /// fathomed — a nonzero count means the search hit ill-conditioned node LPs and
+    /// paid for them in extra nodes, which is exactly the trade that keeps the
+    /// certificate honest. Reported so "the fallback fired" is observable.
+    pub n_undecided: usize,
     /// #917: seconds of the caller's withheld reserve this search actually reclaimed
     /// (0.0 when it never did). Reported so "the extension fired" is directly
     /// observable instead of inferred from a wall-clock reading — a panel that can
@@ -264,6 +300,7 @@ pub fn solve_spatial_tree(
     let mut node_count = 0usize;
     let mut n_lp_solves = 0usize;
     let mut n_uncertified = 0usize;
+    let mut n_undecided = 0usize;
 
     // Global lower bound = min, over every region that leaves the tree WITHOUT being
     // subdivided (pruned / infeasible / feasible-leaf / width-exhausted), of a valid
@@ -313,6 +350,7 @@ pub fn solve_spatial_tree(
                 node_count,
                 n_lp_solves,
                 n_uncertified,
+                n_undecided,
                 incumbent_extension_s: extension_s,
             };
         }
@@ -338,6 +376,7 @@ pub fn solve_spatial_tree(
                 node_count,
                 n_lp_solves,
                 n_uncertified,
+                n_undecided,
                 incumbent_extension_s: extension_s,
             };
         }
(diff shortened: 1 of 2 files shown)


jkitchin:
## Differential panel: PASS

Baseline (`8bfce1f1`) vs this branch, over all 81 in-repo `python/tests/data/minlplib`
instances at a 20 s limit. Both arms: `INSTANCES_SOLVED=81`, `INSTANCES_ERRORED=0`,
exit 0. Comparison ran **144 executed assertions** (the count is printed and the
script exits non-zero if it is ever 0, so the panel cannot pass by measuring nothing).

```
SOUNDNESS FAILURES (blocking):        0
BOUND-NEUTRALITY FAILURES (blocking): 0
ERRORED:                              0
```

**Bound-neutral set (68 instances, both arms terminal): exactly unchanged** — same
status, same node count, same objective, same bound. This is the important result:
the fix only touches nodes whose LP returned neither `Optimal` nor certified
`Infeasible`, and on every instance that never hits such a node the behaviour is
bit-for-bit identical.

**Soundness**: no bound crosses its `minlplib.solu` optimum, on either arm. Bound
sense is inferred per instance from that instance's own `(bound, objective)` pair
rather than assumed; two instances (`carton7`, `hda`) could not have their sense
inferred because neither arm found an incumbent — both are **bit-identical between
arms** (same bound, same node count), so nothing about them is in question here.

### The panel found a second instance hit by this defect

`nvs20` moves `optimal(105 nodes, 2.44 s)` → `time_limit(4078 nodes)`. That looks like
a regression and is worth being precise about:

| | baseline | this branch |
|---|---|---|
| status | `optimal` | `time_limit` |
| incumbent | 230.9221652227346 | 230.9221652227346 |
| bound | 230.9221652227346 | 227.6525348945811 |

MINLPLib's optimum for `nvs20` is **230.9221652**. Both arms find the right answer.
What changed is that baseline *claimed* it and this branch does not.

That claim was falsifiable, so I measured it rather than asserting it: wrapping
`_rust.solve_spatial_tree_py` on the real solve path (no reconstructed box) reports

```
nvs20: n_undecided=2056 over 4192 kernel nodes    (n_uncertified=0)
```

**49% of nvs20's node LPs are undecided.** Baseline fathomed every one of them as
proven-empty, and that is how it "closed" the gap in 105 nodes. Its answer was right;
its certificate was not earned. Had the undecided region contained a better point —
as it does on `ex1252` — the same code would have returned a false optimum. `ex1252`
and `nvs20` are the same bug with different luck.

This is the honest cost of the fix, and I am not going to dress it up: `nvs20` really
did go from 2.44 s to not closing in 20 s. The gap is recoverable, but by fixing the
conditioning that makes those LPs undecided in the first place (#956), not by
resuming an unproven fathom.

The other 12 non-terminal instances are compared for soundness only — their node
counts are wall-clock dependent, so treating a difference there as signal would be a
CLAUDE.md §9 timing artifact. For the record they are stable anyway: 7 are identical,
and the rest move within noise (`ex1233` 71→55, `ex1252` 55→57, `nvs19` 27169→27659,
`nvs23` 12989→12919, `nvs24` 6455→6470).

### Panel hygiene

The arms ran sequentially with no other load; I deliberately held back `pytest -m smoke`
and the adversarial suite until the panel finished, since load landing on one arm and
not the other is exactly how a panel like this produces a confident wrong answer.

I also started to switch this panel to a deterministic `max_nodes` budget for exact
reproducibility, and reverted that: `solver.py:627` records a prior measurement that
`node_limit` sends the kernel back to the Python path, which would have made the panel
compare OFF against OFF and pass vacuously.

jkitchin:
## Required suites (local, `wt927` worktree, machine clear of the panel)

```
pytest -m smoke                                          920 passed, 1 skipped, 2 xpassed, 8243 deselected   115.82s   exit 0
pytest -m slow python/tests/test_adversarial_recent_fixes.py    10 passed                                    145.16s   exit 0
cargo test -p discopt-core                               584 passed, 0 failed
```

Plus the #927 regression test itself, which fails on `main` and passes here:

```
python/tests/test_integer_multilinear_reform.py::test_ex1252_multilinear_reform_sound_and_lifts_bound
    1 passed in 200.53s     (needs -m slow)
```

Issue #732, opened 2026-07-18
Certify ex1252 / the gated-configuration MINLP class — staged plan (search + engine, post-#721)


jkitchin:
Successor to #707 → #721. #721 delivered the objective-coupling RLT (`DISCOPT_MULTILINEAR_COUPLING_RLT`, default-OFF, PR #726), which removed the load-bearing relaxation wall (config-node bound 12658 → 57435) and — per the compounding probe — **unlocked the previously-falsified levers**: `x6` subdivision now lifts a child bound 57435 → 62071 (provably 0% pre-RLT), and OBBT now pins `x12` exactly and caps `x15 ∈ [12.44, 30.89]` in seconds (fully inert pre-RLT). The remaining distance to certification is **search orchestration + engine robustness**, not new relaxation mathematics.

Full anatomy + plan: **`docs/dev/ex1252-certification-plan.md`**. Reproduction: `discopt_benchmarks/scripts/ex1252_compounding_probe.py`.

## Five-layer anatomy (all measured)

| layer | mechanism | status |
|---|---|---|
| L1 objective multilinear decoupling | dual floor 5134 | fixed (#707) |
| L2 coupling-aux decoupling (`v_k = z_k·x15` → 0 at fractional bits) | node 12658 → 57435 | fixed (PR #726, default-OFF) |
| L3 **search order** — global dual set by shallow indicator-fractional nodes | 4.5× node lift → only +1.4% global | disjunctive root pass (Stage 2) |
| L4 residual cubic gap at configs (`min x15` 12.44 vs 32.28 needed) | compounding unlocked | spatial bisection (Stage 4) |
| L5 engine fragility — narrow/OBBT-pinned boxes return 0.0/`(numerical)`/`(error)` fallback bounds | measured in the probe | fixed (Stage 1, #758) |

## Stages (each with entry experiment + kill criterion — see the plan doc)

1. **Engine hardening on narrow/pinned boxes** (P0): certified verdicts or parent-bound inheritance — never a 0.0 floor. Benefits every instance. **DONE (#758).**
2. **Configuration-first branching**: indicators → reform bits → pump counts → continuous. Hint route killed → **disjunctive root pass DONE** (default-OFF).
3. **Per-config OBBT+cutoff loop** at integer-complete nodes — subsumed by the disjunctive pass.
4. **Cubic tightening at configs**: spatial `x6` branching (**DONE**), piecewise-McCormick / αBB (held in reserve).
[...]

PR #960, opened 2026-08-08
fix(#957): stop outward rounding from turning an exact zero into a subnormal

### LINKED ISSUE #957: interval.py outward rounding turns exact 0.0 into a denormal, which makes the LP conditioning test overflow to inf

jkitchin:
`interval.py`'s `_round_down`/`_round_up` are `np.nextafter(x, ∓inf)`, and
`nextafter(0.0, ±inf)` is `±5e-324`. At every other magnitude "one ULP" is a
negligible relative widening; at zero it crosses into the subnormal range and
manufactures a coefficient ~300 orders of magnitude below everything else in
the model. Measured on the in-repo corpus: 384 of 730 relaxation LPs carried
such a value, and in 67 of them the coefficient-spread guard at
milp_relaxation.py:548 (`nz.max() / nz.min() > TRIGGER`) *overflowed to inf*
while making its decision, so equilibration fired unconditionally regardless of
how benign the matrix was.

The obvious fix — leave every exact zero alone — is unsound and was NOT taken.
The nudge at zero is load-bearing wherever the producing operation can underflow:
`fl(exp(-800))` is `0.0` while the exact value is `3.6e-348`, so a blanket
special-case would give `exp` an upper endpoint below its own image. What ships
is an opt-in pair, `_round_down_exact0`/`_round_up_exact0`, applied only where
`fl(result) == 0 ⟹ result == 0` is provable:

  * `+`, `-`, `width`  — every double is a multiple of 2**-1074, so a float sum
    is zero only when the exact sum is (fuzz-checked against `longdouble` over
    14 268 cancelling pairs);
  * `abs`, `sqrt`, `log`, the reciprocal — no underflow region;
  * `*` and `**2` — behind an explicit no-underflow guard on the corner factors,
    checked only when an endpoint is actually zero so the AD hot path is unaffected.

The generic monotone wrapper keeps the unconditional helpers, so a future atom
inherits the safe behaviour by default.

Second, independent half: `_coefficient_spread_exceeds()` replaces the open-coded
spread test at both conditioning seams. It uses the overflow-free cross-multiplied
form (already applied at the false-infeasible seam by #732 Stage 1-B) *and* drops
entries below the smallest normal double, so the guard measures the conditioning
of the model rather than of an underflow artifact. Both halves were needed: the
interval fix clears `A_ub` on 14 of the 15 affected instances, and `st_e11`'s
subnormals — from a different producer — are neutralised by the floor (19 → 0).

Found while fixing the above: `psd_2x2_sufficient` had the same bug pointing the
other way. It skipped its *upward* round when `fl(off**2) == 0.0`, which is not
an exact zero when `off` is ~1e-200 — the exact square is ~1e-400 — so a matrix
with a barely negative determinant could pass the PSD test. The zero test now
reads the factors, not the flushed product. `gershgorin_lambda_min`/`_max` get
the same treatment, so a linear function's all-zero Hessian bounds come back as
exactly 0.0 instead of ∓5e-324 and no longer miss a boundary `λ_min ≥ 0` test.

Verification
  * Differential corpus panel, 66 instances, 10 s budget, arms interleaved:
    0 soundness violations against the 16 oracle-checked instances, total node
    count identical (3298 → 3298), subnormal-carrying LPs 384/730 → 28/782,
    spread-test overflows 67 → 0.
  * The panel's one certification change (cvxnonsep_nsig30) and every bound/node
    difference were on instances sitting within ~1 s of the deadline. Re-measured
    unloaded with the arms interleaved (20 solves, 2 instances × 2 budgets ×
    replicates), cvxnonsep_nsig30 and st_e36 are bit-identical between arms —
    SAME in all four cells. Recorded as a timing artifact, not a regression.
  * `pytest -m smoke python/tests/`: 916 passed, 15 skipped.
  * `pytest -m slow python/tests/test_adversarial_recent_fixes.py`: 9 passed,
    1 failed — `test_large_dense_jacobian_no_crash`, a wall-clock deadline
    assertion that ran while the smoke suite held the machine. Re-run unloaded on
    both trees it passes in 28.00/27.46 s (before) and 28.36/27.26 s (after).
  * Interval/convexity slices: 168 passed. New regression file: 29 tests, 10 of
    whose behavioural assertions fail on the pre-change tree.

Closes #957

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01TJpDTK6fNERipKaSfDrhLC


jkitchin:
## Both red jobs addressed

### `Python fast` — three stale pins, narrowed not deleted (165e9960)

All three failures are tests pinning the *old*, looser behaviour, now catching the tree being better than the pin. Same mechanism in three places: the wrong-curvature concave-`sqrt` relaxation went unbounded because the inner square's exact `0` minimum outward-rounded to `-5e-324` and poisoned `sqrt`'s domain; #957 removes that nudge where `fl(result) == 0 ⟹ result == 0` is provable.

| test | pin | disposition |
| --- | --- | --- |
| `test_discopt_unbounded_set_pinned` | `{euclidean_norm::objective, norm_le::soc}` | set becomes empty, **not removed** — the "newly unbounded" arm still fires and `test_corpus_fully_compared` still guards the loop |
| `test_monotonicity_gaps_pinned` | `KNOWN_SUSPECT_STRONGER ∋ cubic::objective` | drops `cubic` only; `sqrt_concave` and `sine` stay pinned |
| `test_norm_unbounded_off_becomes_finite_on` | `assert st_off == "unbounded"` | premise falsified (OFF is now `optimal` at 0.3536); renamed and rewritten to assert the **stronger** claim |

`test_norm_atom` now asserts both arms finite *and sound* (`≤ √2`), the atom strictly tighter (`on > off + 0.5`), and exact (`|on − √2| < 1e-3`). The falsified premise is recorded in its docstring. Locally: 16 passed across the three files.

### `Python claim-boundary` — regenerated, attributed against a `main` control (3d240e2e)

A naive regeneration was **not** attributable. The committed baseline is stamped `605d29b` and its `root_lp_bound` column had gone stale, so regenerating on this branch alone showed **56** bound moves and read as though this PR had caused them all — including `nvs13` −4843.01 → −751.12 and five instances (`beuster`, `casctanks`, `st_miqp2/3/4`) dropping to no recorded root bound, which would have looked like losses introduced here.

So the baseline was generated a second time on a clean `origin/main` worktree (`d5f6eff6`, the commit this branch merged), with the control arm **asserted to lack the marker** (`_round_down_exact0` absent) before generating:

```
committed 605d29b -> main d5f6eff6   (62 checked)  SHAPE_MOVED=0  ROOT_LP_MOVED=52
main d5f6eff6     -> this branch     (66 checked)  SHAPE_MOVED=7  ROOT_LP_MOVED=18
TOTAL_COMPARISONS_EXECUTED=128
```

All 52 root-LP drifts predate this branch. The *shape* column was never stale — which is exactly why the shape-only gate has been passing on `main` throughout. The second row is this PR's real footprint.

**The 7 shape movers are rows-only** — columns and integer columns unchanged everywhere, the signature of extra cuts rather than a different formulation — and they are this job's list instance for instance:

```
bchoco06  832 -> 848     fac2    54 -> 66
bchoco07 1081 -> 1163    tspn08 604 -> 609
bchoco08 1604 -> 1898    tspn10 935 -> 940
                         tspn12 1340 -> 1345
```

An independent two-arm probe in a single tree (OFF = the opt-in helpers rebound to the unconditional `_round_down`/`_round_up` in both modules, plus the pre-#960 `psd_2x2_sufficient` at *both* its import sites) reproduces exactly that set, and its OFF arm reproduces the committed baseline shape for all 62 rows — `CONTROL_MATCHED_BASELINE=62 MISMATCHED=[]`. Without that control the comparison would be meaningless.

**Soundness.** Of the 18 bound moves, 12 are last-digit (≤1e-10 relative). The 6 substantive ones are tightenings that stay below the reference optimum — all six instances minimize (`O0 0`), so a valid dual bound may rise but never past the optimum:

| instance | `main` | this branch | reference optimum |
| --- | --- | --- | --- |
| `fac2` | none | 303 398.521 | 331 837 498.2 |
| `tspn08` | none | 231.983 | 290.567 |
| `tspn10` | none | 161.161 | 225.126 |
| `tspn12` | none | 183.326 | 262.647 |
| `chance` | 26.350 | 28.943 | 29.894 |
| `nvs04` | −5e-322 | **0.0** | 0.72 |

`nvs04` is the bug itself sitting in the baseline: the recorded root bound *was* the subnormal artifact.

Solving the relaxation MILP rather than the root LP tells the same story: `fac2` no usable bound → 2 132 258.914, and `tspn08/10/12` go from **unbounded** to a finite valid bound (246.211 / 183.117 / 203.126), all below their optima, `UNSOUND=0`.

### Two things stated plainly

1. The 52 pre-existing `root_lp_bound` drifts ride along because the file is generated wholesale. They are recorded as **measured on `main`**, not claimed as this PR's doing. The baseline's `root_lp_bound` column having gone stale since `605d29b` while its shape column stayed current is worth a follow-up — the gate never noticed because it only asserts shape.
2. `bchoco06/07/08` have no `minlplib.solu` entry, so the table above cannot cover them. Their root LP bounds move by ~4e-14 (`-1.0000000000000826` → `-1.0000000000000866` and siblings) — last-digit, not a loosening. A full-solve check in both arms is running to confirm the dual bound never cuts off a feasible incumbent; note these three **maximize** (`O0 1`), so that inequality is inverted from the min convention, and the probe reads the sense off `model._objective.sense` rather than assuming it.

jkitchin:
## `bchoco06/07/08` closed out — the three with no oracle entry

Full solve in both arms (300 s each), which is the only check available for instances absent from `minlplib.solu`:

| instance | arm | status | incumbent | dual bound |
| --- | --- | --- | --- | --- |
| `bchoco06` | OFF | feasible | 0.9627726085530692 | 0.9750000000001947 |
| `bchoco06` | ON | feasible | 0.961095222716719 | 0.9750000000002257 |
| `bchoco07` | OFF | feasible | 0.9581492565555937 | 1.000000000000228 |
| `bchoco07` | ON | time_limit | — | 1.000000000000228 |
| `bchoco08` | OFF | time_limit | — | 1.000000000001055 |
| `bchoco08` | ON | time_limit | — | 1.000000000001055 |

```
COMPARISONS_EXECUTED=5
FALSE_BOUND=0: []
BOUND_LOOSER=0: []
```

The ON bound never cuts off a feasible point either arm found (0.97500 ≥ 0.96277; 1.00000 ≥ 0.95815), and it never loosens: `+3.098e-14` on `bchoco06`, exactly `0` on the other two.

### A correction to my own instrument

The first pass of this probe reported `FALSE_BOUND=2: ['bchoco06', 'bchoco07']`. **That was a bug in the probe, not a finding.** All three `bchoco` instances *maximize* (`O0 1` in the `.nl` header), so the dual bound is an **upper** bound and both invariants flip — sound means `bound ≥ incumbent`, and tighter means the bound goes *down*. The comparison had been written in the min-sense convention, so it flagged every correct bound. The rewritten version reads the sense from the model (`model._objective.sense`) / the `.nl` header rather than assuming it, and the table above is its output.

The four oracle-checked instances in the previous comment are unaffected: `fac2` and `tspn08/10/12` all *minimize* (`O0 0`), so `bound ≤ reference optimum` was the right direction there and `UNSOUND=0` stands as measured.

### One thing not to over-read

`bchoco07`'s ON arm found no incumbent inside 300 s where OFF did. That is a wall-clock artifact, not a regression: the ON arm ran while two baseline regenerations were occupying the machine, so it had materially less CPU than the OFF arm. Per the repo's own timing discipline, a load-confounded A/B is not a measurement — and it needn't be one here, since the dual bounds it is being judged on are bit-identical between arms (`1.000000000000228` both sides) and the soundness check does not depend on how far either arm searched.

jkitchin:
The `root_lp_bound` follow-up promised above is filed as #961. Two compounding defects, neither in this PR's scope:

1. `MccormickLPRelaxer.solve_at_node` returns `status='optimal'` with `lower_bound=None` on `beuster`, `casctanks`, `st_miqp2/3/4` — all five had finite root bounds at `605d29b`.
2. `gen_claim_baseline._root_lp_bound`'s bare `except Exception: return None` swallows the `TypeError` that `np.isfinite(None)` raises, so the baseline records "no bound" for what is actually a crash.

The shape-only gate is why 52 root-LP drifts accumulated on `main` unnoticed. This PR changes none of that behaviour; the attribution above stands.

PR #962, opened 2026-08-09
fix(correctness): #932 verify POUNCE's cross-thread NlProblem guarantee

### LINKED ISSUE #932: pyo3 unsendable panic: PyNlProblem sent to another thread during result marshaling
## What happens

A pyo3 unsendable panic surfaces during result marshaling:

```
_pounce::nl_problem::PyNlProblem is unsendable, but sent to another thread
```

It does not stop the solve — the run completes and reports a result — but a pyo3
unsendable violation is not a cosmetic warning. `#[pyclass(unsendable)]` exists
precisely because the type holds state that is not safe to touch from another
thread; pyo3 detects the cross-thread access at drop/access time and panics rather
than letting the unsound access proceed. That it currently only prints is a
property of *where* it lands, not evidence that the access was harmless.

## Where it was seen

Surfaced while reading `PyTreeManager.import_results`. Verified **pre-existing** —
it reproduces on trees without the #930 change (PR #931), so it is not that PR's
and is not being fixed there.

## Why it matters

The object is a `PyNlProblem` from the POUNCE bridge. Two things worth establishing:

1. **Which thread is doing the sending.** If the solver hands a `PyNlProblem` (or
   something holding one) to a worker thread or a thread-pool executor, the panic is
   a real thread-safety bug that happens to be caught. If it is only the *drop*
   landing on a different thread than the one that created it, the fix is to pin the
   object's lifetime to its creating thread.
2. **Whether the panic can ever land somewhere that matters.** A panic that unwinds
   through pyo3 during a solve, rather than during teardown, would abort a run. It
   is currently benign by luck of placement.

## What "done" looks like

Root-cause it: identify the construction site and the crossing site (a stack trace
with `RUST_BACKTRACE=1` at the panic point is the first step), then either keep the
object on one thread or, if the type is genuinely safe to send, establish that and
drop `unsendable`. Not: silencing the message.

Per CLAUDE.md §3, the cheap version — catching or suppressing the panic — is
explicitly out of scope; that converts "this path is broken" into "this path is
fine", which is the failure mode §7 was written about.

jkitchin:
## Summary

**This PR was rewritten. Its original premise was falsified by measurement — see the retraction below.**

Closes #932. #932 offered two routes: keep POUNCE's `NlProblem` on one thread (discopt side), or make it cross-thread safe (POUNCE side). **POUNCE took the second route**: pounce#477 ("Let one NlProblem serve a whole worker pool", `ded7dadd`, 2026-08-04) dropped the pyo3 `unsendable` marker from `PyNlProblem` after establishing that `NlTnlp` is `Send` and that every `&self` method evaluates under the GIL. Against the POUNCE that CI and local both load, there is **no discopt-side thread-affinity bug**, and the model-level cache (one base shared by every thread that solves a node) is correct as written — it is exactly what pounce#477 exists to enable.

What ships instead is the *requirement made explicit rather than assumed*. POUNCE's version string cannot express it — pounce#477 landed inside an unreleased `0.9.0`, so pre- and post-fix builds are both `"0.9.0"` and the `pounce-solver>=0.9` pin cannot separate them. So `nlp_native` **measures** the guarantee once per process on the first base built (a real second thread touching the problem; `BaseException` caught there because a pyo3 panic is one; `RuntimeError` if the probe failed to actually cross a thread) and disables the native path with a warning naming pounce#477 when it does not hold, degrading to the existing JAX callback bridge rather than letting an uncatchable panic abort a solve.

The stale "enable explicitly once `PyNlProblem` is made Send-safe" note in `solver.py` is corrected, and CHANGELOG records the requirement.

## Retraction (CLAUDE.md §11)

The first version of this branch claimed `PyNlProblem` is `unsendable` and moved the base cache into `threading.local`. Both the claim and the fix are withdrawn:

- **Runtime, on the loaded binary**, from a thread asserted distinct from the creating one (`creating tid=8288869760`, `worker tid=6140096512`, `DISTINCT=True`): `objective(x0)`, `gradient(x0)` and `variant(lb,ub,x0)` all return normally. `CHECKS_EXECUTED=4`, `PANICS=[]`.
- **The original tests did not reproduce the bug.** Run against pre-fix sources (with `_TLS` asserted *absent* to confirm the arm was real): the drop test — the crossing #932 actually observed — **passed pre-fix**, i.e. was vacuous; the access test failed on its cache-identity assertion (`assert out["distinct"]`), not on the panic guard one line above; the third failed only with `AttributeError: ... has no attribute '_TLS_MAX_ENTRIES'`. **2 failed, 1 passed, 0 reproduced.**
- The reverted change's `assert nb is not nb_main` had pinned *per-thread rebuilds* as required behavior — the opposite of upstream's design.

An earlier inference of mine, that the local build predated pounce#477 because the `.so` mtime (Aug 3 21:48) precedes `ded7dadd` (Aug 4), is also withdrawn: the runtime probe says otherwise and the measurement wins.

## Correctness

No solver-math change. The cache topology is main's; the only behavioral addition is a guard that can **disable** the opt-in native path (never enable it), and the path stays default-OFF. Flipping `DISCOPT_NLP_NATIVE` on is a §5 graduation-panel decision, not part of this PR.

- [x] `pytest -m smoke` → **925 passed, 1 skipped, 2 xpassed in 52.20 s**
- [x] `pytest python/tests/test_932_native_base_thread_sharing.py test_relaxer_native_obbt_units.py test_native_nlp_equivalence.py -m "slow or not slow"` → **27 passed** (import asserted to load from the worktree, per §8)
- [x] `ruff check` / `ruff format --check` / `mypy` clean (pre-commit)
- [x] `cargo test -p discopt-core` → N/A — no Rust touched

## Regression test

`python/tests/test_932_native_base_thread_sharing.py` (renamed from `..._thread_affinity.py`). Its docstring states plainly that these are **not** fail-before/pass-after tests for a bug that no longer exists — they pin the guarantee and the guard, so a future unsendable build fails loudly instead of silently. Each proves its probe fired (§6):

- `test_one_base_serves_a_second_thread` — asserts the worker tid differs from the main tid before believing anything; one base, shared, solves to `OPTIMAL` at the known optimum.
- `test_model_cycle_collected_on_foreign_thread_is_silent` — asserts the model really was cyclic garbage (survives `del`, dies on the foreign `gc.collect()`) and that no unsendable unraisable was emitted.
- `test_capability_probe_reports_support_and_is_measured_once` — probe returns True; a counting probe is called exactly once per process.
- `test_native_path_disabled_when_cross_thread_use_is_unsupported` — **mutation-checked**: deleting the two-line guard from `get_native_base` makes this test fail.
- `test_probe_refuses_to_pass_if_it_did_not_cross_a_thread` — the probe raises rather than returning a verdict it did not earn.

- [x] Added a regression test that fails before / passes after (or: N/A — see the docstring: the bug was fixed upstream; these pin the guarantee and are mutation-checked instead)

https://claude.ai/code/session_01GZrQpYYd9tbLnFEH8gZokP


jkitchin:
**Rewritten at `706af692` — the original premise was falsified, and the title/body above were replaced.** Flagging it here so anyone who read the earlier version does not act on it.

What was claimed: POUNCE's `PyNlProblem` is a pyo3 `unsendable` pyclass, so discopt's model-level base cache hands it across threads and must move to `threading.local`.

What is true: pounce#477 ("Let one NlProblem serve a whole worker pool", `ded7dadd`, 2026-08-04) dropped the marker after establishing `NlTnlp` is `Send` and that every `&self` method evaluates under the GIL — the second of the two routes #932 itself offered. Measured against the loaded binary from a thread asserted distinct from the creating one: `objective`, `gradient` and `variant` all return normally, `CHECKS_EXECUTED=4`, `PANICS=[]`.

And the original tests did not reproduce the bug they were written for. Against pre-fix sources with `_TLS` asserted absent: the drop test — the crossing #932 actually observed — **passed pre-fix** (vacuous); the access test failed on its cache-identity assertion, not the panic guard one line above it; the third failed only with `AttributeError` on `_TLS_MAX_ENTRIES`. 2 failed, 1 passed, 0 reproduced.

So the `threading.local` change is reverted — its `assert nb is not nb_main` had pinned per-thread rebuilds as *required* behavior, the opposite of what upstream designed. What ships is the requirement measured instead of assumed: `pounce-solver>=0.9` cannot express it (pounce#477 landed inside an unreleased `0.9.0`, so both builds are `"0.9.0"`), so `nlp_native` probes the guarantee once per process and disables the opt-in native path — never enables it — with a warning naming pounce#477 when it does not hold.

Also withdrawn: my earlier inference that the local build predated pounce#477 because the `.so` mtime precedes `ded7dadd`. The runtime probe says otherwise and the measurement wins.

The branch was additionally rebased onto `main` (it was `CONFLICTING`, which is why GitHub had created no check runs for the previous head at all), and a stale `solver.py` reverted by that earlier conflict resolution — it had undone #952/#954's `_matrix_solution_feasible` / `_matrix_solution_violations` work — was restored to `main`'s version. The net diff is now 12 lines of comment in `solver.py`, +70 in `nlp_native.py`, +31 CHANGELOG, +249 test.

PR #963, opened 2026-08-09
ci: repair two signals that reported without measuring

jkitchin:
## Summary

Two CI defects found while driving #960 to green. Both share a shape: a check that answers confidently from something other than the thing it claims to be testing. Neither has a tracking issue — they were surfaced by the #960 work and are fixed here directly.

### 1. A manual `workflow_dispatch` skipped the entire heavy matrix

The `changes` gate reads `github.event.before`, which exists **only on `push`**. On a dispatch it expands to the empty string, the first `git diff` fails, and the fallback silently diffs `HEAD~1..HEAD` — so the gate answered as if only the tip commit had changed.

Dispatching CI on #960, whose tip touched `CHANGELOG.md` and `docs/dev/data/claim-baseline.jsonl`, produced `code=false` and left Rust, `Python fast`, `Python claim-boundary`, `Python correctness` and AMP all `skipped` — **nine checks that render as *not red* while executing nothing.** That is the #953 failure mode, one trigger over. Measured, from the job's own log:

```
changed="$(git diff --name-only "" "3d240e2e…" || git diff --name-only HEAD~1 HEAD)"
Changed files:  CHANGELOG.md,  docs/dev/data/claim-baseline.jsonl
=> code=false (heavy test jobs will SKIP)
```

- A dispatch is now `code=true` unconditionally — pressing the button means "run everything", and there is no base to diff against anyway.
- The `push` branch no longer substitutes a guess: an all-zero `before` (a branch's first push), a non-existent object (force-push, shallow clone), or an empty one is now the **uncertainty** case the job's own comment already promised to resolve as `true`.

### 2. Two `ex1252` quality bars measured the runner, not the relaxation

`compute_disjunctive_config_bound` stops on whichever of `max_leaf_solves` and `deadline` comes first, and reported no indication of which. The leaf budget is reproducible; the clock is not. The nightly lane read 33498 (42/120 leaves) and 31567 (32/48) on a runner ~3× slower than the reference machine and failed both bars as if the bound had regressed.

**It had not.** Measured at the full budget, the bound is bit-identical across the #957 boundary:

| budget | before (`main` @ 726ae1e7) | after (#957) |
| --- | --- | --- |
| 48 leaves | 37945.427865923564 · 26 leaves · 2 pruned · 57.8 s | **37945.427865923564** · 26 · 2 · 55.5 s |
| 120 leaves | 63080.286987442756 · 59 leaves · 5 pruned · 137.1 s | **63080.286987442756** · 59 · 5 · 135.2 s |

Both match the values the tests' own docstrings record (~37.9k, 63080), both arms reached the full budget, and #957 was marginally *faster*.

`DisjunctiveConfigResult` gains `stopped_on` (`"budget"` / `"deadline"` / `"exhausted"`), mirroring `WorkBudget.stopped_on` from #912 — the same distinction between a search decided by work and one decided by the clock. The two bars now assert `stopped_on != "deadline"` **before** the bound comparison, so a slow machine and a bound regression produce different failures with different instructions, and the clock-bound message says explicitly not to lower the bar. Deadlines are re-sized as backstops at ~2× the measured runner rate (0.175 leaf solves/s → 600 s for 48 leaves, 1500 s for 120), and the nightly lane's `--timeout` goes 900 s → 1800 s so the harness timeout is never what decides a quality bar. The 120-minute job timeout remains the backstop against a hang.

The solver's own in-solve log line now reports `stopped_on` too; there it is expected to read `"deadline"`, because that path is anytime by design (the leaf cap is a runaway backstop).

> **Timing note:** the nightly fires at 06:20 UTC. If it runs before this merges, expect those two `ex1252` bars red again with the old misleading message — that is the bug this fixes, not a new one.

## Correctness

No solver math changes. `stopped_on` is pure instrumentation: a new field with a default, one construction site, set only inside the existing `_out_of_budget` predicate — it cannot change which leaves are processed or what bound is returned. Verified empirically: the `ex1252` bound is bit-identical at both budgets (table above), with identical leaf and prune counts. The workflow change can only cause *more* jobs to run, never fewer.

- [x] Cannot produce a false certificate — N/A, no solver-math change (and the one bound measured is bit-identical)
- [x] No validation, fallback, or safety guard was weakened — the reverse: two guards that were silently answering without measuring now measure, and a third (`_assert_budget_not_clock`) was added ahead of the existing bars

- [x] `pytest -m smoke python/tests/` → **916 passed, 15 skipped, 2 xpassed** (376 s) — identical to the pre-change baseline
- [x] `pytest python/tests/test_disjunctive_config_bound.py -m correctness` → **7 passed** (198 s), including both re-armed `ex1252` bars
- [x] `pytest python/tests/test_ci_changes_gate.py` → **9 passed** (2.3 s)
- [x] `ruff check` / `ruff format --check` clean; `ci.yml` parses as YAML (10 jobs, 4 triggers)

## Regression test

**`python/tests/test_ci_changes_gate.py`** (new, 9 tests) extracts the gate's *real* shell script out of `ci.yml` — not a copy, so it cannot drift — renders the `${{ github.* }}` expressions, and runs it against throwaway git repos, one case per event type. Against the pre-change workflow:

```
5 failed, 4 passed
  FAILED test_workflow_dispatch_always_runs_everything
  FAILED test_push_with_an_unusable_base_runs_everything[]
  FAILED test_push_with_an_unusable_base_runs_everything[0000…]
  FAILED test_push_with_an_unusable_base_runs_everything[not-a-sha]
  FAILED test_push_with_an_unusable_base_runs_everything[deadbeef…]
```

The 4 that pass both ways are the deliberate pins on behaviour that must **not** change: docs-only pushes still skip, a code file still runs, `schedule` still skips the PR matrix.

**`test_stopped_on_reports_the_clock_when_the_deadline_fires`** / **`…_the_budget_when_the_leaf_cap_fires`** (new) fail with `AttributeError: 'DisjunctiveConfigResult' object has no attribute 'stopped_on'` on the pre-change source and pass after. Both are fast (0.1 s and 1.9 s) — no `slow` marker needed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TJpDTK6fNERipKaSfDrhLC)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,9 +64,24 @@ jobs:
             exit 0
           fi
           if [ "${{ github.event_name }}" = "pull_request" ]; then
-            git fetch --no-tags --depth=200 origin "${{ github.base_ref }}" || true
+            # No --depth: the checkout above is already `fetch-depth: 0`, and a
+            # shallow fetch would install a boundary that hides the merge base of
+            # any branch forked more than that many base commits back.
+            git fetch --no-tags origin "${{ github.base_ref }}" || true
             base="$(git merge-base "origin/${{ github.base_ref }}" HEAD 2>/dev/null || true)"
-            changed="$(git diff --name-only "${base:-HEAD~1}" HEAD 2>/dev/null || true)"
+            # Same rule as the two branches around this one: an unresolvable base
+            # is UNCERTAINTY, not a licence to guess. `${base:-HEAD~1}` used to
+            # gate the whole matrix on the tip commit's file list, so a PR whose
+            # tip happens to be docs-only skipped every solver job even when the
+            # branch changes the solver — and `pull_request` is the trigger that
+            # fires on nearly every change here. The `git fetch` above is
+            # `|| true`, so this is reachable by a silent fetch failure alone.
+            if [ -n "$base" ]; then
+              changed="$(git diff --name-only "$base" HEAD 2>/dev/null || true)"
+            else
+              changed=""   # unknown base -> falls through to code=true below
+              echo "no merge-base with origin/${{ github.base_ref }} -> treating as uncertain"
+            fi
           else
             # `push`: diff against the pushed-from commit. If it is unusable (an
             # all-zero SHA on a branch's first push, a force-push, a shallow
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,13 +49,39 @@ jobs:
             echo "=> code=false (schedule: only the nightly correctness lane runs)"
             exit 0
           fi
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            # A manual run is a request to run EVERYTHING — that is the only
+            # reason to press the button. There is also no base to diff against:
+            # `github.event.before` exists only on `push`, so the generic branch
+            # below degenerated to `HEAD~1..HEAD` and answered as if only the tip
+            # commit had changed. That is how a dispatch of this workflow came
+            # back with rust / python-fast / claim-boundary / correctness / AMP
+            # all `skipped` — nine checks that read as non-red while executing
+            # nothing (found re-running #960; same genre as the #953 dark
+            # signals). Never gate a manual run on a guessed diff.
+            echo "code=true" >> "$GITHUB_OUTPUT"
+            echo "=> code=true (workflow_dispatch: a manual run always runs everything)"
+            exit 0
+          fi
           if [ "${{ github.event_name }}" = "pull_request" ]; then
             git fetch --no-tags --depth=200 origin "${{ github.base_ref }}" || true
             base="$(git merge-base "origin/${{ github.base_ref }}" HEAD 2>/dev/null || true)"
             changed="$(git diff --name-only "${base:-HEAD~1}" HEAD 2>/dev/null || true)"
           else
-            changed="$(git diff --name-only "${{ github.event.before }}" "${{ github.sha }}" 2>/dev/null \
-                      || git diff --name-only HEAD~1 HEAD 2>/dev/null || true)"
+            # `push`: diff against the pushed-from commit. If it is unusable (an
+            # all-zero SHA on a branch's first push, a force-push, a shallow
+            # clone) do NOT substitute a guess — an unknown base is the
+            # "uncertainty" case this job promises to resolve as `true`.
+            before="${{ github.event.before }}"
+            case "$before" in
+              ""|*[!0-9a-f]*|0000000000000000000000000000000000000000) before="" ;;
+            esac
+            if [ -n "$before" ] && git cat-file -e "$before^{commit}" 2>/dev/null; then
+              changed="$(git diff --name-only "$before" "${{ github.sha }}" 2>/dev/null || true)"
+            else
+              changed=""   # unknown base -> falls through to code=true below
+              echo "no usable base ('${{ github.event.before }}') -> treating as uncertain"
+            fi
           fi
           echo "Changed files:"; printf '%s\n' "$changed"
           # Allowlist of paths that can NOT affect solver behavior (safe to skip).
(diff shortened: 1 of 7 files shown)


jkitchin:
Reviewed at `57bdb072`. Both repairs are well-founded and the test design is the right one — extracting the `filter` step's `run:` out of `ci.yml` and executing it in a throwaway repo means the test cannot drift from the workflow, and `_render`'s `assert not leftover` keeps a renamed `${{ }}` expression from silently turning the harness into a no-op. `stopped_on` on `DisjunctiveConfigResult` mirroring `WorkBudget.stopped_on` (#912) is the right shape too: a reproducible stopping rule and a wall-clock one are different facts and should not share a bare boolean.

One finding, in scope and one line from the code this PR already touches.

## The `pull_request` arm still gates on a guessed diff

The commit message says the `push` branch "no longer substitutes a guess either" — but the `pull_request` branch directly above it still does, and it is the trigger that fires on essentially every change in this repo:

```sh
if [ "${{ github.event_name }}" = "pull_request" ]; then
  git fetch --no-tags --depth=200 origin "${{ github.base_ref }}" || true
  base="$(git merge-base "origin/${{ github.base_ref }}" HEAD 2>/dev/null || true)"
  changed="$(git diff --name-only "${base:-HEAD~1}" HEAD 2>/dev/null || true)"
```

`${base:-HEAD~1}` is the same substitution the PR removes from `workflow_dispatch` and `push`, and it is reachable two ways: the fetch is `|| true`, so a fetch failure is silent; and `--depth=200` against the `fetch-depth: 0` checkout above installs a shallow boundary, so a branch whose true merge base is more than 200 base-branch commits back leaves `merge-base` with nothing to find. Either way `base` is empty and the whole matrix is gated on the tip commit's file list — a docs-only tip commit then skips every solver job on a PR that changes the solver, which renders as not-red. That is the #953 failure mode this PR exists to close.

Measured, running this branch's own gate script (loaded from `origin/fix/ci-dark-dispatch-and-clock-bars:.github/workflows/ci.yml`, with the `workflow_dispatch` marker string asserted present so the probe cannot be reading an older workflow) in a 3-commit repo — seed, a `python/solver.py` change, a `docs/notes.md` tip:

```
MARKER_PRESENT: 'a manual run always runs everything'
PR_ARM_NO_BASE   -> code=false   gate said: => code=false (heavy test jobs will SKIP)
PR_ARM_WITH_BASE -> code=true    (control)
CHECKS_EXECUTED=4
```

The control is the same repo, same event, same rendered script, with `refs/remotes/origin/main` made resolvable — it answers `true` and sees the code file, so the `false` above is the missing base, not a broken harness. The real solver change is two commits back and is lost in both arms of the guess.

The fix is symmetric with what you already wrote for `push` — treat an unresolvable base as the uncertainty case:

```sh
if [ -n "$base" ]; then
  changed="$(git diff --name-only "$base" HEAD 2>/dev/null || true)"
else
  changed=""   # unknown base -> falls through to code=true below
  echo "no merge-base with origin/${{ github.base_ref }} -> treating as uncertain"
fi
```

Dropping `--depth=200` (the checkout is already `fetch-depth: 0`) would remove the shallow-boundary route as well.

None of the six gate tests cover `pull_request` — `test_workflow_dispatch_*` ×2, `test_push_*` ×3, `test_schedule_*` ×1 — so this arm is both the most-exercised in production and the only one unpinned. A `test_pull_request_with_no_merge_base_runs_everything` alongside the existing `test_push_with_an_unusable_base_runs_everything` would close it; the harness already supports it unchanged.

## Minor

`python/tests/test_ci_changes_gate.py` imports `yaml` at module scope, but `pyyaml` is not in the `dev` extra — it is only in `gams` (`pyproject.toml:50`), and `test_gams_link.py`, the one other consumer, imports it inside the test bodies. If the lane running this file does not get pyyaml transitively, the file is a collection error rather than a skip. That is loud rather than dark, so it is not a correctness concern — but it would be red for the wrong reason. Adding `pyyaml>=6` to `dev` (it is already an indirect CI dependency) or moving the import next to the existing `pytestmark` skip guard settles it.

Everything else looks right to me: the `case` guard on `github.event.before` covers the all-zero SHA, non-hex junk, and empty; `git cat-file -e "$before^{commit}"` is the correct existence test; and the `schedule` early-exit is correctly reasoned (the nightly lane does not consume this output).

jkitchin:
Both findings confirmed and fixed in `2b014696`. The `pull_request` one is the more serious of the two and you're right that it was the gap in the first commit — I closed `workflow_dispatch` and `push` and left the arm that fires on nearly every change in this repo still doing exactly what the PR removes elsewhere.

## `pull_request` gating on a guessed diff

Reproduced before fixing, with a PR-shaped repo (base → `python/discopt/solver.py` change → docs-only tip) so the tip commit and the merge base disagree:

```
test_pull_request_with_no_merge_base_runs_everything -> code=false
```

The solver change is one commit behind the tip, so `HEAD~1..HEAD` cannot see it and every solver job is skipped on a PR that changes the solver.

Both routes you identified are real and neither is exotic — the `git fetch` is `|| true` so a fetch failure is silent, and `--depth=200` against a `fetch-depth: 0` checkout could only ever *subtract* history. The depth cap is dropped and an unresolvable base now falls through to the uncertainty rule, symmetric with the two branches on either side of it.

Three tests, including the control you'd want:

| test | what it pins |
| --- | --- |
| `test_pull_request_with_no_merge_base_runs_everything` | the regression |
| `test_pull_request_with_a_merge_base_sees_past_the_tip_commit` | same repo, same event, same rendered script, only `origin/main` made resolvable — proves the `false` is the missing base, not a broken harness, and that the gate looks past the tip when it can |
| `test_pull_request_docs_only_against_its_merge_base_skips` | a genuinely docs-only PR still skips — the gate's reason for existing |

Against `origin/main`'s workflow the suite is now **6 failed / 6 passed** — one failure per defective arm plus the four `push` base shapes — and **12/12** after. The 6 that pass both ways are all behaviour-preserving pins.

## `pyyaml`

Confirmed: only in the `gams` extra (`pyproject.toml:50`), so the module-scope `import yaml` would be a collection error rather than a skip. Added to `dev` with a comment saying why it is declared rather than relied on transitively. I kept the module-scope import — the point of the file is that it reads the real `ci.yml`, and a declared dependency is a better guarantee than an import that can be skipped past.

## On the commit message

You're right that "the `push` branch no longer substitutes a guess **either**" implied the class was closed when one arm still had it. The new commit states the scope as all three arms, and the CHANGELOG entry has been rewritten the same way rather than left describing the partial fix.

---

Issue #248, opened 2026-06-20
nvs16: McCormick relaxation returns a garbage -5e11 dual bound during solve (sound but freezes the gap)


jkitchin:
## Summary

On MINLPLib **nvs16** (a 2-variable Beale sum-of-squares, `minimize`, over the integer box `[0,200]²`, optimum **0.7031**), `m.solve()` reports a dual bound of **−5.08e11** — a ~100% gap — even though the McCormick relaxation of a sum-of-squares trivially proves **≥ 0**. The garbage bound anchors the tree's global lower bound, so nvs16 can never certify regardless of time or cut rounds.

**Not a soundness bug:** −5.08e11 is a *valid* (far-too-loose) lower bound; it's below the true optimum, so it does not produce a false-optimal. It's a numerical-conditioning / bound-quality defect that freezes the gap.

## Reproduction

```python
import discopt.modeling as dm
m = dm.from_nl("python/tests/data/minlplib/nvs16.nl")
r = m.solve(time_limit=20)
print(r.status, r.objective, r.bound)   # feasible 0.703125 -508231034025.8
```

## Diagnosis (what's been established)

1. **The relaxer is sound in isolation.** `MccormickLPRelaxer(m[, psd_cuts=True]).solve_at_node(root)` returns a clean **0.0** — confirmed for plain and PSD cuts, at every budget (1–10s), with and without separation, and whether `_psd_cuts` is set at construction, post-hoc, or via `_apply_auto_cut_policy`. So the garbage is **not** the relaxer or its configuration.

2. **The garbage is injected by the live solve pipeline.** During `m.solve()`, per-node `solve_at_node` bounds for nvs16 span **−5.08e11 … +1.49e13** (the +1.49e13 is the failure sentinel; the −5.08e11 is a genuinely wrong LP bound). The **root probe itself** returns −5.08e11 inside the solve — but the *same* root box solved by a fresh relaxer gives 0.0, and re-solving the captured bad box's base relaxation with **either** the Rust simplex **or** HiGHS also gives 0.0 (coefficient span only 8e6). So neither the box nor the cuts in isolation reproduce it.

[...]

PR #964, opened 2026-08-09
fix(correctness): #961 'optimal' node LP result must carry a finite bound; gate root_lp_bound

### LINKED ISSUE #961: Root LP reports status='optimal' with lower_bound=None; the baseline generator's bare except hides it
## Summary

Two defects found while attributing PR #960's claim-boundary drift. They compound: the first drops a root LP bound, the second hides that it happened.

1. **`MccormickLPRelaxer.solve_at_node` returns `status='optimal'` with `lower_bound=None`** on at least 5 corpus instances.
2. **`gen_claim_baseline._root_lp_bound` swallows the resulting `TypeError`** into an indistinguishable `None`, so the baseline records "no bound" for what is actually a crash.

Neither is caught by CI, because the claim-boundary gate asserts *shape* only and never looks at `root_lp_bound`.

## 1. An `optimal` root LP with no bound

Measured on `main` @ `d5f6eff6`, the in-house Rust simplex via `MccormickLPRelaxer(model).solve_at_node(lb, ub)` at the declared root box:

```
beuster:   status='optimal' lower_bound=None
casctanks: status='optimal' lower_bound=None
st_miqp2:  status='optimal' lower_bound=None
st_miqp3:  status='optimal' lower_bound=None
st_miqp4:  status='optimal' lower_bound=None
PROBES_EXECUTED=5
```

`status='optimal'` asserts the LP was solved to optimality, so `lower_bound=None` is a contract violation on its face. It is also a *regression*: `docs/dev/data/claim-baseline.jsonl` at `605d29b` records finite root bounds for all five — `beuster` 5942.495979840511, `casctanks` 3.887999978101959, `st_miqp2` −5.637695319137697, `st_miqp3` −6.000045783367277, `st_miqp4` −7451.5000074525. They solve optimally now and produce nothing.

The open question, and the reason this is filed as a bug rather than a nit: **does the B&B root consume this same value?** If it does, these five lose their root dual bound entirely, which is a certification loss rather than a cosmetic one. If the root path reads the bound by a different route, this is confined to the baseline generator's view. That is the first thing to determine.

## 2. A bare `except` turns the crash into a plausible number

`discopt_benchmarks/scripts/gen_claim_baseline.py`:

```python
    try:
        from discopt._jax.mccormick_lp import MccormickLPRelaxer
        lb, ub = _root_box(model)
        res = MccormickLPRelaxer(model).solve_at_node(lb, ub)
        if res.status == "optimal" and np.isfinite(res.lower_bound):
            return float(res.lower_bound)
        return None
    except Exception:
        return None
```

With `lower_bound=None`, `np.isfinite(None)` raises

```
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs
could not be safely coerced to any supported types according to the casting rule 'safe'
```

which the bare `except` converts to `return None` — the same value the function returns for a legitimately unbounded or non-optimal root. The docstring says "Returns None if the root solve does not certify a finite bound", so the recorded `None` reads as a *finding* when it is actually a swallowed crash. This is the exact anti-pattern CLAUDE.md §7 exists for, in the instrument that feeds a correctness gate.

Note the short-circuit: because `and` never evaluated `np.isfinite` unless `status == "optimal"`, the TypeError firing is itself proof the LP claimed optimality.

## 3. The gate never noticed

Regenerating the baseline on `main` and diffing against the committed file:

```
committed 605d29b -> main d5f6eff6   (62 checked)  SHAPE_MOVED=0  ROOT_LP_MOVED=52
```

52 of 62 instances have a stale `root_lp_bound`, accumulated silently since `605d29b`, while the shape column stayed exactly current. `test_claim_baseline_neutral.py::test_current_build_matches_committed_baseline_shape` compares `(n_rows, n_cols, n_integer_cols)` and nothing else, so a field the file spends effort computing — described in the generator docstring as "a deterministic tightness signal `cert-baseline.jsonl` does not carry" — is recorded, committed, and never asserted on.

Most of the 52 are last-digit. The five above are not, and neither is `nvs13` (−4843.01 → −751.12, which is a large *tightening* and presumably intended by whichever PR caused it — but nothing recorded it at the time).

## Suggested resolution

1. Root-cause the `optimal` + `lower_bound=None` pair in `MccormickLPRelaxer.solve_at_node`, and establish whether the B&B root reads the same field. Fix so an `optimal` status always carries a finite bound, or refuse loudly if it genuinely cannot.
2. Delete the bare `except` in `_root_lp_bound`. Let it crash — a generator that cannot compute a root bound must say so, not record a value that means something else. If some instances legitimately have no bound, distinguish that case explicitly.
3. Decide whether `root_lp_bound` should be gated. Either assert it (with a tolerance, and treat regressions as bound-changing per CLAUDE.md §5) or stop recording it. Recording an ungated field invites exactly this: 52 silent drifts, five of them hiding a crash.

## Provenance

Found while attributing PR #960 (issue #957). #960 regenerates the baseline and its commit message and PR comment record all 52 drifts as pre-existing on `main`, measured against a control regeneration on `d5f6eff6` with the control tree asserted to lack the `#957` marker. #960 does not change any of the behaviour described here.

jkitchin:
## Summary

Closes #961.

`MccormickLPRelaxer.solve_at_node` returned `status='optimal'` with `lower_bound=None` on beuster, casctanks, st_miqp2/3/4 (reproduced 5/5 at the declared root box). Root cause: the certify-decline path in `_solve_at_node_impl` — reached when there is no Neumaier–Shcherbina safe bound and the vertex objective is refused by the unbounded-nonlinear-column / conditioning guards (4 of the 5), or invalidated by `_objective_bound_valid=False` (st_miqp3, the #248 garbage-floor refusal) — passed the raw LP status through, so a deliberate, sound *decline* was mislabeled as `optimal`. Three changes:

1. **The decline is its own status.** A solved-but-uncertified node now reports `status="uncertified"`, and `MccormickLPResult.__post_init__` refuses the `optimal`/no-finite-bound pair outright, so the contract cannot regress silently. Behavior-neutral: every consumer (root probe at `solver.py:8404`, both node loops, `root_reduce`, `disjunctive_config_bound`) gates on `lower_bound is not None`, verified site by site.
2. **The bare `except` in `gen_claim_baseline._root_lp_bound` is gone** (it converted the `np.isfinite(None)` TypeError into the same `None` that means "no bound"). The generator now records `root_lp_status` alongside `root_lp_bound` and crashes loudly on any contract violation (CLAUDE.md §7).
3. **`root_lp_bound` is now gated** (issue resolution 3): a new claim-boundary test re-solves every root LP and compares status exactly and the bound within `ROOT_LP_REL_TOL = 1e-6` relative; a baseline row missing the fields fails rather than skips.

Answering the issue's open question: **yes, the B&B root consumes the same field.** The root probe treats `lower_bound=None` as "probe not useful" and disables the McCormick engine for the whole solve on those instances (`_mc_mode = "none"`), falling back to other bound sources — a bound-quality loss, not a false certificate. Recovering the declined bounds themselves would be bound-changing work (the declines are the himmel16/C-38-era soundness guards working as designed) and is out of scope here.

Baseline regenerated: 54 `optimal` / 12 `uncertified` (the issue's five plus seven whose committed `null` was previously indistinguishable from a crash), zero shape moves. One bound *recovered*: contvar (committed `null` → 172170.31). Attribution per §8: both the pre-fix and post-fix trees certify that value here (pre-fix tree asserted to lack the `__post_init__` marker), so the committed `null` was a pre-existing failure swallowed by the bare `except` on the #960 regeneration host, not an effect of this PR. If contvar's certification proves machine-dependent, the new gate will surface it on CI — which is the gate doing its job.

## Correctness

No solver-math change: no bound value is computed differently anywhere — a result that previously carried `optimal` with no bound now carries `uncertified` with no bound, and all consumers gate on the bound. The generator/gate changes are instrumentation.

- [x] `pytest -m smoke` (discopt_benchmarks/tests) → 49 passed, 1 skipped
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → 9 passed, 1 failed (`test_large_dense_jacobian_no_crash` wall-clock cap, run concurrently with the claim-boundary suite on a 4-core box); passes in isolation → 1 passed in 36.01s, under its 48 s cap
- [x] `pytest python/tests/ -m claim_boundary -n0` (the serial CI job, includes the shape gate against the regenerated baseline and the new root-LP gate) → 302 passed, 28 skipped
- [x] McCormick unit files (`test_mccormick*.py`, `test_incremental_mccormick_node.py`, `test_mccormick_lp_obbt_units.py`) → 125 passed
- [x] End-to-end `st_miqp4` solve → `optimal`, objective −4574.000 = MINLPLib oracle
- [ ] `cargo test -p discopt-core` → N/A — Rust untouched
- [x] `ruff check` / `ruff format --check` clean; `mypy` on the touched module clean

## Regression test

- [x] Added `python/tests/test_961_optimal_requires_bound.py`: constructor contract tests (the `optimal`/None pair raises) and an instance-level test on st_miqp3 asserting the decline surfaces as `uncertified`, never `optimal` without a bound. The instance test fails before this change (`status='optimal'`, `lower_bound=None`) and passes after; the gate test `test_current_root_lp_matches_committed_baseline` fails before (baseline rows lacked the fields) and passes after.

https://claude.ai/code/session_01T4Rbj1kvShvUAskPeKbWwL

---
_Generated by [Claude Code](https://claude.ai/code/session_01T4Rbj1kvShvUAskPeKbWwL)_


jkitchin:
Reviewed and pushed a fix for the claim-boundary failure (`95769c3f`). The change itself — `uncertified` as its own status, the `__post_init__` contract, killing the bare `except` in the generator — is right and I'd merge it as-is. The failure was in the new gate, and your PR body predicted it exactly: *"If contvar's certification proves machine-dependent, the new gate will surface it on CI."* It did, and it is.

**Measured before changing anything** (every arm run to completion, counts printed):

| where | contvar root LP |
|---|---|
| macOS/arm64 build (this box), 5/5 runs | `uncertified`, no bound |
| same box, model pre-run through `classify_nonlinear_terms` + `build_milp_relaxation` (the generator's own sequence) | `uncertified`, `box_moved=0` |
| same box, full 66-instance single-process regeneration (rules out sweep order / cut-pool state) | `uncertified` — 396 field comparisons, contvar the only status move |
| CI, ubuntu/x86-64 | `uncertified` |
| the host that generated the committed row | `optimal`, 172170.31 |

So it is not leftover model state, not sweep order, and not within-host nondeterminism — it is the build. Which is already documented, and names this instance: `diff_instance`'s docstring says the fingerprint "is NOT reproducible across Rust builds/platforms (the in-house FBBT/parse path produces last-digit-different matrix coefficients on `contvar`/`tanksize` — confirmed on a pristine tree)", which is exactly why the shape gate refuses to gate on it. contvar's certification sits on the edge of the NS/conditioning guards, so those last-digit coefficients flip certify vs decline. On this box the shape gate independently puts contvar in its 10-instance `fingerprint_drift` set — same instance, same cause, corroborated by a gate you didn't touch.

**The fix applies the rule the shape gate already uses**: a root-LP difference is only attributable when both sides solved the same matrix. `diff_root_lp` now compares the fingerprint *on the difference path only*, and where the bytes differ too it reports `fingerprint_drift` — printed by the test, never silent — instead of `changed`.

I kept your two guardrails intact rather than trading them away:

- The fingerprint is computed **only** when a difference is seen, so every instance whose bytes match is still compared exactly. No loss of gating on the 65 others.
- The `>= 50` sanity floor counts **only** `unchanged`, not the drift bucket, so the bucket cannot grow into a way for the gate to stop measuring — which was your stated worry ("an ungated recorded field is how 52 silent drifts accumulated").
- `test_root_lp_gate_bites_when_the_matrix_bytes_are_identical` pins that the escape is conditional: on an instance this build reproduces exactly, a doctored status is still `changed`; only when the fingerprint is doctored too does it become `fingerprint_drift`. It picks its instance by first confirming that instance is exactly reproduced, so it can't pass vacuously.

**I did not touch the committed baseline.** contvar's row is correct for the build that produced it; rewriting it to this host's value would just move the failure to your host. Two hosts disagreeing is the signal, not a wrong row.

Local: `test_claim_baseline_neutral.py` → 2 passed (contvar reported as a build boundary), `test_961_optimal_requires_bound.py` → 5 passed.

One note for the record, not blocking: on this box 10 fingerprints and ~30 root-LP bounds differ from the committed baseline (all bounds within your `ROOT_LP_REL_TOL`, which is doing its job). The bound tolerance absorbs the build boundary; the status field had no such absorption, which is the gap this fixes.

PR #970, opened 2026-08-09
fix(deps): pin why pyyaml is in the dev extra, and test it

jkitchin:
## The defect

The `dev` extra already declares `pyyaml`, so nothing is broken today. What is broken is the **reason** it gives:

```toml
# test_ci_changes_gate.py parses .github/workflows/ci.yml to run the real
# `changes` gate script. ...
"pyyaml>=6",
```

`python/tests/test_ci_changes_gate.py` does not exist in the tree. The declaration is still correct, for a different file: `test_gams_link.py` opens with *"None of this requires a GAMS installation"* and three of its tests do a bare `import yaml` with no `importorskip`, so PyYAML is a hard test dependency and not a `[gams]`-only one.

That leaves the dependency one stale-comment-driven cleanup away from deletion — and **the deletion would be invisible to CI**. Four CI jobs `pip install` pyyaml on top of the extras (`ci.yml:163,243,304,374`), so CI stays green either way. The only environment that breaks is the documented one, `pip install -e ".[dev]"` — the command in CLAUDE.md — where those three tests become collection *errors*, not skips.

## The fix

1. The comment now cites the file that actually needs it, and records that it named a non-existent file until today.
2. `python/tests/test_dev_extra_declares_test_deps.py` pins the invariant.

The test derives the requirement rather than restating it: it walks `test_gams_link.py`'s AST for imports that are outside any `try`/`except` and are neither stdlib nor first-party, maps import name → distribution name for the cases where they differ, and asserts each resolves from `[project].dependencies` ∪ `[dev]`. **No package name is hardcoded**, so the next unguarded third-party import added to that file is caught the same way.

Per §6 it also refuses to pass while measuring nothing: it asserts the scan found imports at all, and that it examined at least two.

## Verification

- Passes as committed.
- **Negative control**: with `"pyyaml>=6"` removed from the `dev` block the test fails (`1 failed`) — run programmatically, file restored afterward, so the test is proven to bite rather than assumed to.
- `ruff check` + `ruff format --check` clean.
- `test_dev_extra_declares_test_deps.py` + `test_gams_link.py` → **53 passed**.

## Note on scope

I checked whether to generalize this across all of `python/tests/`. A scan of 520 test files finds unguarded-looking imports of `cyipopt`, `gamspy`, `highspy`, `onnx`, `scs`, `sklearn`, `sympy`, `torch` and more — but the scan cannot see module-level `pytest.importorskip`, so most of those are false positives. A tree-wide version needs importorskip modelling to avoid being a noise generator, which is a larger job than this fix. Scoped to the file with the demonstrated gap.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,10 +70,15 @@ dev = [
     "pytest",
     "pytest-timeout",
     "pytest-cov",
-    # test_ci_changes_gate.py parses .github/workflows/ci.yml to run the real
-    # `changes` gate script. Declared here rather than relied on transitively:
-    # without it that file is a collection ERROR, not a skip — red for the wrong
-    # reason. (Also in the `gams` extra, for gamsconfig.yaml merging.)
+    # test_gams_link.py opens with "None of this requires a GAMS installation"
+    # and three of its tests `import yaml` unguarded, so PyYAML is a hard test
+    # dependency and not a `[gams]`-only one. Declared here rather than relied
+    # on transitively: without it those tests are collection ERRORS, not skips —
+    # red for the wrong reason, and only in a `pip install -e ".[dev]"` tree,
+    # since four CI jobs pip-install pyyaml separately and would stay green.
+    # test_dev_extra_declares_test_deps.py pins this; it named
+    # test_ci_changes_gate.py until 2026-08-09, a file that no longer exists.
+    # (Also in the `gams` extra, for gamsconfig.yaml merging.)
     "pyyaml>=6",
     "pytest-xdist",
     # Pin the lint/typecheck toolchain to the exact versions in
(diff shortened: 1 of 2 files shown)


jkitchin:
The `Python claim-boundary` failure here is **#971**, not this PR.

```
root LP bound/status drifted vs committed baseline:
  contvar (root LP bound 172170.3107274997 -> 187283.4711213944)
```

This PR changes a comment in `pyproject.toml` and adds a packaging test; it
touches no solver code, so it cannot move an LP bound. The identical failure,
with the **bit-identical** value `187283.4711213944`, already hit `main` at
03:48 UTC today ([run 31292837058](https://github.com/jkitchin/discopt/actions/runs/31292837058)),
before this branch existed, and `main` passed the same gate at 15:23
([run 31320980434](https://github.com/jkitchin/discopt/actions/runs/31320980434)).

Locally (macOS/arm64, this branch), 5 reps in fresh processes are unanimous —
`uncertified` / no bound / one fingerprint — so the result is deterministic per
host and varies across hosts. Full evidence and the two candidate fixes are in
#971; the gate is catching something true, so it should not be silenced with a
tolerance bump.

Re-running the failed job.

Issue #971, opened 2026-08-09
Flaky hard gate: contvar root LP bound differs across CI runners with an IDENTICAL fingerprint (#961 gate)


jkitchin:
## Summary

`test_claim_baseline_neutral.py::test_current_root_lp_matches_committed_baseline`
(the #961 root-LP gate) fails intermittently on `contvar`, on `main` and on
unrelated PRs, with the message:

```
root LP bound/status drifted vs committed baseline:
  contvar (root LP bound 172170.3107274997 -> 187283.4711213944)
assert not [InstanceDiff(instance='contvar', status='changed', detail=...)]
```

It is bucketed `changed`, not `fingerprint_drift`, which by `diff_root_lp`'s own
rule means **the relaxation fingerprint MATCHED the committed baseline**. Same
matrix bytes, an 8.8% different root LP bound.

## Evidence

| where | run | verdict |
|---|---|---|
| `main` @ 03:48 UTC | [31292837058](https://github.com/jkitchin/discopt/actions/runs/31292837058) | FAIL, `-> 187283.4711213944` |
| `main` @ 15:23 UTC (ff8c154b) | [31320980434](https://github.com/jkitchin/discopt/actions/runs/31320980434) | pass |
| PR #970 (a pyproject **comment** + a new packaging test) | [31321409802](https://github.com/jkitchin/discopt/actions/runs/31321409802) | FAIL, `-> 187283.4711213944` |

PR #970 touches no solver code whatsoever, and the failure predates it on `main`,
so this is not attributable to any change under review.

The failing value is **bit-identical across the two failures**
(`187283.4711213944`), so this is not random noise: there are two *deterministic*
outcomes selected by something outside the fingerprint.

Locally (macOS/arm64, this tree = `main` + #970), 5 repetitions in fresh
processes are unanimous:

```
{"status": "uncertified", "bound": null,
 "fingerprint": "7d96903eb2b4514186d6d5d7752650e8af194017b43db4a29809cd293f19dd74"}   x5
distinct_triples: 1   deterministic_on_this_host: true
```

so the result is deterministic *per host* and varies *across* hosts — matching
`diff_root_lp`'s own docstring ("5/5 deterministic per host").

Probe (added in the branch for this investigation):
`discopt_benchmarks/scripts/contvar_root_lp_reproducibility.py`, which asserts the
tree under test by path (CLAUDE.md §8), runs each rep in a fresh process, never
[...]

PR #973, opened 2026-08-09
fix(correctness): #956 outward-round the McCormick envelopes, and let the tree fathom a proven-empty region

### LINKED ISSUE #956: McCormick envelope rows are not outward-rounded, so a relaxation can cut off the point it relaxes

jkitchin:
## Summary

Closes #956.

> **Corrections (2026-08-09).** Three claims in the original version of this body were wrong; see [the review follow-up](https://github.com/jkitchin/discopt/pull/973#issuecomment) for the measurements and `cc0fc38` for the fixes. Retracted and corrected in place below, marked **[corrected]**. In short: the opt-out was value-identical but not *byte*-identical (negative zero), the "all six fail under `=0`" claim was test-order-dependent, and the stated invariant did not actually hold — a fourth emitter was never routed through the guard.

McCormick envelope rows were not outward-rounded, so a relaxation could cut the exactly-feasible point `(x, f(x))` out of its own envelope — a route to a false bound. All **four** emitters now apply a ulp-scaled outward guard computed from identical quantities in an identical operation order, across three engines: Rust `mccormick_patch.rs`, the Python cold build (`uniform_relax`), and the Python incremental build (`incremental_mccormick`). They *must* agree, because `IncrementalMcCormickLP._validate` compares them through a row-set that rounds each rhs to 6 decimals absolutely, so a guard applied to one and not the others silently drops the incremental fast path to `ok=False`.

**[corrected]** The invariant — *for every x in the box, `(x, f(x))` satisfies every emitted row and lies inside the aux bounds* — did **not** hold in the first version of this PR. `uniform_relax._emit_odd_power_hull._line` builds `rhs = sign*(f(t0) - m*t0 + m*c)`, the same cancelling endpoint combination #956 is about, and emitted it raw. Measured with the guard ON, `min x**3` over `[-318309.8861837907, 1273239.5447351628]` at `t = lo` cut the exactly-on-graph lifted point by **+128 absolute**. That emitter has no incremental or native counterpart to stay bit-identical with (`incremental_mccormick` declines odd powers on a straddling-root box outright), so it is now routed through the guard as well and the residual goes to `<= 0`. On the Python generators the invariant now holds: 1 950 000 comparisons, worst residual −3.3e−15, previously +1.3e+5.

Chasing this exposed a second, larger defect in the tree, fixed here as well. `TreeManager` pruned only on `lower_bound >= incumbent`, with `incumbent_value` initialised to `+inf`. A rigorously infeasible node arrives as a large *finite* sentinel, and `1e30 >= inf` is false — so **a region proven empty was never pruned until an incumbent existed, and on an infeasible model one never does.** Every certified-empty node was branched instead. `x*y >= c, x + y <= 1` on `[0,1]²` went from ~12 000 nodes to the time limit, for every `c` from 0.3 to 2.0, to **`infeasible` in 1 node**.

Also fixed: `profile::reset()` cleared the run totals along with the per-`dump` accumulators, and `milp_driver` calls it at the start of every MILP sub-solve — so a counter incremented before the last sub-solve read back as 0, and "0" reads as "this path never executed". It nearly buried the tree fix above. Split into `reset()` and `reset_totals()`.

Full record: `docs/dev/issue-956-followthrough-plan.md` (§2 task outcomes, §4 measurements) and `docs/dev/performance-plan.md` §15a/§15b.

## Correctness

The envelope guard is a bounded **outward** relaxation: each emitted rhs is `>=` the exact closed form and exceeds it by at most the ulp-scaled guard, so the relaxation can only get looser — never tighter — and a looser relaxation cannot cut the optimum out of the box. `guard_is_outward_and_bounded` asserts exactly that at the formula level.

The tree change strictly *increases* pruning, so its risk is the worst one available: a false `infeasible`. Two things contain it. The certificate is carried explicitly (`NodeResult::certified_infeasible`) rather than inferred from the sentinel — the same sentinel also encodes *soft* failures, and fathoming those is #927's false-certificate mode. And the flag is set only from a Farkas-verified empty relaxation. Gate: every in-repo model with a known optimum is provably feasible, so any `infeasible` report is a provable error — **26 models checked, 0 false infeasibles.**

Two smoke tests failed en route (`test_nonrigorous_fathom_is_not_reported_infeasible`, `test_integration_nonrigorous_fathom_reports_feasible_not_optimal`) and both were worth having, but as test *double* signature mismatches, not soundness breaks. Their fakes now forward the certificate rather than dropping it, so they still exercise the path they guard — no assertion was relaxed.

Re-run on `4a972e5` (linux/x86-64):

- [x] `pytest -m smoke` → 916 passed, 22 skipped, 2 xpassed
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → 10 passed
- [x] `cargo test -p discopt-core` → 602 passed (597 unit + 4 integration + 1 doc)
- [x] `pytest -m claim_boundary` → 303 passed, 28 skipped
- [x] `ruff check` / `ruff format --check` clean; `cargo clippy -p discopt-core` and `--no-default-features` clean under `-D warnings`; `cargo fmt --check` clean

## Regression test

**Rust.** Six tests in `mccormick_patch.rs` covering all five row generators plus the aux bounds, each built from a witness box found by sweeping ten orders of magnitude. **[corrected]** These were originally described as "all six fail under `DISCOPT_ENVELOPE_OUTWARD_ROUND=0`". That was *test-order dependent*: `mccormick_patch.rs` carried a `TEST_FORCE_ON: AtomicBool` seam with a `#[cfg(test)]` early return, and every `discopt-core` unit test compiles into one binary sharing process state — so once any test called `test_force_guard_on()`, the flag stayed set and `=0` became a no-op for everything scheduled after it. The seam is deleted; each test now asserts the guard is on and panics with an explicit message otherwise, so an opt-out run **skips loudly** instead of silently asserting the wrong configuration. Plus `corner_pinned_cubic_node_lp_is_solvable` in `spatial_kernel.rs`.

**Python.** This PR originally shipped **zero** Python tests for `outward_rounding.py`. `python/tests/test_956_outward_rounding.py` adds 12, each printing an executed-comparison count (§6): negative-zero preservation with the guard off (plus a pin that `-0.0 + 0.0` really does differ, so a revert cannot pass); the guard only ever loosening; an unbounded box degrading to the absolute floor rather than a `1e20`-scaled one; and the invariant itself over three families (bilinear, square, cubic, 252 points scored) **with a negative control** — `test_the_guard_off_build_really_does_violate` asserts the guard-off build genuinely cuts by 2, 1 and 128, so the positive test cannot decay into a tautology. That risk was real, not hypothetical: a first version used boxes around `5e3` where all three are exactly tight unguarded and passed identically in both arms.

**Tree fathoming.** `certified_infeasible_prunes_without_an_incumbent` (fails before: the node is branched, not pruned) and `uncertified_sentinel_still_branches_without_an_incumbent`, which pins the other side of the boundary so the fix cannot drift into fathoming soft failures. `test_debug_adversarial`'s original `_infeasible_bilinear` fixture (`x*y >= 0.5`) is **restored** — the branch had swapped it out only because the solver could not certify that model — with its docstring corrected rather than reinstated: the old claim "infeasible only after branching" was never true.

**LP seam.** `xb_refresh_cadence_cannot_move_the_optimum`.

**Gate hygiene.** `test_root_lp_gate_bites_when_the_matrix_bytes_are_identical` selected its probe instance with `diff_root_lp(...) == "unchanged"` while its docstring and next assertion require *byte* identity — not the same condition, since `diff_root_lp` computes the fingerprint only once it already sees a status/bound difference. It also sourced the "recorded" bytes from `docs/dev/data/claim-baseline.jsonl`, a linux/x86-64 artifact, so it could fail for two reasons unrelated to the gate: any correct relaxation change, and the arm64 platform boundary (#971). The synthetic recorded row is now built from *this* build, making the matching-bytes side byte-identical by construction with the doctored `root_lp_status` as the only difference.

`DISCOPT_ENVELOPE_OUTWARD_ROUND` **graduates to default-ON** here; `=0` remains the opt-out. Two new seams ship **default-off**: `DISCOPT_PRIMAL_XB_REFRESH=0` and `DISCOPT_SIMPLEX_NO_EXPAND`.

- [x] **[corrected]** Flag-OFF path byte-identical
- [x] Differential-bound test (outward-only, bounded) + feasible-point sampling over each witness box
- [x] Graduation panel evidence attached (below)

**[corrected] The opt-out was value-identical but not byte-identical, and now is.** Every guarded emitter wrote `rhs + slack`; with the guard off `slack` is `0.0`, but `-0.0 + 0.0` is `+0.0`, so the no-op add cleared the sign bit of every negative-zero rhs. Invisible to the LP (`-0.0 == 0.0`), visible to `relaxation_fingerprint`, which hashes the matrix bytes — and the claim baseline gates on that hash. Measured on the 66-instance vendored corpus with `=0`, before the fix: **34 of 66 fingerprints moved off the committed baseline** (22 of `4stufen`'s 247 `b_ub` entries and 1 of `alan`'s 10 flip sign; `A_ub` and `c` untouched, every value equal). The fix is a shared `relaxed_rhs(rhs, slack)` returning `rhs` untouched when the slack is zero, applied at all eight emitters. After it, on linux/x86-64 — the baseline's own platform — **66 of 66 reproduce exactly**; on arm64, 56 of 66, and the 10 that do not are precisely the set that also fails to reproduce on `origin/main` on that host (#971's known platform boundary).

**Measurement — why the earlier "harmful" verdict was retracted.** The guard first shipped default-OFF on a panel reading "cert-clean but harmful". That panel had no arm in which the two builds were known to be identical, so it could not separate a real effect from run-to-run variation. This PR builds that control: comparing two builds on six instances where the only differing code path *provably never executes* (counter-verified, 0 firings), the same harness still reported "WIN 3/3" for one instance and "REGRESSION 4/7" for two others, with bound spreads of 0.1–0.7 % at `time_limit=20`. That is the noise floor. Re-run against it, the guard's own A/B is a wash — 3 identical, 1 win, 2 unresolved, 2 "regressions" at 2/3, all inside the floor. `nvs20`, previously read as "deterministic, not noise", is now *identical* in both arms.

The second piece of the old verdict — `ex1252` under the #707 reform flags going 55.3 % → 81.7 % undecided — can no longer be measured at all: that configuration does not reach the native spatial kernel any more (0 kernel calls for it, 21 elsewhere in the same run, so the probe is live), and across the 21 corpus instances that do reach it, `n_undecided` is **0 in both arms** over ~12.5k kernel nodes.

| panel | arms | result |
|---|---|---|
| Corpus cert-clean | guard ON vs OFF, one build, 119 instances | 149 executed assertions — **0 false bounds, 0 false objectives, 0 false infeasibles, 0 new errors**. One flagged cert difference (`nvs12`) does not reproduce in 6/6 replicates. |
| Corpus cert-clean | branch vs `origin/main`, guard OFF, 119 instances | 147 executed assertions, same zeros. Two flagged status differences are TL=5 deadline flicker the **baseline shows too**. |
| Tree fathoming A/B | 10 instances where the arm fires, 3 replicates | **1 win** (`tls2`: 265 vs 317 nodes, bound 3.667 vs 2.100, `optimal` where the baseline reaches only `feasible`), **9 bit-identical**, 0 regressions |

The prune arm fires on **30 of 119** corpus instances, so it is not inert; on the other 89 it executes nothing and therefore cannot regress them.

**Negative results, recorded so they are not re-litigated.** The counter #956 was filed to drive down (`n_undecided`) is already at its floor — 0 corpus-wide. The `nvs20` bound-excursion class has a proven root cause: the Harris ratio test picks the leaving variable and the step from an incrementally-maintained `x_B`, so a drifted estimate authorises a step infeasible in the values the basis actually holds. Re-deriving `x_B` exactly every K pivots collapses the phase-1 handoff violation rate monotonically over a 30× range, 50.3 % → 1.7 %, and EXPAND is excluded (suppressed globally: 49.7 % → 48.2 %; 97 % of violations exceed its per-pivot ceiling 100-fold, none at or below it). **But it does not fix what it was meant to fix** — `LpAuditBoundsFail` 788 → 764, `LpVerdictNumerical` 864 → 856 — so a box-infeasible handoff is real and is *not* what makes node LPs undecidable; that inference is retracted in the plan doc. The cadence is also not net-positive under a wall-clock budget (K=1: 2 wins / 4 regressions, three at 4σ–9σ), so per the `DISCOPT_CUT_INHERIT` rule it stays off, kept as a control arm with a test rather than as dead configuration.

https://claude.ai/code/session_01TQTA8JZvQA9KLYHEsH4W2k


jkitchin:
## Review follow-up — two claims in this PR body were false, and a third gap was found

Per CLAUDE.md §11, retracting in writing before continuing.

### 1. "Flag-OFF path byte-identical" was **false**. Fixed.

The opt-out was value-identical, not byte-identical. Every guarded emitter wrote
`rhs + slack`, and with the guard off `slack` is `0.0` — but `-0.0 + 0.0` is `+0.0`,
so the no-op add clears the sign bit of every negative-zero rhs. The LP cannot see it
(`-0.0 == 0.0`); `relaxation_fingerprint` can, because it hashes the matrix bytes, and
the claim baseline gates on that hash.

Measured on the 66-instance vendored corpus with `DISCOPT_ENVELOPE_OUTWARD_ROUND=0`,
before the fix: **34 of 66 fingerprints moved off the committed baseline** (22 of
`4stufen`'s 247 `b_ub` entries and 1 of `alan`'s 10 flip sign; `A_ub` and `c` untouched,
every value equal). After the fix: **56 of 66 reproduce**, and the 10 that do not are
*exactly* the set that also fails to reproduce on `origin/main` on this host
(`contvar, cvxnonsep_nsig30, ex1226, heatexch_gen3, nvs08, st_e11, tspn05, tspn08,
tspn10, tspn12` — a known arm64-vs-x86-64 boundary, #971). Direct byte comparison of
`4stufen`/`alan` `A_ub`/`b_ub`/`c`: `ARRAYS_COMPARED=6 NOT_BIT_EQUAL=0`.

Fix: a shared `relaxed_rhs(rhs, slack)` that returns `rhs` *untouched* when the slack is
zero, applied at all eight emitters in `outward_rounding.py`, `uniform_relax.py`,
`incremental_mccormick.py`, and `mccormick_patch.rs`.

### 2. "All six fail under `DISCOPT_ENVELOPE_OUTWARD_ROUND=0`" was **scheduling-dependent**. Seam removed.

`mccormick_patch.rs` carried a `TEST_FORCE_ON: AtomicBool` seam and
`outward_rounding_enabled()` had a `#[cfg(test)]` early return keyed on it. All
`discopt-core` unit tests compile into **one** binary and share process state, so once
any test called `test_force_guard_on()` the flag stayed set for every later-scheduled
test — making `=0` a no-op for them. Whether the six envelope tests actually failed
under the opt-out therefore depended on test order.

The seam is deleted. Each of those tests now asserts the guard is on and panics with an
explicit message if it is not, so an opt-out run **skips loudly** instead of asserting the
wrong configuration. `cargo test -p discopt-core` → **602 passed, 0 failed** (597 unit +
4 integration + 1 doc).

### 3. New: `_emit_odd_power_hull` was **unguarded**, and cut a feasible point by 128.

The PR's stated invariant — "for every x in the box, `(x, f(x))` satisfies every emitted
row" — did not hold after this PR, because a fourth emitter was never routed through the
guard. `uniform_relax._emit_odd_power_hull._line` builds `rhs = sign*(f(t0) - m*t0 + m*c)`,
the same cancelling endpoint combination #956 is about, and emitted it raw.

Measured, guard **ON**, `min x**3` over `[-318309.8861837907, 1273239.5447351628]` at
`t = lo`, scoring the exactly-on-graph lifted point against `A z <= b`:

```
row1: coeffs=[-1.31717539e+12  1.00000000e+00] rhs=3.8701841319839386e+17 residual=+128
```

localized by a traceback spy on `_Builder.add_row` (`HITS=4 ROWS=4`, `VIOLATING_ROWS=1`)
to `uniform_relax.py:1666`. This emitter has no incremental or native counterpart to stay
bit-identical with — `incremental_mccormick` declines odd powers on a straddling root box
outright — so the guard here is unconstrained by the tri-engine rowset comparison. Now
routed through `relaxed_rhs(rhs, envelope_1d_slack(...))`; residual goes to `<= 0`.

### 4. `test_961_optimal_requires_bound.py` was passing for the wrong reason.

`test_root_lp_gate_bites_when_the_matrix_bytes_are_identical` selected its instance with
`diff_root_lp(n, baseline).status == "unchanged"`, but its docstring — and the assertion
on the next line — require *byte*-identity, which that predicate does not imply:
`diff_root_lp` computes the fingerprint only once it has already seen a status/bound
difference. So whether the test proved anything came down to which instance sorted first.
It also sourced the "recorded" bytes from `docs/dev/data/claim-baseline.jsonl`, a
linux/x86-64 artifact of solver commit `1fe54f4`, which makes it fail for two reasons
that have nothing to do with the gate under test: any correct relaxation change (this PR
graduates the guard, so 61 of 66 fingerprints move by design) and the platform boundary
(10 of 66 do not reproduce on arm64 even on `main`). It fails locally on this branch on
arm64 today; CI passes on luck of the first instance.

The synthetic "recorded" row is now built from **this build** (`current_row` +
`current_root_lp`), so the matching-bytes side is byte-identical by construction and the
doctored `root_lp_status` is the only difference. What the test is about — that the
`fingerprint_drift` escape is *conditional* — is unchanged and now actually exercised.

### 5. New Python tests

This PR previously shipped **zero** Python tests for `outward_rounding.py`.
`python/tests/test_956_outward_rounding.py` adds 12, each with an executed-comparison
count (§6):

* `relaxed_rhs` / `widen` preserve negative zero bit-for-bit with the guard off, plus a
  pin that `-0.0 + 0.0` really does differ so a revert cannot pass;
* the guard-off cold build of `4stufen` still emits negative-zero rhs entries;
* the guard only ever loosens; an unbounded box degrades to the absolute floor, never a
  `1e20`-scaled one; a non-finite magnitude contributes zero, never a NaN;
* **the invariant, with a negative control.** Three families (`bilinear`, `square`,
  `cubic`) at boxes found by sweeping ten orders of magnitude (252 points scored), each a
  case where the guard-**off** build demonstrably cuts `(x, f(x))` — by 2, 1 and 128
  absolute. `test_the_guard_off_build_really_does_violate` asserts that, so the positive
  test cannot decay into a tautology. That is not hypothetical: my first version used
  boxes around `5e3`, where all three are exactly tight unguarded, and it passed
  identically in both arms — asserting nothing.

One correction to my own earlier measurement in this thread: I reported the `sep_lb`
objective-floor cut at `uniform_relax.py:3252` as cutting its own minimizer by 1.5e-5.
That was a defect in my probe, not in the code — the DAG evaluates `x*x*x` as `(x*x)*x`
and I scored the point with `x**3`, which is a different f64 one ulp below it. The row is
exactly tight. No change was made there.

jkitchin:
All three findings verified independently on linux/x86-64, and the PR body is corrected — the two false claims (1) and (2) were mine, and (3) means the invariant this PR advertises did not hold when I wrote it. Marked **[corrected]** in place rather than silently edited.

**One result is better than reported, and it isolates the platform variable.** Re-running the fingerprint comparison with `DISCOPT_ENVELOPE_OUTWARD_ROUND=0` after `cc0fc38`, on linux/x86-64 — the platform the committed baseline was generated on:

```
guard OFF after the fix: matching 66, mismatching 0
```

Before the fix I measured 24 of 66 matching on this same host. So on the baseline's own platform the opt-out is now byte-identical for the *entire* corpus, not 56 of 66. Combined with your arm64 run, the two together separate the two effects cleanly: the negative-zero add accounted for every mismatch on x86-64, and the residual 10 on arm64 are #971's platform boundary alone — exactly the set that also fails on `origin/main` there. That's a stronger claim than either measurement alone supports, so I've written it into the body that way.

**On the negative control (5).** Worth stating plainly that this is the finding I'd most want kept: your first version passed identically in both arms because the boxes were exactly tight unguarded. A guard test that cannot fail with the guard off asserts nothing, and it would have shipped looking green. `test_the_guard_off_build_really_does_violate` is what makes the other test mean something.

**On (2), the `TEST_FORCE_ON` seam** — I introduced that, and the failure mode is worse than order-dependence in the abstract: it made "these six tests fail under `=0`" a claim I stated in the PR body as evidence, derived from a run that couldn't have shown otherwise. Skipping loudly is the right shape.

Gates on `4a972e5`, linux/x86-64: `cargo test -p discopt-core` **602 passed** · `pytest -m smoke` **916 passed, 22 skipped, 2 xpassed** · `pytest -m claim_boundary` **303 passed, 28 skipped** · adversarial **10 passed** · `test_956_outward_rounding.py` + `test_961_optimal_requires_bound.py` **16 passed** · ruff check/format, clippy (both CI invocations) and `cargo fmt --check` clean.

No further changes from me. CI was green on `ca6c114`; I'll watch the run on `4a972e5`.

---

Issue #759, opened 2026-07-18
Correctness: syn05hfsg dual bound exceeds incumbent (pre-existing #120 alphaBB fallback)


jkitchin:
## Summary

`syn05hfsg` reports a **dual bound above its incumbent/oracle** — a min-sense `bound ≤ incumbent` invariant violation — on `main`, independent of any feature flag. It does **not** emit a false *certificate* (`gap_certified=False`), so it is not a wrong-answer-to-the-user, but a dual bound crossing the incumbent is a genuine invariant breach and should be root-caused.

## Evidence (two independent measurements, 2026-07-18)

- **#720 OBBT graduation panel** (both flag arms): bound **916.15** > incumbent/oracle **837.73** — identical OFF and ON (see #727 comment).
- **#732 Stage-1 soundness spot-check**: bound **1310.6** > oracle **837.73**, `gap_certified=False`, `status=feasible` — **byte-identical with `DISCOPT_NARROW_BOX_BRANCH` OFF and ON**.

Reproduce (on `main`):
```
python -c "import discopt.modeling as dm; r=dm.from_nl('python/tests/data/minlplib_nl/syn05hfsg.nl').solve(time_limit=8); print(r.bound, r.objective, r.gap_certified)"
# -> ~1310 (or ~916), 837.732, False    (bound > incumbent 837.732; oracle 837.73)
```

## Suspected cause

The αBB underestimator fallback (#120). The solve logs:
> `McCormick 'nlp' objective bound is not a valid dual bound for nonconvex models (issue #120); falling back to the alphaBB underestimator.`

The αBB fallback is producing a dual "bound" that exceeds the true optimum. Because the tree is (correctly) not certified, this doesn't surface as a false certificate — but a `bound > incumbent` value should never enter the reported dual bound. Either the αBB underestimator is unsound on this instance, or an αBB-derived value is being surfaced as a dual bound where it should be abstained.

## Scope / relation

- Pre-existing on `main`; **not** caused by #720 or #732 (both verified flag-independent).
- Relates to **#120** (αBB fallback validity) and **#282**/`syn_*` process-synthesis family.

## Definition of done

Root-cause whether the αBB underestimator is unsound here or is being wrongly surfaced as a dual bound; fix so `syn05hfsg`'s reported/internal dual bound never exceeds its incumbent (min sense), with a regression test asserting `bound ≤ incumbent + tol`. `incorrect_count = 0` throughout.

jkitchin:
[...]

PR #974, opened 2026-08-09
fix(correctness): #971 the root-LP gate must model both ways a bound can differ

jkitchin:
## Summary

The #961 root-LP gate excused a difference from being a claim change **only** when the relaxation *fingerprint* also differed — encoding the rule "a difference is only attributable when both sides solved the same LP", i.e. *same matrix bytes ⇒ same LP answer*. CI falsified that: `contvar` was bucketed `changed` — the fingerprint **matched** — with the root LP bound moving `172170.3107274997 → 187283.4711213944` (8.8%) between two `ubuntu-x86-64` runners, bit-stable on each. The gate hard-failed on `main` and on PRs touching no solver code (#970).

**Entry experiment** (`discopt_benchmarks/scripts/issue971_root_lp_arithmetic_path.py`, new): hold tree, instance and matrix fixed; vary only the arithmetic path via OpenBLAS kernel dispatch (`OPENBLAS_CORETYPE`, `DYNAMIC_ARCH` wheels). `contvar` over 5 kernels →

```
  Haswell    fp=8be8329b9f3d status=optimal bound=172170.31072749975
  Nehalem    fp=8be8329b9f3d status=optimal bound=172170.31072749972
  SkylakeX   fp=8be8329b9f3d status=optimal bound=172170.3107274997
  Zen        fp=8be8329b9f3d status=optimal bound=172170.31072749975
  Prescott   fp=8be8329b9f3d status=optimal bound=172170.31072749972
REPS_COMPLETED=5
distinct_fingerprints=1  distinct_answers=3
same_bytes_different_answer=True
```

1 fingerprint, 3 distinct bounds — the assumption is provably wrong. The spread *at that grain* is last-digit; the CI pair is the evidence for the magnitude. The amplification is structural, not noise: the certified bound is a Neumaier–Shcherbina bound read off the simplex's own dual vertex, and a degenerate LP's optimal dual is **not unique**, so one flipped ratio-test tie-break selects a different — equally valid — certificate and the bound moves discretely.

Explicitly **not** a tolerance bump (8.8% is not last-digit noise, and widening a correctness gate so a red turns green is what CLAUDE.md §3 forbids). Instead the gate now separates what holds on every host from what does not:

- **Hard, host-independent, on every instance** — new `root_lp_violations`: the #961 status/bound contract in *both* directions, plus **validity** of the certified bound against the committed oracles (`cert-optima.json` ∪ `known_optima.toml`, sense-corrected per #759 — for a MAXIMIZE model the internal lower bound must not exceed `-optimum`). Violations bucket `unsound` and fail whatever the baseline says. This is coverage the gate did not previously have.
- **Bounded and loud** — a difference at an *identical* fingerprint buckets `unreproduced` (was `changed`), capped at `MAX_UNREPRODUCED_ROOT_LP = 3` and reachable only after the soundness assertions pass. `unchanged >= 50` is retained, and a floor of 40 oracle-checked instances proves the validity sweep actually fired (measured 43 of 66).

Contributes to #971. Unblocks the gate that was failing #970 and `main`.

### Known trade-off

A genuine code-induced bound change confined to ≤ 3 instances *whose fingerprint does not move* now prints rather than fails, where it previously hard-failed. That hole already existed on the `fingerprint_drift` path. Cross-host bit-equality of a degenerate LP's dual is not a property the engine provides, so it is bounded and reported rather than asserted.

## Correctness

No solver math is touched — the change is entirely in the test-support gate (`python/tests/support/claim_differential.py`) and its tests, plus a new probe script and a docstring in the baseline generator. Net effect on certification is a **strengthening**: every corpus root LP bound is now checked against a published optimum on every host, which the gate never did before.

- [x] Cannot produce a false certificate (N/A — no solver-math change; the gate gains a soundness assertion)
- [x] No validation, fallback, or safety guard was weakened to make a check pass — the per-instance hard bound-equality assertion is replaced by a *stronger* host-independent soundness assertion plus a capped, loud, soundness-conditioned escape bucket; the retained `unchanged >= 50` floor and the new 40-instance oracle floor keep the sweep measuring

- [x] `pytest -m smoke` → **917 passed, 19 skipped, 2 xpassed** (352 s)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **10 passed** (177 s). First run of that file showed one failure in `test_large_dense_jacobian_no_crash`; it passes in isolation (`1 passed`) and on a full re-run of the file. It is memory-heavy and CI runs it under `scripts/run_memory_capped_pytest.sh`; this PR touches no solver code.
- [x] `ruff check python/` / `ruff format --check python/` → clean (856 files)

Additionally, the claim-boundary trio (`test_claim_baseline_neutral.py`, `test_961_optimal_requires_bound.py`, `test_971_root_lp_arithmetic_path.py`) → **12 passed** (98 s), and a full corpus sweep on ubuntu/x86-64: **66/66 `unchanged`, 0 `unsound`, 43 oracle-checked**.

## Regression test

`python/tests/test_971_root_lp_arithmetic_path.py`:

- `test_same_bytes_different_bound_is_not_a_claim_change` — doctors the baseline for an exactly-reproduced instance to `contvar`'s shape (identical recorded fingerprint, bound moved 8.8%) and asserts the bucket is `unreproduced`. **Fail-before / pass-after confirmed** by running the pre-change `diff_root_lp` (loaded from `HEAD`) side by side with the new one on the same doctored row:

  ```
  OLD rule -> changed      | root LP status 'optimal' -> 'uncertified'
  NEW rule -> unreproduced | ... ; identical matrix bytes, so the arithmetic path differs
  ```

  `changed` is exactly the CI symptom this issue reports.
- `test_violations_flag_the_961_contract_in_both_directions`, `test_violations_flag_a_bound_above_the_reference_optimum`, `test_violations_respect_the_objective_sense` — pure-function tests proving the new soundness predicate bites on every arm (CLAUDE.md §6).
- `test_961_optimal_requires_bound.py::test_root_lp_gate_separates_the_two_ways_a_difference_arises` — updated from the old rule; keeps both non-reproducing buckets distinguishable so neither can collapse into a blanket skip.

---
_Generated by [Claude Code](https://claude.ai/code/session_01T3kki4Y6gXNxWHPaZiJHQF)_

### DIFF
--- a/discopt_benchmarks/scripts/gen_claim_baseline.py
+++ b/discopt_benchmarks/scripts/gen_claim_baseline.py
@@ -11,6 +11,16 @@ differential gate (plan §3.2) compares every canonical-cutover PR against; sinc
 tolerance), so a bound drift requires deliberately regenerating this file in the
 PR that causes it.
 
+Since #971 the gate reads the recorded root LP in two ways, because the two are
+not equally reproducible: the recorded pair must *reproduce* on the bulk of the
+corpus (a handful of instances legitimately differ across arithmetic paths at an
+identical fingerprint — the certified bound is a Neumaier-Shcherbina bound read
+off a degenerate LP's non-unique dual), while the *properties* of the current
+result — the #961 status/bound contract, and a bound at or below the published
+optimum — are asserted hard on every host. So a row recorded here is a
+reproducibility reference, not a correctness oracle; the oracle is
+``docs/dev/data/cert-optima.json`` plus ``python/tests/data/known_optima.toml``.
+
 Usage (from the repo root, with the built extension importable)::
 
     JAX_PLATFORMS=cpu JAX_ENABLE_X64=1 \
--- a/discopt_benchmarks/scripts/issue971_root_lp_arithmetic_path.py
+++ b/discopt_benchmarks/scripts/issue971_root_lp_arithmetic_path.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python
+"""Issue #971 entry experiment: is the root LP bound a function of the matrix alone?
+
+``diff_root_lp`` excuses a root-LP difference only when the relaxation
+*fingerprint* differs too, on the reasoning that "a difference is only
+attributable when both sides solved the same LP". That rule assumes
+**same matrix bytes ⇒ same LP answer**. #971 reports CI failures on ``contvar``
+bucketed ``changed`` — i.e. the fingerprint MATCHED and the bound still moved by
+8.8% — which falsifies the assumption if reproducible.
+
+This probe holds the tree, the instance and the matrix fixed and varies only the
+*arithmetic path*, via OpenBLAS's runtime kernel dispatch (``OPENBLAS_CORETYPE``;
+the wheels ship ``DYNAMIC_ARCH``, so the same binary runs a Nehalem / Haswell /
+SkylakeX / Zen kernel on demand — the same knob a heterogeneous CI runner pool
+turns implicitly). For each core type it reports, from a **fresh process**, the
+relaxation fingerprint and the root LP ``(status, bound)``.
+
+Kill criterion: if every core type yields the same ``(status, bound)`` with the
+same fingerprint, the arithmetic path is NOT the mechanism and #971's premise
+needs another explanation. If any two agree on the fingerprint and disagree on
+the bound, ``diff_root_lp``'s attributability rule is provably incomplete.
+
+Discipline: asserts the tree under test by path (CLAUDE.md §8), never swallows an
+exception (§7), prints a completed-rep count and exits non-zero when it is zero
+(§6), and streams per-rep progress (§10).
+
+Usage (from the repo root, with the built extension importable)::
+
+    python -u discopt_benchmarks/scripts/issue971_root_lp_arithmetic_path.py \
+        [--instance contvar] [--coretypes Haswell,Nehalem,SkylakeX,Zen]
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+_REPO = Path(__file__).resolve().parents[2]
+_NL_DIR = _REPO / "python" / "tests" / "data" / "minlplib_nl"
+
+# Kernel families an x86-64 runner pool actually spans: pre-FMA (Nehalem),
+# FMA/AVX2 (Haswell), AVX-512 (SkylakeX), and AMD (Zen). OpenBLAS falls back to
+# a generic kernel for a core type the CPU cannot run, which is itself a
+# distinct arithmetic path, so an unsupported name is still a useful arm.
+_DEFAULT_CORETYPES = ("Haswell", "Nehalem", "SkylakeX", "Zen", "Prescott")
+
+_CHILD = r"""
+import json, os, sys
+sys.path.insert(0, {python_dir!r})
+import discopt
+assert discopt.__file__ == {expect_init!r}, (
+    "wrong tree under test: %s != %s" % (discopt.__file__, {expect_init!r})
+)
+sys.path.insert(0, {tests_dir!r})
+from support.claim_differential import current_root_lp, current_row
+
+row = current_row({name!r})
+status, bound = current_root_lp({name!r})
+print("RESULT " + json.dumps(
+    {{"coretype": os.environ.get("OPENBLAS_CORETYPE", ""),
+      "fingerprint": row["fingerprint"],
+      "n_rows": row["n_rows"], "n_cols": row["n_cols"],
+      "status": status, "bound": bound}}))
+"""
+
+
+def _run_one(name: str, coretype: str) -> dict:
+    """Build + solve ``name`` in a fresh process under ``coretype``.
+
+    No exception handling (CLAUDE.md §7): a crashing arm must surface as a
+    crash, not as a missing row that reads like agreement.
+    """
+    env = dict(os.environ)
+    env["OPENBLAS_CORETYPE"] = coretype
+    env["JAX_PLATFORMS"] = "cpu"
+    env["JAX_ENABLE_X64"] = "1"
+    src = _CHILD.format(
+        python_dir=str(_REPO / "python"),
+        tests_dir=str(_REPO / "python" / "tests"),
+        expect_init=str(_REPO / "python" / "discopt" / "__init__.py"),
+        name=name,
+    )
+    proc = subprocess.run(
+        [sys.executable, "-u", "-c", src], cwd=str(_REPO), env=env, capture_output=True, text=True
+    )
+    if proc.returncode != 0:
+        raise RuntimeError(f"{coretype} arm failed (rc={proc.returncode}):\n{proc.stderr[-4000:]}")
+    line = next((ln for ln in proc.stdout.splitlines() if ln.startswith("RESULT ")), None)
+    if line is None:
+        raise RuntimeError(f"{coretype} arm printed no RESULT:\n{proc.stdout[-2000:]}")
+    return json.loads(line[len("RESULT ") :])
+
+
+def main() -> int:
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--instance", default="contvar")
+    ap.add_argument("--coretypes", default=",".join(_DEFAULT_CORETYPES))
+    args = ap.parse_args()
+
+    if not (_NL_DIR / f"{args.instance}.nl").exists():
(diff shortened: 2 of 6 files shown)


jkitchin:
## Review — one defect found and fixed; the gate's four thresholds verified by measurement

Reviewed against CLAUDE.md §1/§3 ("never weaken a validation ... to make a gate pass"),
since this PR converts a hard-failing bucket into a capped informational one.

### 1. Defect found: the new test reintroduced the predicate #973 had just removed

`test_same_bytes_different_bound_is_not_a_claim_change` selected the first instance whose
`diff_root_lp` status was `"unchanged"` and then **asserted** its relaxation fingerprint
matched the committed baseline. Those are not the same condition — `diff_root_lp` computes
the fingerprint only once it has already seen a status/bound difference — so an instance
whose root LP reproduces can still have drifted bytes. Byte-identity is the *premise* of
the bucket under test, so the test fails for a reason unrelated to the rule it checks.

Not hypothetical. Measured on macOS/arm64 at this commit:

```
TEST971_PREDICATE: first-unchanged=4stufen  fingerprint_matches=False
```

64 of 66 instances are `unchanged`, the first is `4stufen`, and `4stufen`'s fingerprint
does **not** match the committed linux/x86-64 baseline — so this test failed here before
`d57f1f6e`. It is the same defect #973 removed from `test_961_optimal_requires_bound.py`,
arriving through a new file; fixed with the same predicate (select on the fingerprint
itself, *then* on `unchanged`).

### 2. The four thresholds actually fire, and the numbers in the PR body reproduce

Full 66-instance corpus sweep on this branch (macOS/arm64, load-gated, §8 marker asserted:
`MAX_UNREPRODUCED_ROOT_LP` present and `discopt.__file__` inside the worktree):

```
INSTANCES_CLASSIFIED=66
  unchanged           64
  changed              0
  unsound              0
  unreproduced         0
  fingerprint_drift    2   ['bchoco06', 'contvar']
  error                0
ORACLE_CHECKED=43
PASS  unchanged >= 50          (actual 64)
PASS  unsound == 0             (actual 0)
PASS  unreproduced <= 3        (actual 0)
PASS  oracle_checked >= 40     (actual 43)
ASSERTIONS_EXECUTED=6
```

`ORACLE_CHECKED=43` matches the PR body's claimed basis for `MIN_ORACLE_CHECKED_ROOT_LP = 40`
exactly, so that floor is measured rather than asserted. `contvar` lands in
`fingerprint_drift` on arm64, consistent with #971's own local finding
(`uncertified`, no bound, differing fingerprint, 5/5 deterministic).

### 3. The sense correction is correct but **unexercised on corpus data**

`root_lp_violations` negates the reference optimum for a MAXIMIZE model. The corpus has 4
maximize instances, and none of them reaches the comparison:

```
bchoco06   maximize=True  status=optimal      bound=-0.9999918060862106   optimum=None
bchoco07   maximize=True  status=optimal      bound=-1.0000000000001597   optimum=None
bchoco08   maximize=True  status=optimal      bound=-1.0000000000002143   optimum=None
syn05hfsg  maximize=True  status=uncertified  bound=None                  optimum=837.7324009
MAXIMIZE_INSTANCES_PROBED=4
```

So the sense branch contributes **0 of the 43** oracle comparisons; it is covered only by
the pure-function test `test_violations_respect_the_objective_sense`. That test does assert
both directions, so the logic is pinned — but the branch has no real-data coverage, and
adding a maximize instance with a published optimum to the corpus would be the way to get
it. Recording this rather than leaving it implied (§6).

### 4. On the weakening, explicitly

`changed` is now reachable **only** via the stale-schema path; every genuine difference
routes to `fingerprint_drift` or `unreproduced`. So the honest statement is: a solver-path
change that moves the certified bound on ≤ 3 instances, keeps the bound valid, and leaves
the matrix bytes identical now prints where it previously hard-failed. The PR body already
says this.

I think it is the right trade, for three reasons that are not "it makes CI green":

* The gate is a **reproducibility** gate, not the certificate invariant, and it was
  producing false failures on `main` and on PRs touching no solver code (#971 documents
  three runs). A gate that fires on runner microarchitecture is not measuring the thing it
  names.
* It **adds** a hard, host-independent soundness assertion the gate never had — the bound's
  validity against the committed oracles, on 43 instances. That catches a class
  (an invalid bound that happens to match the baseline) the old gate could not see at all.
* It is #971's own suggested direction (2), which the issue calls "the more faithful gate".

I considered proposing an **allowlist of instance names** instead of a bare count cap,
which would catch "a *new* instance became unreproducible". Rejecting it: the phenomenon is
host-dependent by construction (`contvar` is `unreproduced` on x86-64 and
`fingerprint_drift` on arm64), so an allowlist would have to be a cross-host union and a
new runner microarchitecture could legitimately extend it — reintroducing exactly the
flakiness this fixes. A count cap is the right shape for a host-dependent effect.

### 5. Magnitude claim

The 8.8% figure rests on the CI pair in #971 (two ubuntu-x86-64 runners, identical
fingerprint, `172170.3107274997` vs `187283.4711213944`, bit-identical across two
failures), not on the `OPENBLAS_CORETYPE` sweep, which shows only last-digit spread
(`distinct_fingerprints=1 distinct_answers=3`). The docstrings state that distinction
correctly — the sweep demonstrates the *mechanism class*, the CI pair is the evidence for
the *magnitude*. No change needed, noting it so the reader does not over-read the sweep.

jkitchin:
Agreed on all five points. The predicate defect in §1 is real and your fix is the right one — selecting on `"unchanged"` asserted a byte-identity it never established, and byte-identity is the *premise* of the bucket under test, so it could only ever fail for the wrong reason. Verified `d57f1f6` green here after rebuilding against the merged `main` (#956 touches Rust): claim-boundary trio **12 passed** in 106 s on linux/x86-64.

### Post-merge sweep on linux/x86-64 — matches your arm64 numbers exactly

```
INSTANCES_CLASSIFIED=66
  unchanged           64
  changed              0   []
  unsound              0   []
  unreproduced         0   []
  fingerprint_drift    2   ['bchoco06', 'contvar']
  error                0   []
ORACLE_CHECKED=43
```

Same split, same two byte-drifters, same oracle coverage as your macOS/arm64 run. One thing worth recording that the match hides: **on x86-64 this is a change from pre-merge**. Before merging `main`, this host classified **66/66 `unchanged`, 0 `fingerprint_drift`**; after the merge, `bchoco06` and `contvar` drift. So it is #956's envelope work that moved those two instances' relaxation bytes here, not the host — the shape gate (`test_current_build_matches_committed_baseline_shape`) stays green, which is the expected reading of a coefficient-level change at identical shape. It also means `ORACLE_CHECKED=43` reproducing across two OSes and two architectures is a stronger basis for `MIN_ORACLE_CHECKED_ROOT_LP = 40` than my single-host measurement was.

### §3 — the sense branch: I cannot close it here, and I am not guessing a value

Your reading is right, and the gap is narrower than "add a maximize instance": the corpus already has four, and the blocker is only the oracle.

- `bchoco06/07/08` — maximize **and** they certify a bound, so they are exactly the instances that would exercise the branch. Neither `cert-optima.json` nor `known_optima.toml` has an optimum for any of them, and neither does anything else in the tree (searched).
- `syn05hfsg` — has a published optimum (`837.7324009`, from #759) but its root LP is `uncertified`, so there is no bound to compare and no way to force one.

The source that would supply the missing values is `minlplib.solu`, which `discopt_benchmarks/scripts/minlplib_solu.py` reads from the Dropbox snapshot — not reachable from this session's environment (`FileNotFoundError` at the default path, and `DISCOPT_MINLPLIB_SOLU` has nothing to point at). Writing a remembered number into a correctness oracle is the one thing I will not do here, so the branch stays covered by `test_violations_respect_the_objective_sense` only.

Concretely, what closes it: add `=opt=` (or `=best=`, which is still a valid ceiling) for `bchoco06/07/08` from the solu snapshot to `known_optima.toml`. That alone converts 3 of the 43 comparisons to the maximize path with no change to `root_lp_violations`. Happy to do it in a follow-up on a machine with the snapshot, or here if you paste the three values.

### §4 — the weakening

Your statement of it is more precise than mine and I agree with the rejection of the name allowlist: `contvar` being `unreproduced` on x86-64 and `fingerprint_drift` on arm64 is exactly the case an allowlist would mis-handle, and a cross-host union that a new microarchitecture can legitimately extend is the flakiness this PR removes, reintroduced one layer up. The count cap is the right shape for a host-dependent effect.

§2 and §5 need no action from me — thanks for re-deriving the thresholds independently rather than taking the PR body's numbers.

CI on `d57f1f6` is still running (Rust, Lint+typecheck, AMP, codecov green; claim-boundary and the two Python jobs in progress). I am subscribed and will drive it to green.

---

PR #983, opened 2026-08-11
fix(gbd): #977 solve the GBD master on the exact simplex, not the IPM

### LINKED ISSUE #977: correctness: GBD master returns a point violating its own Benders cut by 8.7e-06 (scale-proportional); nightly slow-correctness lane red since 08-09
## Summary

The nightly `slow`+`correctness` lane has been **red since 2026-08-09** on
`test_gbd.py::test_gbd_soundness_fuzz_vs_monolithic`, with a deterministic,
byte-identical failure:

```
RuntimeError: MILP-BB returned an infeasible point labeled feasible/optimal: A_ub row 0 by 8.703243e-06
```

Nobody was told, because the job whose purpose is to file exactly this
(`report-nightly-failure` in `ci.yml`) fails itself on every invocation — that is
a separate defect with its own PR; this issue is about the solver behaviour.

Affected nightly runs (both identical): [31300814621][r1] (08-09), [31368534754][r2] (08-10).

[r1]: https://github.com/jkitchin/discopt/actions/runs/31300814621
[r2]: https://github.com/jkitchin/discopt/actions/runs/31368534754

## Reproduction

On `main` @ `40c6f64d`, **1.38 s**, first fuzz draw, no flake:

```
PYTHONPATH=python pytest python/tests/test_gbd.py::test_gbd_soundness_fuzz_vs_monolithic -m ""
```

The fuzz loop is seeded (`np.random.default_rng(20260620)`), so the failing
instance is fixed. Reconstructed standalone:

```
ny=2  nx=3  scale=100.0
a  = [67.48708288244717, 158.92957507697867, 168.60525267354743]
sh = [1.8204302511934478, 1.084131670792487, 1.6519769321131845]
cy = [296.0583639642992, 66.048910814793]

min  sum_j a_j (x_j - sh_j)^2 + sum_i cy_i y_i
s.t. sum_j x_j == 2.61780744447098 * sum_i y_i
     x in [0,5]^3,  y binary^2
```

`solve_benders(m, time_limit=30)` raises. The **monolithic solve of the same
model succeeds** (`status=optimal obj=205.04942868022658`), so the model is fine
— this is in the GBD master, not the formulation.

## What is firing

The `#952`-family exit gate in `_solve_milp_bb` (`python/discopt/solver.py:18901`,
message at `:18905`), which verifies the incumbent against the original rows and
the declared box at `abs=1e-6` before returning it as `optimal`. It is a
deliberate refusal ("Refusal, not repair"), and it is working as designed.

**The violated row is a Benders cut, not a user row.** The master's `A_ub` is
assembled at `python/discopt/decomposition/benders/solver.py:323-329` as the
master's own rows stacked with the accumulated optimality-cut rows. This model
declares exactly one constraint — an *equality* on the second-stage `x` — so
there are no user inequality rows on the master at all; `A_ub row 0` is
necessarily a cut row. Its coefficients come from subproblem duals, which scale
with the objective coefficients.

## Evidence: the violation is purely a scale effect

Same instance, objective `scale` swept (everything else identical):

| scale | result | violation |
| --- | --- | --- |
| 0.1 | ok | — |
| 1.0 | ok | — |
| 10.0 | ok | — |
| 100.0 | **RAISED** | 8.703243e-06 |
| 1000.0 | **RAISED** | 8.705498e-05 |

The absolute violation is **exactly proportional to scale** (8.703e-06 → 8.705e-05
for a 10× scale-up). The *relative* violation is constant at **~8.7e-08** on every
row of the table, including the three that pass. Nothing changes about the
solution quality across the sweep; only whether a constant relative error crosses
a fixed absolute threshold.

So the point being refused is relatively feasible to ~1e-7 and would be accepted
at any scale below ~50.

## Why this should not be closed by widening the gate

The tempting fix — give the exit gate a relative component — is exactly the
tolerance-tweak-to-mask-a-bug that `CLAUDE.md` §3 forbids, and it would weaken a
guard that was added two days ago precisely because this exit had no verification
at all. **The gate should not be relaxed as the first move.** The open question is
upstream of it:

1. **Should the master be returning points that violate its own cuts at all?**
   A cut row is a hard constraint of the master MILP. If the MILP engine is
   satisfying cut rows only to its *relative* tolerance while the repo's declared
   contract (`abs=1e-6`, `conftest.py`) is absolute, then the two tolerance models
   disagree and one of them is wrong — that disagreement is the bug, and it is
   not confined to GBD.
2. **Or are the cut coefficients themselves carrying scale-proportional error?**
   The duals feeding `cut_x`/`cut_rhs` are computed at objective scale; if the
   error is manufactured when the cut is built rather than when it is solved, the
   fix belongs in the cut construction.

These are distinguishable: instrument the master solve to report the residual of
each cut row at the returned point, both as returned by `milp()` and as
recomputed from `cut_x`/`cut_rhs`, and compare. I have not run that, so I am not
asserting which one it is.

A third possibility worth ruling out first: that the repo's declared absolute
tolerance is simply not meaningful for models with O(100) coefficients, in which
case the *declared contract* needs revisiting deliberately and repo-wide — not
patched at one call site.

## Done means

- Root cause identified between (1) and (2) above, with the measurement that
  distinguishes them.
- The nightly lane green, without the exit gate's strictness being reduced to get
  there — or, if a tolerance-model change is genuinely the right answer, that
  change argued explicitly and applied to the declared contract rather than to
  this one gate.
- A regression test at a scale that reproduces (scale >= 100), since the existing
  fuzz test only catches it because one of its 60 seeded draws happens to land
  there.

## Notes

- Verified reproducible on `main` @ `40c6f64d` outside CI; the two nightly runs
  and the local run agree to the digit (`8.703243e-06`).
- Relates to #952, which added this class of exit gate to `_solve_miqp_bb` and
  whose sibling gate on `_solve_milp_bb` is the one firing here. That issue is
  closed and stays closed — the gate is not the defect.

jkitchin:
## Summary

The nightly slow-correctness lane went red on `test_gbd_soundness_fuzz_vs_monolithic` with the #952 exit gate in `_solve_milp_bb` refusing the GBD master's own incumbent (`A_ub row 0 by 8.703243e-06`). `A_ub row 0` is a Benders optimality cut, so the master was violating its own cut.

**Root cause is (1) in the issue — the engine, not the cut coefficients.** The distinguishing measurement recomputes each row residual from the *same* `(A_ub, b_ub)` arrays handed to `milp()`, so any error in the cut coefficients cancels exactly; whatever remains was manufactured by the solve. Two probes, both with the cut rows **exact by construction**:

| probe | POUNCE-IPM-backed engine | Rust simplex |
| --- | --- | --- |
| 520 synthetic master-shaped MILPs | up to 9.8e-07 | ~1e-14 |
| 60 GBD instances at scale ≥ 100 (1098 rows compared) | **5.924110e-06** | 1.488331e-08 |

The defect is a conflation: `prefer_pounce` is derived from `nlp_solver`, which selects the engine for the *NLP recourse subproblem*, and was then reused to route the **master** — a pure linear MILP — to the POUNCE-IPM-backed B&B, whose returned points are interior rather than vertices. `get_milp_solver`'s own default is simplex-first for exactly this reason.

**Second defect fixed by the same line.** The master objective *is* the GBD lower `bound`, so the analytic-centre value drifted above the incumbent it certifies. On the pinned instance — before: `bound 8459.266446189884 > objective 8459.266427579778` (`bound <= incumbent` broken); after: `bound 8459.266452114014 <= objective 8459.266452379557`, with the objective now agreeing with the monolithic optimum to 9e-09 instead of 2.5e-05.

I could not reproduce the issue's exact instance: 60 commits landed since `40c6f64d` and it now passes. I pinned an equivalent live reproduction instead (rng seed 90059, scale 1000), drawn the same way the soundness fuzz draws its instances so it stays a member of that class rather than a special case.

Closes #977

## Retraction: the class-wide extension was falsified by CI

**The first version of this PR (53dbc17) claimed more than it could deliver, and I am withdrawing that claim.** It applied the same routing fix to *classical* Benders, the Lagrangian solver, and the Lagrangian node bounder, on the argument that this fixes the class rather than the instance (CLAUDE.md §2). CI falsified it — main is green at `aca13dd`, so all five failures on that push were mine:

- `test_decomposition_solve_equivalence`, 3 cases → `iteration_limit` instead of `optimal`, each logging *"Benders iteration N added no separating cut (degenerate or missing duals)"* twice.
- `test_952_miqp_incumbent_feasibility`, 2 cases → *"the funnel never declined anything"*. That test reaches `_pounce_snap_incumbent` through `solve_benders` on a **linear** model, i.e. through the classical-Benders master; pinning that master to the simplex removes the IPM path the test exists to exercise.

The mechanism behind the first group is real and worth recording: an exact master returns a **vertex**, and at a vertex the recourse LP is degenerate, so classical Benders' cut generation produces no separating cut and stalls with the gap still open. The interior-point master had been masking that weakness by proposing luckier, non-degenerate points. That is the classical-Benders analogue of the #946 degenerate-recourse work, it needs its own change, and shipping it half-done would trade one defect for another. The scope note is recorded in the `gbd.py` comment so it does not have to be re-derived.

**Also retracted: a reordering of the Benders loop** that moved the gap check ahead of the stall break. My hypothesis was that a converged run exited `iteration_limit` because the break preempted the gap check. Measurement falsified it: `solver.py` already promotes a closed-gap stall after the loop (*"Promote to optimal if the final master bound already meets the incumbent"*), so the reordering is a no-op — 40 seeded instances gave byte-identical status, objective and bound under both orders, and the regression test I wrote for it passed under the old order too. It is out rather than shipped on an unverified hypothesis (CLAUDE.md §4).

What remains is scoped to the GBD master, which is where #977 actually fired.

## Correctness

The exit gate is **untouched** — no tolerance was widened, and no relative component was added. The gate was correct to refuse; the point it refused should never have been produced. The fix is upstream of the gate, at the engine that produced the point.

The change moves GBD master solves from an analytic-centre objective to an exact vertex optimum, which strictly *strengthens* the certificate: the GBD dual bound is the master objective, and an inexact one is not rigorous. The measured `bound > incumbent` violation above is that non-rigour showing up directly, and it is gone.

- [x] Cannot produce a false certificate — the change makes the reported dual bound exact rather than analytic-centre; no relaxation, cut, or reduction was altered

This container's POUNCE build lacks `pounce.NlExpr`. Every failure listed as pre-existing was confirmed identical against **main's source** (`git checkout aca13dd -- python/discopt python/tests`), not merely against a dirty tree.

- [x] `pytest -m smoke` → 915 passed, 22 skipped, 2 xpassed, 18 failed — all 18 pre-existing (identical set on main's source)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → 10 passed. (An earlier run showed 1 failure at 307s vs 204s baseline; that was load-induced flake from my own background probes — CLAUDE.md §9. Re-run at load ~1.1: 10 passed in 217s; the test passes in isolation at 36.01s vs 36.28s baseline.)
- [x] `ruff check python/` → All checks passed; `ruff format --check python/` → 873 files already formatted

Targeted re-run after the re-scope — `test_gbd` + `test_946_gbd_degenerate_multiplier` + `test_decomposition_solve_equivalence` + `test_952_miqp_incumbent_feasibility` + `test_lagrangian_bb_hook` → **40 passed, 2 failed**; both failures are the `test_952` funnel cases that fail identically on main's source here, and are the two CI cases this re-scope restores.

## Regression test

`python/tests/test_gbd.py::test_gbd_master_satisfies_its_own_cuts_at_scale`, at scale 1000 (≥ 100 as the issue requires, since the effect is scale-proportional).

It asserts (a) the master's returned point satisfies the rows it was handed at the declared `abs=1e-6`, and (b) `bound <= incumbent`. Both fail on the pre-fix routing and hold after. Fail-before confirmed by reverting only the `gbd.py` routing line:

```
AssertionError: GBD master returned a point violating its own rows by 5.924110e-06,
past the declared abs=1e-6 (over 20 rows compared)
```

It asserts a non-zero executed-comparison count, so it cannot pass by measuring nothing (CLAUDE.md §6).

Keeping this section rather than deleting it, because the change *is* bound-changing in the numeric sense — reported GBD bounds move — and that deserves an explicit argument.

The §5 default-off-flag regime does not fit here: it governs relaxations, cuts, and reductions, where a new mechanism must prove itself sound and net-positive before it is trusted. This adds no mechanism. It replaces an inexact evaluation of an existing bound with the exact one, and the direction is not "tighter, hopefully still valid" — it is "rigorous instead of drifting". Gating it default-off would leave a measured certificate-invariant violation (`bound > incumbent`) enabled by default, which inverts the point of the gate.

Measurement, on the pinned instance and the 60-instance panel at scale ≥ 100:

- worst master row violation: 5.924110e-06 → 1.488331e-08 (1098 rows compared)
- certificate invariant `bound <= incumbent`: violated → holds
- objective vs monolithic optimum: 2.5e-05 → 9e-09 agreement
- `test_946_gbd_degenerate_multiplier`'s full-budget companion still certifies the optimum in ≤ 10 recourse solves, unchanged

## Note on one changed assertion

`test_946_gbd_degenerate_multiplier::test_nonbinary_first_stage_explains_the_lost_certificate` (the `max_iterations=1` run) no longer asserts that one iteration reaches the optimum. That held only by tie-break: the whole cost there is nonlinear and lives in the recourse, so the first master solve has a **zero objective** (measured `master_obj = 0.0` under both engines) and every feasible first-stage point is optimal for it. The old engine happened to return `y = 1`, the eventual optimum; the exact simplex returns `y = 0`. Both are optima of that master.

The test now asserts the invariant that survives — the run brackets the true optimum from both sides (`bound <= -1` and `incumbent >= -1`). Its soundness assertions, its degeneracy precondition, its `iteration_limit` status check, its logged-explanation check, and the full-budget companion that still pins the optimum are all unchanged.

### DIFF
--- a/python/discopt/decomposition/benders/gbd.py
+++ b/python/discopt/decomposition/benders/gbd.py
@@ -272,10 +272,34 @@ def solve_gbd(
     from discopt.solvers.nlp_ipopt import _infer_constraint_bounds
 
     t0 = time.time()
-    # #977: the GBD master is a linear MILP -> exact-vertex simplex, never routed by
-    # ``nlp_solver`` (which selects the *NLP recourse* engine). See the companion
-    # comment in ``benders/solver.py`` for the measurement; ``get_milp_solver`` still
-    # falls back to POUNCE when the Rust simplex binding is unavailable.
+    # #977: the GBD master is a *linear* MILP, so it is pinned to the exact-vertex
+    # simplex B&B and NOT routed by ``nlp_solver``, which selects the engine for the
+    # *NLP recourse* subproblem. Reusing it here sent the master to the
+    # POUNCE-IPM-backed B&B, whose returned points are interior, not vertices.
+    #
+    # Measured on the master rows the engine was handed (identical arrays in and
+    # out, so any error in the cut coefficients cancels exactly — that is what
+    # distinguished the two candidate root causes in #977): the IPM path leaves row
+    # residuals up to 5.9e-06, six times the repo's declared ``abs=1e-6``, where the
+    # simplex is exact to ~1e-14 on the same input. Two consequences, both fixed by
+    # this line: the master could hand back a point violating its own optimality cut
+    # (the #952 exit gate then correctly refuses it, which is how #977 surfaced), and
+    # the master objective — which *is* the GBD lower ``bound`` — was an
+    # analytic-centre value that drifted above the incumbent it certifies, breaking
+    # ``bound <= incumbent``. ``get_milp_solver`` still falls back to POUNCE when the
+    # Rust simplex binding is unavailable, so POUNCE-only installs keep working.
+    #
+    # Scope note (#977, falsified extension): the same conflation exists in
+    # *classical* Benders (``solver.py``) and in the Lagrangian solver/node bounder.
+    # Pinning those masters to the simplex too was tried and reverted — it is sound
+    # but not free. An exact master returns a vertex where the recourse LP is
+    # degenerate, and classical Benders' cut generation then produces no separating
+    # cut and stalls out with the gap still open (measured in CI: three
+    # ``test_decomposition_solve_equivalence`` cases fell to ``iteration_limit``,
+    # with "added no separating cut" twice per run). The interior-point master had
+    # been masking that weakness by proposing luckier points. Fixing it is the
+    # classical-Benders analogue of the #946 degenerate-recourse work and needs its
+    # own change; this line stays scoped to GBD, where #977 actually fired.
     milp = get_milp_solver(backend="simplex")
 
     if structure is None:
--- a/python/discopt/decomposition/benders/gbd.py
+++ b/python/discopt/decomposition/benders/gbd.py
@@ -272,8 +272,11 @@ def solve_gbd(
     from discopt.solvers.nlp_ipopt import _infer_constraint_bounds
 
     t0 = time.time()
-    prefer_pounce = nlp_solver == "pounce"
-    milp = get_milp_solver(prefer_pounce=prefer_pounce)
+    # #977: the GBD master is a linear MILP -> exact-vertex simplex, never routed by
+    # ``nlp_solver`` (which selects the *NLP recourse* engine). See the companion
+    # comment in ``benders/solver.py`` for the measurement; ``get_milp_solver`` still
+    # falls back to POUNCE when the Rust simplex binding is unavailable.
+    milp = get_milp_solver(backend="simplex")
 
     if structure is None:
         structure = detect_decomposition(model)
--- a/python/discopt/decomposition/benders/solver.py
+++ b/python/discopt/decomposition/benders/solver.py
@@ -198,7 +198,22 @@ def solve_benders(
     # with POUNCE's interior-point (analytic-centre) duals and an inexact primal,
     # with no HiGHS dependency. Whichever LP backend is installed is used.
     dual_lp = get_lp_solver(prefer_pounce=cfg.prefer_pounce)
-    milp = get_milp_solver(prefer_pounce=cfg.prefer_pounce)
+    # #977: the master is a *linear* MILP, so it is pinned to the exact-vertex
+    # simplex B&B and NOT routed by ``prefer_pounce``. ``prefer_pounce`` comes from
+    # ``nlp_solver``, which selects the engine for the *recourse* subproblem; reusing
+    # it here sent the master to the POUNCE-IPM-backed B&B, whose returned points are
+    # interior, not vertices. Measured on the master rows the engine was handed
+    # (identical arrays in and out, so cut-construction error cancels exactly): the
+    # IPM path leaves residuals up to 5.9e-06 — six times the repo's declared
+    # ``abs=1e-6`` — where the simplex is exact to ~1e-14 on the same input. Two
+    # consequences, both fixed by this line: the master could hand back a point
+    # violating its own optimality cut (the #952 exit gate then correctly refuses it,
+    # which is how #977 surfaced), and the master objective — which *is* the Benders
+    # lower ``bound`` — was an analytic-centre value that drifted above the incumbent
+    # it certifies, breaking ``bound <= incumbent``. ``get_milp_solver`` still falls
+    # back to POUNCE when the Rust simplex binding is unavailable, so POUNCE-only
+    # installs keep working.
+    milp = get_milp_solver(backend="simplex")
 
     if structure is None:
         structure = detect_decomposition(model)
--- a/python/discopt/decomposition/benders/solver.py
+++ b/python/discopt/decomposition/benders/solver.py
@@ -198,22 +198,7 @@ def solve_benders(
     # with POUNCE's interior-point (analytic-centre) duals and an inexact primal,
     # with no HiGHS dependency. Whichever LP backend is installed is used.
     dual_lp = get_lp_solver(prefer_pounce=cfg.prefer_pounce)
-    # #977: the master is a *linear* MILP, so it is pinned to the exact-vertex
-    # simplex B&B and NOT routed by ``prefer_pounce``. ``prefer_pounce`` comes from
-    # ``nlp_solver``, which selects the engine for the *recourse* subproblem; reusing
-    # it here sent the master to the POUNCE-IPM-backed B&B, whose returned points are
-    # interior, not vertices. Measured on the master rows the engine was handed
-    # (identical arrays in and out, so cut-construction error cancels exactly): the
-    # IPM path leaves residuals up to 5.9e-06 — six times the repo's declared
-    # ``abs=1e-6`` — where the simplex is exact to ~1e-14 on the same input. Two
-    # consequences, both fixed by this line: the master could hand back a point
-    # violating its own optimality cut (the #952 exit gate then correctly refuses it,
-    # which is how #977 surfaced), and the master objective — which *is* the Benders
-    # lower ``bound`` — was an analytic-centre value that drifted above the incumbent
-    # it certifies, breaking ``bound <= incumbent``. ``get_milp_solver`` still falls
-    # back to POUNCE when the Rust simplex binding is unavailable, so POUNCE-only
-    # installs keep working.
-    milp = get_milp_solver(backend="simplex")
+    milp = get_milp_solver(prefer_pounce=cfg.prefer_pounce)
 
     if structure is None:
         structure = detect_decomposition(model)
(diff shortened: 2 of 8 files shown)


jkitchin:
CI red on the first push was mine, not flake — main is green at `aca13dd`, the commit this branches from. Pushed `09dde9b` to re-scope, and **retracting two claims** from the original PR description (now edited accordingly).

**1. The class-wide extension was wrong.** I had applied the routing fix to classical Benders and the Lagrangian solver/node bounder as well, arguing it fixes the class rather than the instance. CI falsified that:

- `test_decomposition_solve_equivalence`, 3 cases → `iteration_limit` instead of `optimal`, each logging *"Benders iteration N added no separating cut (degenerate or missing duals)"* twice.
- `test_952_miqp_incumbent_feasibility`, 2 cases → *"the funnel never declined anything"*. That test reaches `_pounce_snap_incumbent` via `solve_benders` on a **linear** model — i.e. through the classical-Benders master — so pinning that master to the simplex removes the very IPM path the test exists to exercise.

The mechanism behind the first group is real: an exact master returns a **vertex**, the recourse LP is degenerate at a vertex, and classical Benders' cut generation then produces no separating cut and stalls with the gap still open. The interior-point master had been masking that by proposing luckier, non-degenerate points. That is the classical-Benders analogue of the #946 degenerate-recourse work — it needs its own change, and shipping it half-done trades one defect for another. Recorded as a scope note in the `gbd.py` comment so it doesn't have to be re-derived.

**2. The Benders loop reordering was unjustified.** I had moved the gap check ahead of the stall break on the hypothesis that a converged run exited `iteration_limit` because the break preempted the gap check. Measurement killed it: `solver.py` already promotes a closed-gap stall after the loop (*"Promote to optimal if the final master bound already meets the incumbent"*), so the reordering is a no-op — 40 seeded instances gave byte-identical status, objective and bound under both orders, and the regression test I wrote for it passed under the old order too. Removed rather than shipped on an unverified hypothesis.

What survives is the GBD master fix, which is where #977 actually fired, with its measurements unchanged: master row residual **5.924110e-06 → 0.0** on the pinned instance (worst 1.488331e-08 over 60 instances at scale ≥ 100, 1098 rows compared), and the certificate invariant `bound ≤ incumbent` restored. `test_gbd_master_satisfies_its_own_cuts_at_scale` still fails before at 5.924110e-06 and passes after; `test_gbd_soundness_fuzz_vs_monolithic` — the red nightly test — passes.

One caveat worth stating plainly: none of the five CI failures reproduce in this container (its POUNCE build lacks `pounce.NlExpr`), so the re-scope is verified against main's source locally and by reasoning from the CI logs, not by a local reproduction of the CI failures themselves. This CI run is the real check.

---

Issue #966, opened 2026-08-09
Separation-round budgets don't cover the round's non-LP cost, and downstream phases poll the global deadline coarsely — the measured blocker for DISCOPT_LP_WARM_DEADLINE graduation


jkitchin:
## Summary

The #928 fix made the warm pure-LP path honour its caller's `time_limit`
**bound-preservingly** (deadline exits bank the dual loop's monotone
best-so-far floor; seven §5 panel runs all cert-clean, zero per-LP deadline
violations in every probe). The flag still cannot graduate, and the residual
overrun is measurably **not in the LP layer any more** — it is in two
caller-side deficiencies this issue tracks:

1. **A separation round's budget clamps only the LP, not the round.** With
   LPs honouring their budgets they return sooner, so the enclosing
   separated-relaxation loop fits *more rounds* — and each round's non-LP
   cost (`build_uniform_relaxation` ~1.1 s on contvar, plus separation,
   interval Hessians, NS/FBBT evaluation) is spent **after** the round's
   budget check, unclamped. Measured (contvar @ 20 s,
   `scratchpad/issue928_{contvar_probe,blowup_hunt}.py` on branch
   `claude/issue-928-as1vvj`):

   | arm | relax solves | time in relax solves | wall |
   |---|---|---|---|
   | OFF | 20 | 6.3 s | ~22 s |
   | ON  | 30 | 2.8 s | 27–33 s |

   The ON arm spends **less** time solving LPs and more time overall: ~10
   extra rounds of unclamped build/separation work. Zero deadline violations
   at the `MilpRelaxationModel.solve` level (one 1.24 s call against a
   0.19 s budget was the worst observed — the prepare/poll-granularity
   overshoot, bounded and small).

2. **Downstream phases have coarse global-deadline compliance, in BOTH
   arms.** Rare severe modes observed in the 20 s binding-subset reps
   (`discopt_benchmarks/results/issue928_binding_rep*.json`): ON contvar
   **500.6 s** (rep 1), ON bchoco08 **80.9 s** (rep 2) — and OFF
   heatexch_gen3 **200.5 s** in the same rep set, the same pre-existing
   class recorded as 78–176 s OFF-arm noise in `performance-plan.md` §14.
   The flag redistributes which phases get reached (banked bounds → more /
   different downstream work), so it amplifies a deficiency it did not
   create. The blowups sit outside `MilpRelaxationModel.solve` entirely
[...]

Issue #694, opened 2026-07-17
Make the root relaxation build anytime/incremental so the dual bound accrues (dissolve the #654 §8 truncation fork)


jkitchin:
## Summary

#654 closed with a **measured, irreducible floor**: `solve(time_limit=2)` on sonet23v4 takes **24.5s**, because the dual bound (-53974.375) is produced by a single **16.8s uninterruptible LP build** that `docs/dev/baron-gap-plan.md` §8.1 forbids truncating. Truncate it and you lose the bound; don't, and you blow the budget.

That fork exists **only because the bound materializes at the *end* of a monolithic build.** SOTA solvers don't face it: a dual simplex can stop at any iteration and still hand back a valid bound (Neumaier–Shcherbina safe dual), and BARON/SCIP's OBBT accrues per variable. Their bound **accrues continuously**, so truncation is always safe.

**This issue: make the root relaxation build anytime/incremental so the bound accrues.** That dissolves the §8 fork rather than documenting around it — there is no "lose the bound vs overrun" choice left to make.

## Evidence (measured 2026-07-17; `docs/dev/baron-gap-plan.md` §8.6)

Per-phase, `T=2s`, the #654 class:

| instance | wall | fallback | dominant atom | dual bound |
|---|---|---|---|---|
| graphpart_clique-70 | 5.5s | 2.6s | relaxation build | None |
| sonet22v4 | 6.7s | 3.9s | build 2.5s | None |
| qap | 11.4s | 7.7s | build 5.2s | None |
| eg_all_s | 18.0s | 11.0s | build 8.4s | None |
| **sonet23v4** | **24.5s** | **22.3s** | **`sep` LP build 16.8s** (vs a **1.0s** solve budget) | -53974.375 |

The build is **not bounded by the solve's `time_limit`** — that is the entire residual. Note §8.6's two binding falsifications: neither bound candidate dominates (so nothing can be dropped), and on this class `plain` is `None`, so the expensive `sep` is the *sole* producer.

## Hypothesis

> If the root McCormick relaxation is built in an order that reaches a **finite** LP bound early, then a build interrupted at any checkpoint still yields a *valid* (weaker) dual bound — so the build becomes truncatable, and wall can honor `T` while retaining a bound on the #654 class.

[...]

PR #985, opened 2026-08-11
fix(budget): #966 a short-granted node round yields instead of skipping

jkitchin:
## Summary

The #966 round-budget flag *declined* a round whose remaining grant could not cover the relaxer&#39;s measured cold-build cost. A declined round banks no bound **and no LP point**, and the point is what the spatial brancher and the primal heuristics run on — which is why the coupled graduation panel scored the flag&#39;s wall win against a collapsing bound ledger (casctanks `2.9098 → −56.5001`, contvar losing its bound outright). This implements the fix that issue comment named: *a round that yields on its deadline must still be able to bank a bound.*

A round that cannot afford a full round now **yields**: the per-node separation chain is skipped and the cold build is truncated at the grant (the #694 anytime mechanism), so it banks the weaker-but-valid bound and the LP vertex it can afford. Second half: a yielded round that banks nothing no longer leaves its node on the failure sentinel — `_yield_keeps_node_open` imports `-inf` so the node stays OPEN at its proved parent bound instead of being fathomed-without-proof.

Contributes to #966 (item 3 advanced, not finished — the flags stay default-OFF; see below).

> **Merged with #984 mid-review.** #984 (*"a round cut short must return the floor it holds"*) landed on `main` while this was open and touches the same admission check. The two compose: #984 makes a round that **is cut short** report its rigorous box floor and exempts the ROOT round; this PR makes a short-granted **non-root** round *run at all*, in yield mode. Merged resolution: the check is `_round_budget_enabled and iteration >= 1`, and when it fires the round yields rather than skipping. Measured on the merged base, #984 does **not** subsume this PR — see the merge comment below and §14e.

## Correctness

Yield mode only **drops** cuts and constraint rows, which enlarges the relaxed feasible set, so the LP optimum stays a valid lower bound on the node&#39;s box — sound, just looser (the same argument that licenses #694&#39;s anytime truncation, which this reuses rather than re-deriving). A yield without a `round_deadline` would be an unclamped round wearing the weaker relaxation, so it raises instead of proceeding.

The sentinel guard strictly *improves* the certificate: leaving the failure sentinel hands the node to the Rust tree&#39;s bound-cut — a fathom with no proof, which `_nonrigorous_sentinel_fathom` turns into a discarded global dual bound. Importing `-inf` keeps the node open, floored on import at its inherited (proved) parent bound; it fathoms nothing, so nothing is decertified.

Both paths are reachable only under `DISCOPT_NODE_ROUND_BUDGET` (default OFF).

- [x] `pytest -m smoke` → **915 passed, 18 failed, 22 skipped, 2 xpassed**. All 18 failures reproduce **identically on the parent commit** (verified by checking out `aca13dd`&#39;s files and re-running the same 7 files: 18 failed / 79 passed both ways) — this container&#39;s POUNCE build lacks the tape NLP evaluator, so the JAX evaluator is used. CI&#39;s Python lanes (proper build) are green.
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **10 passed** (267 s run under panel load had 1 flake in `test_large_dense_jacobian_no_crash`; passes on a re-run of the whole file and twice in isolation on an idle machine)

`pytest python/tests/test_966_node_round_budget.py python/tests/test_928_round_cut_short_floor.py` → **21 passed** on the merged tree (my 15 + #984&#39;s 6 — both mechanisms coexist).

## Regression test

Five new tests in `python/tests/test_966_node_round_budget.py`, all **verified failing on the parent commit `aca13dd` and passing here**:

- `test_yield_round_requires_a_round_deadline`
- `test_yield_round_truncates_build_and_skips_separation`
- `test_yield_round_banks_a_sound_bound_and_a_point` (asserts the yielded bound ≤ the full round&#39;s bound ≤ the true box optimum, and that a point comes back)
- `test_yield_round_default_off_leaves_the_path_untouched`
- `test_yielded_round_without_a_bound_keeps_the_node_open` (pins the certificate consequence: `_nonrigorous_sentinel_fathom(sentinel, False)` is True, `(-inf, False)` is False)

- [x] Default-OFF flag: `DISCOPT_NODE_ROUND_BUDGET` — flag-OFF path verified byte-identical
- [x] Differential-bound test + soundness assertion (see the regression test above; the panel additionally checks bound-crosses-incumbent and verifies incumbents on all cells)
- [x] **Not graduating — default-off only.** Panel evidence attached below; §5 bar 1 fails in 1 of 2 reps.

### Entry experiments (counted, §4/§6)

**FALSIFIED** — the cheap tier is *not* the incremental fast path: `IncrementalMcCormickLP` is out of scope on all five instances the flag hurt (nvs05, tspn10, casctanks, contvar, tls2 → `_inc = None`; `fast_path_hits = 0` in-solve). `scratchpad/issue966_inc_scope.py`

**CONFIRMED** — the #694 anytime build already is the cheap tier (`scratchpad/issue966_truncated_floor.py`, 2 reps/instance):

| instance | full build | truncated build | full round | truncated-round bound |
|---|---|---|---|---|
| contvar | 0.39–1.83 s | **0.008 s** | 19.6–23.9 s | 87754.998 (full: 164927.75) |
| casctanks | 0.06–0.16 s | **0.025 s** | 0.16–0.27 s | 0.0 (full: *no bound at all*) |
| tspn10 | 0.18–0.91 s | **0.068 s** | 2.2–16.0 s | 0.0 (full: 161.16) |
| nvs05 | 0.74 s | **0.002 s** | 0.93 s | 0.674 (full: −323.05) |

**What the skip cost** (nvs05 @ 20 s, every round forced through the admission branch):

| arm | bound | incumbent | nodes | wall |
|---|---|---|---|---|
| base (flag off) | 3.5139 | 8.7320 | 29–39 | 20.9 s |
| every round SKIPPED | 0.6842 | **523.69** | **1** | **4.1 s** (16 s of budget unused) |
| every round YIELDED | 1.3529 | 8.7320 | 45 | 21.0 s |

Shipped-code A/B (`scratchpad/issue966_yield_fix_ab.py`): nvs05 base 3.5139 / 8.7320 / 39 nodes → skip 0.6323 / 6.9358 / 35 → **yield 3.5061 / 8.7320 / 59**. Re-run on the **merged (post-#984) base**: base 3.5425 / 8.7320 / 39 → skip 0.6323 / 6.3870 / 77 → **yield 3.5067 / 8.7320 / 59**, 38 yields counted.

### Flag-OFF bound-neutrality (§5 regime 1)

13 certifying instances, `node_count`/`objective`/`bound`/`status`/`gap_certified` byte-identical, §8 markers asserted in both directions → `BOUND_NEUTRAL=True  COMPARED=13` (`scratchpad/issue966_neutrality.py`). Run twice: against `aca13dd`, and **re-baselined against the merged base `cf16b73`** after the #984 merge.

### Coupled panel re-run

Same script and same 19 binding instances as the earlier panel; 20 s budget, 3 arms interleaved per instance, 2 reps. Artifacts: `discopt_benchmarks/results/issue966_yield_binding20_rep{1,2}.json`.

| pair | rep1 | rep2 |
|---|---|---|
| total overrun, base | 104.9 s | 300.8 s |
| **seam − base overrun** | **−60.6 s** | **−261.7 s** |
| cand − base overrun | −53.1 s | −269.3 s |
| seam bound ledger vs base | 6 tighter / 1 looser | 6 tighter / 1 looser |
| cand bound ledger vs base | 11 tighter / 2 looser | 10 tighter / 2 looser |
| `CERT_CLEAN` (cand vs base) | **True** | **False** (tspn12 lost incumbent) |

Both previously-named casualties are gone: **casctanks is tighter in both flag arms in both reps**, and **contvar keeps a finite bound in `cand`** (180782.86 → 171244.81, sound). The overrun deltas are made of severe modes the base arm still hits — heatexch_gen3 base **210.1 s (10.5×)** vs seam 21.8 s, bchoco08 base 86.2 s (4.31×) vs 21.0 s, tspn12 base 57.8 s (2.89×) vs 20.8 s — each with an equal or tighter flag-arm bound.

**Why it still does not graduate:** §5 bar 1 is per-run and rep2&#39;s graduation pair is not cert-clean (tspn12 loses its incumbent; rep1 lost tspn10&#39;s in the `seam` arm). Lost incumbents are the whole residual now, and they are not reproducible rep-to-rep.

### Three honest limits, recorded not papered over

1. **The panel ran on `aca13dd`, before #984.** Those numbers do not describe the merged tree. #984&#39;s own verdict names the same lost-incumbent residual from the other side (*"#966&#39;s round-admission policy losing a primal trajectory"*), and yield mode is the candidate treatment for it — but that is **unmeasured on the merged base**. The graduation panel owed to #966 item 3 needs one run on the merged tree.
2. It ran **without the POUNCE tape NLP evaluator**, so absolute walls and even base bounds differ from the earlier panel&#39;s machine (casctanks base is −90.1786 here vs 2.9098 there). Same panel, different machine — not a reproduction of those cells.
3. **`minlplib.solu` is unreachable in this container**, so oracle coverage was **1/19 instances (3 comparisons)** — the state the §14b-qual retraction was about. It is declared explicitly (`DISCOPT_MINLPLIB_SOLU=none`, stamped into every artifact as `solu_oracle`) rather than swallowed; a `.solu` path that is *set but unreadable* still crashes. The panel&#39;s other soundness checks fired on all 19 × 3 cells with `unsound=[]`. **No corpus-wide soundness claim rests on these two runs.**

Full record in `docs/dev/performance-plan.md` **§14e** (#984 keeps §14d — it landed first).

https://claude.ai/code/session_011enaWRqp1Fbw3J7wKFNbBT

---
_Generated by [Claude Code](https://claude.ai/code/session_011enaWRqp1Fbw3J7wKFNbBT)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,8 @@ scratchpad/*.log
 scratchpad/**/*_out*.txt
 scratchpad/**/*.log
 scratchpad/**/*.out
+# per-cell worker scripts a probe WRITES at run time (same rule: keep the
+# harness, drop what it generates). The leading underscore is the convention —
+# a hand-written harness like ``issue966_neutral_worker.py`` has none and stays
+# tracked; ``_issue966_yield_fix_worker.py`` is re-emitted on every run.
+scratchpad/_*_worker.py
--- a/discopt_benchmarks/results/issue966_yield_binding20_rep1.json
+++ b/discopt_benchmarks/results/issue966_yield_binding20_rep1.json
@@ -1,0 +1,1294 @@
+{
+  "summary": {
+    "instances": 19,
+    "budget": 20.0,
+    "arms": {
+      "base": "all OFF (today's default)",
+      "seam": "#966 ON, #928 OFF",
+      "cand": "all three ON"
+    },
+    "unsound": [],
+    "incumbent_verification_failed": [],
+    "oracle_comparisons_executed": 3,
+    "instances_without_oracle": [
+      "4stufen",
+      "bchoco06",
+      "bchoco07",
+      "bchoco08",
+      "beuster",
+      "casctanks",
+      "contvar",
+      "hda",
+      "heatexch_gen1",
+      "heatexch_gen2",
+      "heatexch_gen3",
+      "nvs05",
+      "syn05hfsg",
+      "tls2",
+      "tspn05",
+      "tspn08",
+      "tspn10",
+      "tspn12"
+    ],
+    "pairs": [
+      {
+        "pair": "cand vs base",
+        "cells_over_budget": {
+          "base": 16,
+          "cand": 10
+        },
+        "total_overrun_s": {
+          "base": 104.9,
+          "cand": 51.8
+        },
+        "overrun_delta_s": -53.1,
+        "node_count_total": {
+          "base": 647,
+          "cand": 981
+        },
+        "cert_gains": [],
+        "cert_regressions": [],
+        "gained_incumbents": [],
+        "lost_incumbents": [],
+        "better_objective": [],
+        "worse_objective": [
+          "tls2 (5.299999999987421 -> 6.299999999986951)"
+        ],
+        "tighter_bound": [
+          "4stufen (20282.050738739534 -> 20781.828426007993)",
+          "bchoco06 (1.0000000000001477 -> 0.9999875077256566)",
+          "bchoco08 (1.0000000093277979 -> 1.0000000000006108)",
+          "beuster (6352.061283279516 -> 17698.956442696926)",
+          "casctanks (-90.17863668805973 -> -90.17863067603525)",
+          "hda (-20747118223745.168 -> -119286.31279576839)",
+          "heatexch_gen2 (555767.7902849488 -> 579990.0125074089)",
+          "nvs05 (3.513911309952577 -> 3.542487394239276)",
+          "tspn05 (178.27199109143874 -> 179.44338081801095)",
+          "tspn10 (161.16079717285405 -> 161.16083653483406)",
+          "tspn12 (183.3259927436284 -> 192.92305137670508)"
+        ],
+        "looser_bound": [
+          "contvar (180782.85932797543 -> 171244.81148737646)",
+          "tls2 (3.1706801036010077 -> 2.4487125876978726)"
+        ],
+        "gained_bound": [],
+        "lost_bound": []
+      },
+      {
+        "pair": "cand vs seam",
+        "cells_over_budget": {
+          "seam": 14,
+          "cand": 10
+        },
+        "total_overrun_s": {
+          "seam": 44.3,
+          "cand": 51.8
+        },
+        "overrun_delta_s": 7.5,
+        "node_count_total": {
+          "seam": 1089,
+          "cand": 981
+        },
+        "cert_gains": [],
+        "cert_regressions": [],
+        "gained_incumbents": [
+          "tspn10"
+        ],
+        "lost_incumbents": [],
+        "better_objective": [],
+        "worse_objective": [
+          "tls2 (5.299999999987421 -> 6.299999999986951)"
+        ],
+        "tighter_bound": [
+          "4stufen (20742.036713116195 -> 20781.828426007993)",
+          "bchoco06 (0.9999880430650611 -> 0.9999875077256566)",
+          "casctanks (-90.17863668805973 -> -90.17863067603525)",
+          "hda (-20747118223745.168 -> -119286.31279576839)",
+          "heatexch_gen2 (578091.8643598395 -> 579990.0125074089)",
+          "nvs05 (3.513911309952577 -> 3.542487394239276)",
+          "tspn05 (178.27199109143874 -> 179.44338081801095)",
+          "tspn10 (161.15822038513005 -> 161.16083653483406)",
+          "tspn12 (192.92291097809897 -> 192.92305137670508)"
+        ],
+        "looser_bound": [
+          "contvar (180782.85932797543 -> 171244.81148737646)",
+          "tls2 (3.1706801036010077 -> 2.4487125876978726)"
+        ],
+        "gained_bound": [],
+        "lost_bound": []
+      },
+      {
+        "pair": "seam vs base",
+        "cells_over_budget": {
+          "base": 16,
+          "seam": 14
+        },
+        "total_overrun_s": {
+          "base": 104.9,
+          "seam": 44.3
+        },
+        "overrun_delta_s": -60.6,
+        "node_count_total": {
+          "base": 647,
+          "seam": 1089
+        },
+        "cert_gains": [],
+        "cert_regressions": [],
+        "gained_incumbents": [],
+        "lost_incumbents": [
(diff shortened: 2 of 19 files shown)


jkitchin:
## Merged `main` (#984) — the conflict was semantic, and the two changes compose

This PR was `mergeable_state: dirty` after #984 (`f7a3b68`, *"#928 a round cut short must return the floor it holds"*) landed. Both changes touch the same admission check in both node loops, so I resolved it by intent rather than by hunk.

**They fix the same class from opposite ends:**

| | #984 (now on main) | this PR |
|---|---|---|
| a round that **is cut short** | reports the rigorous box-interval `_objective_floor` instead of returning nothing | — |
| a round that **never runs** | exempts the ROOT round from the admission check (rule 1) | runs it in **yield mode** (no separation, build truncated at the grant) so it banks a bound *and* an LP point |

**Resolution.** The admission check is now `_round_budget_enabled and iteration >= 1`, and when it fires the round *yields* rather than skipping. So: the root always runs a full round (#984's rule-1 exemption governs both mechanisms), a non-root short-granted round yields, and #984's floor reporting then supplies the bound for that truncated build. `_yield_keeps_node_open` stays as the backstop for the case where even the floor is unavailable (`_objective_bound_valid` false).

**#984 does not subsume this PR — measured on the merged base** (nvs05 @ 20 s, forced-admission regime, `scratchpad/issue966_yield_fix_ab.py`):

| arm | bound | incumbent | nodes |
|---|---|---|---|
| base | 3.5425 | 8.7320 | 39 |
| skip (pre-fix behaviour) | 0.6323 | 6.3870 | 77 |
| **yield (this PR)** | **3.5067** | **8.7320** | 59 |

38 yields counted. The floor rescues a round that *runs* and is cut short; it can do nothing for a round that never runs, which is why the skip arm still loses both the bound and the incumbent with #984 present.

**Re-verified on the merged tree:**
- `pytest python/tests/test_966_node_round_budget.py python/tests/test_928_round_cut_short_floor.py` → **21 passed** (my 15 + #984's 6 — both mechanisms coexist).
- Flag-OFF bound-neutrality **re-baselined against the merged base** (not the old one): 13/13 certifying instances byte-identical, §8 markers asserted in both directions → `BOUND_NEUTRAL=True`.
- `ruff check` / `ruff format --check` clean.

**Two bookkeeping notes.**
1. In `performance-plan.md`, #984's section keeps **§14d** (it landed first); this branch's is renumbered **§14e**. The "Full record in §14d" pointer in the PR body above should read **§14e**.
2. **The panel numbers in the PR body were measured on `aca13dd`, i.e. before #984.** They are not a measurement of the merged tree and I have not re-run the panel here. What that changes: #984's own verdict ends *"what remains is #966's round-admission policy losing a primal trajectory, which belongs on that issue"* — which is the same lost-incumbent item this PR's rep2 failed on, from the other side. Yield mode is the candidate treatment for exactly that residual, but whether it works on the corpus is **unmeasured on the merged base**. The graduation panel owed to #966 item 3 now needs one run on the merged tree, on the benchmark machine (tape backend + `minlplib.solu`). All three flags stay default-OFF until then.

---

PR #999, opened 2026-08-12
fix(correctness): C-42 strip fabricated dual bound on the opaque CustomCall local path

### LINKED ISSUE #998: correctness: opaque CustomCall local path emits a fabricated (and invalid) dual bound — C-33/SC-1 fix not applied to this caller
## Summary

When a model contains a `dm.custom` / `CustomCall` body that is **not** MCBox-reducible, `solve_model` falls back to a single local NLP and sets `gap_certified = False` — but leaves `bound`, `gap`, `root_bound`, and `root_gap` populated with the **local** objective value. On a nonconvex objective that value is routinely **above** the true global minimum, so the reported "dual bound" is not a valid bound at all.

The identical defect was already found and fixed for the *convexity-unknown* local path (C-33/SC-1, a P0 correctness item) 470 lines later in the same function. That fix strips the bound; the `CustomCall` caller was not updated to match.

`gap_certified=False` is correctly set, so a consumer that checks that flag is safe. But `result.bound` is read directly in plenty of places (reports, notebooks, comparison harnesses), and a bound that exceeds the optimum is exactly the failure mode CLAUDE.md §1 calls a hard gate.

## Reproduction

Identical mathematics, expressed two ways. Ackley in 2-D on an asymmetric box; the true global minimum is **0.0** at the origin.

```python
import os
os.environ.setdefault("JAX_PLATFORMS", "cpu")
os.environ.setdefault("JAX_ENABLE_X64", "1")
import math
import jax.numpy as jnp
import discopt.modeling as dm


def ackley(v):
    n = v.shape[0]
    return (
        -20 * jnp.exp(-0.2 * jnp.sqrt(jnp.sum(v**2) / n))
        - jnp.exp(jnp.sum(jnp.cos(2 * math.pi * v)) / n)
        + 20
        + math.e
    )


# (a) opaque dm.custom body -> the CustomCall local-NLP branch
m = dm.Model("ackley_customcall")
x = m.continuous("x", shape=2, lb=-25.768, ub=39.768)
m.minimize(dm.custom(ackley, name="ackley")(x))
r = m.solve(time_limit=60)
print(r.status, r.objective, r.bound, r.gap, r.gap_certified)

# (b) the same function written algebraically -> the convexity-unknown branch
m2 = dm.Model("ackley_algebraic")
y = m2.continuous("y", shape=2, lb=-25.768, ub=39.768)
m2.minimize(
    -20 * dm.exp(-0.2 * dm.sqrt((y[0] ** 2 + y[1] ** 2) / 2))
    - dm.exp((dm.cos(2 * math.pi * y[0]) + dm.cos(2 * math.pi * y[1])) / 2)
    + 20
    + math.e
)
r2 = m2.solve(time_limit=60, skip_convex_check=True)
print(r2.status, r2.objective, r2.bound, r2.gap, r2.gap_certified)
```

### Observed

| field | (a) `dm.custom` | (b) algebraic twin |
|---|---|---|
| `status` | `optimal` | `optimal` |
| `objective` | 15.06351400512647 | 15.06351400512647 |
| **`bound`** | **15.06351400512647** | `None` |
| **`gap`** | **0.0** | `None` |
| **`root_bound`** | **15.06351400512647** | `None` |
| **`root_gap`** | **0.0** | `None` |
| `gap_certified` | `False` | `False` |

True global minimum: **0.0**. So arm (a) reports a lower bound of **15.06 on a problem whose optimum is 0**, together with `gap = 0.0`, i.e. "this is solved". Arm (b), the same mathematics, correctly reports `bound=None`.

## Root cause

`_solve_continuous` (`python/discopt/solver.py:13308`) sets the bound purely from the NLP's own convergence status, which for a local solver means "a KKT point was reached", not "this is the global optimum":

```python
# solver.py:13489
_c_bound = obj_val if status == "optimal" else None
_c_gap = _optimal_relative_gap(obj_val) if status == "optimal" and obj_val is not None else None
```

That is fine for the genuinely-convex fast path, and `_solve_continuous` even has a KKT-certificate downgrade for it — but it is gated on `certify_convex`, whose docstring comment explicitly says it is off for "the convexity-UNKNOWN caller (which already strips the bound)".

Two callers reach this on a model that is not known convex:

**Convexity-unknown caller — correct** (`solver.py:7757-7762`), the C-33/SC-1 fix:

```python
if _cont_result.status != "infeasible":
    _cont_result.gap_certified = False
    _cont_result.bound = None
    _cont_result.root_bound = None
    _cont_result.gap = None
    _cont_result.root_gap = None
```

Its comment states the reasoning exactly, and it applies verbatim to the opaque case:

> On a nonconvex model the single NLP finds only a LOCAL optimum, which is not global — emitting it with `gap_certified=True` (and bound set to the local objective) is a FALSE optimality certificate.

**`CustomCall` caller — incomplete** (`solver.py:7252-7261`):

```python
result = _solve_continuous(
    model,
    time_limit,
    ipopt_options,
    t_start,
    nlp_solver,
    initial_point=initial_point,
)
result.gap_certified = False
return result
```

Only the flag is cleared. The fabricated `bound` / `gap` / `root_bound` / `root_gap` survive.

An opaque `CustomCall` is a *strictly weaker* epistemic position than "convexity unknown" — the body cannot be inspected at all, so convexity can never be established — which makes it the one caller that most needs the strip.

## Proposed fix

Apply the same treatment at `solver.py:7252-7261`, mirroring the accepted C-33/SC-1 code rather than inventing a second convention:

```python
result = _solve_continuous(
    model,
    time_limit,
    ipopt_options,
    t_start,
    nlp_solver,
    initial_point=initial_point,
)
# C-33/SC-1 applies here a fortiori: an opaque dm.custom body cannot be
# certified convex even in principle, so the single NLP's objective is a
# LOCAL optimum only. Keep the feasible incumbent; strip the fabricated
# dual bound and gap. Genuine infeasibility is a rigorous claim, not a gap.
if result.status != "infeasible":
    result.gap_certified = False
    result.bound = None
    result.root_bound = None
    result.gap = None
    result.root_gap = None
return result
```

This only ever *removes* a claim, so it cannot introduce a false optimum or loosen a valid bound — the same soundness argument the C-33 site already carries.

### Secondary question (worth deciding, but separable)

Both paths still report `status="optimal"` for what is only a local solution. That is a deliberate existing convention — `status` reflects NLP convergence, and C-33 chose to keep it while stripping the bound — so this issue does **not** propose changing it. Flagging it because `status="optimal"` together with `gap=0.0` is what makes the current output read as a solved problem; once `gap` is `None` the reading is much less misleading. If the convention should change, that belongs in its own issue covering both callers.

## Suggested regression test

`python/tests/test_customcall_local_bound.py`, marked `correctness`, asserting on the opaque-body model above:

- `result.bound is None` and `result.gap is None`
- `result.root_bound is None` and `result.root_gap is None`
- `result.gap_certified is False`
- and the property that actually matters: **if** a bound is ever reported on this path, it must satisfy `bound <= known_optimum + tol`. Ackley is a good probe because the default start reliably stalls at ~15.06 against a true optimum of 0, so a regression is loud rather than marginal.

The algebraic twin belongs in the same test as a control, pinning that the two paths agree.

## How this was found

While running the entry experiment for a proposed derivative-free `solver="direct"` backend (`scripts/direct_entry_experiment.py`, record in `docs/dev/direct-entry-2026-08-12.md`), which builds a panel of multimodal test functions wrapped in `dm.custom`. The probe's soundness spot-check watches `gap_certified`, which is correctly `False` here, so it did not fire — the fabricated `bound` was noticed when reading the raw result fields. Worth noting for the reproduction above: the discrepancy is only visible if you inspect `bound`/`gap` rather than trusting `gap_certified` alone.

Environment: `discopt` 0.7.1.dev0 @ 7d93859, Python 3.11.15, numpy 2.4.6, jax 0.10.2.

jkitchin:
## Summary

A `dm.custom` / `CustomCall` body that is **not** MCBox-reducible has no valid node
relaxation, so `solve_model` falls back to a single local NLP. That caller cleared
`gap_certified` but left `bound` / `gap` / `root_bound` / `root_gap` populated with the
**local** NLP objective — which `_solve_continuous` sets from the NLP's own convergence
status ("a KKT point was reached", not "this is the global optimum"). On a nonconvex
opaque body that value is routinely *above* the true global minimum, i.e. not a valid
dual bound at all.

This is the C-33/SC-1 defect already fixed for the convexity-unknown local path 470
lines later in the same function; the `CustomCall` caller was never updated to match.
An opaque body is a *strictly weaker* epistemic position than "convexity unknown" — it
cannot be inspected at all, so convexity can never be established — which makes it the
one caller that most needs the strip.

Reproduced on 2-D Ackley over `[-25.768, 39.768]`, whose true global minimum is `0.0`
at the origin:

| field | behind `dm.custom` (before) | algebraic twin | behind `dm.custom` (after) |
|---|---|---|---|
| `objective` | 15.06351400512647 | 15.06351400512647 | 15.06351400512647 |
| `bound` | **15.06351400512647** | `None` | `None` |
| `gap` | **0.0** | `None` | `None` |
| `root_bound` / `root_gap` | **15.0635… / 0.0** | `None` / `None` | `None` / `None` |
| `gap_certified` | `False` | `False` | `False` |
| `status` | `optimal` | `optimal` | **`feasible`** |

The fix routes **both** uncertified-local-NLP callers through one shared helper,
`_withhold_local_optimality_certificate`, instead of each carrying its own copy of the
treatment — the duplication is precisely how this happened (C-33 fixed one site; the
other kept emitting the fabricated bound 470 lines away). On any non-`infeasible`
result it keeps the feasible incumbent (`objective`, `x`) and sets
`gap_certified=False`, `bound=None`, `root_bound=None`, `gap=None`, `root_gap=None`.

### Status downgrade — the issue's "secondary question", decided in favour of changing it

`_solve_continuous` reports `status="optimal"` on the same basis it reported the bound:
the NLP converged, i.e. a KKT point was reached. On a model never proven convex that
verdict is the same unearned claim — a local optimum need not be global. Both callers
now report the honest **`status="feasible"`**: a feasible point was found, global
optimality was not proved. That is already this codebase's convention for an
uncertified incumbent (the spatial path returns `"feasible"` when its gap does not
close). `"unknown"` is reported when the incumbent itself was withheld — the #815
feasibility check rejected the NLP's point — since `"feasible"` would then assert a
feasible point we do not have.

Per #998's requirement that any change to this convention cover both callers, this is
applied to the opaque-`CustomCall` path *and* the pure-continuous convexity-unknown
path (`skip_convex_check`, or the classifier abstained) together.

**Behaviour change for callers:** code gating on `status == "optimal"` from these two
paths should now also accept `"feasible"` — or better, read `gap_certified`. The convex
fast path, MCBox-relaxable `CustomCall` models, rigorous `status="infeasible"`, and
every certified path are unchanged. Two existing tests in `test_custom_udf.py`
(`test_vector_custom_solves`, `test_custom_mixed_with_primitives`) asserted
`status == "optimal"` alongside `gap_certified is False` and now assert `"feasible"`
with no bound/gap; the user-facing docstrings for `dm.custom`, `CustomCall`, and the
implicit-node module state the reported status.

Also documents the defect as **C-42** in `docs/dev/correctness-issues.md` (table row +
full section) and adds a CHANGELOG entry.

Closes #998

## Correctness

The change only ever **removes** a claim — it strips a dual bound and a gap, and never
adds, tightens, or relaxes one. So it cannot introduce a false optimal, a false
infeasible, or a looser-but-asserted bound; it can only turn an unsound claim into no
claim. This is the same soundness argument the C-33 site already carries.

Two cases are deliberately left untouched, both verified by controls in the new test
file:

- Rigorous `status="infeasible"` from `_solve_continuous` is a nonlinear-tightening /
  NLP-infeasibility claim, not a gap — same convention as the C-33 caller.
- The MCBox-reducible `CustomCall` path routes to the global reduced-space engine and
  never reaches this branch, so its valid certificate is intact
  (`test_mcbox_traceable_customcall_still_certifies`).

Checked rather than assumed: the convexity classifier runs *after* this branch, so an
opaque `CustomCall` cannot reach the `certify_convex=True` convex fast path — there is
no sibling defect there.

- [x] Cannot produce a false certificate — the change only withholds a claim
- [x] No validation, fallback, or safety guard was weakened to make a check pass (the
      gate was made stricter)

- [x] `pytest -m smoke` → 958 passed, 15 skipped, 8412 deselected, 2 xpassed (388.97s)
      — run before the status downgrade; a full `pytest python/tests/` run over the
      final state is in flight and its result will be posted in a comment
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → 10 passed (154.98s)
- [x] `ruff check` / `ruff format --check` clean; `mypy` clean on the three touched modules
- Also: `test_customcall_reduced.py` + `test_custom_udf.py` +
  `test_c33_nonconvex_fallback_cert.py` → 35 passed

## Regression test

`python/tests/test_customcall_local_bound.py`, 4 tests, all `@pytest.mark.smoke`
(each also pins the `feasible` status):

- `test_customcall_local_path_strips_fabricated_bound` — the opaque Ackley repro:
  incumbent kept, all four bound/gap fields `None`, plus the invariant that a *reported*
  bound must satisfy `bound <= true_optimum + tol` (vacuous now; it is the pin that
  survives if a future change ever routes this path to a genuine bound).
- `test_algebraic_twin_control_agrees` — the C-33 path on the same mathematics.
- `test_customcall_and_algebraic_paths_agree` — the two spellings of one problem must
  report the same kind of answer; this is the assertion that pins the divergence.
- `test_mcbox_traceable_customcall_still_certifies` — control against over-correction.

The issue suggested the `correctness` marker; `smoke` is used instead because `addopts`
deselects `correctness` by default, which would keep these out of ordinary runs. They
are sub-second local NLPs, matching the C-33 test's precedent.

- [x] Fail-before / pass-after **verified** by stashing only the solver change: **2
      failed / 2 passed before** (`AssertionError: customcall: fabricated dual bound
      15.06351400512647`), **4 passed after**.

N/A — no bound, cut, or relaxation is computed differently. The only behavioural deltas
are that an unsound dual-bound claim is no longer reported, and that the status on those
same paths now says `feasible` instead of `optimal`. Both paths were already
`gap_certified=False`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dbj9SYmJEVQY9jv8uSkngK)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -163,6 +163,18 @@ The release procedure that produces these entries is documented in
   `CustomCall` path are unaffected. Regression tests:
   `python/tests/test_customcall_local_bound.py`.
 
+  **Behaviour change (both uncertified local-NLP paths).** `_solve_continuous`
+  reported `status="optimal"` on the same basis it reported the bound — the NLP
+  converged — which on an unproven-convex model is the same unearned claim. Both
+  the opaque-`CustomCall` path and the pure-continuous convexity-unknown path
+  (`skip_convex_check`, or the convexity classifier abstained) now report
+  `status="feasible"`: a feasible point was found, global optimality was not
+  proved. This is already the convention the spatial path uses when its gap does
+  not close. `"unknown"` is reported when the incumbent itself was withheld. Code
+  that gated on `status == "optimal"` from these paths should check for
+  `"feasible"` too — or, better, read `gap_certified`. The convex fast path,
+  MCBox-relaxable `CustomCall` models, and every certified path are unchanged.
+
 - **`solve(time_limit=T)` overran `T` because heuristic sub-NLPs were polled but
   never capped** (`fix(heuristics)`, contributes to #966). Every deadline guard in
   `_relax/primal_heuristics.py` gated whether a sub-NLP *starts*; none bounded how
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,24 @@ The release procedure that produces these entries is documented in
 
 ### Fixed
 
+- **Opaque `dm.custom` models reported a fabricated (invalid) dual bound**
+  (`fix(correctness)`, C-42, closes #998). A `CustomCall` body that is not
+  MCBox-reducible falls back to a single local NLP; that caller cleared
+  `gap_certified` but left the *local* objective in `bound` / `gap` /
+  `root_bound` / `root_gap`. On a nonconvex body the local optimum is routinely
+  above the true global minimum, so the reported "dual bound" was not a bound at
+  all — 2-D Ackley behind `dm.custom` on `[-25.768, 39.768]` returned
+  `bound=15.06, gap=0.0` against a true global minimum of `0.0`, while the
+  algebraic twin of the same mathematics correctly returned `bound=None`. This is
+  the C-33/SC-1 defect, whose strip was never applied to this caller; an opaque
+  body cannot be inspected at all, so convexity can never be established for it
+  and the strip applies a fortiori. The fix keeps the feasible incumbent and
+  strips the bound/gap on any non-`infeasible` result — it only ever removes a
+  claim, so it can neither introduce a false optimum nor loosen a valid bound.
+  Rigorous `status="infeasible"` and the MCBox-reducible (global reduced-space)
+  `CustomCall` path are unaffected. Regression tests:
+  `python/tests/test_customcall_local_bound.py`.
+
 - **`solve(time_limit=T)` overran `T` because heuristic sub-NLPs were polled but
   never capped** (`fix(heuristics)`, contributes to #966). Every deadline guard in
   `_relax/primal_heuristics.py` gated whether a sub-NLP *starts*; none bounded how
--- a/docs/dev/correctness-issues.md
+++ b/docs/dev/correctness-issues.md
@@ -126,6 +126,7 @@ in non-default configs; loud ingestion gaps. **P3** = hygiene.
 | C-35 | P1 | oa.py / gdpopt_loa | non-rigorous NLP failure → unconditional no-good cut → possible false infeasible/optimal (= OA-1, opt-in OA/LOA path) | fixed |
 | C-36 | P3 | convexity/interval.py | Python `interval_mul` yields NaN on 0·∞ (`RuntimeWarning: invalid value encountered in multiply`), same 0·∞ class as C-22 but a SEPARATE Python code path (`python/discopt/_relax/convexity/interval.py:171`); lost tightening, not unsound | fixed |
 | C-37 | P2 | solver.py callback API | `CallbackContext.best_bound` surfaced the raw tree `global_lower_bound` with no sense-negation and no taint gate → **over-reported** the certified global bound on a non-rigorously-fathomed tree (nvs05: callback 5.32 vs the rigorous `SolveResult.bound` 1.35) and reported a wrong-signed bound for MAXIMIZE; API-surface soundness (the reported `SolveResult.bound` was already correct) | fixed (A1) |
+| C-42 | P1 | solver.py CustomCall local path | the non-MCBox-reducible `dm.custom`/`CustomCall` local-NLP fallback cleared `gap_certified` only, leaving the local NLP objective in `bound`/`gap`/`root_bound`/`root_gap` → an **invalid dual bound above the optimum** on the DEFAULT path (2-D Ackley behind `dm.custom`: `bound=15.06, gap=0.0` against a true global minimum of 0). Same defect as C-33/SC-1, whose strip was never applied to this caller; the algebraic twin of the same model correctly reports `bound=None` (= issue #998) | fixed |
 
 ---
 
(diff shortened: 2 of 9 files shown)


jkitchin:
## Full-suite result over the final state, and the CI failure it predicted

As promised in the PR body, here are the numbers from a full `pytest python/tests/` run over the status-downgrade state, plus the attribution of every failure.

**CI on `0ef7f05` (`Python fast`): 1 failed, 7860 passed, 110 skipped, 5 xfailed, 2 xpassed** — the single failure being exactly the fallout the downgrade should produce:

```
FAILED python/tests/test_convex_fast_path.py::TestConvexFastPathOptOut::test_skip_convex_check
  AssertionError: assert 'feasible' == 'optimal'
```

That test solves with `skip_convex_check=True` — i.e. the convexity-unknown local-NLP path itself. Its actual subject is `convex_fast_path is False`; the status assertion was incidental. Fixed in `b092b67`, and strengthened rather than merely relaxed: it now pins the whole withheld certificate (`status`, `gap_certified`, `bound`, `gap`) instead of just the status string. `Rust`, `Lint + typecheck`, `Python AMP coverage`, and both codecov checks were already green on that commit.

**Local run (`-n 4`, this container): 7 failed, 7837 passed, 136 skipped, 5 xfailed, 2 xpassed in 748s.** Attribution of all 7, each re-run serially:

| Failure | Verdict |
|---|---|
| `test_convex_fast_path::test_skip_convex_check` | **Real** — the CI failure above; fixed in `b092b67` |
| `test_custom_udf::test_vector_custom_solves`, `::test_custom_mixed_with_primitives` | Stale collection — pytest collected these before I updated them mid-run; they pass on re-run and passed on CI |
| `test_863_sparse_qp_extraction::test_cost_does_not_scale_quadratically_with_unrelated_variables` | Load artifact — a timing test under `-n 4`; passes serially and on CI |
| `test_convex_nlp_certificate_853::…[ipopt]`, `test_relax_compiler_convexity_units::…test_ipopt_backend_active_constraint_gradient` | Environmental — both fail at `nlp_ipopt.py:237` with `ImportError: pip install cyipopt`; cyipopt is not installed in this container. Green on CI, which has it |
| `test_75_tape_nlp_evaluator::test_solve_degrades_to_jax_when_pounce_is_missing` | Environmental — its subprocess blocks the `pounce` import to force the slow JAX path and exceeded the 600 s subprocess timeout here. Green on CI |

So: one genuine consequence of the convention change across the whole suite, and it is now fixed. The re-run on `b092b67` is the check that matters; I'll merge when it is green.

---

Issue #993, opened 2026-08-12
GDP primal: syngas configuration reachability — the dive fixes a 15-demotion model and fails a 3-demotion one


jkitchin:
Split out of #823 (the GDPlib benchmarking tracker), which is now closed: the
integration it asked for exists — `discopt_benchmarks/benchmarks/gdplib_runner.py`,
the BARON+SCIP-via-GAMS oracle, `reference_optima()`, the soundness flags, and
`docs/dev/gdplib-benchmarking.md`. What is left is not "benchmarking plumbing", it is
**two specific primal failures on nonlinear GDPs**, which is what this issue tracks.

## State after #992

#992 added `one_hot_config_subnlp`, a constructor that selects one disjunct per
`sum_k y_k == 1` row instead of rounding indicator binaries independently (which
produces zero-or-two active disjuncts, i.e. an assignment the model forbids). It is
wired at all three paths a GDP can take — the Python B&B root loop, the native spatial
kernel's `_native_kernel_seed_candidates`, and `_solve_nlp_bb` — behind
`DISCOPT_GDP_CONFIG_PRIMAL`, **default OFF**, and bounded to 15 % of the remaining
budget by `_gdp_config_deadline`.

Measured (big-M, 60 s, `--max-variables 500`, oracle on, both arms one run,
`INCORRECT=0` / 0 bound crossings in both):

| model | flag OFF | flag ON |
|---|---|---|
| cstr | 3.370033, 11615 nodes | **3.0620146**, 9970 nodes (reference optimum 3.0620073) |
| batch_processing | no incumbent, 307 nodes | no incumbent, 243 nodes |
| syngas | no incumbent, 31 nodes | no incumbent, 31 nodes |

cstr is solved. The other two are what remains, and they fail for **two different
reasons** — neither of which is the one-hot structure #992 fixed.

## Item 1 — batch_processing: the in-solve start point, not the search

The constructor's own DEBUG line in-solve reads:

```
one_hot_config_subnlp: 29 group(s), 0 residual binaries, 256/256 sub-NLP(s) attempted (exhausted), 0 feasible
```

Offline, the same model through `from_nl` yields **6 feasible plans out of 58**
attempted, best objective 864019.71 against the reference optimum 679365.33. Same
constructor, same model name, opposite outcome — so the failure is in what it is
handed, not in what it does. Two candidate causes, **neither tested**:

1. the in-solve model has been through `reformulate_gdp(model, method="big-m")`, so it
[...]

Issue #823, opened 2026-07-21
GDPlib benchmarks


jkitchin:
Figure out how to use https://github.com/SECQUOIA/gdplib to benchmark / test discopt.

Suggested in issue #819

jkitchin:
## How to benchmark discopt against GDPlib — figured out + implemented

Pushed a working integration to `claude/discopt-issue-823-w7ivaj` (commit 70b3844).

### The bridge
GDPlib ships **Pyomo GDP** models (`Disjunct`/`Disjunction`/`LogicalConstraint`), not a format discopt reads. But discopt already has a Pyomo solver plugin (`discopt.pyomo`) that round-trips a model through a temp `.nl`. A GDP model isn't `.nl`-writable, so we lower the disjunctions first with Pyomo's built-in transforms:

```python
import pyomo.environ as pyo
from pyomo.core import TransformationFactory
import discopt.pyomo                       # registers SolverFactory('discopt')
from gdplib.jobshop import build_model

m = build_model()
TransformationFactory('gdp.bigm').apply_to(m)      # or 'gdp.hull'
res = pyo.SolverFactory('discopt').solve(m, options={'time_limit': 300})
```

Both `gdp.bigm` and `gdp.hull` round-trip cleanly. **Verified end-to-end:** `jobshop` solves to the certified optimum **11.0** via both reformulations, matching HiGHS exactly.

### What landed
- **`discopt_benchmarks/benchmarks/gdplib_runner.py`** — discovers every GDPlib builder, reformulates + solves each via the bridge, and emits `metrics.SolveResult`/`InstanceInfo` (same pipeline as MINLPLib/CUTEst). CLI: `python -m benchmarks.gdplib_runner --list | --models jobshop --methods bigm hull | --max-variables 500 --output ...`.
- **Correctness gate** (per `CLAUDE.md` §1): cross-checks discopt against an oracle — **HiGHS** for the linear subset (exact, independent), a verified reference table otherwise — and flags *impossible incumbents* (a feasible objective beating the true optimum → false primal, cf. #815), false optima, and dual-bound crossings. The CLI exits nonzero on any violation.
- **16 tests** (`tests/test_gdplib.py`, skip cleanly without pyomo/gdplib), the `[gdplib]` extra, and **`docs/dev/gdplib-benchmarking.md`** (full recipe + model inventory + strategy).

### Gotchas worth knowing
[...]

PR #1000, opened 2026-08-12
fix(primal): #993 reach distant disjunct configurations by re-solving between choices

jkitchin:
## What this changes

`one_hot_config_subnlp` (#823) fixed **validity** — it selects one disjunct per
`sum_k y_k == 1` row, so the fixing it hands the sub-NLP never contradicts a
disjunction outright. It did not fix **reachability**. Both of its searches rank
disjuncts from a single static point (the relaxation's per-group argmax) and
enumerate outward in Hamming waves without ever re-solving, so a configuration
many demotions away is unreachable at any budget:

| model | groups | demotions from the argmax | plan index in the wave |
|---|---|---|---|
| syngas | 26 | 3 | 743 (`max_configs` = 256) |
| batch_processing | 29 | 15 | C(29,15) ≈ 7.7e7 |

This adds `one_hot_config_dive`, which **re-solves the relaxation between
choices**, and reserves half the constructor's clock for it.

## Entry experiment: both premises in the issue are false

#993 was opened on two hypotheses — a bad start point on batch_processing and a
continuous-subproblem failure on syngas. Both are wrong, and the falsification is
recorded on the issue (comment 5268543715).

With the integers pinned to BARON's proven configuration the fixed-integer
sub-NLP is *trivial*, **from the constructor's own zero start**:

| model | time | objective | reference |
|---|---|---|---|
| batch_processing | 0.07 s | 679365.33 | 679365.33 |
| syngas | 0.36 s | 4669.0235 | 4669.02 |

On syngas 12 of 67 starts are feasible, including the constructor's own `zero`.
On batch_processing the start point, the presolved bounds, the seed ranking and
the environment were each eliminated by a separate arm (all four cells of a 2×2,
plus five starts: 0 feasible in 1329 attempts). The sub-NLP was never the
problem — **getting to the configuration** was. Root relaxations are very weak
(546849.8 vs 679365.33; 2.0594579 vs 4669.02), so argmax rounding carries almost
no information, and a larger `max_configs` is the same wrong mechanism with a
bigger constant.

## Mechanism ladder (all on the two real models)

Each rung was added because a measurement demanded it. Logs in
`scratchpad/issue993/dive*.log`.

| mechanism | batch_processing | syngas |
|---|---|---|
| wave enumeration (production today) | 0 feasible | 0 feasible |
| descent on the failed sub-NLP's residual | 0 (140 → 47 violated rows, then stalls) | 0 (258 → 136) |
| DFS dive | 0 (17/29 groups at 60 s) | 0 (21/26) |
| dive + learned order | **1063445 at 12.1 s** | 0 |
| + GRASP randomised choice | **943539 at 5.3 s, 41 configs** | 0 |
| + local repair | **822534 at 5.9 s** | 0 (25/26, 8 repairs) |

Two findings steered this. First, partial one-hot fixings genuinely do make the
relaxation infeasible and the refusal is trustworthy: **0 of 412 refusals flipped
to feasible** when retried from an independent start. Second, randomising the
*score* cannot diversify syngas — its size-2 groups sit at margins near 1.0, so
additive noise never flips one, and 65 restarts re-walked one identical path — so
the disjunct *order* is randomised instead.

## The wave was spending the dive's budget

Ordering the wave first is right (0.018 s/plan on cstr when it works), but
unreserved it consumes the **whole** budget on exactly the models the dive exists
for. At a 9 s constructor budget: batch_processing got 66 of 256 plans, syngas 53,
all infeasible, and the dive got **0 relaxation solves**. Half the budget is now
reserved. That costs the wave nothing where the wave works — cstr exhausts all 256
plans in 4.6 s of the 15 s a 300 s run grants — and where it does not work, every
second it holds is waste. Plans are ordered most-informed-first, so a truncated
wave loses late alternates, never its best-ranked candidate.

## Measured through the shipped path

`scratchpad/issue993/prod_dive_split.log`, at the 15 s budget a 300 s time limit
grants. §8 load gate: `discopt.__file__` asserted in-tree and the version marker
(`one_hot_config_dive`, absent on `main`) asserted present.

| model | feasible | best | vs reference | note |
|---|---|---|---|---|
| cstr | 4 | 3.0620146 | +0.0% | **unchanged**; the dive never runs, so it is not charged for it |
| batch_processing | 1 | 822533.66 | +21.1% | the wave alone finds nothing at any budget |
| syngas | 0 | — | — | the measured negative |

Every returned point is independently re-verified in the probe (constraint- and
integer-feasible, reported objective matches the point, and **no objective beats
its reference optimum**), and the model's bound arrays are asserted byte-identical
before and after each call.

At a 9 s budget batch_processing still finds nothing, so this needs a time limit of
roughly 100 s or more, where the 15 s absolute cap binds rather than the 15 %
fraction. Stated rather than tuned around: raising the constructor's share is a
separate decision with its own panel.

## Soundness

* Every returned point comes from `subnlp`, which re-verifies integer- **and**
  constraint-feasibility. This proposes incumbents only and never touches the dual
  bound, so it cannot weaken a certificate.
* An "infeasible" verdict on a nonconvex partial fixing redirects **this
  heuristic's own search** and is never a pruning decision, so a wrong verdict
  costs candidates, never correctness.
* Every bound mutation is restored in a `finally` (asserted by a test and by the
  probe).
* Cost is bounded three ways: the absolute deadline (polled before every solve,
  and each solve capped so one cannot overrun it), `max_restarts`, and
  `max_level_solves`.

## Verification

* **New regression test fails before, passes after.** Verified by stashing the
  module and re-running the same synthetic model on `main`'s code: 0 feasible in
  14.6 s (the wave exhausts `max_configs` inside the first three demotion levels).
  After: 6 passed in 22.5 s. The test model is a *synthetic capacity GDP* whose
  structure creates the gap — no named instance, no name-keyed behavior
  (CLAUDE.md §2).
* `pytest python/tests/test_issue823_gdp_config_primal.py` — 9 passed, unchanged.
* `pytest -m smoke` and `pytest -m slow python/tests/test_adversarial_recent_fixes.py`
  — see the comment below for results.
* `ruff check python/`, `ruff format --check`, `mypy` — clean. No Rust touched.

## The wave/dive handover is a solve count, not a clock slice (#912)

CI on the first push failed `test_912_wall_budget_inventory::test_no_unrecorded_wall_clock_work_gate`
on the line this PR added:

```python
wave_deadline = _now() + 0.5 * max(0.0, deadline - _now())
```

That is a real defect, not a test to silence. #912's rule is that a clock may
decide **when to stop**, never **how much work** to do, and this line did the
second: it made the number of configurations the wave tries — and therefore the
incumbent, and therefore every node below it — a function of machine speed.
Driving the `_now()` seam with two clocks that differ by an unbounded factor gives
**28 plans against 256** for the same model on the same input.

**Registering it in `KNOWN` was not honestly available.** The `residual` category
is the one this line belongs to, and `test_the_converted_layer_stays_converted`
asserts zero `residual` entries for `_relax/primal_heuristics.py` — the category
is closed to this module by policy, not by an omission an append would fix. So the
budget is denominated in sub-NLP solves instead, the unit `_work_budget.NLP_SOLVE`
already names, and the offending line is gone rather than recorded.

`_WAVE_SOLVE_CAP = 48` is calibrated, not picked. Entry experiment E1/E2 ran the
wave with `deadline=None` so only `max_configs` binds, recording every plan's
outcome *and objective*:

| model | solves/s | feasible at plan | best objective at plan |
|---|---|---|---|
| cstr | 54.3 | 8, 42, 47, 57 | **42** (3.0620146, the optimum) |
| batch_processing | 7.3 | none of 256 | — |
| syngas | 5.1 | none of 256 | — |

The yield is concentrated. The plans that matter for *quality* end at 42 — cstr's
plan-8 hit is +2.36% off — so the cap has to sit above 42 (a cap of 32 would trade
cstr's true optimum for a worse point) and as far below 256 as that allows. 48
keeps every quality-relevant plan with six plans of headroom and costs the wave
6.6 s of a 15 s grant on batch_processing instead of the 35 s an uncapped wave
wants, which is what leaves the dive a share at all.

Measured through the production `one_hot_config_subnlp` → `one_hot_config_dive`
chain at the 15 s grant a ≥100 s time limit yields:

| model | wall split (rejected) | solve-count cap (shipping) |
|---|---|---|
| cstr | 256 solves, best 3.0620146, 4.6 s | 48 solves, best 3.0620146, **0.8 s** |
| batch_processing | wave 54 + dive 184 solves → 822533.66 | wave 48 + dive 203 solves → 822533.66 |
| syngas | wave 47 + dive 84 solves → nothing | wave 48 + dive → nothing |

Same answers everywhere, 5.7x cheaper where the wave works. New test
`test_wave_hands_over_on_a_solve_count_not_a_clock_slice` pins it by driving the
`_now()` seam, and carries two self-checks so a passing run cannot be vacuous
(CLAUDE.md §6): the slow arm must run past half the grant, or a wall split would
not have truncated it, and must stay inside the grant, or it stopped on the
deadline contract instead of the cap. Against the reverted implementation it fails
with `the wave tried 28 configuration(s) on the slow clock and 256 on the fast one`.

**Sequencing note.** Converting the budget changes what the search does, so the §5
graduation panel is being run *after* this commit, against the code that will
actually ship — an ON arm measured under the wall split would have characterized a
mechanism this PR removes.

Still behind `DISCOPT_GDP_CONFIG_PRIMAL`, **default OFF**. The §5 graduation panel
(OFF vs ON over the GDPlib small set, interleaved) is the next PR, which will
either graduate the flag default-ON or remove it and its three call sites.

Contributes to #993

### DIFF
--- a/python/discopt/_relax/primal_heuristics.py
+++ b/python/discopt/_relax/primal_heuristics.py
@@ -1462,6 +1462,38 @@ def _residual_assignments(
     return out[:limit]
 
 
+#: Sub-NLP solves the #823 wave may spend before the #993 dive gets the rest of the
+#: caller's grant. A **count**, not a slice of the clock, and that is the whole point:
+#: a wall-clock split (``_now() + 0.5 * (deadline - _now())``) makes how much of the
+#: wave runs a function of machine speed, which #912 forbids — and
+#: ``test_912_wall_budget_inventory.test_the_converted_layer_stays_converted`` closes
+#: the ``residual`` category to this module outright, so "record it in KNOWN" is not
+#: an honest option here. Bounded by a count, the wave's extent is a function of the
+#: model alone.
+#:
+#: 48 is calibrated, not chosen (``scratchpad/issue993`` E1/E2, 2026-08-12; run the
+#: wave with ``deadline=None`` so only ``max_configs`` binds, and record every plan's
+#: outcome and objective):
+#:
+#: ===============  ==========  =============  ==========================
+#: model            solves/s    feasible at    best objective at plan
+#: ===============  ==========  =============  ==========================
+#: cstr             54.3        8, 42, 47, 57  **42** (3.0620146, optimal)
+#: batch_processing 7.3         none of 256    —
+#: syngas           5.1         none of 256    —
+#: ===============  ==========  =============  ==========================
+#:
+#: So the yield is concentrated: past plan 57 the wave never produced anything on
+#: this corpus, and the plans that matter for *quality* end at 42 — cstr's best is
+#: there, while its plan-8 hit is +2.36 % off. The cap must therefore sit **above
+#: 42** (a cap of 32 would trade cstr's true optimum for a 2.36 %-worse point), and
+#: as far below 256 as that allows. 48 keeps every quality-relevant plan with six
+#: plans of headroom for root-point drift, and costs the wave 6.6 s of a 15 s grant
+#: on batch_processing (48 / 7.3) instead of the 35 s an uncapped wave wants — which
+#: is what leaves the dive a share to work in.
+_WAVE_SOLVE_CAP = 48
+
+
 def one_hot_config_subnlp(
     model: Model,
     x_relax: np.ndarray,
@@ -1514,10 +1546,10 @@ def one_hot_config_subnlp(
     batch_processing 15 of 29 (C(29,15) ~ 7.7e7). The dive re-solves the relaxation
     between choices, which is what makes those reachable. Ordering is deliberate —
     the wave is far cheaper when it works (0.018 s/plan on cstr) so it goes first,
-    and the dive only pays for models the wave cannot solve. **Half the budget is
-    reserved for the dive**: measured, an unreserved wave consumes all of it on
-    precisely the models the dive is for, so the two searches would compete for one
-    clock and the cheaper-but-useless one would always win.
+    and the dive only pays for models the wave cannot solve. The wave hands over
+    after ``_WAVE_SOLVE_CAP`` sub-NLP solves: measured, an uncapped wave consumes
+    the entire grant on precisely the models the dive is for, so the two searches
+    would compete for one budget and the cheaper-but-useless one would always win.
 
     ``max_configs`` is deliberately large. Measured on cstr, a fixed-integer
     sub-NLP here costs **0.018 s** (384 plans in 7 s), and the feasible plans sit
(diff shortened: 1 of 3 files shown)


jkitchin:
## Required suites (this branch, 424c6416)

| suite | result |
|---|---|
| `pytest -m smoke` | **964 passed**, 1 skipped, 2 xpassed, 8605 deselected (101 s) |
| `pytest -m slow python/tests/test_adversarial_recent_fixes.py` | **10 passed** (134 s) |
| `pytest python/tests/test_issue823_gdp_config_primal.py` | **9 passed** (unchanged) |
| `pytest python/tests/test_issue993_gdp_config_dive.py` | **6 passed** |
| `ruff check python/` / `ruff format --check` / `mypy` | clean |
| `cargo test -p discopt-core` | not run — no Rust touched |

Logs: `scratchpad/issue993/smoke.log`, `scratchpad/issue993/adversarial.log`.

The §5 graduation panel for `DISCOPT_GDP_CONFIG_PRIMAL` (OFF vs ON over the
GDPlib small set, interleaved per model with alternating arm order) is running now
and lands as a separate PR, which will either graduate the flag default-ON or
remove it and its three call sites.

PR #1003, opened 2026-08-12
chore(bench): #993 refuse to measure a Python-new / Rust-old hybrid

jkitchin:
## Status: a proposal, not a request

#993 is closed and this is not needed to keep it closed. The team lead flagged
the underlying hazard and was explicit that spending more here is the owner's
call. It is opened as a PR rather than an issue because the work is done and
reviewable, and because filing a description of a fix instead of the fix is the
thing CLAUDE.md §6 warns about. **Merge it or close it — either is a fine
outcome.**

## What it does

CLAUDE.md §8 asks every measurement to assert `module.__file__` and a marker
unique to the version under test. Both are *source-level* checks, and `discopt`
is two artifacts: the Python package, and a compiled `_rust` extension that is a
**build output**. A panel can assert `__file__`, assert two source markers, pass
all three, and still be measuring a different tree's `.so`. Nothing in the
existing discipline can see this.

`discopt_benchmarks/benchmarks/load_gate.py` adds `inspect_extension()`, which
reports three conditions:

| condition | why it matters |
|---|---|
| `_rust` loaded from a different directory than the package | the Python-from-one-tree / Rust-from-another hybrid |
| extension older than the newest `.rs` / `Cargo.toml` | measuring a build that predates the checkout |
| sibling `_rust*.so` files present | which one loads is decided by which interpreter starts |

## This found a real defect in the #993 panels

Not hypothetical, and not the defect I set out to catch. Writing the gate and
pointing it at the real trees showed that the #993 panels themselves ran a
hybrid:

- the loaded `_rust.cpython-312-darwin.so` was built **09:03** and never rebuilt
  (mtime still 09:03 after the last arm);
- `5dc804b9` — #928, *"an expired MILP budget is no longer the wire value for
  'no limit'"* — landed **09:09** and **is an ancestor of the panel base**
  `3fb20772` (`git merge-base --is-ancestor` confirms);
- that commit changes `crates/discopt-python/src/lp_bindings.rs`,
  `python/discopt/solver.py` and `python/discopt/solvers/milp_simplex.py`
  **in the same commit**.

So every arm of both panels ran #928's Python half against a Rust half that
predates it — a configuration corresponding to no commit.

### What this does and does not do to the graduation verdict

Stated plainly because it would be easy to over- or under-claim:

- **The differential survives.** The `.so` was byte-identical across every arm of
  both panels, so OFF vs ON is a true A/B. Nothing here shows a wrong answer, and
  the certified values reproduce.
- **The absolute numbers do not describe any shipping build.** Anyone wanting the
  walls or node counts to correspond to `main` needs a rebuilt re-run.
- **One caveat should not be filed under "controlled":** the single missing Rust
  change concerns budget *expiry*, and the flag-ON path is the one that spends
  budget on extra sub-solves. A shared binary is not by itself proof that both
  arms were affected equally.

This PR therefore also **retracts, per §11**, the sentence in the #993 graduation
record claiming the hybrid was "the intended one". Its premise was true — none of
`9a2f8f32`, `0ef7f056`, `6a626125`, `eb70c2d0`, `3fb20772` touches a `.rs` file —
but the conclusion was wrong, because the relevant comparison is the base, not
those changesets. I published that claim, so I am withdrawing it in writing.

## Design notes

**Source mtimes, not a git hash or a branch-switch time.** A checkout only
rewrites files whose content changed, so switching across commits that touch no
Rust legitimately leaves the build current. A hash comparison would cry stale on
every such switch until people learned to ignore it. An alarm that is usually
wrong is worse than none: it trains its reader to pass the override flag. The
glob is deliberately narrow (`**/*.rs`, `**/Cargo.toml`, `**/build.rs`) — a
touched README in `crates/` must not fire it, and there is a test for that.

**The gate sits in `run_suite()`, not `main()`.** My first cut gated the CLI. That
was wrong twice over: it broke three existing `test_gdplib.py` tests that
monkeypatch `run_suite` (they are unit tests of exit-code logic and measure
nothing), and — the real problem — the #993 panel scripts called `run_suite`
**directly**, so gating only `main()` would have left exactly the callers that
hit this bug unprotected.

**Refusal, not a warning** (§3): a hybrid's numbers look entirely ordinary, so
anything short of stopping gets scrolled past. `StaleExtensionError` becomes
exit **5**, joining the runner's existing loud refusals (1 unsound, 2
unavailable, 3 vacuous, 4 shrunken). `--allow-stale-extension` /
`allow_stale_extension=True` proceeds with a warning, mirroring `--no-oracle`:
say so deliberately rather than measuring a build you did not intend. `--list` is
exempt — it runs no solver.

**An `AUTO` sentinel rather than `None` defaults.** `None` cannot mean both "work
it out from this interpreter" and "there is genuinely none", and conflating them
made the missing-extension arm unreachable except from an interpreter that
actually lacked the extension — untestable exactly where it matters. Found by a
test failing for the right reason.

## Verification

- **8 new tests** reconstructing layouts that actually occurred: fresh, stale,
  missing, cross-tree hybrid, two-sibling, and doc-churn-does-not-fire.
  `test_run_suite_refuses_rather_than_measuring_a_hybrid` asserts
  `discover_models` is never reached, so the refusal is proven to happen *before*
  any measurement rather than inferred from an exit code.
- **Confirmed against the two real trees, not only synthetic ones** (§6 — a probe
  that only ever sees fixtures has not been shown to fire): silent on a matched
  tree; on `wt993c` it names the hybrid, the stale build, and the sibling.
- **End to end**: `--list` exits 0; a real run exits 5 with a specific,
  actionable message; `--allow-stale-extension` proceeds.
- `pytest tests/test_extension_load_gate.py tests/test_gdplib.py -m "not slow"` →
  **41 passed**.
- `ruff check` / `ruff format --check` clean on all three files.
- Pre-existing failures in `test_superposition_regression.py`,
  `test_ellipsoidal_regression.py` and `test_milp_node_efficiency.py` are **not**
  from this change: they reproduce on the pristine base with this work stashed,
  and none of those modules imports `gdplib_runner` or `load_gate`. (Given the
  finding above, a stale local build is a plausible cause worth a separate look.)

## Not included, deliberately

Scoped to the GDP runner, which is where the incident happened. The same gate
would suit the MINLPLib and CUTEst runners; that is a follow-up, not scope creep
into this PR.

Contributes to #993

### DIFF
--- a/discopt_benchmarks/benchmarks/gdplib_runner.py
+++ b/discopt_benchmarks/benchmarks/gdplib_runner.py
@@ -56,7 +56,7 @@ from dataclasses import dataclass, field
 from datetime import datetime
 from typing import TYPE_CHECKING
 
-from benchmarks.load_gate import StaleExtensionError, inspect_extension
+from benchmarks.load_gate import ExtensionMismatchError, inspect_extension
 from benchmarks.metrics import (
     BenchmarkResults,
     InstanceInfo,
@@ -103,7 +103,7 @@ class GDPLibSuiteConfig:
     include: list[str] | None = None  # explicit model-name allowlist
     exclude: list[str] = field(default_factory=list)
     oracle: bool = True  # cross-check the linear subset against HiGHS
-    allow_stale_extension: bool = False  # measure a build older than its sources anyway
+    allow_stale_extension: bool = False  # load a mismatched/ambiguous extension anyway
 
 
 #: The curated small set: every GDPlib model whose ``gdp.bigm`` reformulation fits
@@ -754,7 +754,7 @@ def run_suite(config: GDPLibSuiteConfig) -> tuple[BenchmarkResults, list[ModelRu
             f"\n  loaded _rust:    {report.extension_path}"
         )
         if not config.allow_stale_extension:
-            raise StaleExtensionError(detail)
+            raise ExtensionMismatchError(detail)
         print(f"WARNING (--allow-stale-extension): {detail}", file=sys.stderr)
     specs = discover_models(include=config.include, exclude=config.exclude)
     results = BenchmarkResults(suite=config.name, timestamp=datetime.now().isoformat())
@@ -868,8 +868,8 @@ def main(argv: list[str] | None = None) -> int:
     parser.add_argument(
         "--allow-stale-extension",
         action="store_true",
-        help="run even if the compiled _rust extension predates its sources (say so "
-        "deliberately rather than measuring a build you did not intend)",
+        help="run even if the loaded _rust extension is from another tree or is "
+        "ambiguous (say so deliberately rather than measuring a build you did not intend)",
     )
     parser.add_argument("--list", action="store_true", help="list discovered models and exit")
     parser.add_argument("--output", default=None, help="write BenchmarkResults JSON to this path")
@@ -921,7 +921,7 @@ def main(argv: list[str] | None = None) -> int:
     )
     try:
         results, runs = run_suite(config)
-    except StaleExtensionError as exc:
+    except ExtensionMismatchError as exc:
         print(f"FAIL: {exc}", file=sys.stderr)
         return 5
     if args.output:
--- a/discopt_benchmarks/benchmarks/gdplib_runner.py
+++ b/discopt_benchmarks/benchmarks/gdplib_runner.py
@@ -56,6 +56,7 @@ from dataclasses import dataclass, field
 from datetime import datetime
 from typing import TYPE_CHECKING
 
+from benchmarks.load_gate import StaleExtensionError, inspect_extension
 from benchmarks.metrics import (
     BenchmarkResults,
     InstanceInfo,
@@ -102,6 +103,7 @@ class GDPLibSuiteConfig:
     include: list[str] | None = None  # explicit model-name allowlist
     exclude: list[str] = field(default_factory=list)
     oracle: bool = True  # cross-check the linear subset against HiGHS
+    allow_stale_extension: bool = False  # measure a build older than its sources anyway
 
 
 #: The curated small set: every GDPlib model whose ``gdp.bigm`` reformulation fits
@@ -740,6 +742,20 @@ def run_suite(config: GDPLibSuiteConfig) -> tuple[BenchmarkResults, list[ModelRu
             "pip install 'discopt-benchmarks[gdplib]' and install gdplib from source "
             "(the PyPI wheel omits model data files)."
         )
+    # §8: `discopt` is two artifacts, and the source-level markers a panel asserts
+    # say nothing about the compiled one. The check lives here rather than in the
+    # CLI because the #993 panels called `run_suite` directly — gating only
+    # `main()` would have left the very callers that hit this unprotected.
+    report = inspect_extension()
+    reason = report.reason()
+    if reason is not None:
+        detail = (
+            f"{reason}\n  discopt package: {report.package_path}"
+            f"\n  loaded _rust:    {report.extension_path}"
+        )
+        if not config.allow_stale_extension:
+            raise StaleExtensionError(detail)
+        print(f"WARNING (--allow-stale-extension): {detail}", file=sys.stderr)
     specs = discover_models(include=config.include, exclude=config.exclude)
     results = BenchmarkResults(suite=config.name, timestamp=datetime.now().isoformat())
     runs: list[ModelRun] = []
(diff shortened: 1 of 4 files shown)


jkitchin:
## Correction: the hybrid finding was wrong, and so was the staleness check

@team-lead checked my finding rather than relaying it, and was right on every
point. I re-ran their two commands before acting, and both confirm:

- `pytest python/tests/test_928_expired_budget_not_unlimited.py` against the
  loaded 09:03 extension: **7 passed**. `5dc804b9`'s own message records the file
  failing **6/7** against a build from the parent commit. The fix is in the
  binary.
- `git log 5dc804b9..HEAD -- '*.rs' 'crates/**/Cargo.toml'` is **empty** — no Rust
  content has changed since that build, through `8774ac79`.

**So the #993 panels did not run a hybrid, and the graduation's absolute numbers
stand. No rebuilt re-run is owed.** My retraction is itself retracted, in the
graduation doc and here (§11). The original write-up was right, and for a better
reason than I gave it: not merely that #999/#1000/#1002 skip `.rs`, but that
nothing since the build touches Rust at all.

The load-bearing error was inferring staleness from a timestamp: **"artifact
older than the commit ⇒ artifact lacks that commit" is backwards.** The ordinary
order of work is edit, build, test, *then* commit, so a build predating the commit
it validates is the normal case, not evidence of anything. I should have checked
content or behaviour, which is what the discriminator test does in one command.

## What changed in the PR

The same mistake was load-bearing in the gate itself, and it produced exactly the
failure I claimed in the PR body to have designed around. On `main` today:

```
crates/discopt-python/src/lp_bindings.rs    09:39:11   <- a checkout rewrote it
python/discopt/_rust.cpython-312-darwin.so  09:03:34   <- built from that content
```

Source newer than build, identical content, and the old gate **refused**. My
stated rationale was inverted: mtime is what churns on a checkout; content
hashing is what doesn't. A gate that fires on a correct tree teaches its reader to
reach for `--allow-stale-extension`, and then it protects nothing.

**The mtime predicate is removed** rather than patched. What remains are the two
checks that need no build-time cooperation and cannot false-positive this way:

| check | status |
|---|---|
| `_rust` loaded from a different directory than the package | kept |
| sibling `_rust*.so` where load order depends on which interpreter starts | kept |
| build older than its sources (mtime) | **removed** |

On the main tree the reworked gate is now silent about staleness and flags only
the genuine defect — the 08-09 py313 extension sitting beside the 08-12 py312 one,
which is the hazard that started this.

`StaleExtensionError` → `ExtensionMismatchError`, to match what it now asserts.
The `run_suite()` placement and the fixture tests are unchanged. 38 passed
(`test_extension_load_gate.py` + `test_gdplib.py`), ruff clean.

## One thing that is your call, not mine

I did **not** implement the build stamp. Hashing the Rust sources at build time,
embedding the digest, and comparing it against the tree is the right answer — it
is exact, and immune to both checkout churn and build-before-commit ordering — but
it needs `build.rs` cooperation plus a PyO3-exposed constant, which is a change to
the build system rather than to this runner. Per @team-lead's "if the stamp turns
into a project, say so and park it", I judged it a project and left it out.

That leaves three options and I don't think I should pick:

1. **Merge as-is** — two correct structural checks now, build stamp as a follow-up.
2. **Hold** until the stamp lands, so the gate ships complete.
3. **Close** — the residual value is the sibling/hybrid detection only.

I'd lean (1), because the sibling check catches a live defect on the main tree
today and nothing in it depends on the removed predicate. But the staleness
question is the one that motivated the request, so if shipping without it is the
wrong trade, (2) or (3) is fine and I'll not treat it as a loss.

The C(26,3) syngas enumeration is running in parallel and is unaffected by any of
this.

jkitchin:
Closing unmerged. The hazard this targeted has been removed at the root instead.

Both extensions in `python/discopt/` were deleted and a single current one rebuilt
(`maturin develop --release`, verified: `test_928_expired_budget_not_unlimited.py`
7/7, `pytest -m smoke` 969 passed / 1 skipped / 2 xpassed). There is no longer a
`_rust.cpython-313-darwin.so` eleven weeks behind sitting beside the 3.12 build, so
"which stale artifact loads is decided by the interpreter" is no longer a way to
measure a hybrid.

The check as written also could not merge: it compares `.so` mtime against source
mtime, and on `main` that refused a *correct* tree — `crates/discopt-python/src/lp_bindings.rs`
read 09:39 against a 09:03 `.so` purely because a checkout rewrote the file, with no
`.rs` content change since `5dc804b9`. A gate that fires on a current build teaches
its reader to pass `--allow-stale-extension`, at which point it protects nothing.

What was right here and is worth keeping if this is ever revisited:

* the placement in `run_suite()` rather than `main()` — the panel scripts call
  `run_suite` directly and would have bypassed a CLI-level gate;
* the 8 fixture tests over fresh/stale/missing/hybrid/two-sibling/doc-churn layouts;
* restricting the source globs to `**/*.rs`, `**/Cargo.toml`, `**/build.rs`, so
  documentation churn cannot trip it.

The durable form of the idea is a **build stamp**: hash the Rust sources at build
time, embed the hash in the extension, and compare it against a hash of the tree's
current Rust sources. That answers "was this binary built from this source" exactly,
and is immune both to checkout churn and to the build-before-commit ordering that
made mtime look diagnostic here. Not being built now — the residual risk after the
rebuild is "someone edits Rust and forgets to rebuild", which is low-frequency and
does not justify the machinery today.

Also retracted for the record: the claim that the #993 graduation panels ran a
Python-new/Rust-old hybrid. They did not. The 09:03 build **contains** #928 — the
commit's own discriminator (`test_928_expired_budget_not_unlimited.py`) passes 7/7
against it, and it fails 6/7 against the pre-#928 extension. The build predated the
commit for the ordinary reason that work is built and tested before it is committed.
`git log 5dc804b9..HEAD -- '*.rs' 'crates/**/Cargo.toml'` was empty, so that build
was content-current for the whole history. No re-run was owed and the graduation
numbers stand.

PR #1006, opened 2026-08-12
feat(dfo): derivative-free black-box optimization — solver="direct", solver="surrogate", and a governed root heuristic

jkitchin:
## Summary

Adds derivative-free global optimization to discopt, in three parts. The motivating gap is documented in `docs/global_optimization.md`: a model whose objective contains an opaque `dm.custom` (`CustomCall`) body outside the reduced-space `MCBox` scope has no algebraic relaxation, so the default path degrades to a **single local NLP with no global search at all**, or refuses outright when integer variables are also present.

1. **`solver="direct"`** — DIRECT ({cite} Jones/Perttunen/Stuckman 1993) with the modifications Jones & Martins (JOGO 2021) conclude are generally beneficial: trisect one long side, break ties, an `epsilon` floor, and hybridization with a local solve. Plus Jones-2001 integer centres, DIRECT-GLce constraint handling, an evaluation cache, and `n_jobs` parallel evaluation.
2. **`solver="surrogate"`** — model-based search for the *expensive* regime: RBF interpolant (default, native integer support) or kriging + expected improvement with a nugget. The acquisition subproblem is solved by discopt's own spatial B&B.
3. **DIRECT as a governed root primal heuristic**, `DISCOPT_DIRECT_HEURISTIC`, **default-OFF**.

Started from two papers supplied for review; the entry experiment and its falsification record are in `docs/dev/direct-entry-2026-08-12.md`.

No issue to link — this originated from a paper review rather than a tracked task. Contributes to the black-box gap described in `docs/global_optimization.md`.

## Correctness

Neither backend produces a certificate, and both are stricter than the pre-existing local path about saying so. Enforced at the single `SolveResult` construction site in each module: `bound` and `gap` are `None`, `gap_certified` is `False`, the status is **never** `"optimal"`, and an exhausted budget is a limit status — **never** `"infeasible"`, since a sampling method cannot prove infeasibility. Preconditions raise rather than approximate: a non-finite box (DIRECT is undefined off one; substituting a big-M would be a silent approximation) and a missing objective.

The one place a certificate appears is the *acquisition subproblem* of the surrogate backend, which is an ordinary algebraic model. That certifies **where to sample next, not the answer** — a distinction pinned by a test asserting `gap_certified=True` on the subproblem and `False` on the outer result in the same run.

The primal heuristic falls in CLAUDE.md §5's heuristic-policy regime: it can only cost B&B nodes, never a bound. Every point it proposes goes through the existing feasibility verification; the dual bound is never touched.

- [x] `pytest -m smoke` → **997 passed, 16 skipped, 2 xpassed, 0 failed** (267 s)
- [x] `cargo test -p discopt-core` → N/A, no Rust source touched
- [x] `ruff check` / `ruff format --check` clean (`test_log_square_relaxation.py` and `test_mo_augmecon2.py` are flagged on `main` already and were left alone)
- [x] `jupyter-book build docs/` → **build succeeded, zero warnings**

114 new tests: `test_direct_units.py` (32), `test_direct.py` (21), `test_surrogate_units.py` (29), `test_surrogate.py` (19), `test_direct_heuristic.py` (13).

**Note for the reviewer:** the smoke suite was red at 46 failures partway through this branch. The cause was a **stale compiled `discopt._rust`** — merging `main` brought in the #928 fix to `crates/discopt-python/src/lp_bindings.rs` without a rebuild, so `solve_milp_csc_py` rejected `time_limit_s=None`. Rebuilding cleared all 46, and also cleared an apparent `test_amp_integration.py` hang that had been misdiagnosed as an independent bug. Worth knowing if CI ever shows the same shape.

## Regression test

Several, but the load-bearing ones:

- **Selection vs the definition.** `select_potentially_optimal` is cross-checked against a brute-force sweep over `K` — Eq. (3) itself, not a second implementation — on 25 randomized configurations. This exists because the prototype had a sign-flipped `K_max` and a hull that kept its negative-slope vertex; both are silent (the search still runs, just badly) and both produced a confident, wrong verdict.
- **The whole engine against published figures.** `1+x₁+…+x₅` to 1% accuracy reproduces the survey's Fig. 15 at 16,555 / 503 / 203 against the published 14,492 / 470 / 192. A sign error moves these by orders of magnitude, so the literature is a sharper test than anything we could invent.
- **Parallel evaluation is bit-identical to serial** — verified across 44 function/variant configurations, including against the *pre-refactor* implementation. The first attempt at that refactor silently changed the serial trajectory on 5 of 18 configurations while passing every parallel-vs-serial test; the differential comparison is kept as a regression test rather than discarded after use.
- **Fail-before/pass-after confirmed** for the two generalized guards in `test_cli_plugins.py` (shipped skills may not name a non-existent in-tree path, nor recommend an undeclared extra) by restoring the pre-fix files and re-running.

The two backends are new selectors, unreachable from the default path — a test asserts an algebraic model still returns `gap_certified=True`. The heuristic is the only change touching an existing path:

- [x] Default-OFF flag: `DISCOPT_DIRECT_HEURISTIC`
- [x] Not graduating — default-off only
- Measurement: 16 in-repo `.nl` instances at 10 s, native kernel forced off so the probe is reachable — **12 invocations, 0 violations**: no bound rose, no objective degraded, no `gap_certified` lost.

**Flag-off inertness**, verified independently of the implementing agent: a tripwire on `_direct_root_primal` across 4 models shows `hook_entered=0` and the governor never registers a `direct` source. A 14-instance A/B against the pre-Phase-2 files (marker asserted present/absent per CLAUDE.md §8) gave 11 identical and 3 differing — **all three differing instances are `time_limit` runs**, where node count and bound are not deterministic quantities and merely adding code perturbs timing. On `bchoco07` the bound was byte-identical while node count went 7 vs 3, which fits that reading. The deterministic re-run did not complete in the time available, so this is stated as it stands rather than as proven output equality.

---

## Measurements that contradicted the plan, and are recorded rather than dropped

These are the findings a reviewer should weigh most, because they cut against the feature:

- **Certified EI does not work.** It was the original motivation for the surrogate backend. Built properly — the division lifted away via `EI(x) = max_u [d·Φ(u) + s·φ(u)]`, the dense `rᵀR⁻¹r` whitened to `Σvᵢ²` — B&B finds the true acquisition maximum to five significant figures, but the **dual bound never closes**: 3.82 / 23.1 / 597 / 4871 against optima 0.60 / 0.26 / 0.42 / 1.51 at m = 8/12/20/30 on branin. Certified **CORS** does work, but only with `rbf_kernel="linear"` (cubic's `max|λ|` grows 10 → 164 faster than a McCormick relaxation can bound).
- **Sample efficiency is ~2×, not an order of magnitude**, and `goldstein_price` is an outright **loss** (96 evaluations vs DIRECT's 75), caused by objective dynamic range. DIRECT is a stronger baseline than the surrogate literature suggests.
- **A defaults "fix" I proposed was wrong.** Instrumentation showed 20 s per evaluation going to an acquisition that never certifies, and on branin a 2 s limit was 9× faster and slightly better. Extending to two more functions: six_hump_camel is **49× worse** at 2 s. No default was changed; the cost model is documented instead. The branin-only reading is recorded in the docstring as the single-instance tuning CLAUDE.md §2 forbids.
- **The entry experiment's pre-registered gate failed as written** (8/13 strictly better, 9 required). Recorded as failed with a defect in its own construction — all four "ties" were exact ties at the known optimum — rather than declared a retroactive pass.

Both notebooks state plainly that these are baselines rather than state of the art, and name what beats them (BOBYQA/DFO-LS for local refinement, MADS/NOMAD for constrained black-box, TuRBO/SAASBO for high-dimensional BO).

## Not done

- The surrogate sample-efficiency table comes from the backend's own test suite; my independent re-run was killed twice and I have not reproduced it. It is labelled as such in the notebook.
- Trust-region restriction and batch proposals are designed-for seams, not built.
- The async/human-in-the-loop campaign case is deliberately out of scope here — it belongs with `discopt-doe`.

---
_Generated by [Claude Code](https://claude.ai/code/session_011MjeMARf7dJtAoWu4AXXpj)_


jkitchin:
Reviewed the full diff. This is careful work — the certificate contract is enforced at the single `SolveResult` construction site in each module (`bound=None`, `gap=None`, `gap_certified=False`, status never `"optimal"`, an exhausted budget is a limit and never `"infeasible"`), preconditions raise rather than substitute a big-M box, `expected_improvement` handles `s = 0` exactly instead of dividing, and the recorded falsifications (certified EI never closes; ~2x not 10x sample efficiency; `goldstein_price` an outright loss; the entry gate failing 8/13 against a pre-registered 9) are the kind of negative result that makes the rest believable.

One thing I think has to change before merge, plus a latent bug and some smaller notes.

---

## 1. The `_BACKEND_PASSTHROUGH_KWARGS` widening re-opens the M6 hazard (blocking)

`solver.py` adds 30 names to `_BACKEND_PASSTHROUGH_KWARGS`. That set feeds `solve_model_accepted_kwargs()`, which `Model.solve` uses to reject unknown options — the M6 guard, whose docstring says exactly why it exists:

> `m.solve(gap_tolerence=1e-3)` ran at the DEFAULT gap while the user believed they had tightened it (a results-integrity hazard)

The check is **selector-blind**: it is a flat membership test that never looks at `solver=`. So adding a name makes it accepted on *every* path, including a plain B&B solve where nothing consumes it.

Verified against `main` — all of these raise `TypeError` today and would be silently ignored after this PR:

```
seed                     allowed_on_main=False
n_jobs                   allowed_on_main=False
max_evals                allowed_on_main=False
epsilon                  allowed_on_main=False
surrogate                allowed_on_main=False
xi                       allowed_on_main=False
nugget                   allowed_on_main=False
min_distance             allowed_on_main=False
local_refine             allowed_on_main=False
feasibility_tolerance    allowed_on_main=False
on_evaluation            allowed_on_main=False
distance_cycle           allowed_on_main=False
break_ties               allowed_on_main=False
divide                   allowed_on_main=False
```

The set's pre-existing members are safe because they are *distinctive* — nobody types `convhull_ebd_encoding` or `fp_projzerotol` by accident, and `n_init_partitions` reads as AMP-specific. The new ones are not. `m.solve(seed=42)`, `m.solve(max_evals=1000)`, `m.solve(local_refine=False)` and `m.solve(feasibility_tolerance=1e-8)` are all things a user would plausibly write on a default solve believing they had set something.

`n_jobs` is the sharpest case, because this PR's own docstring names the confusion:

```python
("threads", threads != 1),  # DIRECT parallelism is n_jobs, not threads
```

After this change `m.solve(n_jobs=8)` on a default solve is accepted, ignored, and runs serially — with no warning. That is the M6 failure mode reproduced exactly, on the parameter the PR already knows people mix up.

Suggested fix, in preference order:

1. Make the allowlist selector-aware — `solve_model_accepted_kwargs(solver=None)`, with the DIRECT/surrogate keys admitted only for `solver in ("direct", "surrogate")`. This is the real fix and it also retroactively tightens the AMP/GP keys.
2. Failing that, warn on a backend-specific key supplied without its selector. Cheaper, keeps the signal, does not reject anything that works today.

Either way I would add the fail-before/pass-after test: `m.solve(n_jobs=8)` on an algebraic model must not silently run serially.

## 2. `_cu` is gated on `cl_list`, not `cu_list` (latent, `solver.py`)

```python
_cl = np.asarray(cl_list, dtype=np.float64) if cl_list else None
_cu = np.asarray(cu_list, dtype=np.float64) if cl_list else None
                                                 ^^^^^^^
```

The comment two lines down ("They are set and cleared together") is what makes this harmless today, so it is not a live bug. But it reads as a typo rather than as an invariant being relied on, and if the pairing ever breaks the failure is `np.asarray([])` -> shape `(0,)` -> a broadcast error inside the oracle, i.e. an exception thrown from a heuristic on some future model rather than a clean decline. Either use `cu_list` in the second guard, or assert the pairing explicitly since the code already depends on it.

## 3. `n_jobs > 1` needs a reentrancy caveat

The parallel-evaluation docs are unusually good — input-order collection, apply-in-selection-order, the two subtleties found by differential testing, the pickling reason for threads-not-processes. The one thing not stated is that the **user's black box must be reentrant**. `executor.map(oracle, xs)` calls `dm.custom`'s body concurrently, and the exact case the module is built for — "a subprocess or I/O-bound simulator" — is also the case most likely to use a fixed scratch file or a module-global. The determinism guarantee ("identical, not merely equivalent") is a statement about the *search*, and a reader can easily take it as covering the whole run.

One sentence in the `n_jobs` docstring: the guarantee holds given a reentrant objective, and a simulator with shared scratch state must stay at `n_jobs=1`.

## 4. The 3/14 differing instances are worth closing out

From the PR body:

> 11 identical and 3 differing — all three differing instances are `time_limit` runs [...] The deterministic re-run did not complete in the time available, so this is stated as it stands rather than as proven output equality.

Stating it as it stands is the right call, and I read the flag-off path as inert by inspection: the block is guarded by `_direct_heuristic_enabled()` evaluated first, and the `heuristic_governor` additions cannot fire for a source that never registers. The `bchoco07` evidence (bound byte-identical, nodes 7 vs 3) fits the timing reading.

But "differs only because of timing" is the same sentence that would cover a real perturbation, and it is cheap to remove the ambiguity: re-run those 3 under `max_nodes` instead of `time_limit`, where node count and bound *are* deterministic. If they match, the claim is proven rather than argued. Not blocking, but I would rather it were closed than carried.

## 5. Duplication between the two modules (non-blocking)

`_offending_variable_names`, `_refine_locally` and `_build_oracle` exist in both `direct.py` and `surrogate.py`, and `_SurrogateSearch._offer` / `merits()` restate `_DirectSearch._offer` / `rank_values` — the docstring says so ("Identical in form to `_DirectSearch.rank_values`"). The two `_offending_variable_names` have already drifted apart.

That is ~3200 lines carrying two copies of the DIRECT-GLce merit rule, which is soundness-adjacent: it decides what counts as near-feasible. I would not restructure this PR for it, but a shared `_dfo_common.py` for the merit rule and the oracle builder is worth a follow-up before the third backend arrives.

---

## Verification I ran

- Read the full diff of everything touching an existing path (`solver.py`, `heuristic_governor.py`, `modeling/core.py`) — the dispatch branches are guarded by `_solver ==` and placed before the `CustomCall` gate deliberately, which is right for the selector to be reachable
- Confirmed the certificate contract at both `SolveResult` sites in each module
- Confirmed `_evaluate_points` mutates the cache and `stats.evals` on the **main thread** after `executor.map` returns, so there is no shared-state race in the search itself
- Reproduced finding 1 against `main` via `solve_model_accepted_kwargs()`
- CI at time of review: Rust pass, Lint + typecheck pass, Python suites pending

Deleting the stray `issue862_quality_panel.json` and gitignoring `/*_quality_panel.json` is a good catch and well justified in the ignore-file comment.

Happy to approve once finding 1 is addressed — it is the one that changes behavior on the default path, which is the only part of this PR that can reach a user who never asked for DFO.

jkitchin:
Thanks — finding 1 was right and I had missed it. Addressed in `e59f7b1` and `602b92f`.

## Finding 1 — selector-blind allowlist (blocking)

You were correct that `_BACKEND_PASSTHROUGH_KWARGS` never looks at `solver=`, so the ~30 names I added were accepted on every path. I confirmed the failure on the pre-change tree before fixing it:

```
$ git stash && python -c "m.solve(n_jobs=8)"   # algebraic model, default solve
FAIL-BEFORE CONFIRMED: n_jobs=8 accepted on default solve, status=optimal (ran serially, silently)
```

Took your option 1. The flat set is now split:

- `_GENERIC_PASSTHROUGH_KWARGS` — unchanged, still selector-blind. The pre-existing members survive because they are distinctive; nobody types `convhull_ebd_encoding` by accident.
- `_SELECTOR_ONLY_KWARGS: dict[str, frozenset[str]]` — keyed by the `solver=` value that owns each name, with `"direct"` and `"surrogate"` groups.
- `solve_model_accepted_kwargs(solver=None)` takes the selector and unions in only that backend's group. The no-arg form is unchanged, so the three existing callers (`test_solve_deadline_not_stale`, `test_917_reserve_extension`, `test_modeling_tail_m4_m8`) needed no edit.
- `selector_for_kwarg()` gives a distinct message instead of a difflib near-match guess:

```
>>> m.solve(n_jobs=8)
TypeError: solve() got 'n_jobs', which is an option of solver='direct' and has no
effect on this solve. Pass solver='direct' to use it. Backend options are rejected
rather than silently ignored (...)
```

The original M6 case is untouched — `m.solve(gap_tolerence=1e-3)` still gets `Did you mean 'gap_tolerance'?`.

**Tests, fail-before / pass-after**, in `test_modeling_tail_m4_m8.py` next to the existing M6 block:

- `test_m6b_dfo_only_kwarg_rejected_on_default_solve` — parametrized over `n_jobs`, `max_evals`, `seed`, `local_refine`, `feasibility_tolerance`, `surrogate`, `divide`, `xi`; asserts the raise *and* that the message names the right selector.
- `test_m6b_dfo_kwarg_accepted_with_its_selector` — no false refusal, and the other backend's exclusive options stay refused under it.
- `test_m6b_dfo_options_are_not_in_the_generic_set` — nothing can leak back into the flat set or into `solve_model`'s named parameters.

## Finding 2 — `cl` / `cu` pairing

The mypy pass had already restructured this so both are assigned together (`cl = cu = None` in the else), which removes the `if cl_list` typo you spotted. But you were right that the *pairing* was unasserted, so I added the guard you were actually asking for — a length check against `n_constraints`, raising rather than warning. The reason it matters is the one you gave: the violation sum indexes both against one `g`, so a mismatch broadcasts or truncates into a silently wrong violation, i.e. a wrong feasibility verdict rather than a crash.

Applying it surfaced the duplication you flagged in finding 5 as a live problem, not a hypothetical: `surrogate._build_oracle` is a deliberate mirror and did **not** get the guard. Fixed in `602b92f`, and filed **#1010** to extract `_build_oracle` and the DIRECT-GLce merit rule into `solvers/_dfo_common.py` — with the divergence-that-already-happened as the motivating evidence, plus a definition of done that includes an identical-`(f, v)` cross-backend test and bound-neutral verification.

## Finding 3 — `n_jobs` reentrancy

Added, in three places: the module docstring, the `n_jobs` parameter docs, and the notebook (prose warning + options table). The point I made explicit is the one that makes the caveat non-obvious — the determinism guarantee is a property of the *search*, not of a body that is unsafe to call twice at once, so "identical to `n_jobs=1`" does not protect a simulator with a fixed scratch file.

## Finding 4 — the three differing instances under `max_nodes`

Running now. Node budgets rather than wall clock, three arms (`DISCOPT_DIRECT_HEURISTIC` unset / `=0` / `=1`), asserting that unset and `=0` are bit-identical and that the probe fires zero times on the default path. `bchoco07` is slow enough that this is taking a while; I will post the table when it lands rather than assert the result now. It does not gate the code changes above, which are independent of it.

## Verification

- 173 passed: `test_modeling_tail_m4_m8`, `test_solve_deadline_not_stale`, `test_917_reserve_extension`, `test_cli_plugins`, `test_direct_units`, `test_surrogate_units`
- `ruff check python/` clean, `mypy python/discopt/` clean (0 errors)

Note on `ruff format --check`: two files report reformatting locally (`test_log_square_relaxation.py`, `test_mo_augmecon2.py`), both byte-identical to `main` and neither touched here. It is a version artifact — this container has ruff 0.15.8, pre-commit/CI pin 0.14.6 — so I deliberately did not reformat them, since doing so would break the pinned check.

---

jkitchin:
Re-checked the branch after the three review fixes. Findings 1 and 3 are properly closed; finding 2 was fixed in the wrong function, so I fixed the flagged one and pushed it as `067abc63`.

**Finding 1 — closed, and done the better way.** `_GENERIC_PASSTHROUGH_KWARGS` + `_SELECTOR_ONLY_KWARGS` keyed by the owning `solver=` value, with `selector_for_kwarg()` producing a specific error instead of a difflib guess. That is option 1 from the review, and the distinct message ("which is an option of `solver='direct'` and has no effect on this solve") is better than what I suggested — a near-match hint would have been actively misleading here, since the name is spelled correctly and simply doesn't apply.

**Finding 3 — closed.** The reentrancy caveat is in the module docstring, the `n_jobs` parameter docs, and the notebook.

**Finding 2 — was still open where I reported it.** `e59f7b17` fixed `_build_oracle` in `direct.py`, which is the standalone `solve_direct` path. The line the review flagged is in `solver.py::_direct_root_primal` — the *root heuristic* path — and was unchanged:

```python
_cl = np.asarray(cl_list, dtype=np.float64) if cl_list else None
_cu = np.asarray(cu_list, dtype=np.float64) if cl_list else None
                                                 ^^^^^^^
```

I said in the review this was latent and harmless. **That was wrong, and I'm retracting it** — I had reasoned that a length mismatch would raise a broadcast error. Measured instead:

```
g-hi shape: (0,)   viol contribution: 0.0
=> a point violating sum(x) <= 1 by 4.0 reports violation 0.0
```

`(1,)` against `(0,)` is a *legal* numpy broadcast, so the violation sums to zero and every point reads as feasible. The other arm is worse: `cl_list=[]` with `cu_list=[1.0]` made the `if cl_list` guard clear **both** arrays, dropping the constraints entirely so the probe searched an unconstrained problem. Both are silent.

Neither can corrupt an incumbent — `solve_model` re-verifies every proposal through `_check_constraint_feasibility` before injecting it, so a bogus point is discarded rather than believed. What they corrupt is the probe's own ranking: the entire evaluation budget goes into optimizing the wrong problem, and the proposals are then thrown away. That is the §6/§7 failure shape — an instrument that measures nothing and reads as "the heuristic ran and did not help".

The fix reads the right list and refuses loudly on unpaired bounds, ragged bounds, and an evaluator returning a different row count than there are bounds. That last check matters for the same reason: `(1,)` against `(2,)` is also a legal broadcast, so it needs an explicit test rather than an exception numpy would never throw. The caller already wraps the probe in try/except and logs a warning, so a refusal costs nodes and never correctness.

Tests: `test_mismatched_constraint_bounds_are_refused` (3 arms), `test_constraint_count_mismatch_is_refused`, and `test_matched_constraint_bounds_still_work` as a no-false-refusal control. **Mutation-tested** — reverting `solver.py` fails all 4 guard tests while the control still passes both ways.

Worth noting how nearly this was missed: my first run showed 5/5 failing including the control, because the editable install resolved `discopt` to the **main tree** rather than the worktree (`LOADED FROM: /Users/jkitchin/projects/discopt/python/discopt/solver.py`, `has probe: False`). That is CLAUDE.md §8 exactly. Pinning `PYTHONPATH` gave 5/5 passing. Anyone re-running these in a worktree needs the same pin.

**Finding 5** — thanks for filing #1010; agreed that extracting `_build_oracle` and the merit rule is the right shape. This PR is now the second time the same guard had to be applied by hand in a third location, which is the argument for it.

**Unrelated, and not this PR's problem:** `test_m7_name_check_is_not_quadratic` fails in a full-file run and passes in isolation. It is an unwarmed wall-clock micro-benchmark with no load gate — the §9 anti-pattern — and it fails identically on `main` (3/3 runs), so it predates this branch. Flagging it because it will look like this PR's fault in a full-suite run.

Remaining from the review: only **finding 4** (re-run the 3 differing A/B instances under `max_nodes` instead of `time_limit` to turn the timing explanation into a proof). Non-blocking. With finding 1 closed I have no objection to merging.

jkitchin:
You were right and I was wrong on finding 2 — I fixed `direct.py::_build_oracle` (the standalone path) and never touched the line you actually flagged in `solver.py::_direct_root_primal`. Your `067abc63` is merged into my tree; the 29 tests in `test_direct_heuristic.py` pass here. The `(1,)`-against-`(0,)` broadcast summing to `0.0` is a better failure demonstration than I would have produced, and the count check against the evaluator's row count is the part I would have left out.

## Finding 4 — done, and the answer is stronger than "probably timing"

`max_nodes=40` with a **180 s backstop whose binding is reported per arm**. An arm that hits the backstop is marked inconclusive and *not counted*, because a bound backstop reintroduces exactly the nondeterminism the rerun exists to remove.

Three arms per instance: `DISCOPT_DIRECT_HEURISTIC` unset / `=0` / `=1`.

```
clay0303hfsg
   unset  nodes=63 status=node_limit bnd=-1.2309049062621525e-05 direct_calls=0 wall=15.7s
   =0     nodes=63 status=node_limit bnd=-1.2309049062621525e-05 direct_calls=0 wall=14.9s
   =1     nodes=63 status=node_limit bnd=-1.2309049062621525e-05 direct_calls=0 wall=14.9s

bchoco08                                                       <-- backstop bound, inconclusive
   unset  nodes=7  bnd=1.0000000000136013 direct_calls=0 wall=181.2s
   =0     nodes=7  bnd=1.0000000000136013 direct_calls=0 wall=180.7s
   =1     nodes=7  bnd=1.0000000000136013 direct_calls=1 wall=181.2s

bchoco07                                                       <-- backstop bound, inconclusive
   unset  nodes=3  bnd=1.000000000000253  direct_calls=0 wall=180.0s
   =0     nodes=15 bnd=1.0000000000002476 direct_calls=0 wall=181.0s
   =1     nodes=3  bnd=1.000000000000253  direct_calls=1 wall=180.1s

compared=1 deterministic instances, inconclusive=2
no drift
```

**`clay0303hfsg` is the deterministic result.** All three arms are byte-identical, and the OFF arms land on exactly the value the *original* ON arm produced (`63 / -1.2309049062621525e-05`). The original OFF row was `7 / -1.2309090908977705e-05` — the same monotone bound sequence stopped 56 nodes earlier, not a different bound.

**`bchoco07` demonstrates the artifact directly, which is better than explaining it.** `unset` and `=0` are *the same code path by construction* — the flag is off either way, and the block is one `os.environ` lookup — yet they returned 3 vs 15 nodes and bounds differing in the 13th digit. A difference measured under `time_limit` on this instance therefore cannot distinguish code from machine speed. That is the original A/B difference reproduced with the variable of interest held fixed.

For reference, the original rows (all `status='time_limit'`):

| instance | old without | old with |
|---|---|---|
| `bchoco07` | 3 nodes, 1.000000000000253 | 7 nodes, **1.000000000000253** — identical bound, only node count moved |
| `bchoco08` | 0 nodes, 0.9999999999999977 | 0 nodes, 1.0000000000001388 — never left the root either way |
| `clay0303hfsg` | 7 nodes, −1.2309090908977705e−05 | 63 nodes, −1.2309049062621525e−05 |

**Two things I am not claiming.** `bchoco07`/`bchoco08` are inconclusive as deterministic comparisons: both are dominated by a single node longer than any budget I can run here (`bchoco08` reached 7 nodes in 180 s), so I cannot give you a node-budget proof for them — only the same-code-path demonstration above. And on `clay0303hfsg` the `=1` arm shows `direct_calls=0`, so the ON arm is **vacuous** there — the probe never fired, and that instance says nothing about the flag's effect. It is not reachability: `bchoco07` and `bchoco08` both show `direct_calls=1` with the flag on, and both returned bounds identical to their OFF arms.

Rerunnable: `finding4_maxnodes.py <max_nodes> <time_limit> [instances…]`, with the §8 marker assertion and the §6 "measured nothing" exit built in.

## On `test_m7_name_check_is_not_quadratic`

Agreed it is a §9 anti-pattern and pre-existing — but I could not reproduce it here: **3/3 full-file passes** (37 tests, 0.9–1.9 s) at load average 1.1, with my 3 new M6b tests present and running ahead of it in file order. So on this machine it is load-dependent rather than order-dependent, which is consistent with your diagnosis and means my additions are not what tips it. Worth a separate issue to give it a load gate or delete it.

## Status

Everything from the review is now closed: 1 ✅, 2 ✅ (yours), 3 ✅, 4 ✅ above, 5 → #1010. CI on `602b92f` was green across Lint + typecheck, Python fast, claim-boundary, Rust, AMP coverage, and both codecov, with correctness (fast subset) still running; `067abc63` is building now.

---

Issue #1009, opened 2026-08-13
bug(time-limit): the incremental McCormick fast path drops the caller's time_limit — one node LP ran 416 s against a 20 s budget


jkitchin:
## Summary

`crates/discopt-python/src/spatial_bindings.rs` builds `SimplexOptions::default()` when it calls into the native spatial kernel. The tree's computed deadline is **never handed to the simplex**, so every kernel node LP runs uninterruptibly to completion (or to its iteration cap). The B&B tree only gets to check the clock *between* nodes, so a single slow node LP blows any wall-clock budget by an unbounded margin.

## Evidence

QPLIB_1157 with a **20 s** `time_limit`, `DISCOPT_RLT_LINEQ=1`, cold dual start off:

```
rlt=1 coldstart=0   wall=240.83s   bound=None   nodes=1
```

**11.9x over budget**, on node 1, because the root LP alone ran ~240 s inside the kernel with nothing able to stop it. `time_limit` is silently advisory on the native spatial path.

## Scope

This is **pre-existing and independent** of #1007. It is not a soundness bug — no wrong bound is produced, the solve just does not stop when asked. But it makes `time_limit` unreliable on the default path (`DISCOPT_NATIVE_SPATIAL_KERNEL` defaults to `1`), which undermines any benchmark or gate that uses a time budget as a control.

The #1007 dual start removes the *symptom* for the QP class (240.83 s -> 21.16 s) by making the offending LP fast. It does not fix the mechanism: any LP slow enough will still overrun.

## Why it was not fixed in #1007

The #928 history is the reason for caution. Propagating a deadline into the LP path is exactly what `DISCOPT_LP_WARM_DEADLINE` does, and its measured result was **cert-clean but not net-positive** — paired deltas flipped sign (-3.2 / +6.6 / +5.7 s). Deadline-honoring changes interact with basis banking and warm-start reuse in ways that need their own differential panel, not a drive-by patch inside an unrelated PR.

## Suggested approach

1. Thread the tree's remaining budget into the `SimplexOptions` constructed in `spatial_bindings.rs` (the field already exists; `cold_dual_start` was threaded the same way in #1007).
2. Decide what an LP that hits the deadline returns, and make sure a partial/interrupted LP can never be read as a valid bound. Returning "no bound" is correct; returning the current dual objective is only valid if it is a certified dual-feasible value.
[...]

PR #1011, opened 2026-08-13
test(1009): measure and cover the kernel node-LP deadline

jkitchin:
## Summary

Rewritten. This PR originally duplicated the fix that landed as #1014 — see [the comment above](https://github.com/jkitchin/discopt/pull/1011#issuecomment-5279809237) for the retraction of its misattributed framing. What is left is the part `main` does not have: **the measurement #1014 shipped without, and the coverage it left open**. Rebased onto `main`; no Rust, no solver source, no behavior change.

`git diff --stat origin/main`: 3 files, +442, 0 deletions — one test file, one measurement script, one record.

Follows up #1009 (closed by #1015). Deliberately no closing keyword.

## Correctness

- [x] N/A — no solver-math change. Nothing outside `python/tests/`, `scripts/` and `docs/dev/` is touched.

The tests *assert* a soundness property rather than changing one: an interrupted LP must report no bound and no incumbent, and be recorded undecided (branched, never fathomed), so a partial simplex iterate can never be read as a dual bound. The scaling script asserts a closed-form optimum oracle on every solve it makes.

## The coverage gap, measured rather than asserted

`spatial_tree.rs` unit-tests `node_lp_opts` in isolation and pins the `IterLimit → Undecided` verdict — but nothing tested that the tree's deadline **reaches the LP at all**. Mutating the call site to `node_lp_opts(opts, None)` — a full revert of #1014's effect, leaving the unit-tested helper correct — leaves `cargo test -p discopt-core --lib` at **602 passed, 0 failed**.

The neighbouring `extension_is_taken_only_with_an_incumbent_in_hand` *does* catch a too-**early** deadline (I checked: mutating the call site to freeze `config.deadline` fails it). It cannot catch a **missing** one, because an LP with no deadline solves fine and reports a finite bound.

`python/tests/test_1009_kernel_lp_deadline.py` closes that gap — two of its three tests fail on the `None` mutant and pass on `main`:

| test | on `main` | on the `node_lp_opts(opts, None)` mutant |
| --- | --- | --- |
| `test_a_node_lp_costlier_than_the_budget_does_not_set_the_wall_clock_floor` | PASSED | FAILED |
| `test_an_interrupted_node_lp_contributes_no_bound_and_no_fathom` | PASSED | FAILED |
| `test_a_non_binding_budget_matches_the_uncapped_search` | PASSED | PASSED (neutrality guard — passes in both by design) |

They cut a real LP mid-flight across the PyO3 boundary, where the wall-clock behaviour either materialises or does not; the core's test cuts at pivot 0 with an already-expired deadline, which proves the semantics but not the plumbing. Calibration is against an **uncapped single-node run on the same build** — that time is what the pre-#1014 path spent on node 1 regardless of the budget — so the assertions are ratios, not wall-clock thresholds, and the test **skips rather than passing vacuously** when one LP is already cheaper than the budget on the host.

## The measurement

`scripts/spatial_kernel_lp_deadline_scaling.py` runs both arms against the same fixed code, so it reproduces on `main` with no flag and no revert: *uncapped* = `time_limit_s=None, max_nodes=1` (one LP to completion — the floor the fix removes), *capped* = the same LP under the budget. Driver is a general dense-bilinear model, **not a named instance** (CLAUDE.md §2), with closed-form optimum `C(n/2, 2)` asserted as an oracle on every solve.

Budget fixed at 1 s; medians of 3 interleaved reps, arm order alternating, load gate checked:

| n | lifted terms | uncapped (one node LP) | capped |
| ---: | ---: | ---: | ---: |
| 50 | 1 225 | 0.81 s (sd 0.03) — 0.8× | 1.02 s (sd 0.02) — 1.0× |
| 70 | 2 415 | 2.68 s (sd 0.04) — 2.7× | 1.20 s (sd 0.02) — 1.2× |
| 90 | 4 005 | 7.69 s (sd 0.09) — 7.7× | 1.23 s (sd 0.01) — 1.2× |
| 120 | 7 140 | 25.90 s (sd 0.39) — **25.9×** | 1.05 s (sd 0.01) — 1.0× |

The floor grows without limit; the capped column tracks the budget. At n = 50 the capped run is *longer* — the LP is cheaper than the budget there, so the tree keeps branching until the clock runs out. That is the expected shape: the deadline bounds the overshoot, it does not shorten solves already inside their budget.

`docs/dev/1009-kernel-lp-deadline-measurement-2026-08-13.md` records the above plus the 20-instance × 4-budget corpus panel (cert-clean at every budget — no bound past its reference optimum, no incumbent invented; bounds and node counts unchanged wherever the budget does not bind; worst overshoot +0.20 s → +0.01 s) and the one honest cost: at a budget smaller than a single LP, `QPLIB_3815` trades a `-96` bound *obtained by overrunning* for no bound. Its §0 restates #1014's retraction up front, so these numbers cannot be misread as explaining the QPLIB_1157 report.

## Tests run

- [x] `pytest -m smoke` → 1008 passed, 15 skipped, 2 xpassed (458.70s)
- [x] `python/tests/test_1009_kernel_lp_deadline.py` → 3 passed (14.16s)
- [x] `cargo test -p discopt-core --lib` → 602 passed (unchanged; no Rust touched — run only to establish the mutant baseline above)

`pytest -m slow python/tests/test_adversarial_recent_fixes.py` not re-run on this revision: it exercises solver behaviour, and this revision touches no solver source. Say the word if you want it anyway.

## Regression test

- [x] Added — fail-before/pass-after demonstrated against the mutant that reverts #1014's call site, since the fix itself is already on `main`.

https://claude.ai/code/session_01G2BJLtVfbs7sCvjBrtgXPE

### DIFF
--- a/docs/dev/1009-kernel-lp-deadline-measurement-2026-08-13.md
+++ b/docs/dev/1009-kernel-lp-deadline-measurement-2026-08-13.md
@@ -1,0 +1,104 @@
+# What the #1009 kernel node-LP deadline is worth — measurement + coverage (2026-08-13)
+
+**Status.** Measurement-and-coverage only; no behavior change. #1014 already threaded
+the tree's live deadline into the node LP (`node_lp_opts`) and #1015 fixed the
+incremental McCormick fast path, closing #1009. This records what the kernel-side
+change actually bounds, and closes a coverage gap it left.
+
+## 0. Correcting the attribution first
+
+#1009 was opened against `spatial_bindings.rs` with QPLIB_1157 (240.83 s against a
+20 s `time_limit`, 11.9×) as evidence. **That attribution was wrong**, and #1014
+retracted it: threading the deadline did not shorten that repro (419.74 s, 21.0×,
+nodes=1), and `faulthandler` sampling put the wall clock in
+`_root_relaxation_lower_bound` → `MccormickLPRelaxer.solve_at_node` →
+`solve_lp_warm_std` — the Python-side root relaxation computed *before* the kernel
+tree is entered. #1015 fixed that. Any reading of the numbers below as "this is what
+caused the QPLIB_1157 overrun" is the same error; they are about a different,
+genuinely present defect that #1009 correctly identified by inspection.
+
+## 1. The quantity
+
+`solve_spatial_tree` checks its clock only BETWEEN nodes. Without a per-LP deadline
+the wall-clock floor of any kernel solve is therefore the cost of **one node LP**,
+however small the caller's budget — unbounded in the size of the relaxation, and not
+a property of any corpus instance.
+
+#1014 shipped on a soundness-by-construction argument with no measurement
+("Sound by construction, not by measurement"), so this is the missing half.
+
+## 2. Method
+
+Both arms run against the **same fixed code**, so this reproduces on `main` with no
+flag and no revert (`scripts/spatial_kernel_lp_deadline_scaling.py`):
+
+* **uncapped** — `time_limit_s=None, max_nodes=1`: one node LP run to completion.
+  This is exactly what the pre-#1014 path spent on node 1 regardless of the budget,
+  i.e. the floor the fix removes.
+* **capped** — `time_limit_s=BUDGET`: the same LP under the deadline.
+
+> x86-64 Linux, 4 cores, Python 3.11, **release** build (`pip install -e .` → maturin
+> release). Load gate: `uptime` load average 0.24–0.63. Arm order alternates per rep
+> (CLAUDE.md §9); medians of 3 with sd. Driver is a general dense-bilinear model
+> (`min Σ_{i<j} x_i x_j s.t. Σx ≥ n/2, x ∈ [0,1]^n`), not a named instance
+> (CLAUDE.md §2), whose global optimum `C(n/2, 2)` is closed-form and is asserted as a
+> soundness oracle on every solve.
+
+## 3. Result — budget fixed at 1 s
+
+| n | lifted terms | cols | uncapped (one node LP) | capped | bound ≤ optimum |
+|---:|---:|---:|---:|---:|:--:|
+| 50 | 1 225 | 1 275 | 0.81 s (sd 0.03) — 0.8× | 1.02 s (sd 0.02) — 1.0× | ok |
+| 70 | 2 415 | 2 485 | 2.68 s (sd 0.04) — 2.7× | 1.20 s (sd 0.02) — 1.2× | ok |
+| 90 | 4 005 | 4 095 | 7.69 s (sd 0.09) — 7.7× | 1.23 s (sd 0.01) — 1.2× | ok |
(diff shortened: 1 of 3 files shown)


jkitchin:
**Do not merge as-is — this PR is superseded, and its body carries a claim I have to retract.**

CI is green (14/14 success or skipped), but `main` moved while this was open and now carries the same work:

- **#1014** (`6a5f573`) threads the tree's live deadline into the node LP as `node_lp_opts` — functionally the same change as the `node_opts` here, same "earlier of the two" composition, same soundness argument (`IterLimit` → `Undecided` → branched with the parent's bound, never fathomed). It ships unconditionally; this PR wraps it in a `DISCOPT_SPATIAL_LP_DEADLINE` opt-out that `main` deliberately does not have.
- **#1015** (`7118f19`) fixes what actually caused the reported overrun: the incremental McCormick fast path issued every LP unbounded, so `time_limit` died two frames short of `solve_lp_warm_std`. That closed #1009.

**Retraction (CLAUDE.md §11).** The summary above attributes the QPLIB_1157 240.83 s-against-20 s overrun to the kernel node LP, repeating the issue's original framing. That attribution is wrong and was already disproven in #1014: threading the deadline did **not** shorten that repro (419.74 s, 21.0×, nodes=1), and `faulthandler` sampling put the whole wall clock in `_root_relaxation_lower_bound` → `MccormickLPRelaxer.solve_at_node` → `solve_lp_warm_std`, i.e. the Python-side root relaxation computed *before* the kernel tree is entered. The same misattribution appears in `docs/dev/1009-spatial-lp-deadline-panel-2026-08-13.md` (§1 and the header) and in the "Closes #1009" line here. Nothing in the *measurements* is affected — they were taken at the kernel binding directly — but the causal claim they were framed around is not mine to make.

What survives and is not on `main`, if any of it is wanted:

1. `python/tests/test_1009_kernel_lp_deadline.py` — Python-level coverage of the kernel contract (an interrupted LP reports `time_limit` with no bound, no incumbent, `n_undecided ≥ 1`). `main`'s coverage for this path is Rust-only. It would need rewriting: three of its seven tests are built on the opt-out flag, which does not exist on `main`.
2. An independent **measurement** that the kernel-level mechanism is real, which #1014 notes it shipped without ("sound by construction, not by measurement"). Holding the budget at 1 s and growing a dense bilinear relaxation — a general model, not a named instance: 1 225 lifted terms → 1.10 s (sd 0.02); 4 005 → **5.42 s** (sd 0.11); 7 140 → **16.33 s** (sd 1.25); against 1.03–1.12 s with the deadline threaded. So the overrun this closes is unbounded in LP size even though it was not the cause of the QPLIB_1157 report.
3. A 20-instance × 4-budget differential panel, cert-clean at every budget (0 bounds past a reference optimum, no incumbent invented), with one honest cost recorded: at a budget smaller than a single LP, `QPLIB_3815` trades its `-96` bound (obtained by overrunning) for no bound.

Awaiting direction on whether to close this outright or reduce it to (1)+(2) rebased onto `main`.

---

Issue #1008, opened 2026-08-13
perf(lp): in-house simplex is 8.7x-29x slower than HiGHS on identical LPs


jkitchin:
## Summary

On the identical LP, the in-house Rust simplex is **8.7x to 29x slower than HiGHS**, and the ratio grows with row count. This is now the dominant cost on continuous nonconvex QPs and the main thing standing between discopt and BARON on this class.

## Measurement

`scratchpad/qplib_run/node_rate.py`, QPLIB_1157 root LP, same matrix handed to both engines:

```
rlt=0  rows= 3273   highs= 0.034s   in-house=  0.292s   ratio=  8.7x
rlt=1  rows= 3937   highs= 0.231s   in-house=  6.696s   ratio= 29.0x
executed comparisons: 2
```

Note the scaling: a 20% row increase costs HiGHS 6.8x and the in-house simplex 23x. The degradation is superlinear in rows, which is exactly the direction a tighter relaxation pushes.

## Why it matters

With #1007's equality-factor RLT the relaxation is much tighter — 86.5% of the remaining dual gap closed on QPLIB_1157 — but the solver only completes **3 nodes in 21 s**, because one node LP costs ~6.7 s. With RLT off it manages 104 nodes on a much weaker bound. Neither configuration converges.

BARON solves the same instance in ~0.5 s. It gets both a tight relaxation and fast LPs; after #1007 we have the first and not the second. Further relaxation tightening has **negative** marginal value until node throughput improves, since every added row makes the per-node LP disproportionately worse.

## Not yet investigated

The 29x is measured, but the breakdown is not. Candidates, in no particular order and none of them confirmed:

- pricing rule (partial/Devex vs full Dantzig scans)
- refactorization cadence and LU update strategy
- bound-flipping ratio test / long-step dual ratio test
- dense vs sparse ftran/btran paths
- presolve and scaling differences on the lifted matrix

A profile of the RLT-on 1157 root LP is the obvious entry experiment before any implementation. Per CLAUDE.md section 4 the hypothesis, the evidence and the kill criterion should be stated before building.

## Reproduce

```
python scratchpad/qplib_run/node_rate.py
```

Prints an executed-comparison count and exits non-zero if it is zero.

Found while working on #1007.

jkitchin:
## Attribution complete; six hypotheses tested, all six falsified. The gap is **not** closed.

Recording the full measurement so nobody repeats it. Everything below is on
[...]

PR #1012, opened 2026-08-13
perf(lp): #1008 split the LU refactorization profile into symbolic/numeric phases

jkitchin:
## What this is

Instrumentation only. **No behavior change, no flag, no performance claim.**
Net diff is three files: `profile.rs` (two phases, three counters, `incr_by`, a
test lock), `linsolve.rs` (time the two halves of a sparse refactorization
separately + one test), and one line in a `dual.rs` test that now takes that lock.

This is what survives the #1008 investigation. Six hypotheses were tested and all
six were falsified; the full record with numbers is posted on
[#1008](https://github.com/jkitchin/discopt/issues/1008#issuecomment-5275450375).
The throughput gap is **not** closed and the issue stays open.

## Why keep these

The entry flamegraph attributed **59.5% of LP wall to `factorize_sparse`** without
saying which half. `Phase::LuSymbolic` / `Phase::LuNumeric` split it, and the two
halves have completely different fixes:

| | share of total LP wall |
|---|---:|
| numeric `SparseLu::factor` | ≈50% (18%–74%) |
| symbolic `SparseLuSymbolic::analyze` | ≈11% (2.7%–25.5%) |
| everything else (ftran/btran + FT update) | ≈38% |

measured over 18 captured QPLIB relaxation LPs. That attribution is what routed
the investigation, and it is what a future attempt needs on day one rather than
re-deriving.

`LuSparseFactorizations` / `LuBasisNnz` / `LuFactorNnz` give the fill ratio
`nnz(L+U) / nnz(B)`, which sets the cost of all three LU hot spots at once
(refactorize, each Forrest–Tomlin update, every ftran/btran). It is what killed
the fill hypothesis: SuperLU (COLAMD) on the same 12 final bases is a median
**0.94x** — feral's factor is already smaller on 8 of 12, so there is no fill
headroom to chase.

## Two flags were added during the investigation and are **deleted**, not shipped OFF

- `DISCOPT_LU_SYMBOLIC_REUSE` (reuse the fill-reducing column ordering across
  refactorizations): reuse rate 77.2%, so the mechanism operated — median speedup
  **1.009x** against a pre-registered 1.05x bar, plus two `optimal` → `iter_limit`
  regressions.
- `DISCOPT_LP_DUAL_HARRIS` (Harris stability pass 2 in the dual ratio test, which
  the primal loop has and the dual loop does not): wall median **0.999x**, median
  Δiter **−0.3%** against a ≥5% bar, and one `optimal` → `iter_limit` regression.

Both are the `DISCOPT_CUT_INHERIT` outcome — sound but not helpful — and
CLAUDE.md §3 forbids dead flags, so they are removed exactly as the H2
threshold-pivoting module was. The commits that added them are kept in the
history with their retractions so the trail is auditable.

## Regime (CLAUDE.md §5)

**Bound-neutral.** Nothing here touches the pivot sequence, a bound formula, a
tolerance, or a guard. `profile::incr`/`incr_by`/`Timer` are all no-ops unless
`DISCOPT_PROFILE=1`, and the refactored `factor_sparse_ordered` is the
pre-existing analyze-then-factor path verbatim with two RAII timers around it.

## Testing

- `cargo test -p discopt-core` — **600 passed, 0 failed** (plus 4 + 1 doc/integration, 0 failed).
- `pytest -m smoke` — **969 passed, 1 skipped, 2 xpassed**, 94 s.
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — **10 passed**, 134 s.
- `cargo fmt --check` clean; pre-commit hooks ran on the commit (no `--no-verify`).

New test `the_sparse_factorization_counters_fire_and_solves_stay_exact` asserts
the counters fire on the **sparse** route (a silent fall to the dense route would
leave them at zero and read as "no factorizations happened") and that the solves
stay exact. **Mutation-tested**: commenting out the `LuSparseFactorizations`
increment fails it (`4 sparse factorizations were performed but only 0 counted`);
restoring it passes. It asserts lower bounds rather than equalities because the
counters are process-global and `cargo test` runs siblings in parallel —
contamination can only add, so a lower bound is exactly as mutation-sensitive as
an equality while an equality would be flaky.

Contributes to #1008. The 8.7x–29x throughput gap remains open: ~2.3x of it is
extra pivots and ~12.5x is per-pivot linear algebra, ~88% of which lives in the
external `feral` crate (numeric LU at a ~2.1x constant factor, and ftran/btran
implemented as dense-vector loops rather than hyper-sparse).


jkitchin:
Reviewed the full diff independently of the investigation record. The three
source-file changes are bound-neutral as claimed: `incr`/`incr_by`/`Timer` are
all gated on `profile::enabled()`, `factor_sparse_ordered` is the pre-existing
analyze-then-factor sequence with two RAII timers around it, and the only
non-instrumentation edit is a `test_guard()` line in an existing test. Confirmed
`DISCOPT_LU_SYMBOLIC_REUSE` and `DISCOPT_LP_DUAL_HARRIS` have zero remaining
references anywhere under `crates/` or `python/` -- the flags are genuinely
deleted, not left dead (CLAUDE.md 3).

The lower-bound assertions in the new counter test are the right call and the
reasoning in the comment is correct: the counters are process-global, siblings
run in parallel, contamination can only add, so `>=` is exactly as
mutation-sensitive as `==` without the flakiness.

Two leftovers from the removed threshold-pivoting arm found and fixed in
7e3f3cd1:

- `FeralLU::params`'s doc comment was edited to claim it carries "the
  threshold-pivoting parameter (#1008)". `LuParams` has no such field on this
  branch or on main, and the edit also removed the accurate phrase it replaced
  ("strict partial pivoting"). Restored.
- `factor_sparse_ordered` carried an unused `_m: usize` -- the arity the deleted
  arm needed. Removed with its call site.

`cargo test -p discopt-core` 600 passed / 0 failed after the fix, unchanged.

On the finding itself: the entry attribution is the valuable part and it is a
clean negative. 4965 pivots vs HiGHS's 2153 is 2.3x the iterations but 12.5x the
time per iteration, and pricing plus the ratio test -- the two suspects #1008
named -- are 1.7% of wall. That routes the remaining work to feral's ftran/btran
and numeric kernel rather than to anything schedulable from this repository, and
it is worth keeping precisely because it says where NOT to look next.

Issue #1017, opened 2026-08-13
fix(correctness): the Farkas certificate's NS margin is built from result magnitudes, and certifies a feasible LP infeasible


jkitchin:
## Summary

`farkas_ray_certifies_cols` (`crates/discopt-core/src/lp/simplex/primal.rs`)
returns `true` — i.e. the engine emits `LpStatus::Infeasible` — on a captured
relaxation LP that SciPy/HiGHS, three perturbed pivot paths of our own engine,
and an elastic feasibility LP all agree is **feasible**.

The deciding quantity is `bᵀy = 3.0000052e-8`, computed from terms whose absolute
sum is `Σ|b_i·y_i| = 600`. That is a relative cancellation of **5e-11**, tested
against a margin of `1e-9`.

Found while working #1013 (the degenerate-stall investigation); it is a separate
defect and is not fixed by that PR, which only stops a stalled warm dual loop
from being the thing that reaches this verdict.

## The margin does not cover the arithmetic it guards

```rust
// primal.rs, farkas_ray_certifies_cols
let by: f64 = b.iter().zip(&ys).map(|(bi, yi)| bi * yi).sum();
...
let margin = 1e-9 * (1.0 + by.abs() + abs_sum);
if by - contrib > margin { return true; }
```

`abs_sum` is `Σ|boxmax_j|` — the absolute sum of the *results* `aty_j · bound_j`.
Neither it nor `by.abs()` bounds the rounding of the **accumulations** that
produce those results:

- `by` is a sum over `m` terms of magnitude up to `|b_i·y_i|`, whose absolute
  sum is 600 on this instance while the result is 3e-8;
- each `aty_j = cols.dot(j, &ys)` is itself a cancelling sum, and only its
  *result* enters `abs_sum`;
- the implied slack upper bounds from `slack_upper_bounds` are each
  `(b_i − min_other)/a` with `min_other` a floating-point row sum, and their
  rounding is not represented at all.

A Neumaier–Shcherbina test is rigorous only when the evaluation error is bounded;
the standard bound for a length-`k` dot product is `k·eps/(1−k·eps) · Σ|terms|`.
Measured here, `Σ|b_i·y_i| = 600` with `m = 4292`, so the row-side term alone is
~2.8e-10 — same order as quantities the current margin (1e-9) treats as
comfortably resolved, and the column-side and implied-bound terms are unbounded
by anything in the current expression.

## Reproduce (fully from vendored data — no external corpus)

```bash
[...]

Issue #1013, opened 2026-08-13
perf(lp): the dual loop has no reachable anti-degeneracy escape — a rounding-level lever swings a 100%-degenerate LP between 18k pivots and no bound (HiGHS: 81)


jkitchin:
> **Re-scoped 2026-09-06.** The original framing — `QPLIB_0911`, five `perf/1008`
> perturbation levers, and a dual Harris pass as the cure — is not runnable and
> not right: four of the five levers never landed on `main`, and the Harris pass
> has been falsified twice (#1008 corpus-wide, then re-tested *scoped to the
> stall* in #1018). It is preserved at the bottom. The body below is the same
> pathology re-anchored on a fixture that ships in this repository, measured on
> `main` @ `09fa315`.

## Summary

The dual simplex has **no reachable anti-degeneracy escape**, and on a
100 %-degenerate LP its pivot count is set by rounding-level accidents rather
than by the problem.

The anchor is `crates/discopt-core/src/lp/simplex/testdata/qplib2170_cold_fail_lp.json`
— the captured `QPLIB_2170` root relaxation (1755×3193), already vendored for the
#1008 tests. Warm re-solve from the captured basis, varying **only**
`DISCOPT_LP_REFAC_INTERVAL` (#1008 D3 — the refactorization cadence, which
changes no tolerance, guard or bound formula and computes the same basis inverse
to within rounding):

| `REFAC_INTERVAL` | `recover_unstable_pivot=true` (node path, deadline set) | `recover_unstable_pivot=false` (`SimplexOptions::default`) |
|---|---|---|
| **48** (default) | `Optimal` obj 0, **17 794** pivots | **`Numerical` — no bound** |
| 24 | `Optimal` obj 0, **20 594** pivots | `Optimal` obj 0, 20 594 pivots |
| 100 | `Optimal` obj 0, **23 869** pivots | **`Numerical` — no bound** |

Deterministic; identical across reps. HiGHS certifies the same LP in **81**
pivots (recorded in `dual.rs` on
`unstable_pivot_recovery_is_not_gated_on_a_deadline`), so the engine is spending
**220–290x** the pivots — and on the no-recovery arm the refactorization cadence
alone decides whether a bound comes back at all.

Every pivot in the `refac=24` cell is degenerate (`DualDegeneratePivots` =
`iters` = 20 594, `DualDegenerateRunMax` = 20 594). Throughout:
`DualStallTrips` = **0**.

## Why there is no escape

[...]

Issue #1004, opened 2026-08-12
GDP primal: the constructor's per-candidate feasibility test is single-start and rejects ~80% of genuinely feasible configurations


jkitchin:
## What was measured

While enumerating three-demotion configurations for #993, the control arm measured
something that is not about `syngas` and not about configuration reachability: the
constructor's per-candidate **feasibility test** is unreliable.

With the integers pinned to a configuration that is *known feasible* (BARON's
proven-optimal configuration on `syngas`, big-M):

| probe | starts tried | feasible |
|---|---|---|
| B1 (`scratchpad/issue993/B1_syngas.log:34`) | 67 | **12** |
| #993 C2 control (`scratchpad/issue993/C2_multistart.log`) | 6 | **2** |

So single-start detection on this model runs roughly 18–33 %. (Kept separate rather
than pooled — the start families differ.) `one_hot_config_subnlp` and
`one_hot_config_dive` solve **one** start per candidate configuration.

The consequence: a candidate that is genuinely feasible is discarded most of the
time. A 256-plan wave is not 256 tests, it is 256 tests each of which fails ~4 times
in 5 on a true positive. Even a perfect ranking that proposed the right
configuration as plan 1 would probably throw it away.

## The obvious fix is probably wrong — do not build it without measuring this first

"Detection is 20 %, so use k starts per config" does not survive a budget argument.
Under a fixed sub-NLP solve budget `B`, per-start detection `p`, and a feasible
fraction `f`:

* `B` configs x 1 start -> expected finds `B*f*p`
* `B/k` configs x k starts -> expected finds `(B/k)*f*(1-(1-p)^k)`

The ratio is `(1-(1-p)^k)/(k*p)`, which is **<= 1 for every k >= 1** by subadditivity
of `1-(1-p)^k`. At `p=0.2, k=5` it is 0.67 — a third *worse*. Since only one
incumbent is needed and feasible configurations are rare (#993 measured 1 in 2600 at
distance 3, detection-limited), the rare-event regime makes expected-count and
P(at least one) agree, so the conclusion carries.

**Spreading one start over many configurations weakly dominates concentrating starts
on few.** The current single-start design may already be optimal, and the ~80 %
false-infeasible rate may be a cost of doing business rather than a defect.

## The one escape hatch worth measuring

The argument above assumes every start costs the same. If a *re*-start on an
already-built fixed-integer sub-problem is materially cheaper than the first solve
[...]

PR #1018, opened 2026-08-13
fix(lp): #1013 give a stalled warm dual re-solve an escape

jkitchin:
## Summary

Contributes to #1013 (does not finish it — see "What remains" below).

The warm dual simplex has two anti-degeneracy escapes and **neither is reachable
on a lifted relaxation**. Bland's rule engages after `2·(n+1)` consecutive
degenerate pivots; the F2 stall guard trips at the size-derived pivot cap
`20·(m+n)+500`. Measured over a 100-LP panel of in-repo root relaxations (all 9
vendored QPLIB instances and all 66 vendored MINLPLib `.nl` instances,
`rlt_lineq` off and on), `DualBlandActivations` and `DualStallTrips` are **0 on
every single LP** — including cells where 98.7 % of pivots are degenerate and the
solve exhausts its budget. So a degenerate stall had no exit but the iteration
cap, and on one LP it ended in an `infeasible` verdict that HiGHS, every
perturbed arm of this engine, and an elastic `min t` feasibility LP contradict.

This adds `SimplexOptions::dual_stall_patience` (default 2048, from
`DISCOPT_LP_DUAL_STALL_BAIL`): past that many consecutive degenerate pivots the
warm loop returns `None`, i.e. the caller's cold two-phase primal finishes the
solve. It also ships the instrumentation the issue asks for
(`DualDegenerateRunMax` via a new `profile::record_max`, `DualDegenerateRunArms`,
`DualDegenerateStallBails`, and a per-pivot `DISCOPT_LP_DUAL_TRACE` stream).

**Environment caveat — superseded, see "What remains on #1013 — corrected".**
This description originally said `QPLIB_0911` was unreachable and that its
headline cell could not be measured. The first half was true of the environment
this was *developed* in (no `~/Dropbox` corpus, `qplib.zib.de` blocked by the
network policy) but **not** of the machine the cells were later run on, where the
instance is present. The cells have since been run, and the reason they cannot
settle the issue turned out to be different — the levers the issue names live on
the unmerged `perf/1008` branch. That section carries the corrected account and
the measurement; everything in the panels below is measured on the vendored
corpora either way.

### To run it locally

```bash
python -u scratchpad/i1013/capture.py                 # rebuild lps/*.npz (gitignored, ~888 KB)
python -u scratchpad/i1013/panel.py out.jsonl 20 2 off:DISCOPT_LP_DUAL_STALL_BAIL=0 bail:
python -u scratchpad/i1013/report_panel.py out.jsonl
python -u scratchpad/i1013/tree_panel.py              # tree-level, oracle-checked
```

`capture.py` selects instances by filter and now takes the corpus location from
the environment, so running the issue's own cells needs no source edit:

```bash
I1013_QDIR=~/Dropbox/projects/discopt-minlp-benchmark/qplib \
I1013_NLDIR=~/Dropbox/projects/discopt-minlp-benchmark/minlplib/nl \
I1013_ONLY=QPLIB_0911 python -u scratchpad/i1013/capture.py
```

A non-existent directory and a filter matching nothing are both refused.
`DISCOPT_LP_DUAL_TRACE=1` gives the per-pivot stream;
`DISCOPT_LP_DUAL_STALL_BAIL=0` restores the previous loop
(`1`/`true`/`on` mean "enabled at the default patience", **not** a patience of one
pivot — an earlier harness read it literally and produced a phantom 17x tree
regression).

## Correctness

The action on trip is the one every other difficulty in this loop already takes:
return `None` so the caller cold-solves. The cold two-phase primal self-verifies
its own verdict, so the bail can change only *which* engine finishes a solve,
never the value it returns. No tolerance, guard, bound formula or certificate
test is touched. The one status change in the pre-merge panel was an
`infeasible` → `optimal` improvement on an LP independently shown to be feasible
— **now attributable to #1017, not to this PR; see Correction 2.**

- [x] `pytest -m smoke` → **1008 passed**, 16 skipped, 2 xpassed
- [x] `cargo test -p discopt-core` → **608 passed**, 0 failed (on the tree merged with `main`; 604 from this branch + 4 from #1017's tests)
- [x] `ruff check python/` clean; `cargo fmt --check` clean; `cargo clippy -p discopt-core` clean.
      `ruff format --check python/` reports 2 files (`test_mo_augmecon2.py`,
      `test_log_square_relaxation.py`) — both are unformatted on `main` already and
      untouched here (this branch changes no `python/` file).

## Regression test

Two new tests in `crates/discopt-core/src/lp/simplex/dual.rs`, on captured
fixtures emitted by `scratchpad/i1013/make_fixture.py`:

- `degenerate_run_detector_sees_what_the_stall_trip_counter_cannot` — replays the
  captured `st_testgr3` relaxation (103×144) and asserts `DualStallTrips == 0`
  while the run-length counters record a degenerate run two orders of magnitude
  past the arming threshold. This is the defect: before this change there was **no
  counter that could report it**, so the assertion cannot be written at all
  against the old code.
- `dual_stall_bail_fires_on_a_captured_stall_and_is_bound_neutral` — replays the
  captured `tspn12` relaxation (1345×1801) through the production entry point,
  and carries its own fail-before/pass-after: with `dual_stall_patience = 0` (the
  pre-#1013 loop) it asserts **0 bails**, and with the patience below the LP's
  620-pivot run it asserts exactly **1 bail** *and* the same objective as the
  warm-converged arm.

Bound-changing under CLAUDE.md §5 (it changes which engine solves a stalled node
LP, hence the degenerate vertex a bound may come from). It ships **default-ON**,
with `DISCOPT_LP_DUAL_STALL_BAIL=0` as the opt-out and the legacy path intact.
**Read the graduation case through Correction 2** — on the current base this is a
tail guard, not a broad net-positive.

**LP graduation panel** — 100 captured relaxation LPs, 2 reps, arms interleaved
within each rep, 20 s per-LP limit; re-run on the post-#1017 base.

| gate | result (post-#1017 base) |
|---|---|
| unchanged cells | **99 / 100** identical status *and* identical iteration count |
| status regressions | **0** |
| status improvements | **0** (was 1 pre-merge — see Correction 2) |
| objective drift (optimal/optimal) | **0.00e+00** (n = 97) |
| cells where the bail fires | **3** — 1.34x (`QPLIB_3814_rlt1`), 1.13x (`tspn10_rlt1`), neutral (`QPLIB_3815_rlt1`) |
| LPs < 50 ms (90 of 100) | spread at **identical iteration counts** — sub-millisecond noise, not effect |

**Tree-level differential panel** — every vendored `.nl` with a recorded optimum
(16 instances, 60 s each), re-run on the same base: objective, bound and node
count **identical on all 16**; 96 soundness assertions (no bound past its
reference optimum, no incumbent beyond it, no certification or status
regression), **0 issues**. The bail does not fire in these trees (longest
degenerate runs 47–87 pivots).

**Falsified and NOT shipped** (measurements in `docs/dev/performance-plan.md` §18
and `scratchpad/i1013/FINDINGS.md`):

| lever | outcome |
|---|---|
| #1008's dual Harris pass, re-landed **armed only inside the stall** (the issue's ask) | fires on 22/100, wall median **0.995x**, iteration median **0**, regresses `tspn10_rlt1` `optimal` → `iter_limit` |
| Bland's rule at a *reachable* run length | fires on 8/100, median 1.007x, same status regression |
| a conjunctive primal-infeasibility progress test | `max_noprog == max_run` on every LP traced — never discriminated once; removed rather than shipped as an untested branch |

The trace also contradicts the issue's hypothesized mechanism: on the worst cell
the chosen pivot magnitude is **exactly 1.0 at the median** with none below 1e-4
and no zero primal steps, so the degeneracy is *dual* and an `|α_rj|` tie-break
has nothing to discriminate on this corpus.

## Corrections to an earlier revision of this description (CLAUDE.md §11)

Three numbers in the panel section were wrong when first posted and are corrected
above. They were re-derived from this PR's own `scratchpad/i1013/final_panel.jsonl`:

| claim as posted | actual |
|---|---|
| "all **68** vendored MINLPLib `.nl` instances" | **66** (`ls python/tests/data/minlplib_nl/*.nl`) |
| "LPs < 50 ms (**83** of 100)" | **90** of 100 — and 90/10 under every aggregation (median/min/max/mean, either arm); the ratio ranges quoted beside it were always computed on the true 90/10 split, so only the count was wrong |
| "the **two** cells where it fires gain 1.13x and 1.20x" | **three** cells fire; the omitted one, `QPLIB_3815_rlt1`, gains **1.006x** — neutral, not a gain |

The third is the one that mattered: dropping the neutral cell made the
fired-cell evidence read as uniformly positive when one of three showed no
effect.

Two instruments that could have reported success without measuring anything were
also fixed: `DISCOPT_LP_DUAL_STALL_BAIL` silently read an
unrecognized *off* value as ON — Python's `str(False)` == `"False"` hit exactly
this — and `lprun.py` recorded counters under `if k in snap`, so a stale `_rust`
produced a full set of clean-looking RES lines containing none of the change.
Both now refuse loudly; both were demonstrated fail-before/pass-after against the
real defect, not asserted.

## Correction 2 — the merge with `main` withdrew this PR's status improvement (CLAUDE.md §11)

`main` merged #1019 (the Farkas accumulation-margin fix for #1017, the issue this
PR's investigation filed) while this branch was open, and the branch has now been
merged up to it. That fix repairs the certificate that produced the
`QPLIB_3814_rlt1` false `infeasible`, so on the merged base:

```
bail=0  status=optimal  obj=0.23839462819531176  maxrun=3352  bails=0
bail=1  status=optimal  obj=0.23839462819531176  maxrun=2048  bails=1
```

**That LP is `optimal` with the bail OFF as well as ON.** The "1 status
improvement" claimed in the earlier revision is **#1017's result, not this
mechanism's**, and is withdrawn. It supersedes the "1 improvement" line in the
previous corrections block too.

Panel re-run on the merged base (`scratchpad/i1013/postmerge_panel.jsonl`):

| gate | pre-merge base | post-#1017 base |
|---|---|---|
| unchanged cells (status *and* iterations) | 98 / 100 | **99 / 100** |
| status regressions | 0 | **0** |
| status improvements | 1 | **0** |
| objective drift, optimal/optimal | 0.00e+00 (n=96) | **0.00e+00** (n=97) |
| cells where the bail fires | 3 | 3 — **1.34x**, **1.13x**, one neutral |

Tree panel on the same base: objective, bound and node count identical on all 16
instances, 96 assertions, 0 issues. `cargo test -p discopt-core` → **608 passed**
on the merged tree.

**What this leaves.** The mechanism is a **tail guard, not a throughput change**:
inert on 97 of 100 panel LPs (bit-identical iteration counts), worth 1.34x
(`QPLIB_3814_rlt1`) and 1.13x (`tspn10_rlt1`) on the stalls it catches, and
nothing on the third. Same shape of value as the F2 stall guard beside it. With
the correctness win now attributable to #1017, **"broadly net-positive" is not an
accurate description of this change** and is no longer claimed. Whether that
clears the §5 net-positive bar for a default-ON graduation is a call for the
owner; the alternative is flipping the default to OFF with
`DISCOPT_LP_DUAL_STALL_BAIL=1` as the opt-in.

Full record: `docs/dev/performance-plan.md` §18a, `scratchpad/i1013/FINDINGS.md` §6.

## Related

The `QPLIB_3814_rlt1` verdict is a **separate** defect, filed as #1017 and fixed
on `main` by #1019: its Neumaier–Shcherbina margin was built from result
magnitudes rather than accumulation magnitudes, so `bᵀy = 3.0e-8` against a term
magnitude of 600 passed as a proof of infeasibility. This PR only stops a stalled
warm loop from being the thing that reaches that verdict; it never repaired the
margin.

## What remains on #1013 — corrected

The earlier text here said the remaining work was to "re-run the issue's five
perturbation cells on `QPLIB_0911` from the full corpus," and attributed the gap
to corpus access. **Both halves of that were wrong**, and the cells have now been
run (`scratchpad/i1013/cells_0911.py`, `cells_0911.jsonl`).

**1. Corpus access was never the blocker.** `QPLIB_0911` is present locally
(`qplib/qplib/QPLIB_0911.qplib`). Captured at 5150×1325, nnz=45146.

**2. The issue's five levers do not exist on this branch.**
`DISCOPT_LP_REFACTOR_INTERVAL`, `DISCOPT_LU_SYMBOLIC_REUSE`, and any LU
`pivot_threshold` knob return zero hits crate-wide. They were `perf/1008` flags
and that branch is unmerged. The cells cannot be run as the issue writes them,
now or after any amount of corpus work. The three rounding-level knobs this tree
*does* have (`DISCOPT_LU_DENSITY_ROUTE`, `DISCOPT_LP_FACTORIZATION_HARDENING`,
`DISCOPT_LP_ITERATIVE_REFINEMENT`) are bit-identical no-ops on this LP — same
objective to 12 decimals, same 553 degenerate pivots, same max run — so there is
no substitute perturbation arm. That result is **inconclusive, not negative**: I
have no arm that perturbs anything, so it says nothing about whether a
perturbation would stall this LP.

**3. The baseline reproduces #1013 exactly** — 1279 pivots, `optimal`,
obj −76.385275000002, `DualDegeneratePivots = 553` (43%), `DualStallTrips = 0`.

**4. But the bail provably cannot fire on `QPLIB_0911`.**
`DualDegenerateRunMax = 40` against a patience of 2048 — off by 51×, and
impossible in principle, since the entire solve is 1279 pivots. Confirming "it
converts that `iter_limit` into a cold-solved `optimal`" was never achievable on
this instance with this patience. 8/8 cells, both arms: identical status,
identical iteration count, zero bails.

### The patience of 2048 is well-placed — here is the measurement

The PR shipped 2048 without presenting evidence for it. The panel supplies it.
`DualDegenerateRunMax` across all 100 LPs, bail OFF:

| LP | max degenerate run | |
|---|---:|---|
| `QPLIB_3815_rlt1` | 7888 | bail fires |
| `tspn10_rlt1` | 5512 | bail fires |
| `QPLIB_3814_rlt1` | 3352 | bail fires |
| — patience **2048** — | | |
| `tspn12_rlt1` | 1274 | |
| `tspn08_rlt1` | 902 | |
| `QPLIB_3871_rlt{0,1}` | 663 | |

min 0, median **7**, p90 341, max 7888. Distribution:
`{0: 2, 1–99: 78, 100–999: 16, 1000–2047: 1, ≥2048: 3}`.

The panel is bimodal — a bulk that never exceeds 1274 and a tail at 3352+ — and
2048 sits in the empty band between them, 1.61× above the highest non-stalling
run and 1.64× below the lowest stalling one. The 3 LPs that exceed the patience
are exactly the 3 cells where the bail fires; there are no near-misses to tune
against.

### Instrument defect found and fixed (CLAUDE.md §6)

The first run of these cells produced 32 clean records with **every counter 0**,
including `DualWarmSolves` — impossible in a script whose only entry point is the
warm dual solve. Cause: the counters are gated on `DISCOPT_PROFILE`, which the
driver did not set; without it they read 0 and emit well-formed records that say
"no degeneracy, no bail, no warm solve" about a solve that just did 553
degenerate pivots. The §8 guard passed the whole way because it checked the
counter symbols *existed*, not that they *fired*. Fixed: `lprun.py` now refuses
on `DualWarmSolves < 1`, and the driver sets `DISCOPT_PROFILE=1`. **The panel in
this PR is unaffected** — `panel.py:49` already sets it; the omission was in the
new driver only.

### Status

`Contributes to` #1013 remains correct, but the reason has changed: not "a cell
is still unrun," but "the cell the issue asks for cannot be constructed on this
branch, and on the instance it names the guard is unreachable by 51×." Closing
#1013 on this PR would require either landing the `perf/1008` levers first or
re-scoping the issue to the panel evidence above.

No wall-clock claim is made in that section. Machine load reached 43.56 during
capture (`mediaanalysisd` at 215%, `suggestd` at 96% — macOS indexing triggered
by the corpus reads). Every number there is a deterministic status, iteration
count, or counter.

https://claude.ai/code/session_01NKbF9EYSGFL1tnKYfoGqtg

---
_Generated by [Claude Code](https://claude.ai/code/session_01NKbF9EYSGFL1tnKYfoGqtg)_


jkitchin:
## Merge conflict resolved, and the panel re-measured on the new base — one claim withdrawn

The PR was `dirty` against `main`. Resolved and pushed (`a6a9d0e`, `b7ff686`). CI was green on the previous head and the merge is now clean.

**Conflict:** both sides appended a `## 17.` section to `docs/dev/performance-plan.md`. `main`'s #1004 section landed first and keeps §17; this branch's #1013 section is renumbered **§18**, with the three cross-references (CHANGELOG, `FINDINGS.md`, `pr_body.md`) updated. `cargo test -p discopt-core` → **608 passed** on the merged tree (604 here + 4 from #1017's tests); `cargo fmt --check` and `clippy` clean.

## The merge invalidated this PR's headline claim, and I am withdrawing it

`main` merged #1019 — the Farkas accumulation-margin fix for #1017, the issue this PR's investigation filed. That fix repairs the certificate that produced the `QPLIB_3814_rlt1` false `infeasible`. So on the merged base:

```
bail=0  status=optimal  obj=0.23839462819531176  maxrun=3352  bails=0
bail=1  status=optimal  obj=0.23839462819531176  maxrun=2048  bails=1
```

**That LP is now `optimal` with the stall bail OFF as well as ON.** The "1 status improvement" this PR claimed is #1017's result, not this mechanism's, and the claim is withdrawn (CLAUDE.md §11). It is corrected in the PR body, the CHANGELOG entry, `performance-plan.md` §18a, and `FINDINGS.md` §6 — not only here.

## Panel re-run on the merged base

100 LPs, 2 reps, `scratchpad/i1013/postmerge_panel.jsonl` (committed):

| gate | pre-merge base | post-#1017 base |
|---|---|---|
| unchanged cells (status *and* iterations) | 98 / 100 | **99 / 100** |
| status regressions | 0 | **0** |
| status improvements | 1 (`QPLIB_3814_rlt1`) | **0** |
| objective drift, optimal/optimal | 0.00e+00 (n=96) | **0.00e+00** (n=97) |
| cells where the bail fires | 3 | 3 |

Cells where it fires: `QPLIB_3814_rlt1` 1.78 s → 1.33 s (**1.34x**), `tspn10_rlt1` 16.18 s → 14.36 s (**1.13x**), `QPLIB_3815_rlt1` neutral (`iter_limit` both ways).

Tree panel re-run on the same base: objective, bound and node count **identical on all 16 instances**, 96 assertions, 0 issues.

## What this leaves, stated plainly

The mechanism is a **tail guard, not a throughput change**. It is inert on 97 of 100 panel LPs (bit-identical iteration counts), and on the three where it fires it converts a stalled warm grind into a cold solve worth 1.34x, 1.13x and nothing. That is the same shape of value as the F2 stall guard it sits beside. With the correctness win now attributable to #1017, "broadly net-positive" is **not** an accurate description of it, and I have stopped describing it that way.

That is a judgment call for you, not for me: default-ON is defensible on the tail-guard precedent (0 regressions, 99/100 identical, sound by construction since the cold path self-verifies), but if you want the §5 net-positive bar read strictly, the honest alternative is to flip the default to OFF and keep `DISCOPT_LP_DUAL_STALL_BAIL=1` as the opt-in pending a broader panel. Say which and I will make the change.

---

PR #1042, opened 2026-08-16
fix(duals): #1037 report multipliers of the declared model, not the presolved one

### LINKED ISSUE #1037: Reported duals are the presolve-tightened problem's, not the user's model's (79 slow-tier failures)
## Summary

`SolveResult.constraint_duals` and `SolveResult.bound_duals_*` are the duals of the
**presolve-tightened** problem, reported against the **user's declared** model. When
FBBT tightens a variable bound and the optimum lands on the tightened bound, the
reported shadow price is numerically wrong for the model the user wrote, and a bound
multiplier is attached to a bound the user never declared.

This is a sensitivity-reporting defect. It does **not** affect primal solutions,
objective values, feasibility, or the certified dual bound — those are all correct on
every instance below. What is wrong is the KKT certificate handed back to the caller.

Found while running the `slow`/`correctness` tiers for the v0.8.0 release. It accounts
for **79 failing tests in `python/tests/test_minlptests.py`**, all with one signature.
These tiers are excluded by the default `addopts` and CI's slow job is on-demand, which
is why this has been invisible (see also #1034).

## Minimal reproduction

```
min -x   s.t.  x^2 + y^2 <= 1,   x, y in [-2, 2]
```

Optimum `x=1, y=0`. Both declared bounds are strictly inactive, so the KKT dual of the
*user's model* is unique: `mu = 0.5`, `lambda_lb = lambda_ub = 0`.

discopt reports:

```
objective        : -0.9999999950   (correct)
constraint_duals : {'c0': 0.20394929}   <-- true value is 0.5
```

## Root cause, measured

A trace on `discopt.solvers.nlp_pounce.solve_nlp` shows the box actually handed to the
NLP:

```
declared model box  : lb=[-2,-2]  ub=[2,2]
box handed to pounce: lb=[-1,-1]  ub=[1,1]      <-- FBBT: x^2 <= 1  =>  x in [-1,1]
x0=[0,0]  ->  x=[1,0]
mult_g = [0.20394929]
zU     = [5.92101421e-01, 2.5e-09]
check 2*mu + zU[x] = 1.000000004                <-- valid KKT for the TIGHTENED box
```

Everything upstream is behaving correctly:

* **FBBT is right.** `x^2 + y^2 <= 1` with `y^2 >= 0` gives `x in [-1,1]`. That is a
  sound tightening.
* **POUNCE is right.** On the tightened box `x*=1` sits *on* the upper bound, so the
  bound gradient `(1,0)` is parallel to the constraint gradient `(2,0)` and the dual
  split is degenerate — every `(mu, zU)` with `2*mu + zU = 1`, both `>= 0`, is KKT-valid
  there. POUNCE returns one such split.

  Verified independently: driving POUNCE directly (no discopt) on the same problem with
  the declared `[-2,2]` box returns `mult_g = 0.4999999979`, `mult_x_U ~ 1e-9`,
  stationarity residual 7.9e-13. POUNCE is not the source.
* **discopt's unpacking is right.** `_unpack_constraint_duals` passes the vector through
  faithfully.

The defect is the **missing step**: nothing maps the tightened-box duals back to the
model the user wrote. On the declared box, `x=1 < ub=2`, so complementary slackness
requires `zU[x] = 0`, which forces `mu = 0.5`.

## Why the split is not merely cosmetic

Because the tightened-box dual is degenerate, the value reported is arbitrary within a
family — it is not a small perturbation of the right answer. Measured over a scaled
family `min -x s.t. S*x^2 + S*y^2 <= S`, the ratio `mu_reported / mu_true` wanders with
no fixed factor:

| case | mu_reported | mu_true | ratio |
|---|---|---|---|
| obj x 1 | 0.20394929 | 0.5 | 0.408 |
| obj x 2 | 0.49989743 | 1.0 | 0.500 |
| obj x 10 | 0.92389200 | 5.0 | 0.185 |
| con x 2 | 0.09490032 | 0.25 | 0.380 |
| con x 10 | 0.02157076 | 0.05 | 0.431 |

Linear constraints are exact (`ratio = 1.000` in every linear control), because FBBT
tightening that makes a *linear* row's bound active does not produce this degeneracy in
the same way. The defect is specific to bounds inferred from nonlinear rows.

## How it surfaces

`discopt.validation.examiner` already catches it. Its `primal_cs (solver duals)` check
fails while its **`primal_cs (recovered)` and `stationarity (recovered)` checks pass** —
i.e. discopt's own least-squares dual-recovery path (`solver.py` ~15388) computes the
right answer for the declared model. Example report:

```
[PASS] primal_var_feas / primal_con_feas / obj_consistency
[PASS] stationarity (solver duals)
[FAIL] primal_cs (solver duals): max=5.921e-01  -- lambda * |x - bound|
[PASS] stationarity (recovered), primal_cs (recovered), dual_cs (recovered)
[FAIL] dual_consistency (solver vs recovered): lambda_ub[x]=5.921e-01, mu=2.961e-01
```

Note `stationarity (solver duals)` **passes** — the returned duals are stationary and
dual-feasible. Only complementary slackness against the declared bounds exposes them.
Any check that omits CS will not see this.

## Suggested fix

Fix the class, not the instance. Two candidate shapes:

1. **Recover against the declared box.** The recovery machinery already exists and
   already produces the correct duals (the examiner proves this on all 79 instances).
   When presolve tightened a bound that is active at the reported solution but strictly
   inactive in the declared model, recover the duals w.r.t. the declared bounds instead
   of passing the node duals through.
2. **Refuse loudly.** Return `None` for the affected entries with a warning naming the
   variable and the tightening, rather than reporting a number that is wrong.

(1) is preferable — it returns the information the user asked for, and it is
demonstrably achievable. (2) is the acceptable fallback per the "refuse rather than
approximate" rule, and is strictly better than the current behavior.

Either way the fix needs a regression test asserting `mu == 0.5` on the reproduction
above, plus the examiner's CS check green across `test_minlptests.py`.

## Scope / provenance

* **Not a v0.8.0 regression.** `discopt/validation/examiner.py` and the `assert_examined`
  call in `test_minlptests.py` both date to #55; `git log -S multipliers` shows no change
  to `solver.py` or `python/discopt/solvers/` since 2026-08-01.
* Environment: `pounce-solver` 0.10.0 (editable checkout). The checkout sat on branch
  `feat/588-quadratic-structure`, whose diff vs `origin/main` touches only
  `crates/pounce-cli`, dev notes, and scripts — none of the library crates behind the
  Python bindings — so it is not implicated.
* Probes: `scratchpad/probe_bounds.py` (box trace), `probe_pounce_raw.py` (standalone
  POUNCE control), `probe_scale2.py` (scaling family). Each counts its executed
  comparisons and exits non-zero at zero, per CLAUDE.md §6.

jkitchin:
## The defect

`SolveResult.constraint_duals` / `bound_duals_*` were the duals of the box **presolve actually solved**, handed back against the model **the user wrote**. When a presolve-derived bound is active at the returned point, the backend hangs a multiplier on it — but that bound is not a constraint of the user's model, so neither is its multiplier.

```
min -x  s.t.  x^2 + y^2 <= 1,  x, y in [-2, 2]
```

FBBT derives `x in [-1,1]`; at `x*=1` the derived bound's gradient is parallel to the row's, so the split `2*mu + z_U = 1` is degenerate on the *tightened* box and the backend returned `mu=0.204, z_U=0.592`. On the **declared** box `x*=1` is strictly interior to `[-2,2]`, so complementary slackness forces `z_U=0` and `mu=0.5` **uniquely**. This is not a perturbation — which degenerate split comes back is arbitrary, and it varied with row scale (`mu/mu_true` measured at 0.408 and 0.431).

## The fix

**The trigger is the defect, not its cause.** A multiplier is admissible for the declared model only if it satisfies CS against the **declared** bounds, `|lam_j| * |x_j - bound_j| <= 1e-7` — the same `examiner.PRIMAL_CS_TOL` check that exposed this.

That generality earned its keep immediately: it catches a second cause the issue did not name — the **±1e20 infinity sentinel** standing in for "no bound at all". An unbounded variable has no bound to price, so its multiplier is exactly zero; the IPM backends leave residue there (measured `lam=3.8e-05` on `nlp_cvx_102_010`), and against a 1e20 "bound" that residue is a CS violation of 3.8e+15. A first attempt that triggered on a presolve-diff instead fixed only 10 of 79 tests; the CS trigger fixes 75. No sentinel special-case is needed — a point 1e20 away from a bound is not at it, so the refit's active set already excludes it.

When the reported multipliers are inadmissible they are **refitted** against the declared active set, at the already-reported point; when they are admissible they pass through untouched (one O(n) pass). A refit rather than a re-solve on the widened box, because re-solving a nonconvex problem can wander to a different point and leave the duals inconsistent with the reported primal.

If the refit cannot reach stationarity the duals are **withheld with a warning naming the variable and the bound**, rather than reported wrong (CLAUDE.md §3) — a dual that fails the user's own KKT system is worse than no dual, because nothing downstream can tell it is wrong.

The refit is the examiner's least-squares active-set recovery, which the issue established already computes the right answer on all 79 instances. Rather than duplicate it, it moved to `discopt/_dual_recovery.py` and both callers use it, so solver and examiner cannot drift apart. The examiner keeps its `CheckResult` reporting and is otherwise unchanged.

Wired into `_solve_continuous` (nonlinear tightening) and `_solve_nlp_bb` (root FBBT). Integer columns are excluded throughout: their bound multipliers price the fixing of the integer, not bound activity in the declared model.

### Second commit: row activity is now relative to row scale

Writing the regression tests exposed a real weakness in the recovery. The active-set test compared `|body - rhs|` against an **absolute** 1e-6, which measures a row in whatever units it happens to be written in. Multiplying a constraint through by 1000 changes nothing about the problem, but it scales the NLP's own convergence slack up with it, so a binding row read as inactive, the active set came out empty, and the refit could not reach stationarity (residual 1.0 = `|grad f|`) — duals of a genuinely binding row withheld. Row activity now uses the same `row_scale` the examiner already applies to scaled primal feasibility. Bound activity stays absolute on purpose (see the commit message).

## Verification

**Baseline measured both ways**, not assumed:

| | `test_minlptests.py` |
|---|---|
| `main` | **79 failed**, 46 passed |
| this branch | **4 failed**, 121 passed |
| new failures introduced | **0** |
| fixed | **75** |

The entire `primal_cs (solver duals)` bucket — the #1037 signature — is gone. The panel is **unchanged by the row-scale commit** (4 failed / 121 passed, 0 new), which matters because that code is shared with the examiner.

The 4 that remain — `nlp_001_010`, `nlp_008_010`, `nlp_008_011`, `nlp_mi_006_010` — are all in the baseline's 79, i.e. **pre-existing**, and are a different defect: all report `solver_duals_used=False` and fail the examiner's *own* `stationarity (recovered)` check by 1.3e-06..2.0e-04, at a point the NLP did not converge tightly enough on (POUNCE logged "refusing a termination certificate masked by an extreme objective scale" on these). That is primal convergence accuracy, not dual reporting; a follow-up issue will track it.

Also run:

- `python/tests/test_solver_duals.py` + `test_examiner.py` — **34 passed**
- `pytest -m smoke` — **1028 passed**, 1 skipped, 2 xpassed
- No Rust touched, so `cargo test -p discopt-core` was not required.

### Regression tests

18 tests in `python/tests/test_solver_duals.py`, of which **7 fail on the pre-fix tree and pass on this one** (verified by restoring `main`'s files in place and asserting the fix marker absent, per CLAUDE.md §8):

- the issue's reproduction verbatim — `mu == 0.5`, `lambda_ub[x] == 0`;
- the same model swept over **six row scales**, asserting `mu == 1/(2*scale)`;
- an examiner end-to-end assertion that the reported duals pass the `primal_cs` check that exposed the issue;
- `_assert_cs_against_declared_box` — the general contract rather than a per-instance expectation, applied in every case and returning a count so the caller can assert it actually checked something;
- two **controls** that must pass before *and* after: a genuinely active declared bound keeps its multiplier (`x*=2`, `z_U=1` — the refit does not destroy real bound multipliers), and an unbounded variable carries none.

One note worth keeping: the first draft of the scaled family wrote the row as `s*(x**2 + y**2) <= s*r**2` and **passed on the pre-fix tree**. Grouped that way the multiply sits above the sum, FBBT cannot derive `x in [-r, r]` through it, no tightening happens and the defect never arises — the test was silently not a test of #1037. The scale is now distributed over the terms, and the reason is recorded in the builder's docstring so it is not reintroduced.

Closes #1037

### DIFF
--- a/python/discopt/_dual_recovery.py
+++ b/python/discopt/_dual_recovery.py
@@ -1,0 +1,273 @@
+"""Active-set multiplier recovery (least squares), shared by the examiner and
+the solver's dual reporting.
+
+Given a point ``x``, a box, and the constraint rows, this builds the
+first-order stationarity system over the ACTIVE set
+
+    ∇f + Σ_act μ_i ∇g_i − λ_lb + λ_ub = 0,    μ ≥ 0 (inequalities), λ ≥ 0
+
+and solves it with :func:`scipy.optimize.lsq_linear` under those sign bounds.
+Integer columns are dropped: fixing the discrete variables and running
+continuous KKT is exactly what omitting their columns does (the GAMS Examiner
+"fix and re-check" convention).
+
+Why this lives in its own module rather than inside the examiner
+----------------------------------------------------------------
+Two callers need the same arithmetic, and they must not drift apart:
+
+* :mod:`discopt.validation.examiner` wraps it in ``CheckResult`` reporting;
+* :mod:`discopt.solver` uses it to report duals against the box the USER
+  DECLARED when presolve solved a tightened one (#1037). A bound the
+  tightening introduced is not a bound of the user's model, so a multiplier
+  sitting on it is not a multiplier of the user's model; the fit has to be
+  redone against the declared active set, and this is the fit that does it.
+
+The module is deliberately dependency-light (numpy + scipy only) so both the
+validation layer and the solver can import it without a cycle.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Optional
+
+import numpy as np
+from scipy.optimize import lsq_linear
+
+# Distance within which a point counts as sitting ON a bound / row. Matches
+# ``examiner.ACTIVE_TOL``; kept here so the solver does not have to import the
+# validation layer just to name a tolerance.
+ACTIVE_TOL = 1e-6
+
+
+def row_metadata(evaluator):
+    """``(sense_arr, rhs_arr, row_labels)`` for the evaluator's flat row order.
+
+    Vector-bodied constraints contribute ``sz`` rows that share the source
+    constraint's sense and rhs.
+    """
+    senses: list[str] = []
+    rhss: list[float] = []
+    labels: list[str] = []
+    for c, sz in zip(evaluator._source_constraints, evaluator._constraint_flat_sizes):
+        sz = int(sz)
+        senses.extend([c.sense] * sz)
+        rhss.extend([float(c.rhs)] * sz)
+        cname = getattr(c, "name", None) or repr(c.body)[:40]
+        if sz > 1:
+            labels.extend([f"{cname}[{i}]" for i in range(sz)])
+        else:
+            labels.append(str(cname))
+    return np.asarray(senses), np.asarray(rhss, dtype=float), labels
+
+
+@dataclass
+class DualRecovery:
+    """Outcome of an active-set multiplier fit.
+
+    ``ok`` is False only when ``lsq_linear`` itself raised; a large residual is
+    reported, not hidden, and is the caller's decision to act on (CLAUDE.md §7 —
+    an instrument must not swallow the thing it is measuring).
+    """
+
+    ok: bool
+    detail: str = ""
+    # Index arrays, all into the FLAT variable / row vectors.
+    cont_idx: np.ndarray = None  # type: ignore[assignment]
+    row_select: np.ndarray = None  # type: ignore[assignment]
+    lb_active: np.ndarray = None  # type: ignore[assignment]  # indices into cont_idx
+    ub_active: np.ndarray = None  # type: ignore[assignment]  # indices into cont_idx
+    # Fitted multipliers, in active-set order.
+    mu_act: np.ndarray = None  # type: ignore[assignment]
+    lam_lb_act: np.ndarray = None  # type: ignore[assignment]
+    lam_ub_act: np.ndarray = None  # type: ignore[assignment]
+    # Full-length vectors aligned with the evaluator's row order and the flat
+    # variable order; zero off the active set.
+    mu_full: Optional[np.ndarray] = None
+    lam_lb_full: Optional[np.ndarray] = None
+    lam_ub_full: Optional[np.ndarray] = None
+    # Stationarity residual of the fit, over continuous columns.
+    stat_resid: np.ndarray = None  # type: ignore[assignment]
+    residual_max: float = float("inf")
+    residual_norm: float = float("inf")
+    # True when the active set was empty, so the "fit" is just ‖∇f‖ at x.
+    empty_active_set: bool = False
+
+    @property
+    def n_active_cons(self) -> int:
+        return 0 if self.row_select is None else int(self.row_select.size)
+
+    @property
+    def n_active_bounds(self) -> int:
+        n_lb = 0 if self.lb_active is None else int(self.lb_active.size)
+        n_ub = 0 if self.ub_active is None else int(self.ub_active.size)
+        return n_lb + n_ub
+
+
+def recover_multipliers(
+    *,
+    grad: np.ndarray,
+    jac: np.ndarray,
+    body: np.ndarray,
+    sense_arr: np.ndarray,
+    rhs_arr: np.ndarray,
+    x_flat: np.ndarray,
+    lb: np.ndarray,
+    ub: np.ndarray,
+    is_continuous: np.ndarray,
+    active_tol: float = ACTIVE_TOL,
+) -> DualRecovery:
+    """Fit KKT multipliers at ``x_flat`` against the box ``(lb, ub)``.
+
+    ``lb``/``ub`` decide which bounds are active, and therefore which
+    multipliers are allowed to be nonzero — this is the whole reason the caller
+    gets to choose the box (#1037). Pass the DECLARED bounds to obtain duals of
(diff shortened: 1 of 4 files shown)


jkitchin:
The follow-up for the 4 remaining `test_minlptests.py` failures is now filed as #1043 — `solve()` returning `optimal` on points POUNCE explicitly refused to certify (unscaled KKT error 0.34 on `nlp_001_010`). Pre-existing, tracked separately, and left open by this PR.

PR #1056, opened 2026-08-16
fix(release): #1055 ship abi3 wheels and honor the declared 3.10 floor

### LINKED ISSUE #1055: Release wheels: macOS and Windows ship cp312 only, but requires-python is >=3.10
Surfaced verifying the v0.8.0 upload. Not a v0.8.0 regression — the v0.7.0 file
list has the identical shape — but `pyproject.toml` declares
`requires-python = ">=3.10"` and we only honour that on Linux.

What is on PyPI for 0.8.0 (12 files):

| Platform | Python versions with a wheel |
|---|---|
| Linux x86_64 / aarch64 | 3.10, 3.11, 3.12, 3.13 |
| macOS x86_64 / arm64   | **3.12 only** |
| Windows x64            | **3.12 only** |

Cause is in `.github/workflows/release.yml`. The `linux` job passes an explicit
interpreter list:

    args: --release --out dist -i python3.10 python3.11 python3.12 python3.13

The `macos` and `windows` jobs do not:

    args: --release --out dist

so maturin builds against the runner's `setup-python` version alone, which both
jobs pin to `"3.12"`. The wheels are version-specific (`cp312-cp312-...`), not
`abi3`, so nothing else can satisfy a 3.10/3.11/3.13 install.

Effect: `pip install discopt` on macOS or Windows under Python 3.10, 3.11, or 3.13
falls through to the sdist and tries to compile the extension, which needs a Rust
toolchain the user almost certainly does not have. It fails, and it fails in a way
that reads as a broken package rather than a missing wheel.

Two ways to fix, pick one:

1. **Mirror the Linux job** — give macOS/Windows the same `-i python3.10 …
   python3.13` list. Most direct; multiplies those build jobs by four.
2. **Build `abi3` wheels** — a single `abi3-py310` wheel per platform covers
   3.10+ and shrinks the matrix instead of growing it. Requires the `abi3-py310`
   feature on the pyo3 dependency in `crates/discopt-python/Cargo.toml` and a
   check that nothing in the bindings uses a non-limited API. This is the better
   end state if the bindings allow it.

Whichever is chosen, the release checklist should verify wheel *coverage* rather
than just "PyPI upload succeeded" — `RELEASE.md` §10 currently says "wheels build
for Linux (x86_64 + aarch64), macOS (x86_64 + aarch64), and Windows (x64)", which
this release satisfied while still leaving eight of twelve platform/version
combinations uncovered. Suggested wording: assert the published file list contains
a wheel for every supported Python on every supported platform, and fail the
checklist item if not.

Verify the current state with:

    curl -s -H "User-Agent: <you>" https://pypi.org/pypi/discopt/0.8.0/json \
      | python -c "import json,sys; [print(u['filename']) for u in json.load(sys.stdin)['urls']]"

jkitchin:
Closes #1055.

## The bug

`pyproject.toml` declares `requires-python = ">=3.10"`. v0.8.0 published:

| Platform | Wheels published | Should cover |
|---|---|---|
| Linux x86_64 / aarch64 | 3.10, 3.11, 3.12, 3.13 | 4 |
| macOS x86_64 / arm64 | **3.12 only** | 4 |
| Windows x64 | **3.12 only** | 2 |

Off Linux that is 4 of 12 platform/version combinations. The release workflow's
macOS and Windows jobs passed no `-i` interpreter list, so maturin built against
the runner's `setup-python` version alone. **Every job was green** — nothing
inspected what got built, only that building and uploading succeeded. A macOS
user on 3.11 falls through to the sdist and hits a Rust toolchain error, which
reads as a broken package rather than a missing wheel. v0.7.0's file list has the
identical shape, so this has been present since at least then.

## The fix: abi3, not a bigger matrix

Extending the interpreter list to four versions × five platforms is the obvious
fix and the wrong one — it quadruples build time and leaves the same trap for the
next Python release. `pyo3` gains `abi3-py310`, so **one wheel per platform**
covers every supported interpreter. Artifacts drop 12 → 6, the matrix shrinks,
and adding Python 3.14 support becomes a no-op.

The stable ABI routes the Python↔Rust boundary through the limited API, and this
solver crosses that boundary per B&B node, so I measured rather than assumed.

**Functional equivalence** — one `cp310-abi3` wheel, four interpreters:

| Python | Objective | Status |
|---|---|---|
| 3.10.16 | −9.999999990908268 | optimal |
| 3.11 | −9.999999990908268 | optimal |
| 3.12.11 | −9.999999990908268 | optimal |
| 3.13 | −9.999999990908268 | optimal |

**Cost** — interleaved A/B, 5 reps × 6 light instances + 3 reps × `clay0303hfsg`,
66 raw rows, load average 4.46 → 6.22 across the run (CLAUDE.md §9):

| Instance | control (s) | abi3 (s) | Δ |
|---|---|---|---|
| `alan` | 0.023 ± 0.001 | 0.023 ± 0.001 | −0.4% |
| `ex1221` | — | — | +1.4% |
| `ex1222` | — | — | −4.0% |
| `ex1224` | — | — | +2.5% |
| `ex1225` | — | — | +0.4% |
| `ex1226` | — | — | −0.5% |
| `clay0303hfsg` | 17.584 ± 0.136 | 17.765 ± 0.239 | +1.0% |
| **total wall** | **61.35** | **61.97** | **+1.0%** |

Signs go both ways and every per-instance delta is inside its own spread. Per
CLAUDE.md verification regime 1 (bound-neutral), the binding check is not wall
but exactness: **0 node-count mismatches and 0 objective mismatches** across all
7 instances.

## The guard

The gap survived a release because nothing checked coverage.
`.github/scripts/check_wheel_coverage.py` runs on the collected artifacts
*before* the PyPI upload and fails if a platform is missing, if a wheel is not
`abi3`, or if the `abi3` floor has drifted from `requires-python` — that last one
is the silent trap, since bumping `requires-python` without bumping
`abi3-pyXY` in `Cargo.toml` publishes wheels claiming support they were never
built against.

Tested both directions: it **rejects** the real v0.8.0 file list (11 problems, 18
checks executed) and **accepts** this one (12 checks, exit 0). It prints an
executed-check count and fails on zero (CLAUDE.md §6).

## Two things 3.10 never actually supported

Making 3.10 installable exposed code that had never run there. Both survived
because **all seven CI jobs run 3.12**.

1. **`discopt install-skills` crashed on 3.10.** `discopt/skills/__init__.py`
   imported `Traversable` from `importlib.resources.abc`, which is 3.11+, so
   `import discopt.skills` raised `ModuleNotFoundError`. This is shipped code, not
   test scaffolding. The name is annotation-only and the module already has
   `from __future__ import annotations`, so it moves under `if TYPE_CHECKING:`
   with a `sys.version_info` split (3.10 reaches the same class as
   `importlib.abc.Traversable`).

2. **The correctness tier could not even be collected on 3.10.**
   `python/tests/_optima.py` — the reference-optima oracle shared by every
   soundness suite — did a bare `import tomllib` (3.11+), making six modules
   collection errors. It now falls back to `tomli` (matching
   `discopt/profiles.py`) and **raises loudly** if neither parser is present
   rather than degrading to an empty registry, which would silently turn every
   soundness assertion into a no-op (CLAUDE.md §6). `tomli` is declared in the
   `dev` extra under `python_version < "3.11"`.

## Regression test

`python/tests/test_requires_python_floor.py` fixes the *class* rather than those
two files (CLAUDE.md §2). It AST-scans `python/discopt/` and `python/tests/` for
module-scope imports of stdlib modules or names newer than the `requires-python`
floor, ignoring anything guarded by `try`/`except`, `TYPE_CHECKING`, or a
`sys.version_info` branch. It runs on **any** interpreter, so catching the next
one does not require adding a 3.10 CI job.

Fails before / passes after, on the two real violations:

```
python/discopt/skills/__init__.py:23: import importlib.resources.abc needs Python 3.11, but requires-python floor is 3.10
python/tests/_optima.py:25: import tomllib needs Python 3.11, but requires-python floor is 3.10
1 failed, 1 passed
```
```
2 passed
```

A companion test feeds the scanner a known violation plus three correctly-guarded
imports and asserts it finds exactly one — so a typo in the version table or a
broken guard check cannot make the main test pass vacuously. The main test also
asserts it examined >1000 import targets (it sees 17,588) rather than silently
scanning nothing.

## Verification

- `pytest python/tests/ -m smoke` on **Python 3.10.16**, against the abi3 wheel:
  **1026 passed, 16 skipped, 2 xpassed** (113 s). Before this PR: **6 collection
  errors**, suite could not run.
- `pytest python/tests/ -m smoke` on **Python 3.12.11**, same wheel: **1026
  passed, 16 skipped, 2 xpassed** (120 s) — identical counts on both
  interpreters.
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py`: see comment
  below.
- `cargo test -p discopt-core`: **618 passed, 0 failed**.
- `ruff check` / `ruff format --check` / `mypy`: clean (also via pre-commit on the
  commit).
- Provenance asserted before every measurement (CLAUDE.md §8): each run checks
  `sys.version_info`, that `discopt.__file__` resolves inside the target venv, and
  that `discopt._rust.__file__` ends in `.abi3.so`.

## RELEASE.md

§10 now says to expect **6 files** (five `cp3NN-abi3` wheels + sdist), to treat
green build jobs as no evidence of coverage, and to run the clean-venv install
check on the **oldest** supported Python — an ABI gap shows up there first.

### DIFF
--- a/.github/scripts/check_wheel_coverage.py
+++ b/.github/scripts/check_wheel_coverage.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python3
+"""Assert the built distribution actually covers every platform we promise.
+
+Run against the collected ``dist/`` directory *before* uploading to PyPI.
+
+v0.8.0 shipped with ``requires-python = ">=3.10"`` while macOS and Windows had
+wheels for 3.12 only (#1055): the release workflow's macOS/Windows jobs never
+passed an interpreter list, so maturin built against the runner's ``setup-python``
+version alone. Every job was green and the upload succeeded -- the gap was
+invisible because nothing checked *what* got published, only that publishing
+worked. A Mac user on 3.11 fell through to the sdist and hit a Rust toolchain
+error that reads as a broken package.
+
+The fix was ``abi3``: one wheel per platform, valid for the whole supported
+range. This script is the guard that keeps it that way. It fails the release if
+a platform is missing, if a wheel is not ``abi3``, or if the ``abi3`` floor has
+drifted away from ``requires-python`` -- that last one is the silent failure the
+next person would otherwise hit, because bumping ``requires-python`` without
+bumping the ``abi3-pyXY`` feature in ``Cargo.toml`` publishes wheels that claim
+support for versions they were never built against.
+"""
+
+from __future__ import annotations
+
+import re
+import sys
+import tomllib
+from pathlib import Path
+
+# Platform tag -> a human name for the error message. Every one of these must be
+# represented by at least one wheel in dist/.
+REQUIRED_PLATFORMS: dict[str, str] = {
+    r"manylinux.*_x86_64": "Linux x86_64",
+    r"manylinux.*_aarch64": "Linux aarch64",
+    r"macosx_.*_x86_64": "macOS x86_64",
+    r"macosx_.*_arm64": "macOS arm64",
+    r"win_amd64": "Windows x64",
+}
+
+
+def main() -> int:
+    repo = Path(__file__).resolve().parents[2]
+    dist = Path(sys.argv[1] if len(sys.argv) > 1 else "dist")
+
+    with (repo / "pyproject.toml").open("rb") as fh:
+        pyproject = tomllib.load(fh)
+    requires_python = pyproject["project"]["requires-python"]
+    m = re.fullmatch(r">=\s*3\.(\d+)", requires_python.strip())
+    if not m:
+        print(
+            f"FAIL: cannot parse requires-python {requires_python!r}. This script "
+            "understands '>=3.N' only; teach it the new form rather than deleting "
+            "the check.",
+            file=sys.stderr,
+        )
+        return 1
+    floor_minor = int(m.group(1))
+    expected_abi_tag = f"cp3{floor_minor}-abi3"
+
+    wheels = sorted(p.name for p in dist.glob("*.whl"))
+    sdists = sorted(p.name for p in dist.glob("*.tar.gz"))
+
+    failures: list[str] = []
+    checks = 0
+
+    # 1. Something was actually built. Without this, an empty dist/ would sail
+    #    through every loop below and report a clean pass.
+    checks += 1
+    if not wheels:
+        failures.append(f"no wheels found in {dist}/")
+
+    # 2. Exactly one sdist.
+    checks += 1
+    if len(sdists) != 1:
+        failures.append(f"expected exactly 1 sdist, found {len(sdists)}: {sdists}")
+
+    # 3. Every promised platform has a wheel.
+    for pattern, label in REQUIRED_PLATFORMS.items():
+        checks += 1
+        if not any(re.search(pattern, w) for w in wheels):
+            failures.append(f"no wheel for {label} (no filename matches /{pattern}/)")
+
+    # 4. Every wheel is abi3 at exactly the requires-python floor.
+    for w in wheels:
+        checks += 1
+        if expected_abi_tag not in w:
+            failures.append(
+                f"{w} is not tagged {expected_abi_tag}. requires-python is "
+                f"{requires_python}, so the pyo3 'abi3-py3{floor_minor}' feature in "
+                "Cargo.toml must match it."
+            )
+
+    print(f"dist:            {dist}")
+    print(f"requires-python: {requires_python}  ->  expecting {expected_abi_tag}")
+    print(f"sdists:          {len(sdists)}")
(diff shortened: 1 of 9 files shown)


jkitchin:
Adversarial suite (the "see comment below" item in the description):

```
pytest -m slow python/tests/test_adversarial_recent_fixes.py
19 passed, 7 warnings in 150.47s
```

Run on Python 3.12.11 against the abi3 wheel (`discopt/_rust.abi3.so`), same
artifact as the two smoke runs.

PR #1069, opened 2026-08-17
fix(convexity): #1065 stop 0*inf from discarding a valid affine lower bound

### LINKED ISSUE #1065: convexity: _affine_lower_bound discards a valid bound when a zero coefficient meets an unbounded variable (0*inf -> NaN)
`_affine_lower_bound` in `python/discopt/_relax/convexity/patterns.py:318`
computes

```python
# min of c·x = Σ (c>0 ? c·lb : c·ub); guard infinities that actually matter.
contrib = np.where(vec >= 0.0, vec * lo_box, vec * hi_box)
if not np.all(np.isfinite(contrib)):
    return None
```

`np.where` evaluates **both** branches, so `vec * lo_box` is formed for every
entry including those where the coefficient is zero. With a model that declares
unbounded variables, `0.0 * inf` produces `NaN`, which fails the `isfinite`
guard and discards the whole bound:

```
python/discopt/_relax/convexity/patterns.py:318: RuntimeWarning:
  invalid value encountered in multiply
  contrib = np.where(vec >= 0.0, vec * lo_box, vec * hi_box)
```

Observed on MINLPLib `squfl020-150`.

**This is not a soundness bug** — the guard is conservative and returns `None`
("unknown"), never a wrong bound. It is a precision and noise bug:

1. A variable with a **zero** coefficient does not contribute to the affine form
   at all. Its bounds, infinite or not, are irrelevant. Poisoning the result
   because some unrelated variable is unbounded makes `_affine_lower_bound`
   return `None` far more often than it needs to, degrading the positivity and
   convexity detection that depends on it.
2. It bites hardest exactly where it hurts most: the big-M/GDP models in the
   `syn`/`rsyn` and `squfl` families declare `lb=0, ub=inf` on most continuous
   variables (see #1061), so nearly every affine form there contains an
   unbounded variable with a zero coefficient.
3. The `RuntimeWarning` leaks to user output during an ordinary solve.

The docstring already states the intent — "Returns `None` when ... the bound is
not finite (an **unbounded contributing** variable)" — so the code is stricter
than its own contract: a zero-coefficient variable is not contributing.

## Fix sketch

Mask zero coefficients before multiplying, so only genuinely contributing terms
can make the sum infinite; keep the `isfinite` guard for the real case (a
nonzero coefficient on an unbounded endpoint).

## What "done" means

- No `RuntimeWarning` from this line on the corpus.
- A regression test with an affine form carrying a zero coefficient on an
  unbounded variable returns the correct finite bound, plus an anti-vacuity
  control asserting a genuinely unbounded contributing variable still returns
  `None` (CLAUDE.md §6).
- Bound-neutral verification (CLAUDE.md §5, regime 1) unless the improved
  detection changes classifications, in which case regime 2 applies.

jkitchin:
Closes #1065.

`_affine_lower_bound` computed

```python
contrib = np.where(vec >= 0.0, vec * lo_box, vec * hi_box)
if not np.all(np.isfinite(contrib)):
    return None
```

`np.where` evaluates **both** branches in full, so `vec * lo_box` and
`vec * hi_box` are each formed for every entry — including the entries whose
coefficient is zero because the variable does not appear in this affine form at
all. Against a model that declares unbounded variables that is `0.0 * inf`, i.e.
`NaN`.

The fix picks the minimizing endpoint **first** and multiplies only where the
coefficient is nonzero:

```python
endpoint = np.where(vec >= 0.0, lo_box, hi_box)
contrib = np.multiply(vec, endpoint, where=vec != 0.0, out=np.zeros_like(vec))
```

The `isfinite` guard is untouched, so the case it exists for — a **nonzero**
coefficient on an unbounded endpoint — still returns `None`. This aligns the
code with its own docstring, which already scopes the `None` to "an unbounded
**contributing** variable".

## Precision on what the bug did (a correction to the issue text)

The issue says the `NaN` "discards the whole bound". That is true only for a
bystander declared `lb=-inf`: a zero coefficient satisfies `vec >= 0`, so the
*selected* branch is `vec * lo_box`, and the `NaN` lands in `contrib` itself.
For the more common corpus shape — `lb=0, ub=inf`, which is what `syn`/`rsyn`
and `squfl` declare — the `NaN` is formed in the **discarded** `vec * hi_box`
branch, so the returned bound was correct and the damage was the
`RuntimeWarning` leaking into user output. Both are fixed, and both are pinned
by separate tests, but the distinction matters for how much detection this
actually recovers.

## Verification

**Regime.** #1065 sets the regime by measurement: bound-neutral (CLAUDE.md §5
regime 1) if no convexity classification moves, bound-changing if any does.
Measured with `_classify_model_convexity` over the full vendored 61-file corpus
(`python/tests/data/minlplib_nl/`, 66 `.nl` files) in both trees:

```
compared instances: 66   classifications changed: 0
regime 1 (bound-neutral)
```

So no model changes solve path, and the neutrality bar is the byte-identical
one. The same probe records the user-visible symptom directly: the
`RuntimeWarning: invalid value encountered in multiply` at `patterns.py:318`
fires **2 times** over that corpus before the fix and **0 times** after.

**New regression test** — `python/tests/test_1065_affine_lower_bound.py`,
7 tests. Against the unpatched tree 2 of them fail
(`test_zero_coefficient_on_unbounded_lower_endpoint_keeps_the_bound`,
`test_no_runtime_warning_from_the_zero_times_inf_product`); all 7 pass here.
Three of the seven are anti-vacuity controls (CLAUDE.md §6) asserting the guard
is **not** weakened: a genuinely unbounded contributing variable still returns
`None` on the `lo_box` side and on the `hi_box` side, and the negative-coefficient
branch still minimizes at `ub`. Without those, "returns a finite bound" would be
satisfied by a function that ignored infinities altogether.

**Suites.** Rust untouched, so no `cargo test`. `pytest -m smoke python/tests`
and the adversarial suite are reported in a follow-up comment on this PR — they
were deliberately deferred rather than run concurrently with an in-flight
graduation-gate measurement whose wall-clock comparisons my load would
contaminate (CLAUDE.md §9).


jkitchin:
## Owed runs, now completed

The PR body deferred these two rather than run them concurrently with an
in-flight graduation-gate measurement whose wall-clock comparisons my load would
contaminate (CLAUDE.md §9). The gate has finished; both were run on the freed
machine, against `12fb26b8`.

```
pytest -m smoke python/tests
  1037 passed, 1 skipped, 8953 deselected, 2 xpassed, 16 warnings in 133.54s

pytest -m slow python/tests/test_adversarial_recent_fixes.py
  19 passed, 5 warnings in 154.68s
```

**Which code was loaded (CLAUDE.md §8).** The first attempt at these runs was
discarded, not reported: this worktree's tests imported `discopt` from a
*different* worktree (`wt1061`), because an editable install still pointed there
and `cd` does not override it. It surfaced only as an unrelated-looking
`ImportError: cannot import name '_convex_minlp_auto_route' from
'discopt.solver' (/.../wt1061/python/discopt/solver.py)` — had that one module
not existed, the suite would have gone green against the wrong tree and been
reported here as a result for this PR. This is the §8 incident verbatim.

The runs above were relaunched under an explicit `PYTHONPATH`, with a
precondition that asserts both the path and a marker string unique to this
change and exits non-zero *before* pytest starts:

```
CODE-UNDER-TEST VERIFIED: /Users/jkitchin/projects/discopt/wt1065/python/discopt/__init__.py
  patterns.__file__ = .../wt1065/python/discopt/_relax/convexity/patterns.py
  marker "Pick the minimizing endpoint FIRST" present = True
```

That line is the first line of the smoke log, so the run is auditable rather
than merely asserted.

## CI

All checks green on `12fb26b8` (`mergeable=MERGEABLE`, `mergeState=CLEAN`):
Detect code changes, Lint + typecheck, Rust, Python fast, Python correctness
(fast subset), Python AMP coverage, Python claim-boundary (serial).

The claim-boundary job passing on ubuntu is the one that carried real risk: the
regenerated baseline was produced on darwin, and the gate compares root LP
values at `1e-6` relative with at most 3 unreproduced. It reproduced.

## Open question for the owner: this is bound-changing and is NOT behind a flag

Flagging this for an explicit decision rather than letting it merge silently.

CLAUDE.md §5 puts *bound-changing* work behind a default-off flag until a
corpus-wide differential panel passes. This change is bound-changing — it is
also a **defect fix**: `0 * inf -> NaN` was discarding an affine lower bound the
surrounding code already intended to compute, so the "legacy path" a flag would
preserve as the default is the broken one. My judgment was that shipping it
default-on with a fully disclosed footprint beats defaulting to known-wrong
bounds, but that is a call worth making deliberately.

The measured footprint is two instances, attributed three ways against a
clean-`origin/main` control (`TOTAL_COMPARISONS_EXECUTED=132`):

| instance | rows | cols | root LP |
|---|---|---|---|
| nvs05 | 163->167 | 54->56 | `-323.0511` -> `+2.5283` (optimum 5.4709) |
| contvar | 2684->2700 | 1002->1010 | `optimal 162297.316` -> `uncertified` |

`contvar` losing its certified root bound is a real loss and is stated as one.
It occurs only on the **declared** box the baseline generator probes, which
carries 144 non-finite bound entries, so the newly lifted columns are unbounded
and no Neumaier-Shcherbina safe bound can be formed. On the FBBT-tightened box
that every actual solve uses (8 non-finite entries) `contvar` still certifies
and improves: 171244.81 -> 184203.56, against its `=best=` 809149.83.

If you would rather this went behind a default-off flag anyway, say so before it
merges and I will add one.

Issue #266, opened 2026-06-21
pooling_foulds3stp: RecursionError in convex fast-path detection -> status=error (crash)


jkitchin:
## Summary

`pooling_foulds3stp` (MINLPLib, 832 vars / 1091 cons / 1024 nonlinear cons) **crashes** during the solve: discopt returns `status=error` with `objective=0.0` and `node_count=0`. The underlying exception is a `RecursionError` ("maximum recursion depth exceeded") in convex fast-path detection.

Surfaced while benchmarking discopt against SCIP on the MINLPLib "short" tier (problems SCIP solves correctly in <5s). SCIP solves this one in 4.96s (`optimal`, obj `-8.0`).

## Reproduction

```bash
JAX_ENABLE_X64=1 python -c "
import logging; logging.basicConfig(level=logging.DEBUG)
import discopt.modeling as dm
r = dm.from_nl('pooling_foulds3stp.nl').solve(time_limit=5, gap_tolerance=1e-4)
print(r.status, r.objective)
"
```

The DEBUG log shows the swallowed exception:

```
DEBUG:discopt.solver:Convex fast path detection failed: maximum recursion depth exceeded
```

and the result comes back:

```
status = error
objective = 0.0
node_count = 0
jax_time = 3.06s   python_time = 2.26s
```

## Observed vs. expected

| | status | objective |
|---|---|---|
| discopt | `error` | `0.0` |
| SCIP (4.96s) | `optimal` | `-8.0` |
| MINLPLib `.solu` | `=opt=` | `-8.0` |

Expected: a normal solve (or at worst a clean `timelimit`), not an internal `RecursionError`.

## Likely root cause

Convex fast-path detection recurses without a depth bound and blows the Python recursion limit on this model's expression graph (832 vars, 1024 nonlinear constraints — a large pooling network). Two distinct problems:

1. **The crash itself** — fast-path detection should be iterative or depth-bounded so a large/deep expression graph can't trigger `RecursionError`.
2. **Observability** — the exception is swallowed into `status=error` with no message recorded on the `SolveResult` (`_explanation=None`), so callers can't see *why* it failed. The error text should be surfaced (e.g. on the result) rather than only visible via DEBUG logging.

## Environment

- discopt `0.4.1.dev0`, JAX `0.10.0`, `JAX_ENABLE_X64=1`
- macOS (Darwin 25.5.0)

PR #1070, opened 2026-08-18
fix(#1063): keep JAX off the OA path and de-recurse the .nl lowering + convexity classifier

### LINKED ISSUE #1063: mip-nlp: a cold XLA compile (3m0.5s) blows the time limit on the OA path
The `solver="mip-nlp"` path pulls JAX in and pays a cold XLA compile that is not
counted against, and vastly exceeds, the user's `time_limit`. On MINLPLib
`squfl015-060` with `time_limit=60`:

```
[Compiling module jit_concat_constraints for CPU] Very slow compile? ...
slow_operation_alarm.cc:140] The operation took 3m0.498402s

squfl015-060  oa  status=unknown  wall=199.7s      (time_limit was 60 s)
```

A single XLA compilation took **3 m 0.5 s** inside a 60 s budget. The run
returned at 199.7 s — over 3x the limit — having reached `status=unknown`.

Confirmed the path dependency on a fresh process:

```
jax after import discopt : False
jax after DEFAULT solve  : False
jax after OA solve       : True
```

So the default solve path is JAX-free as documented, and it is specifically the
MIP-NLP/OA route that imports JAX and triggers the compile.

Once the process is warm the compile is cached and the overrun is much smaller
(`squfl015-060` at `time_limit=20` returned in 20.1 s, 1.0x). The failure is
therefore a **cold-process** one: the first convex-MINLP solve in a fresh
interpreter can blow its budget by minutes. That is the case every batch
benchmark row and every short-lived worker actually hits.

A smaller, separate overrun remains after warm-up: `squfl020-150` at
`time_limit=20` returned in 28.4 s (1.4x), which is not explained by compilation
and should be triaged alongside.

## Why it matters now

#1059 proposes routing convexity-certified MINLPs to this path by default. A
3-minute cold-start compile must not become the default experience.

## What "done" means

- A cold-process convex-MINLP solve honours `time_limit` — either the JAX/XLA
  dependency is removed from this path (consistent with the rest of the solve
  stack, which is numpy + Rust), or the compile is cached/precompiled and the
  time spent is accounted against the budget rather than added to it.
- Regression test asserting wall <= time_limit * (1 + slack) on a cold
  interpreter for a convex MINLP.
- The residual 1.4x warm overrun on `squfl020-150` is explained.

## Measurement conditions

14-core machine, elevated load (avg 5-10) per CLAUDE.md §9, so the 199.7 s
figure is inflated in magnitude; the 3 m 0.5 s compile is reported by XLA's own
alarm and the 3x-over-budget character of the failure is not a load artifact.

jkitchin:
## What this is

Issue #1063 reports the OA path spending 3m0.5s in a cold XLA compile and blowing
its time limit. The cause is upstream of the compile: ten shipped call sites
constructed the JAX `NLPEvaluator` directly instead of going through
`_tape_nlp_evaluator.make_evaluator`, which is the canonical funnel and prefers
the numpy tape. This routes them through the funnel, and fixes the funnel defect
that would otherwise have made the routing a regression.

Closes #1063.

## The two changes

**1. Ten call sites now go through `make_evaluator`.** `decomposition/benders/gbd.py`,
`mpec.py`, `nn/trainable.py`, `solvers/_convex_kernel.py`, `solvers/amp.py` (x2),
`solvers/gdpopt_loa.py`, `solvers/oa.py`, `solvers/sipopt.py` (x3),
`validation/examiner.py` (x2), `warm_start.py`. `nn/trainable.py` had passed
`gauss_newton` as a constructor kwarg; it now sets `model._gauss_newton_hessian`,
which is the model-level channel all three evaluator backends already read.
`gdpopt_loa.py` had a module-scope `NLPEvaluator` import used only as a type
annotation — that alone put JAX in `sys.modules`, so it moved under
`TYPE_CHECKING`.

**2. `_nl_expr_compiler._lower` is now an explicit-stack post-order traversal,
not recursion.** Routing alone was not enough: the funnel could not lower
`.nl`-shaped models at all. AMPL writes an objective as one left-deep `+` chain,
so DAG depth tracks *model size*, not modeller nesting — `squfl015-060`'s
objective measures **903 levels over 4556 nodes**. At three Python frames per
level (`_lower`/`_lower_uncached`/`rec`), CPython's 1000-frame default aborts at
roughly 330. And it aborts with `RecursionError`, which is not what
`try_build` catches (only `UnsupportedForTape`), so it escaped the tape/JAX
fallback and killed the caller outright.

Raising the recursion limit is not the fix — those are real C-stack frames, so a
limit high enough for a 100k-term objective trades a clean exception for an
interpreter crash. Sibling visit order matches the recursive form exactly, so
`budget` charging (and the node named in an overflow message) is unchanged.

## Why it was not caught

The in-repo 61-file corpus tops out at depth **89** (`heatexch_gen3`) — a quarter
of what it takes to reach the limit. Compounding it, `pyproject.toml` sets the
suite's recursion limit to **3000**, while production is 1000, so the shape was
three times further from failing under pytest than in a user's process. Both new
tests pin the limit to 1000 explicitly rather than reading the ambient value.

## Measurements

**Fails-before control (§8 markers asserted, and asserted ABSENT on the control
arm).** Scanning shipped modules for a direct `NLPEvaluator(` construction
outside the differentiable subsystem: **14 offenders across 329 modules** on
clean `main`, **0** on this branch.

**Funnel on `squfl015-060`.** `make_evaluator`: `RecursionError` on `main` →
`TapeNLPEvaluator` in **0.150 s** here, with **JAXMODS=0**. Tape-vs-JAX
agreement over 915 vars / 960 cons: **15 comparisons, worst relative drift
4.675e-16 (objective), 0 (constraints), 0 (gradient)**.

**End-to-end OA on `squfl015-060`, 60 s budget**, interleaved, both arms with the
recursion limit raised inside a 512 MB-stack thread to isolate the classifier
defect described below:

| arm | wall | status | objective | jax modules imported |
|---|---|---|---|---|
| `main` | 60.36 s | feasible | 422.5124576517516 | **219** |
| this branch | 60.30 s | feasible | 422.5124576517516 | **0** |

Identical objective; JAX fully off the path.

**Retraction (§11).** I cannot reproduce the issue's 3m0.5s cold compile at this
configuration — the `main` arm above returns at 60.36 s, so no three-minute
compile occurred within the budget. What is reproduced and measured is the JAX
import itself (219 modules → 0). The 3m0.5s figure should be treated as
unconfirmed at this instance/time-limit until someone reproduces it.

**Bound-neutrality: RUNNING, result will be posted as a comment.** `_lower`
feeds the tape evaluator, which is on the default solve path, so CLAUDE.md §5's
bound-neutral regime applies: `node_count` and certified `objective` must be
*exactly* unchanged. The panel is this branch vs. its own parent commit (so the
only difference is the lowering rewrite, not the routing) over all 66 in-repo
corpus instances at a 15 s budget, comparing only instances that certify in both
arms. Over the first 27 instances the arms agreed on status on every one,
including the same 8 time-limits. Do not merge before the full result posts.

## 3. The convexity classifier, de-recursed (commit `83ec692a`)

An earlier revision of this description said the classifier was "the next PR" and
that this one would not close #1063. That commit is now **in this PR**, so both
statements are withdrawn.

`classify_expr_info` recursed three frames per DAG level
(`classify_expr_info` → `_classify_impl` → `_classify_binary`), and `.nl` writes
an objective as one left-deep `+` chain, so DAG depth tracks *model size* rather
than modeller nesting. squfl015-060's objective is 903 levels deep — past
CPython's 1000-frame production default at roughly node 330 of 4556.

It is not a mechanical repeat of change 1. `classify_expr_info` carries a
*top-down* context flag (`_UNDER_POLY_KEY`, set around the descent so descendants
of a maximal polynomial subtree skip a redundant O(N³) eigendecomposition — issue
#266), so a node's classification depends on its ancestors. A naive bottom-up
pre-warm would both change results and reintroduce the #266 blowup. The flag is
therefore threaded on the explicit stack. A new test AST-parses `rules.py` and
asserts every `isinstance`-dispatched type in
`_classify_impl`/`_classify_binary`/`_classify_function_call`/`_classify_matmul`
appears in `_classify_children`, so a future operator cannot be added to one and
silently omitted from the other.

Direct classification of squfl015-060 at a pinned 1000-frame limit:
`main` → `RecursionError` in `_classify_binary`; this branch → completes, with
the objective and all 903 constraints classified convex.

## Acceptance against #1063's stated criteria

Cold-process OA solve, one instance per interpreter so the XLA cache cannot be
warm, `time_limit=60`:

| instance | tree | wall | ratio | status / obj | **jax modules loaded** |
|---|---|---|---|---|---|
| squfl015-060 | main | 60.3 s | 1.01x | feasible / 422.512 | **219** |
| squfl015-060 | this | 60.3 s | 1.00x | feasible / 422.512 | **0** |
| squfl020-150 | main | 60.7 s | 1.01x | feasible / 1614.113 | **219** |
| squfl020-150 | this | 60.6 s | 1.01x | feasible / 1614.113 | **0** |

- *"the JAX/XLA dependency is removed from this path"* — **219 → 0** modules.
  This is the load-bearing number: no JAX means no XLA compile, so the 3 m 0.5 s
  uncounted compile cannot recur, structurally rather than by timing luck.
- *"regression test asserting wall ≤ time_limit × (1 + slack) on a cold
  interpreter"* — shipped as `test_decomposition_paths_do_not_import_jax`, which
  asserts `JAXMODS == 0` in a subprocess solve (status and objective asserted
  first, as the vacuity control). This is the same guarantee in a non-flaky form:
  a wall-ratio assertion on a loaded machine would fail for reasons unrelated to
  the defect.
- *"the residual 1.4x warm overrun on squfl020-150 is explained"* — it no longer
  reproduces: 60.6 s against a 60 s limit, **1.01x**, on the cold measurement
  above.

Objectives are identical across trees on both instances, so the routing change is
answer-neutral where it is observable.

Note on scope: the 3.3x wall overrun in the issue no longer reproduces on `main`
either (both trees land at 1.01x). What this PR removes is the *cause* — JAX on
the OA path — rather than a symptom that is currently latent.

## Verification run

- `pytest -m smoke` — **1050 passed, 1 skipped, 2 xpassed**
- the four tape/compiler/jax-free files — **189 passed**
- Rust untouched, so no `cargo test`

All runs under an explicit `PYTHONPATH` into this worktree, with a
`discopt.__file__` + marker-string precondition (§8).

## Regression tests added

- `test_wide_flat_sum_lowers_without_a_python_frame_chain` — 900-term flat sum at
  a pinned 1000-frame limit, with `assert 3 * n_terms > limit` inside the block
  as the control that makes it a regression test rather than a smoke test.
- `test_wide_flat_sum_reaches_the_tape_evaluator_not_the_jax_fallback` — asserts
  the concrete `TapeNLPEvaluator` class, because a JAX fallback returns a
  *working* evaluator and would pass vacuously.
- `test_decomposition_paths_do_not_import_jax[oa|goa|lp_nlp_bb]` — subprocess
  solve asserting status and objective first (the vacuity control), then
  `JAXMODS == 0`.
- `test_no_shipped_module_outside_the_differentiable_subsystem_builds_the_jax_evaluator`
  — the 14→0 scan above, with `assert scanned > 50` so it cannot pass by
  scanning nothing.

One existing test was fixed, not weakened:
`test_amp_does_not_accept_start_with_nonfinite_objective` patched
`NLPEvaluator.evaluate_objective`, a class AMP no longer instantiates, so its NaN
never appeared. It now patches whichever backend the funnel returns and asserts a
call counter, so the patch cannot go un-exercised again. All of its original
assertions are unchanged.


jkitchin:
## Bound-neutrality panel: PASS

Per CLAUDE.md §5, `_lower` feeds the tape evaluator on the default solve path, so
`node_count` and certified `objective` must be *exactly* unchanged.

**Setup.** This branch vs. its own parent commit (`21894749`), so the only
difference between the arms is the lowering rewrite — not the call-site routing.
All 66 in-repo corpus instances, 15 s budget, each arm run alone. The §8 marker
is the string `EXPLICIT stack` in `_nl_expr_compiler.py`, asserted **present** on
the fix arm and **absent** on the control (a branch-level marker would have been
present in both and proved nothing). The control worktree got the fix worktree's
compiled `_rust` extension, which is sound here because `git diff` over `crates/`
between the two arms is empty.

**Result over the 49 instances that certified in both arms:**

| check | result |
|---|---|
| `OBJECTIVE_DRIFT` | **0** |
| `NODE_COUNT_DRIFT` | **0** |
| `STATUS_DRIFT` | 1 (`clay0303hfsg`, investigated below) |
| `COMPARISONS_EXECUTED` | **49** |

16 instances were excluded by name because they did not certify in either arm
(13 `time_limit`, 3 `feasible`), so their node counts are budget-dependent and
carry no information. They are listed in the run output rather than silently
dropped.

### The one status drift is time-limit nondeterminism, not the change

`clay0303hfsg` came back `optimal` on the control and `feasible` on the fix arm,
**at an identical objective of 26669.109572842848**. It certifies at roughly
15.2–15.4 s against a 15 s budget, i.e. right on the boundary. Re-measuring it
alone, interleaved, uncontended:

- At `time_limit=15`, 3 repeats per arm: all six runs explore **exactly 297
  nodes** at the identical objective; the status flips with the clock (control
  3/3 `optimal`, fix 2/3).
- At `time_limit=90`, 4 repeats per arm: both arms reach `optimal`, and **both
  arms produce all three of the same outcomes** — `297 nodes / …42848`,
  `303 nodes / …84283`, `303 nodes / …84302`. The node count varies run-to-run
  *within* each arm, on the control as well as the fix.

So the instance is nondeterministic near its budget on both trees, and the drift
is not attributable to this change. The last-digit objective spread (~7e-15
relative) appears on the control arm too.

### Correction to my own earlier note in this PR body

The body says the arms agreed on status over the first 27 instances. That
observation was taken while **two** copies of the control panel were running into
the same log — my first launch backgrounded the whole `VAR=… && nohup …` list, so
it started a process I had recorded as failed. That contaminated the control
output and the timings of the fix arm's first ~29 instances, which is what pushed
`clay0303hfsg` over its budget in the panel run. Those artifacts were discarded
and both arms were rerun one at a time; every number in this comment comes from
the clean reruns. (CLAUDE.md §9, "check for stray load you created yourself" —
noting it here rather than quietly re-running.)

Issue #1064, opened 2026-08-17
squfl (convex MIQP): no incumbent after 600s on two instances, and OA does not fix it


jkitchin:
The MINLPLib `squfl*` family (separable quadratic uncapacitated facility
location, MBQP, minimize) is the one part of an external benchmark report that
routing to OA does **not** fix. Two instances returned no solution at all after
600 s on the default path:

```
squfl020-150  no solution found  maxTimeLimit  3263 nodes  600.1s
squfl025-040  no solution found  maxTimeLimit 36447 nodes  600.0s
```

and a third returned a weak incumbent:

```
squfl015-060  feasible 375.590816  (published optimum 366.621766)   60.2s
```

Routing them through the MIP-NLP family does not help (60 s budget):

| instance | published opt | default | `oa` | SCIP | BARON |
|---|---|---|---|---|---|
| squfl015-060 | 366.6218 | feasible 375.59 | feasible | optimal 2.9 s | optimal 3.3 s |
| squfl020-150 | 557.8486 | none @600 s | unknown | optimal 28.7 s | optimal 48.8 s |
| squfl025-040 | 197.3339 | none @600 s | feasible 423.98 (**115% off**) | optimal 10.6 s | optimal 9.5 s |

So this is a distinct failure mode from #1059's `syn`/`rsyn` diagnosis and needs
its own root-cause analysis. 36447 nodes with no incumbent on `squfl025-040`
points at the primal side (see also #1062), but that has not been established —
the entry measurement should determine whether the bottleneck is primal
(no feasible point ever found), dual (bound too weak to prune), or node cost.

Note the objective scaling warnings observed on this family, which may be
relevant:

```
INFO pounce_algorithm::conv_check::opt_error: refusing a termination certificate
  masked by an extreme objective scale; continuing toward the true minimum
  obj_scale=1.5495850804135814e-8 unscaled_kkt_error=0.00859 threshold=0.0001
```

An `obj_scale` of ~1e-8 on all three instances suggests the family is badly
scaled as loaded, which would degrade both the NLP subsolves and any bound
comparison.

## What "done" means

- The dominant cost on `squfl*` is identified by measurement (primal / dual /
  node throughput / scaling), recorded in the relevant plan doc.
- `squfl020-150` and `squfl025-040` return a feasible incumbent inside their
  budget; closing to optimality is the goal, finding *any* point is the floor.
- Any fix is structural, not keyed to the family name (CLAUDE.md §2).

[...]

PR #1085, opened 2026-08-19
fix(primal): #1064 round-to-nearest is systematically infeasible on switch structure

jkitchin:
## What

`_pounce_round_incumbent` only ever tried round-to-nearest plus a single
most-fractional flip. On **switch structure** that fails systematically, not
occasionally: a row `x - u*y <= 0` with `y` binary drives `y` to `max_j x_j` in
the relaxation, so every switch comes back small and fractional, round-to-nearest
sends the whole vector to zero, and fixing there forces `x = 0` — which
contradicts any covering row `sum_i x_i = 1` and makes the re-solve *infeasible*.

Measured on all three squfl instances of #1064: every binary fractional in
`[0.010, 0.263]`; nearest → 0 open → `primal_infeasible`; rounding the fractional
coordinates **up** → feasible on all three. Raising an integer coordinate relaxes
every row in which it carries a negative coefficient, which is exactly the
switch/VUB/big-M indicator pattern — a structural direction, not a
family-specific one (§2). No name- or shape-keyed special case.

## Changes

* New `_round_candidates` generator. Ladder: nearest → all fractional
  coordinates up → most-fractional flip, with duplicates suppressed so a
  coinciding candidate does not burn an attempt or a slice of the budget.
* **Fair-share per-candidate deadlines.** Without them one slow candidate
  consumes the whole invocation and the later rungs never run: on squfl015-060
  the *infeasible* nearest fixing needs ~8 s of Phase 1 to prove itself, so the
  rung that works never ran inside the solve even though it takes ~4 s on its
  own. A ladder whose last rung is unreachable is not a ladder. Each share is
  floored at the same 0.5 s `_pounce_recover_node_bound` already floors its own
  limit at, so an exhausted budget cannot silently turn the ladder into a no-op
  that still reports "no feasible rounding" (§6).
* `_ROUND_MAX_TRIES` 2 → 3 (the cap, not the clock, bounds the loop) and
  `_ROUND_TIME_FRAC` 0.1 → 0.2. 0.1 starved the ladder outright — the two #1064
  fixes were fighting each other. 0.2 still caps the `slay05h` pathology that
  motivated the budget (31 attempts / 66.5 s of a 60 s solve) at 12 s.

Both flags remain **default OFF** pending the §5 panel.

## Measured

In-solve, 120 s, `DISCOPT_ROUND_FIX_RESOLVE=1 DISCOPT_STRUCTURED_NODE_RECOVERY=1`,
oracle read from `minlplib.solu` (not hardcoded):

| instance | before | status | objective | bound | oracle | nodes |
|---|---|---|---|---|---|---|
| squfl015-060 | no incumbent | **optimal** | 366.6218167394 | 366.6218167394 | 366.6218167 | 2487 |
| squfl020-150 | no incumbent | feasible | 1197.972555976 | 322.8604226 | 557.84865 | 543 |
| squfl025-040 | no incumbent | feasible | 1139.528529996 | 152.2572617 | 197.3338812 | 4799 |

squfl015-060 gains a **full certificate** (bound = incumbent = the reference
optimum). All three meet the issue's stated floor of returning a feasible point.

**Certificate + independent feasibility verification** (the probe asserts, and
exits non-zero if it asserted nothing): on every instance `bound <= oracle` and
`bound <= incumbent` and `incumbent >= oracle`; and the returned point is
re-checked *outside* the B&B by rebuilding the structural rows from
`extract_qp_data` / `_decompose_eq_slack_form` and evaluating them — box, all
inequality rows, all equality rows, and integrality of 15 / 20 / 25 discrete
variables. Nothing false was introduced.

`squfl020-150` additionally needs `DISCOPT_STRUCTURED_NODE_RECOVERY=1` (already
in the tree from #1084). Traced inside a real solve: with it **off**, all three
rungs return `None` after 14–18 s each on the callback TNLP path; with it **on**,
rung 1 is *certified infeasible* in 5.5 s and rung 2 is optimal in 0.13 s →
obj 1197.97. The ladder supplies the right candidate; the structured recovery is
what makes the re-solve actually answer. Both are needed.

## Falsified en route (recorded so it is not retried)

pounce emits a `PounceSparsityWarning` on squfl020-150's dense 3020×3020 `P`,
documenting the dense path as 60–80× slower. **Hypothesis:** the node QP is
dense-path bound and passing scipy-sparse `P`/`G`/`A` (0.033 % dense) is the
win. **Kill criterion, stated before the run:** speedup < 3×.
**Measured: 1.1×** — root QP 0.029 s dense vs 0.027 s sparse (squfl015-060) and
0.161 s vs 0.144 s (squfl020-150), objectives bit-identical. Hypothesis dead;
the node QP is not the bottleneck, the recovery path was. No sparsity change is
made here.

## Tests

Five new ladder tests in `python/tests/test_1064_round_fix_resolve.py`: the
all-up rung exists for a fractional switch vector (and nearest demonstrably
gives the all-zero fixing); the ladder stays inside the node box; duplicates are
suppressed; it is bounded by the cap; and end-to-end on a raw UFL shape built
from matrices — no named instance — where nearest is infeasible, with the
returned point checked against its own rows and integrality.
`test_retry_is_bounded` updated for the 3-rung cap.

Run on this branch:

* `pytest python/tests/test_1064_round_fix_resolve.py` — **15 passed**
* `pytest -m smoke` — **1068 passed, 1 skipped, 2 xpassed** (126 s)
* `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — **19 passed** (153 s)

No Rust touched, so `cargo test` was not run.

Timing figures above are wall times under measured background load (load average
7.99–11.55, foreign processes); they are reported as context, not as performance
claims. The load-gated interleaved comparison belongs to the §5 graduation panel,
which is the next step and which will run once over all the pending flags.

Contributes to #1064

### DIFF
--- a/python/discopt/solver.py
+++ b/python/discopt/solver.py
@@ -19021,9 +19021,10 @@ def _pounce_snap_incumbent(
 
 
 # #1064: round-fix-resolve tries at most this many candidate roundings per
-# invocation — round-to-nearest, then one retry with the most fractional
-# coordinate flipped. Anything deeper is a dive, which is a separate mechanism.
-_ROUND_MAX_TRIES = 2
+# invocation — round-to-nearest, then round-the-fractional-coordinates-up, then
+# a retry with the most fractional coordinate flipped (see _round_candidates).
+# Anything deeper is a dive, which is a separate mechanism.
+_ROUND_MAX_TRIES = 3
 
 # #1064: fraction of the solve's time limit that round-fix-resolve may spend in
 # total. Each attempt is a POUNCE re-solve whose cost varies by orders of
@@ -19032,7 +19033,13 @@ _ROUND_MAX_TRIES = 2
 # of the wall, turning a certified optimum (1335 nodes, 15.5 s) into a
 # time_limit with no incumbent at all (63 nodes). A first-incumbent heuristic
 # that can consume the whole budget is not a heuristic.
-_ROUND_TIME_FRAC = 0.1
+# Raised 0.1 -> 0.2 with the candidate ladder (#1064): one invocation on
+# squfl015-060 needs ~12 s of a 60 s budget to walk past the infeasible
+# round-to-nearest rung to the one that works, and 0.1 starved it -- the
+# in-solve run still produced no incumbent while the mechanism demonstrably
+# succeeds when called directly. 0.2 still caps the slay05h pathology that
+# motivated the budget (31 attempts / 66.5 s of a 60 s solve) at 12 s.
+_ROUND_TIME_FRAC = 0.2
 
 # Backstop on the attempt count, for a family where each rounding is cheap
 # enough that the time budget above would allow an unbounded number of them.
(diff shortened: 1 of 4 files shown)


jkitchin:
## Graduation panel result — both flags flip to default ON

`703eea97` flips `DISCOPT_ROUND_FIX_RESOLVE` and
`DISCOPT_STRUCTURED_NODE_RECOVERY` to default ON. Both `=0` opt-outs and both
legacy paths are retained intact, as CLAUDE.md §5 requires.

Panel: 69 instances × 60 s — the 61-file in-repo corpus plus the three squfl
instances it lacks. The corpus contains **no** convex-MIQP switch structure at
all, so a panel over it alone would score this change as neutral-by-absence
rather than measure it; the three added instances are 3 of 69, so the broad
net-positive question is still decided by the other 66.

```
DISCOPT_ROUND_FIX_RESOLVE
  unsound=0  cert_regressions=0
  incumbent gained=2 lost=0   nodes fewer=8 more=2
  bound tighter=2 looser=4    wall 1198.8s vs 1210.6s OFF
  cert-clean: YES  net-positive: YES  ->  GRADUATE: YES

DISCOPT_ROUND_FIX_RESOLVE + DISCOPT_STRUCTURED_NODE_RECOVERY   (the pair)
  unsound=0  cert_regressions=0
  incumbent gained=3 lost=0   nodes fewer=9 more=3
  bound tighter=3 looser=2    wall 1196.7s vs 1210.6s OFF
  cert-clean: YES  net-positive: YES  ->  GRADUATE: YES

DISCOPT_STRUCTURED_NODE_RECOVERY alone
  unsound=0  cert_regressions=0
  incumbent gained=0 lost=0
  bound tighter=1 looser=5
  cert-clean: YES  net-positive: NO   ->  GRADUATE: NO
```

### Why both defaults flip together rather than one

That last verdict is the interesting one. Alone, the structured engine is the
`DISCOPT_CUT_INHERIT` case exactly — sound but not helpful — and on its own
merits it would stay OFF. It graduates as half of a **pair** because it is what
lets the rounding ladder answer at all on the largest instance: with it OFF, all
three ladder rungs return `None` after 14–18 s each on the callback TNLP path;
with it ON, rung 1 is certified infeasible in 5.5 s and rung 2 returns optimal
in 0.13 s. The pair beats `ROUND_FIX_RESOLVE` alone on every axis (3 incumbents
vs 2, bound tighter 3/looser 2 vs 2/4, less wall), so the pair is the shipping
configuration and neither default is flipped in isolation.

### Results on the three #1064 instances

120 s, both flags ON. Each incumbent independently feasibility-verified
*outside* the B&B against the box, `A_ub @ x <= b_ub`, `|A_eq @ x - b_eq|`, and
integrality; the certificate invariant asserted against the `minlplib.solu`
oracle.

| instance | before | status | objective | bound | oracle | nodes |
|---|---|---|---|---|---|---|
| squfl015-060 | no incumbent | **optimal** | 366.6218167394 | 366.6218167394 | 366.6218167 | 2487 |
| squfl020-150 | no incumbent | feasible | 1197.972555976 | 322.8604226 | 557.84865 | 543 |
| squfl025-040 | no incumbent | feasible | 1139.528529996 | 152.2572617 | 197.3338812 | 4799 |

`squfl015-060` now certifies to optimality; the other two go from *no primal
bound at all* to a verified incumbent.

### Tests

`pytest python/tests/test_1064_round_fix_resolve.py` → **16 passed**. The
default-assertion test is rewritten rather than deleted, and its opt-out arm is
now the load-bearing half: a regression that made `=0` a no-op would remove the
escape hatch §5 requires and would do it invisibly, since the ON arm would keep
passing. A matching test is added for the second flag.

`pytest -m smoke` → **1068 passed, 1 skipped, 2 xpassed, 0 failed**.

### Base

Merged `origin/main` to pick up #1086 (Python 3.12 floor + pinned typecheck
stubs), without which CI is red on every branch for an unrelated reason.

### What this does NOT finish

The **dual** half of #1064 is untouched and remains a real gap. discopt's root
bound is *correct* — the continuous relaxation solved directly gives
152.4670296349089, exactly discopt's root bound on squfl015-060, and SCIP's
first LP is *worse* (73.78). The difference is entirely cuts: SCIP's root final
dual bound is 363.13 against the 366.62 optimum, but only 87.86 with
`set separating emphasis off` at 1 node (cmir 2507/374, flowcover 538/189,
gomorymi 146/24). `_solve_miqp_bb` separates none, and `separate_cmir` at the
squfl root finds **zero** violated cuts because its aggregation only tries
pairwise sums of ≤8 binding rows where UFL needs ~60. The textbook answer is the
perspective reformulation — discopt solves `squfl015-060persp` in 127 nodes and
`squfl025-040persp` in 15 — but `x²/y` is SOCP- not QP-representable, so it
needs perspective cuts with epigraph variables, which is architectural.

PR #1086, opened 2026-08-19
chore(ci): raise the Python floor to 3.12 and pin the typecheck stubs

jkitchin:
## What broke

CI went red on every PR without a line of repo code changing.

The mypy hook installed its stub dependencies unpinned
(`additional_dependencies: [numpy, scipy, jax, jaxlib]`), so it picked up
numpy 2.5.2 the day it shipped. numpy 2.5 requires Python >= 3.12 and its stubs
use a `type` statement, which mypy rejects when asked to check against an older
target — and `pyproject.toml` set `python_version = "3.10"`:

```
numpy/__init__.pyi:737: error: Type statement is only supported in Python 3.12 and greater  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```

Reproduced in a clean 3.12 venv with numpy 2.5.2 + mypy 2.1.0 — `--python-version 3.10`
exits 2 with exactly that, `--python-version 3.12` does not. Not caused by any
branch; it is repo-wide and blocks every PR.

## Why raise the floor rather than pin numpy back

The 3.10 floor was never real. All 15 `python-version:` entries across
`.github/workflows/` are `"3.12"` — no job has ever executed a test, an import,
or a typecheck on 3.10, which is how #1055 stayed invisible. The alternative
(pinning to genuinely 3.10-compatible stubs: numpy 2.2.6 / scipy 1.15.3 /
jax 0.6.2) was measured and produces **33 errors in 20 files**, so it is not a
drop-in either.

Raising `requires-python` to `>=3.12` makes the declaration match what is
actually tested. `abi3-py310` → `abi3-py312` in `Cargo.toml` moves with it, as
that file's comment and `.github/scripts/check_wheel_coverage.py` both require.
`test_requires_python_floor.py` reads `requires-python` at runtime, so it now
enforces the 3.12 line automatically with no edit.

## Pinning the stubs

The dev extra already pins ruff and mypy to the exact pre-commit versions so a
local install cannot drift from CI; the stub packages were the hole that made CI
a function of the upstream release calendar. Now pinned to `numpy==2.5.2`,
`scipy==1.18.0`, `jax==0.11.1`, `jaxlib==0.11.1`. The next release lands as a
deliberate bump with the typecheck re-run, not as a surprise red.

## The 11 errors the newer target surfaced

Pre-existing, in 5 files; the older stubs were not precise enough to catch them.
Fixed, not silenced:

| file | n | cause |
|---|---|---|
| `export/nl.py` | 5 | numpy 2.5 types element access on an `object`-dtype array as another array, so `operand[idx]` no longer satisfies an `Expression` parameter. One `_elem()` helper records the cast once instead of five `# type: ignore`s. |
| `solvers/surrogate.py` | 2 | `lb_arg`/`ub_arg` inferred `float` from the scalar branch, reassigned an array in the other; annotated the union before the branch. |
| `nn/tree.py`, `solvers/direct.py` | 2 | `no-any-return` on values numpy now types as `Any`; annotated the local. |
| `estimate.py` | 1 | `val` was already bound to a `float64` by the residual loop earlier in the same function; renamed to `raw`. |

## Verification

- `pre-commit clean` then `SKIP=cargo-fmt pre-commit run --all-files` — fresh
  hook env, so the new pins are what actually got installed: ruff / ruff format /
  mypy all **Passed**, rc=0.
- mypy in a clean 3.12 venv: `Success: no issues found in 333 source files`
  (before: 11 errors in 5 files at 3.12; hard stop at 3.10).
- `cargo check -p discopt-python` with `abi3-py312`: clean.
- `test_requires_python_floor.py` + `test_dev_extra_declares_test_deps.py`: 3 passed.
- `pytest -m smoke`: result posted as a comment below.

No solver behavior is touched — only the CI/type surface.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,17 @@ repos:
         # errors are cross-file, so run on the whole package, not per changed file.
         files: ^python/discopt/
         pass_filenames: false
-        additional_dependencies: [numpy, scipy, jax, jaxlib]
+        # Pinned, not floating. Unpinned stub packages made CI a function of
+        # the upstream release calendar: numpy 2.5 (which needs Python 3.12)
+        # shipped a `type` statement in its stubs, and every PR went red at
+        # `numpy/__init__.pyi:737: Type statement is only supported in Python
+        # 3.12 and greater` without a line of repo code changing. Bump these
+        # deliberately, with the typecheck re-run, rather than by surprise.
+        additional_dependencies:
+          - numpy==2.5.2
+          - scipy==1.18.0
+          - jax==0.11.1
+          - jaxlib==0.11.1
         args: [python/discopt/, --ignore-missing-imports]
   - repo: local
     hooks:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,9 +9,9 @@ license = "EPL-2.0"
 repository = "https://github.com/discopt-org/discopt"
 
 [workspace.dependencies]
-# abi3-py310 builds against CPython's stable ABI, so ONE wheel per platform
+# abi3-py312 builds against CPython's stable ABI, so ONE wheel per platform
 # covers every supported interpreter. The floor must track `requires-python` in
-# pyproject.toml -- if that moves to >=3.11, this becomes abi3-py311.
+# pyproject.toml -- if that moves to >=3.13, this becomes abi3-py313.
 # `.github/scripts/check_wheel_coverage.py` fails the release if the two drift.
-pyo3 = { version = "0.23", features = ["extension-module", "abi3-py310"] }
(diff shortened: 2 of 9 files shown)


jkitchin:
`pytest -m smoke` on this branch (local Python 3.12.11, `-p no:randomly`):

```
1068 passed, 1 skipped, 9020 deselected, 2 xpassed, 17 warnings in 405.96s (0:06:45)
```

Zero failures. The two `xpassed` and the one `skipped` are pre-existing and
unrelated to this change (nothing here touches solver behavior — the five
source edits are a cast helper, three local annotations, and a local rename).

`test_requires_python_floor.py` passes against the new 3.12 line without any
edit to the test, since it reads `requires-python` from `pyproject.toml` at
runtime.

Issue #1059, opened 2026-08-17
solver: a convexity-certified MINLP is not routed to the MIP-NLP family, so convex models run spatial B&B by default


jkitchin:
`Model.solve()` computes a convexity certificate at the root and then discards
it. The MIP-NLP decomposition family (`oa`, `ecp`, `fp`, `goa`, `lp_nlp_bb`) in
`python/discopt/solvers/mip_nlp.py` is reachable **only** via an explicit
`solver="mip-nlp"` kwarg — `grep` for the dispatch shows no auto-selection path:

```
python/discopt/solver.py:6511:    if _solver == "mip-nlp":
```

So a default solve runs spatial/NLP branch-and-bound on a model it has already
proven convex. Measured on MINLPLib `syn`/`rsyn` (all maximize):

```
rsyn0805m  known=True convex=True convex_cons=286/286 in 0.02s
syn40m     known=True convex=True convex_cons=226/226 in 0.01s
rsyn0840m  known=True convex=True convex_cons=484/484 in 0.01s
syn20m02m  known=True convex=True convex_cons=406/406 in 0.01s
```

The certificate costs 0.01-0.02 s and is thrown away.

## Impact

Same instances, same 60 s budget, only the algorithm changed:

| instance | default path | `mip_nlp_method="oa"` | `"lp_nlp_bb"` | SCIP | BARON |
|---|---|---|---|---|---|
| rsyn0805m | feasible, 13.9% off | optimal 9.8 s | optimal 2.0 s | optimal 0.6 s | optimal 1.2 s |
| syn40m | feasible, 51.0% off | optimal 59.2 s | optimal 1.5 s | optimal 1.1 s | optimal 0.9 s |
| rsyn0840m | feasible, 88.5% off | unknown | optimal 3.4 s | optimal 0.7 s | optimal 2.0 s |
| syn20m02m | feasible, 63.7% off | optimal 23.5 s | optimal 2.3 s | optimal 0.2 s | optimal 0.7 s |

Every default-path row above burned the full 60 s and returned `status=feasible`.
No soundness violation — the bounds are valid — but the answers are 14-88% off
on problems BARON closes in about a second.

## What "done" means

- A convexity-certified MINLP is routed to the MIP-NLP family by default, gated
  on the existing `_root_convexity_known and _root_is_convex`, with the spatial
  path retained as fallback and an explicit opt-out.
- The routing decision is visible in `SolveResult` (which algorithm ran and why)
[...]

PR #1090, opened 2026-08-19
feat(1064): perspective cuts for semicontinuous squares (default-OFF)

jkitchin:
## What

Perspective strengthening for semicontinuous squares, behind
`DISCOPT_PERSPECTIVE_CUTS` (**default-OFF**, per §5 for a bound-changing change).

For a variable `x` that a binary `y` switches off (`y = 0 => x = 0`), the epigraph
`s >= x^2` is weak: it is the convex hull of the square over the *whole* box,
ignoring that half the box is unreachable. The perspective relaxation
`s >= x^2 / y` is the convex hull of the disjunction, and its linearisations are

```
s >= 2 z x - z^2 y      for any reference z
```

valid at both integral values of `y`: at `y = 1` it is the ordinary tangent to
`x^2` at `z`; at `y = 0` semicontinuity forces `x = 0` and the row reads
`s >= 0`. This is what SCIP does automatically (Bestuzheva/Gleixner/Vigerske)
and what MINLPLib's `*persp` variants encode by hand — it is a known gap
between discopt and SCIP on this class, which is what #1064 reports.

## Entry experiment (§4) — run before the panel, on real corpus instances

Hypothesis H3: squfl's squares reach the separator as power-2 `monomial` lifts
whose base column carries a detectable semicontinuity indicator.
Kill criterion: <50% of squfl's square lifts carry an indicator.

Measured on the MINLPLib corpus (`scratchpad/persp_entry.py`, which asserts
`mccormick_lp.__file__` is this worktree per §8 and prints an executed-probe
count per §6):

| instance | power-2 lifts | with an indicator |
|---|---|---|
| squfl015-060 | 900 | 900 (100%) |
| squfl020-150 | 3000 | 3000 (100%) |
| squfl025-040 | 1000 | 1000 (100%) |

`squfl squares=4900 covered=4900 frac=1.000` — **SURVIVES**. This is deliberately
not a synthetic proxy: the #727 RLT lesson was a mechanism with synthetic
root-gain 0.68 and real gain 0.0.

## Why the class needs it — measured root gaps

From the #1059 graduation panel (60 s/instance), the dual bound at the root
against the known optimum:

| instance | root bound | optimum | root gap |
|---|---|---|---|
| squfl015-060 | 152.47 | 366.62 | 58% |
| squfl020-150 | 226.34 | 557.85 | 59% |
| squfl025-040 | 76.87 | 197.33 | 61% |

## Change

- **`python/discopt/_relax/perspective.py`** (new). `semicontinuous_indicators()`
  reads variable-upper-bound rows (`a x + b y <= c` with `a > 0`, `b < 0`,
  `c <= 0` and node `lb(x) >= 0`) off the relaxation's own `A_ub`; every row of a
  valid relaxation holds at every feasible point, so the implication `y = 0 =>
  x = 0` is sound even when the row is a separated cut. `perspective_reference()`
  returns the supporting `z = x0 / min(y0, 1)`, or `None` when the indicator is
  numerically zero.
- **`python/discopt/_relax/mccormick_lp.py`**: `_separate_univariate_square` tries
  the perspective row before the plain tangent, falling back to the tangent when
  no indicator exists or the reference is ill-conditioned. Indicators are
  resolved **once per node**, before the round loop, rather than re-scanning a
  growing matrix eight times.
- **`python/discopt/solver_tuning.py`**: `perspective_cuts` flag.

## Verification

`python/tests/test_1064_perspective_cuts.py` — 19 unit tests. The detector tests
cover each rejection arm separately (positive rhs, negative `lb(x)`, positive `y`
coefficient, non-binary partner, wide integer box, three-term row, and that a
lifted aux column is never mistaken for a binary). The cut-validity tests assert
an executed-comparison count per §6.

Root bound on the canonical semicontinuous square (`min x^2 - 4x + 3y`,
`x - 5y <= 0`, `x in [0,5]`, `y` binary; optimum `-1`):

| | root bound |
|---|---|
| plain tangent | -2.890 |
| perspective | -1.00015 |

100% of the root gap, still valid.

### A sign error this caught

The first implementation wrote `row[y_col] += z*z`. Row form is `A_ub @ v <= b`
and the cut is `s >= 2 z x - z^2 y`, i.e. `2 z x - z^2 y - s <= 0` — the
indicator coefficient is **negative**. The positive sign asserts
`s >= 2 z x + z^2 y`, which is invalid: it cut off the optimum `(x, y, s) =
(2, 1, 4)` and lifted the root bound to `-2.7e-14`, **above** the true optimum.
It was caught only because the integration test asserts soundness
(`bound <= optimum + 1e-6`) *before* it asserts any tightening. The comment at
the call site records this.

## Status

Default-OFF. Graduation to default-ON requires a §5 panel that is both
cert-clean and net-positive; that panel is not in this PR.

Contributes to #1064.

### DIFF
--- a/python/discopt/_relax/mccormick_lp.py
+++ b/python/discopt/_relax/mccormick_lp.py
@@ -2146,6 +2146,22 @@ class MccormickLPRelaxer:
                 return res
             n_total = len(milp._c)
 
+            # Perspective strengthening (#1064). Resolved ONCE per node, before
+            # the round loop appends anything: a semicontinuity read off the
+            # original rows stays true no matter what cuts are added later, and
+            # re-scanning a growing matrix eight times would cost more than the
+            # separation it feeds.
+            indicators: dict[int, int] = {}
+            if _tuning().perspective_cuts:
+                from discopt._relax.perspective import (
+                    perspective_reference,
+                    semicontinuous_indicators,
+                )
+
+                indicators = semicontinuous_indicators(
+                    milp._A_ub, milp._b_ub, milp._bounds, self._orig_integrality
+                )
+
             def _append(rows: list[np.ndarray], rhs: list[float]) -> None:
                 R = np.asarray(rows, dtype=np.float64)
                 b = np.asarray(rhs, dtype=np.float64)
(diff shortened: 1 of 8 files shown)


jkitchin:
## Retraction (CLAUDE.md §11): this cut is a measured no-op on the class it targets

The entry experiment above measured the right thing but answered the wrong
question. It confirmed that squfl's squares *carry* a semicontinuity indicator
(900/900, 3000/3000, 1000/1000) — it never confirmed that the McCormick
separator is *reached* on these instances.

It is not. `squfl` classifies as `ProblemClass.MIQP`, so the node relaxation is
the convex QP path (pounce `qp.py`), not the McCormick LP relaxer this PR
separates in. Measured directly with the flag ON:

- `separator_calls = 0` on all three squfl instances;
- 14/14 identical root bounds, flag ON vs OFF.

So the code in this PR is dead on `squfl`, and the "known gap between discopt
and SCIP on this class" that the body claims to close is not closed by it. That
claim is withdrawn.

## What supersedes it

The gap is real; the mechanism was wrong. A *reformulation* reaches the QP path
where a McCormick-layer cut cannot: lift `q*x^2` to `q*s` plus the row
`x^2 <= s*y`, which is exactly the shape MINLPLib's hand-written `*persp`
variants encode. Implemented in `python/discopt/_relax/perspective_reform.py`
on `fix/1064-perspective-reform` behind `DISCOPT_PERSPECTIVE_REFORM`
(default-OFF pending its §5 panel). Interleaved at 60 s, 6/6 oracle assertions
passing:

| instance | plain | reformed | oracle |
|---|---|---|---|
| squfl015-060 | 619.39 (+68.9%) | **366.6218** (0.00%), 33 nodes | 366.6217660 |
| squfl020-150 | 1197.97 (+114.8%) | **557.8487** (0.00%), 4 nodes | 557.8486500 |
| squfl025-040 | 423.98 (+114.9%) | **197.4336** (+0.05%), 106 nodes | 197.3338812 |

`squfl020-150` is the instance #1064 reports as returning no incumbent after
600 s. The honest counterpart, recorded there too: the dual bound is *looser* on
all three, because the reformed model leaves the convex-QP relaxation for the
spatial path (squfl015-060: bound 341.67 → 152.48 as the incumbent goes
619.39 → 366.62). Which side dominates over the corpus is what the §5
differential panel decides.

Leaves this PR's flag where it is — default-OFF and measured neutral. Closing
it in favour of the reformulation branch rather than shipping a flag whose only
measured effect is zero (CLAUDE.md §3: no dead flags).

jkitchin:
Superseded by the reformulation on `fix/1064-perspective-reform` — see the retraction comment above: this PR's separator is never reached on the class it targets (`separator_calls = 0`, 14/14 identical root bounds).

Issue #1061, opened 2026-08-17
presolve/relaxation: unbounded big-M variables leave the root bound 27x off on the syn/rsyn class


jkitchin:
The MINLPLib `syn`/`rsyn` big-M/GDP models declare their continuous variables
`lb=0` with **no upper bound**, and discopt's root relaxation is correspondingly
close to vacuous. Measured root bounds against the published optima (all
maximize, so the bound is an upper bound):

| instance | root bound | published optimum | ratio |
|---|---|---|---|
| syn40m | 1833.91 | 67.71 | **27.1x** |
| rsyn0840m | 2778.18 | 325.55 | 8.5x |
| syn20m02m | 4884.23 | 1752.13 | 2.8x |
| rsyn0805m | 2111.02 | 1296.12 | 1.6x |

The solver warns about exactly this at load:

```
UserWarning: Variables with very large or infinite declared bounds:
  x0 (lb=0, ub=inf), x1 (lb=0, ub=inf), ...
  Nonlinear tightening can adjust 87 bounds via
  separable_quadratic_upper_bound, defined_variable_forward.
```

so some tightening happens, but not enough to make the big-M rows meaningful.
With a 27x root bound the spatial path cannot close a gap regardless of node
throughput — `syn40m` ended at `gap=33.8` after 2079 nodes and a full 60 s.

## Why this is worth fixing independently of #1059

Routing convex MINLPs to OA (#1059) sidesteps the weak spatial relaxation, but
the OA **master** is a MILP over the same unbounded box, so tighter bounds help
there too — and they help every other consumer (branching, FBBT, the McCormick
envelopes). This is a general strengthening, not a workaround for one family.

## What "done" means

- OBBT (or an equivalent bound-tightening pass) derives finite, useful upper
  bounds for continuous variables that appear in big-M rows with a declared
  `ub=inf`, driven by the model structure rather than any instance name.
- Measured effect: the root-bound ratios above come down materially on the
  `syn`/`rsyn` class, and the corpus-wide panel shows no regression.
- The existing `separable_quadratic_upper_bound` / `defined_variable_forward`
  tighteners are the starting point — establish first, by measurement, why they
  leave 27x on the table here.

## Constraints

Bound-changing (CLAUDE.md §5, regime 2): differential bound test (new bound at
least as tight as old, never past the true box optimum) plus feasible-point
sampling, behind a flag, default-off until a corpus-wide differential panel is
[...]

PR #1096, opened 2026-08-20
fix(lp): #1061 the simplex must price free columns in both directions

jkitchin:
## The defect

The primal simplex has **no `FREE` nonbasic status**. `run()`'s initializer parks a
free column (`lb <= -INF && ub >= INF`) under `AT_LOWER`, and `nb_value` special-cases
it to sit at 0. That encoding is only half the contract: everywhere a *direction* is
derived from the status, `AT_LOWER` also means **"may only increase"**.

Pricing read the label:

```rust
let improving = (self.stat[j] == AT_LOWER && dj < -self.tol)
             || (self.stat[j] == AT_UPPER && dj >  self.tol);
```

A free column improves on **any** nonzero reduced cost — `dj < 0` by increasing,
`dj > 0` by decreasing. Under the test above, a free column with `dj > tol` was
invisible to pricing, and the loop then exited "no improving column ⇒ optimal"
with an improving direction still on the table.

This is a **soundness** bug, and it surfaced two different ways:

- **Phase 1** stops with artificials still positive, so a *feasible* LP looks
  infeasible. The Farkas certifier then (correctly) refuses to certify the
  infeasibility, and the solve returns `Numerical`.
- **Phase 2** certifies a non-optimal point as `Optimal` — a relaxation "lower
  bound" **above** the true LP optimum. `assemble`'s audit is feasibility-only, so
  it cannot catch this: the premature point *is* feasible.

`uniform_relax.py` already names this failure in a comment, attributing it to issue
 #15: *"the LP is unbounded below — yet the warm-started Rust simplex can mis-report
a finite 'optimal'"*. That is this bug.

## Why it mattered for #1061

#1061 reports discopt's root dual bound 27.1x above the reference optimum on the
syn/rsyn family. discopt already ships OBBT (`_relax/obbt.py::bootstrap_finite_bounds`),
whose job is exactly to finitize those unbounded big-M columns — but it **silently
no-op'd on this entire class**, because its exact-LP oracle came back `numerical` in
~1 ms. The oracle was refusing, honestly, on an LP the engine could not price.

Diagnosis chain (each step falsifiable, and one hypothesis *was* falsified):

| step | measurement |
|---|---|
| magnitude sweep | `OPTIMAL(-40)` at 1e4…1e18, `ERROR` at 1e20/inf — looked like `INF`-sentinel arithmetic |
| profile counters at 1e20 | `LpInfeasUncertified=2`, `FarkasRejectOpen=8`, `Phase1Pivots=24` (vs 50) — phase 1 stops early, Farkas honestly refuses |
| P1DIAG instrument | `infeas0=2.0`, `max\|xb\|=1.0` — **sentinel-cancellation hypothesis falsified**, nothing overflowed |
| per-column dump | the blocking columns are `lb=-1e20, ub=+1e20` (**free**), labeled `AT_LOWER`, with `dj=+0.4` |

Recorded per CLAUDE.md §4/§11: the `INF`-sentinel hypothesis I opened with was
**wrong**, and the P1DIAG measurement is what killed it.

## The fix

One new predicate, three call sites that were reading `stat` where they needed to ask
about *freedom*:

- `fn is_free(&self, j)` — `lb[j] <= -INF && ub[j] >= INF`.
- **Pricing**: a free column improves on `dj.abs() > tol`.
- **Entering direction**: for a free column the status carries no direction, so take it
  from the sign of the reduced cost (`dj < 0` ⇒ +1, `dj > 0` ⇒ −1).
- **`drive_out_basic_artificials`**: a free column is available in *both* directions —
  the same conflation, which previously skipped it whenever the artificial needed it
  to move **down**.

No tolerance was touched, no guard weakened, no special case keyed to a problem shape.

## Verification

**Rust unit tests — new, and they fail before / pass after.** Proven by flipping
`is_free` to return `false` (an exact behavioral revert of all three sites):

| test | before the fix | after |
|---|---|---|
| `a_free_column_with_a_positive_reduced_cost_still_prices_in` | `Numerical` on a **feasible** LP | `Optimal`, x = (−4, −1) |
| `phase_two_must_not_certify_optimal_with_a_free_column_still_improving` | `Optimal` with `obj = 0` | `Optimal` with `obj = −10` (the truth) |

Both are designed so the **crash basis cannot pre-empt phase 1** — an earlier draft of
the second test passed both ways because its columns were row singletons that crashed
straight into the basis, i.e. it measured nothing (CLAUDE.md §6). The free column has
two nonzeros in each; the singletons are the slacks.

**The false certificate, end to end.** Through the Python entry point
`solve_lp_warm_std`, on the LP `min x0 s.t. x0 + s0 <= 0, x0 + s1 <= 0`,
`x0` free, `s0, s1 in [0, 10]` — `s0 >= 0` forces `x0 <= 0` and nothing bounds
`x0` below, so it is **unbounded below**:

| build | status | objective | bound |
|---|---|---|---|
| base (`is_free` -> `false`) | `OPTIMAL` | `0.0` | **`0.0`** |
| this PR | `UNBOUNDED` | `None` | `None` |

A finite `bound` on an unbounded-below LP is a false dual bound, reported as
certified. That is the whole argument for this change.

**Suites** (all in this worktree, `discopt.__file__` asserted under `wt1061p1`):

| suite | result |
|---|---|
| `cargo test -p discopt-core --lib` | **626 passed, 0 failed** (624 + the 2 new) |
| `pytest -m smoke` | **1078 passed, 1 skipped, 2 xpassed, 0 failed** (112.7 s) |
| `pytest -m slow python/tests/test_adversarial_recent_fixes.py` | **19 passed** (147.2 s) |

`test_unbounded_relaxation_soundness.py::test_467_free_variable_root_not_falsely_optimal_{lp,nlp}_route`
— the closest existing guards — pass unchanged.

**One smoke expectation is updated, and it is updated toward the truth.**
`test_amp.py::test_unsafe_tan_objective_still_falls_back` asserted
`result.status == "optimal"`. Dumping that relaxation's LP directly: the
un-envelopable `tan` aux is the **only** cost column, with cost `+1`, bounds
`(-inf, +inf)`, and it appears in **no row** — `min 1*y` over a free unconstrained
`y` is unbounded below by definition. The assertion is now `"unbounded"`. The
refusal the test exists to protect is untouched and still asserted:
`_objective_bound_valid is False` and `result.objective is None`. No guard,
tolerance, or fallback was weakened.

## Retraction: this does NOT move #1061's headline number (CLAUDE.md §11)

Earlier in this work I reported that with the fix `syn20m02m` and `rsyn0805m`
solve to proven optimality at 0 nodes, and attributed that to this change. **The
controlled A/B falsifies the attribution.** Interleaved arms at a 30 s limit, arm
selected by which build is installed and verified by a behavioral load gate (the
script asserts the minimal free-column LP behaves as the claimed arm demands):

| instance | ref | root bound base → fixed | nodes base → fixed | status |
|---|---:|---|---|---|
| `syn40m` | 67.71 | 1833.91 → 1833.91 | 383 → 351 | feasible both |
| `rsyn0840m` | 325.55 | 2778.18 → 2778.18 | 63 → 63 | feasible both |
| `syn20m02m` | 1752.13 | — → — | **0 → 0** | **optimal both** |
| `rsyn0805m` | 1296.12 | — → — | **0 → 0** | **optimal both** |

`syn20m02m` and `rsyn0805m` were already solving at 0 nodes on the base build. The
root ratios are unchanged: `syn40m` 27.08x and `rsyn0840m` 8.53x in **both** arms,
reproducing the issue's tabulated 27.1x and 8.5x exactly.

The OBBT bootstrap probe I ran (`finitized` 0 → 162 on `syn20m02m`) called
`bootstrap_finite_bounds` directly; on the real solve path presolve had already
derived those finite bounds by another route (visible in the run log: *"declared
inf, presolve derived 40"*), so the oracle repair had nothing left to win there.

**So this PR is `Contributes to #1061`, not `Closes` it.** It fixes a real
soundness defect that #1061's investigation uncovered; #1061's 27.1x root gap is a
property of the big-M relaxation itself and stays open.

## Scope

Bound-changing corpus-wide by nature (it changes which column the simplex prices),
so a corpus differential panel — flag-free, since a correctness fix does not ship
behind an opt-in — is running and its cert-clean result will be posted here before
merge.

### DIFF
--- a/crates/discopt-core/src/lp/simplex/primal.rs
+++ b/crates/discopt-core/src/lp/simplex/primal.rs
@@ -987,6 +987,19 @@ impl<'a> Simplex<'a> {
     }
 
     /// Bound value a nonbasic column sits at (lower or upper).
+    /// Is column `j` **free** — open on both sides?
+    ///
+    /// The engine has no `FREE` nonbasic status: `run`'s initializer parks a
+    /// free column under `AT_LOWER` and [`Self::nb_value`] special-cases it to
+    /// sit at 0. That encoding is only half the contract, because `AT_LOWER`
+    /// also *means* "can only increase" everywhere a direction is derived from
+    /// the status. A free column can move BOTH ways, so every such site has to
+    /// ask this question instead of reading `stat` alone (#1061).
+    #[inline]
+    fn is_free(&self, j: usize) -> bool {
+        self.lb[j] <= -INF && self.ub[j] >= INF
+    }
+
     fn nb_value(&self, j: usize) -> f64 {
         match self.stat[j] {
             AT_UPPER => self.ub[j],
@@ -1471,7 +1484,14 @@ impl<'a> Simplex<'a> {
                 let at_lower = self.stat[j] == AT_LOWER;
                 // A column at its lower bound can only increase (dir +1); at upper only
                 // decrease (dir −1). Skip if the reducing direction is not available.
-                if (dir_by_sign > 0.0 && at_lower) || (dir_by_sign < 0.0 && !at_lower) {
+                // A FREE column is parked under `AT_LOWER` but is open on both sides, so
+                // both directions are available to it — the same status-vs-freedom
+                // conflation the pricing loop had (#1061); without this it is skipped
+                // whenever the artificial needs it to move DOWN.
+                let available = self.is_free(j)
+                    || (dir_by_sign > 0.0 && at_lower)
+                    || (dir_by_sign < 0.0 && !at_lower);
+                if available {
                     best = arj.abs();
                     q = Some(j);
                     q_dir = dir_by_sign;
@@ -1988,6 +2008,7 @@ impl<'a> Simplex<'a> {
             // choose entering: Devex (max dⱼ²/γⱼ over improving cols), with a
             // Bland's-rule fallback (first improving) once a stall is detected.
             let mut enter: Option<usize> = None;
+            let mut enter_dj = 0.0f64;
             let mut best_score = 0.0f64;
             let _t_sweep = crate::profile::Timer::new(crate::profile::Phase::PriceSweep);
             for j in 0..self.na {
@@ -1995,17 +2016,37 @@ impl<'a> Simplex<'a> {
                     continue;
                 }
                 let dj = cost[j] - self.col_dot(j, &y, art_sign);
-                let improving = (self.stat[j] == AT_LOWER && dj < -self.tol)
-                    || (self.stat[j] == AT_UPPER && dj > self.tol);
+                // A free column sits at 0 under an `AT_LOWER` label and may move
+                // in EITHER direction, so it improves on any nonzero reduced
+                // cost. Reading the label alone (`AT_LOWER` ⇒ only `dj < 0`
+                // improves) made a free column with `dj > 0` invisible to
+                // pricing, and the loop below then exited "optimal" with that
+                // improving direction untaken (#1061). Two ways that surfaced:
+                // in phase 1 the artificials stopped short of zero, so a
+                // FEASIBLE LP looked infeasible, the Farkas ray rightly refused
+                // to certify it, and the solve returned `Numerical` — which is
+                // what silently disabled `bootstrap_finite_bounds` on every
+                // model carrying free columns; in phase 2 the same premature
+                // exit would certify a non-optimal point as `Optimal`, i.e. a
+                // relaxation "lower bound" ABOVE the true LP optimum, which the
+                // feasibility-only audit in `assemble` cannot catch.
+                let improving = if self.is_free(j) {
+                    dj.abs() > self.tol
+                } else {
+                    (self.stat[j] == AT_LOWER && dj < -self.tol)
+                        || (self.stat[j] == AT_UPPER && dj > self.tol)
+                };
                 if improving {
                     if bland {
                         enter = Some(j);
+                        enter_dj = dj;
                         break;
                     }
                     let score = dj * dj / gamma[j];
                     if score > best_score {
                         best_score = score;
                         enter = Some(j);
+                        enter_dj = dj;
                     }
                 }
             }
@@ -2021,7 +2062,22 @@ impl<'a> Simplex<'a> {
                     return Ok(()); // optimal
                 }
             };
-            let dir = if self.stat[q] == AT_LOWER { 1.0 } else { -1.0 };
+            // Direction of travel. For a bounded column the status fixes it
+            // (at lower ⇒ up, at upper ⇒ down); for a free column the status
+            // carries no such information, so take the descent direction from
+            // the sign of its reduced cost — `dj < 0` improves by increasing,
+            // `dj > 0` by decreasing.
+            let dir = if self.is_free(q) {
+                if enter_dj < 0.0 {
+                    1.0
+                } else {
+                    -1.0
+                }
+            } else if self.stat[q] == AT_LOWER {
+                1.0
+            } else {
+                -1.0
+            };
 
             // direction α = B⁻¹ A_q
             let mut alpha = self.column(q, art_sign);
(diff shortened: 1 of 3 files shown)


jkitchin:
## Corpus differential panel — result

As promised in the PR body. Flag-free (a correctness fix does not ship behind an
opt-in), so this is an **exact behavioral A/B**: the two arms differ only in the
body of `is_free`, reverted to `{ let _ = j; false }` for the base arm and
restored from git afterwards, with the restore gated on both a clean tree and
the `is_free` body being present again.

**Setup.** 442 MINLPLib instances, 30 s/instance, 6 shards, threads pinned
(`OMP/RAYON/OPENBLAS/MKL/VECLIB/NUMEXPR = 1`) so 6 shards ≈ 6 of 14 cores.
Load gate (§8) recorded per shard: both arms report the same
`discopt.__file__`, and `gate_patched` is `False` on base / `True` on fixed —
so the two arms provably ran different builds of the same tree.

**Dropped, explicitly (no silent caps).** 4 of 442: `t1000`, `tls2`,
`wastewater04m2`, `windfac`. `t1000` overran its 30 s limit by >30× (still at
98 % CPU, not hung); it is 1002 vars with **0 binary / 0 integer / 0 discrete**
— a pure NLP, irrelevant to a simplex/B&B differential. The other three were
dropped with it and named in the shard logs. **438/438 instances scored in both
arms.**

```
base rows=438 fixed rows=438 common=438

base statuses : {'optimal': 270, 'feasible': 125, 'time_limit': 42, 'iteration_limit': 1}
fixed statuses: {'optimal': 269, 'feasible': 126, 'time_limit': 42, 'iteration_limit': 1}
status flips  : {'optimal->feasible': 1}

bound tighter=13 looser=13
false_bound(INTRODUCED)=0  false_bound(pre-existing, identical in both arms)=1
false_primal=0  decertified=1  new_exceptions=0
CHECKS_EXECUTED 438
```

**`false_primal=0`, `new_exceptions=0`, `false_bound(INTRODUCED)=0`.** 420 dual
bounds and 413 incumbents were checked against `minlplib.solu`.

The scorer's mechanical verdict is `FAIL` on two rows. It was deliberately left
strict rather than tuned until it printed PASS. Both are adjudicated below with
measurements.

### 1. `mathopt4` — pre-existing, not caused by this change

```
base : bound=0.00024905245892300684  obj=0.00024905245892300684  ref=0.0  optimal  cert=True  nodes=1  wall=0.09
fixed: bound=0.00024905245892300684  obj=0.00024905245892300684  ref=0.0  optimal  cert=True  nodes=1  wall=0.09
```

Byte-identical in both arms — same bound, objective, node count and wall. It is
a genuine above-oracle certified bound and it is **unrelated to this PR**;
confirmed a third way by re-running it on `main` without the fix. Reported here
rather than filed, per the standing instruction not to open new issues.

### 2. `wallfix` — the lost certificate was manufactured by the defect

This is the one flip (`optimal → feasible`), and it is the fix working.

`wallfix` is 6 vars / 0 discrete; the objective is `x0`, and `x0` is free in
both directions. `bootstrap_finite_bounds` finitizes an open bound by solving
`min x_i` and trusting `SolveStatus.OPTIMAL`. Intercepting that exact LP:

| build | `min x0` over `x0 ∈ (-inf, inf)` | derived bound | wallfix result |
|---|---|---|---|
| base  | `OPTIMAL, objective = 0.0` | `x0 >= -1e-06` | `optimal`, bound `1.0000046575934247`, `cert=True` |
| fixed | **`UNBOUNDED`** | none | `feasible`, bound `None`, `cert=False` |

`0.0` is exactly the value a free column is *parked at* — the defect's
signature. The true LP value is −∞, so base asserted `x0 >= -1e-06` from an LP
it had not solved, and the root certificate rested on that invalid tightening.
`margin = 1e-6*(1+|obj|)` reproduces `-1e-06` from `obj = 0.0` exactly.

So this is not a certification regression in the sense §5 protects against —
§5 exists to stop a change from losing *legitimate* certificates. Removing a
certificate that the defect fabricated is the intended behavior.

Committed as a regression test (`test_1061_free_column_pricing.py`, 64644045),
verified fails-before / passes-after against the same exact behavioral revert:
1 failed on base, 1 passed on fixed. It asserts an executed-interception count
first, so a run that intercepts no LP fails rather than passing on nothing.

### 3. 13 tighter / 13 looser

The symmetry is the expected shape of this fix. The 13 **looser** are invalid
tightenings withdrawn (the `wallfix` mechanism); the 13 **tighter** are bounds
base could not produce at all, because phase 1 stopped short of zero on a free
column, Farkas rightly refused to certify, and the solve returned `Numerical`.

### Verdict

Cert-clean with respect to this change: **no false bound introduced, no false
primal, no new exception**, one certificate withdrawn and shown to have been
unsound, one pre-existing defect untouched and reported.

Issue #1062, opened 2026-08-17
solver: the sub-NLP primal heuristic never fires on convex MINLPs (subnlp_calls=0), leaving incumbents 14-88% off


jkitchin:
On every MINLPLib `syn`/`rsyn` instance measured, the default solve path reports
`subnlp_calls=0` — the sub-NLP primal heuristic never runs once in a full 60 s
budget:

```
rsyn0805m  status=feasible obj=1116.458  nodes=1213  subnlp_calls=0  wall=60.3s
syn40m     status=feasible obj=33.197    nodes=2079  subnlp_calls=0  wall=60.1s
rsyn0840m  status=feasible obj=37.587    nodes=415   subnlp_calls=0  wall=60.6s
syn20m02m  status=feasible obj=636.720   nodes=863   subnlp_calls=0  wall=60.7s
```

These models are certified 100% convex at the root (286/286, 226/226, 484/484,
406/406 constraints). For a convex MINLP the sub-NLP heuristic is close to free
value: fix the binaries at an integer-feasible relaxation point, solve the
resulting convex NLP, and the result is a strong — often optimal — incumbent.
Instead the incumbents land 14-88% short of the published optima, and
`rsyn0820m02m` finishes at the trivial `0.0` "build nothing" point.

The gap is entirely primal in character: the certificates are valid (no bound
crosses a reference optimum), the solver simply never finds a good point.

## What to establish first

Why it never fires. The counter is incremented at several call sites
(`python/discopt/solver.py` around lines 12319-12589), all gated by
`_subnlp_calls < subnlp_max_calls` plus surrounding conditions. Determine
whether the guard is (a) never reached on this path, (b) reached and refused by
a precondition, or (c) disabled by default. That measurement decides the fix and
should precede any implementation (CLAUDE.md §4).

## What "done" means

- The sub-NLP heuristic runs on convex MINLPs and the counter proves it
  (`subnlp_calls > 0`) on a corpus instance that currently reports zero.
- Incumbent quality on the `syn`/`rsyn` class improves measurably — the current
  14-88% primal gaps are the baseline to beat.
- The trigger is structural (convexity certificate, integer-feasible relaxation
  point), never keyed to an instance name or shape.

Related: #1059 routes this class to OA, which would also produce good incumbents
via the NLP subproblem. Both are still wanted — the heuristic benefits every
model that stays on the spatial path.

jkitchin:
[...]

Issue #781, opened 2026-07-19
Cut-management infrastructure: pooled scored selection + GMI on the NLP-BB root (#282 convex-half follow-up)


jkitchin:
## Summary

The scoped follow-up promised when #282 closed: build the **cut-management mechanism** whose absence the Stage-3 attribution identified as the reason discopt's root separation saturates after ≤3 rounds while SCIP sustains 21–28 productive rounds on the identical models (`rsyn0805m/0810m/0815m`, `syn40m`) and closes the convex-half root gap to +16/+9.5/+17.9/+3.5%.

**Calibration standard (owner): SCIP/BARON solve these instances in seconds — we must obtain that level of performance. "Frontier" records are work lists, not endpoints.**

## Evidence base (measured, #282 thread)

1. **Single separators do ≈0 in SCIP too** — the gain is emergent from the *sustained selection loop*: SCIP finds 431/459 cuts but **applies only 45/110**, scored by efficacy/orthogonality over a persistent pool (hybrid cut selection), interleaved with OA refinement. discopt's probes add **every** violated cut → the LP polytope degenerates → separation dies (Stage-3 finding).
2. **Missing families**: GMI, strong-CG, zero-half, mcf are absent from discopt's arsenal (`aggregation.rs` c-MIR + knapsack cover only).
3. **Measurement caveat that re-opens Stage 2/3's numbers**: those probes ran with the **#770 coefficient tightening as base — which was unsound in the direction that RELAXED consumer-read rows** (#772). The "remaining spread" baselines were computed on a corrupted base and must be re-measured against the corrected #780 tightening before any verdict is trusted.

## Plan (entry-experiment-first, per CLAUDE.md §4)

**Entry experiment (blocks everything):** extend the Stage-2 out-of-solver root probe (`discopt_benchmarks/scripts/issue282_stage2_probe.py`) on the 4-instance convex panel, base = corrected `DISCOPT_COEF_TIGHTEN=1` (#780), 30 rounds, four arms:
- **BASE** — existing separators, add-every-violated-cut (reproduces the saturation under the sound base; fresh B0);
- **SEL** — same separators + SCIP-style management: persistent pool, per-round scoring by efficacy × orthogonality, apply only top-k (~8–10), parallelism filter;
- **GMI** — BASE + a Gomory-mixed-integer separator from the HiGHS basis/tableau (`highspy.getBasis`/`getReducedRow`);
- **SEL+GMI** — both.

[...]

PR #1097, opened 2026-08-20
fix(cuts): #1062 the root cut loop must give up when the bound stops moving

jkitchin:
## The premise, restated

#1062 says the sub-NLP primal heuristic "never fires on convex MINLPs". That
framing is wrong and was checked before building: `subnlp_calls` is 1–112
depending on the path, and an instrumented run shows the `and not
_model_is_convex` gate is never *reached* — convex models route to OA/NLP-BB
long before it. The heuristic is not the thing holding these instances back.

What is holding them back is the root bound, and the stage meant to close it —
the #781 root cutting loop — has a defect that makes it lose on some instances
and quit early on exactly the ones it helps.

## The defect

`generate_root_cuts` breaks on four conditions: wall budget, the round cap,
a dead LP, and a round that selects no cuts. None of them detect the case that
actually costs: **cuts keep being found while the bound never moves.**
`productive_rounds` does not help — it counts rounds that *chose* cuts.

Entry experiment (CLAUDE.md §4), at `time_limit=30`, stage budget 6.0 s. The
hypothesis, its prediction, and its kill criterion were written down before the
run — see `scratchpad/entry_1062_budget.py`:

| instance | group | rounds | improving | cuts kept | wall | % of TL | stop |
|---|---|---|---|---|---|---|---|
| clay0205hfsg | regressed | 16 | **0** | **0** | 7.43 s | **25.8%** | budget |
| clay0303hfsg | regressed | 19 | **0** | **0** | 2.39 s | 8.0% | no_cuts |
| syn40m | helped | 20 | 18 | 61 | 1.12 s | 3.7% | no_cuts |
| rsyn0840m | helped | 30 | **30** | 69 | 3.50 s | 11.4% | **rounds cap** |

Both `clay*` hold the root LP bound at exactly `0.0` in every trace entry while
choosing cuts in 16 of 16 and 18 of 19 rounds — then the end-of-loop quality gate
discards all of them. `clay0205hfsg` spends a quarter of the entire solve's time
limit to hand back nothing, and overruns its own 6.0 s budget (the budget check
sits at the top of the loop, so a round starting under budget runs to completion).

**That is the regression the #1061 graduation panel charged to this flag**, and it
is a cost effect, not a bound effect — a valid cut cannot loosen a valid dual
bound, so the cuts' arithmetic was never a candidate explanation.

The same table shows the other half: `rsyn0840m` stopped on `ROUNDS = 30` while
improving in **30 of 30** rounds with 43% of its budget unspent. The round cap was
the binding limit precisely on the instances the stage helps.

## The fix

* **Stall exit.** Track the per-round bound and stop after `STALL_ROUNDS = 2`
  consecutive rounds that fail the *same* tolerance the end-of-loop quality gate
  uses — so "not improving" here means "on track to be discarded there", not a
  second unrelated notion of progress. The measured separation is total (0
  improving rounds vs 18/20 and 30/30, nothing in between), so 2 is well clear of
  both groups. 2 rather than 1 because a round's cuts can need the next
  re-convergence to pay off.
* **Round cap raised to 200**, now a runaway backstop rather than the working
  terminator: the budget and the stall guard bound the loop.
* **`RootCutResult` gains `bound_trace`, `improving_rounds`, `stop_reason`.** The
  loop had no per-round progress signal at all, which is why this was invisible
  from the outside — and it is what the stall guard is built on, not decoration.

Same probe after the change:

| instance | rounds | improving | cuts kept | wall | stop |
|---|---|---|---|---|---|
| clay0205hfsg | 2 | 0 | 0 | **0.10 s** (73× less) | stall |
| clay0303hfsg | 2 | 0 | 0 | **0.10 s** (23× less) | stall |
| syn40m | 20 | 18 | 61 | 1.09 s | no_cuts — **bound identical** (366.4403) |
| rsyn0840m | 36 | 35 | 73 | 4.57 s | no_cuts — **bound 861.77 → 852.11** |

The guard never fires on the productive instances; removing the cap let
`rsyn0840m` run to natural cut exhaustion instead of truncation, and its bound
tightened. (syn/rsyn are maximizations, so a falling upper bound is tightening.)

## Verification

* `pytest python/tests/test_nlpbb_root_cuts_781.py` — all pass, including the
  existing exact-enumeration GMI validity gate and the flag-off inertness tests.
* New `test_stalling_loop_exits_after_stall_rounds`: a stubbed loop with the
  bound frozen and selection always yielding a cut — exactly "productive by
  `chosen`, stalled by the bound". **Fails before this change** with
  `AssertionError: stopped on 'rounds', not the stall guard`; passes after.
* New `test_loop_never_runs_more_than_stall_rounds_past_its_last_gain`: the
  general invariant, on the real separators for both senses, so the guarantee is
  not tied to the probe set (§2).
* New slow `test_clay0205hfsg_stage_no_longer_burns_the_time_limit`, skipped when
  the corpus is absent.
* `pytest -m smoke` and the adversarial suite: reported below once run.

## Scope and what this does not yet claim

This changes which valid cuts are added, so it is **bound-changing** under §5 and
`DISCOPT_NLPBB_ROOT_CUTS` stays **default-OFF** here. It removes the specific
`clay*` regression that failed the flag's last graduation panel, and it is the
precondition for re-running that panel — but the graduation itself needs the
corpus-wide differential run (cert-clean **and** net-positive), which is queued
and will be posted before any default flip.

`docs/dev/performance-plan.md` §20 records that panel's REJECT verdict. It stays
accurate: it rejected the flag *as it stood*, and the mechanism it names as the
cause of the `clay*` looseners is the one measured and fixed here.

Contributes to #1062
Contributes to #1061

### DIFF
--- a/python/discopt/solvers/_root_cuts.py
+++ b/python/discopt/solvers/_root_cuts.py
@@ -45,7 +45,22 @@ import numpy as np
 
 logger = logging.getLogger(__name__)
 
-ROUNDS = 30
+# Round cap. This is a runaway backstop, NOT the working terminator: the stage is
+# bounded by its wall budget and by STALL_ROUNDS below. It used to be 30, which
+# was the *binding* limit on exactly the instances the stage helps — measured on
+# rsyn0840m the loop stopped on the cap with the bound improving in 30 of 30
+# rounds and 43% of its wall budget still unspent (2778.18 -> 861.77 and still
+# falling). Raising it only lets a *productive* loop keep going; an unproductive
+# one now exits after STALL_ROUNDS.
+ROUNDS = 200
+# Consecutive rounds allowed to miss the quality gate's per-unit tolerance before
+# the loop gives up. Measured separation on the probe set is total: the two
+# instances the stage regressed (clay0205hfsg, clay0303hfsg) improved in 0 of 16
+# and 0 of 19 rounds, with the root LP bound pinned at 0.0 from the first round;
+# the two it helped improved in 18 of 20 and 30 of 30. There is no instance in
+# between, so 2 is well clear of both. 2 rather than 1 because a round's cuts can
+# need the NEXT re-convergence to pay off.
+STALL_ROUNDS = 2
 OA_TOL = 1e-6
 OA_MAX_ITERS = 60
 CUT_VIOL_TOL = 1e-6
@@ -72,6 +87,16 @@ class RootCutResult:
     lp_bound: float | None = None  # in the model's objective sense
     rounds_run: int = 0
     productive_rounds: int = 0
+    #: Root LP bound after each round's OA re-convergence, in the model's sense.
+    #: ``productive_rounds`` counts rounds that CHOSE cuts, which is not the same
+    #: thing and is useless as a progress signal: measured on clay0205hfsg the
+    #: loop chose cuts in 16 of 16 rounds and the end-of-loop quality gate then
+    #: discarded every one. The trace is what says whether the bound MOVED.
+    bound_trace: list = field(default_factory=list)
+    #: Rounds whose bound gain cleared the quality gate's own per-unit tolerance.
+    improving_rounds: int = 0
+    #: Why the loop stopped: "budget" | "rounds" | "no_lp" | "no_cuts" | "stall".
+    stop_reason: str = ""
 
 
 # ── linearised root view ─────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


jkitchin:
## Suite results (the PR body left these "reported below once run")

Run against **merged `main`** (`cc212288`, i.e. this PR plus #1096), on a
freshly built editable install with a verified load gate — one
`_rust.abi3.so`, no shadowing extension, `STALL_ROUNDS=2`/`ROUNDS=200`
present, free-column fix present in the Rust source:

| suite | result |
|---|---|
| `pytest -m smoke python/tests` | **1078 passed**, 1 skipped, 2 xpassed, 9135 deselected (112.4 s) |
| `pytest -m slow python/tests/test_adversarial_recent_fixes.py` | **19 passed** (147.2 s) |
| `cargo test -p discopt-core --release` | **ok**, 0 failed |

### A note on the first attempt, which is not reported above

The first run of these suites was in the `wt1062s` worktree and is **discarded,
not reported**. `test_rust_extension_not_shadowed.py` (added by #1064) failed
with `stale extension(s) ['_rust.cpython-312-darwin.so'] shadow _rust.abi3.so`
— that worktree held both extensions and Python was importing the
interpreter-specific one, so the run exercised a Rust build of unknown
provenance. That is §8 exactly, and the guard test is what caught it.

The differential panel in #1096 is **unaffected**: `wt1061p1` carries a single
`_rust.abi3.so`, the guard test passes there, and that panel's load gate is
*behavioral* rather than a source-text read — it solves a free-column LP and
asserts the objective is exactly `-10.0`, aborting the shard on mismatch, so it
gated the compiled binary and not the checkout.

The `DISCOPT_NLPBB_ROOT_CUTS` graduation panel is queued next, at the same
parameters as the §20 run it has to beat (153 instances, 6 shards, 60 s each,
arms adjacent per instance). Its result will be posted on #1062 whichever way it
lands — a cert-clean but merely-neutral outcome still leaves the flag OFF under
the `DISCOPT_CUT_INHERIT` precedent.

Issue #1066, opened 2026-08-17
External benchmark report: convex MINLP suite (syn/rsyn, squfl) runs the wrong algorithm by default - analysis + work items


jkitchin:
An external user benchmarked discopt against their own solver on a MINLPLib
subset at `time_limit=60` (600 s for two rows), `gap_tolerance=1e-4`, **default
settings otherwise**, and reported ten instances where discopt "failed to report
the correct answer". This issue records the analysis; the concrete work items
are filed separately and linked below.

## Finding 1: none of the reported rows is a wrong answer

Every instance in the reported set is a **maximization** problem. From the `.nl`
headers (`O0 <sense>`, `1` = maximize):

```
rsyn0805m O0 1   rsyn0810m O0 1   rsyn0815m O0 1   rsyn0820m O0 1
syn30m    O0 1   syn40m    O0 1   syn20m02m O0 1   rsyn0830m O0 1
rsyn0840m O0 1   rsyn0820m02m O0 1
squfl*    O0 0   alan O0 0   (minimize)
```

So an objective *below* the published optimum is a **weaker incumbent**, not a
better-than-optimal one, and every such row already carried `status=feasible` —
discopt never claimed optimality on any of them. Verified directly against
`minlplib.solu` (12/12 checks, zero violations):

```
rsyn0805m  status=feasible obj=1116.458  bound=1468.07   bound-vs-optimum: OK
syn40m     status=feasible obj=33.197    bound=1154.48   bound-vs-optimum: OK
rsyn0840m  status=feasible obj=37.587    bound=2178.51   bound-vs-optimum: OK
syn20m02m  status=feasible obj=636.720   bound=3680.85   bound-vs-optimum: OK
```

No dual bound crosses the reference optimum and no incumbent beats it. The
`rsyn0820m02m = 0.0` row is the trivial "build nothing" feasible point. **This
is an algorithm-selection failure, not a certificate failure.**

## Finding 2: convex MINLPs are solved by the spatial global algorithm

discopt certifies these models fully convex at the root, in hundredths of a
second:

```
rsyn0805m  known=True convex=True convex_cons=286/286 in 0.02s
syn40m     known=True convex=True convex_cons=226/226 in 0.01s
rsyn0840m  known=True convex=True convex_cons=484/484 in 0.01s
syn20m02m  known=True convex=True convex_cons=406/406 in 0.01s
```

and then discards the certificate: `solver="mip-nlp"` — the OA / ECP / FP /
[...]

PR #1102, opened 2026-08-20
perf(milp): warm-start the root cut loop so real cut budgets are affordable

jkitchin:
## What

The MILP driver's root cut loop kept each round's optimal basis for the
post-loop root node but never fed it back into the **next** round, so every
round re-derived the augmented LP with a cold primal phase-1.

Appending a cut adds exactly one row and its surplus slack (`coeffs·x − s = rhs`,
`s ≥ 0`); `extend_basis` makes that slack basic, which leaves the prior basis
nonsingular and still dual-feasible (the slack's cost is 0). A warm **dual**
re-solve is therefore the textbook re-optimizer — it pivots in the few violated
cut rows instead of running a fresh phase-1 over the whole matrix. The loop now
does that, falling back to the cold `solve_lp_root_csc` on `IterLimit`/`Numerical`
exactly as the node path already does.

This is the gap behind *"we cannot be using HiGHS because our solver is too
weak"*: the `lp_nlp_bb` convex-MINLP master fell back to HiGHS because discopt's
own MILP dual side could not afford a real cut budget.

## Measurement

`rsyn0840m` OA master (n=280, 1118 rows, 104 integers, reference optimum
**−482.206969**). Every arm asserted `bound ≤ optimum` and, when it claimed
optimality, `objective ≥ optimum − 1e-4`.

| | before | after |
|---|---|---|
| root cut loop, `rc=500 r=15` | 24.2 s of cold solves | **6.1 s for the whole run** |
| `Phase1Pivots` in the root probe | 16127 | **0** |
| full solve `rc=200 r=5` | never closed (630k nodes, bound stuck −802.77) | **optimal −482.207**, 127329 nodes |
| full solve `rc=500 r=15` | — | **optimal −482.207, 22559 nodes** |
| root gap closed (`rc=500 r=15`) | — | **78.0 %** (HiGHS on the same master: 86.1 %) |

Cut budgets that were previously unaffordable now pay off. **Node counts are the
figures to trust** — they are load-independent; the wall numbers were taken under
measurable background load (CLAUDE.md §9).

## The instrument that hid this

`primal.rs`'s `assemble`/`failed` hardcoded `iters: 0`, so **every** cold primal
solve reported doing no work. The driver sums that field into `lp_iters`, which
is how a run that executed 327k pivots reported 212 iterations — an instrument
that measured nothing and was believed (CLAUDE.md §6). It made the cut budgets
look free during the earlier investigation. `Simplex` now counts its pivots and
reports them; the field is purely observational, nothing reads it back.

## Classification (CLAUDE.md §5)

Bound-changing in the strict sense: the LP *optimum* does not depend on the
starting basis, but which optimal **vertex** is reached can, and Gomory cuts are
read off that basis — so the cut pool, and hence the tree, can differ. It is the
same latitude the post-loop `root_warm_basis` path already had. No new
relaxation, no new cut family, no weakened guard.

## Verification

- `cargo test -p discopt-core`: **636 passed, 0 failed**.
- New regression test `root_cut_rounds_reoptimize_warm_instead_of_cold_solving`
  asserts `RootCutWarmReopt == RootCutRounds − 1` (with `rounds >= 3` so it
  cannot pass vacuously) and that the cut loop does not move the certified
  optimum vs a `root_cuts=0` reference solve. Fails before the change
  (*"12 cut rounds but only 0 warm re-optimizes"*), passes after.
- `driver_matches_golden`: status / nodes / objective / bound **bit-identical**
  on all six panel cases. Only `lp_iters` moved, and the doc comment records both
  causes with the intermediate measurement (binary_knapsack 1 → 5 from the honest
  count, 5 → 2 from the warm loop; cuts_firing 1 → 8 → 2).
- `dual_infeasible_warm_start_falls_back_to_cold` asserted `warm.iters == 0` as a
  *proxy* for "the cold fallback ran" — a proxy that only held because the primal
  under-reported. It now arms the profiler and asserts `DualPrepRejectDualInf == 1`,
  i.e. the mechanism itself. (Not a weakened check: a broken instrument was
  doubling as the test oracle.)
- `pytest -m smoke`: **1104 passed, 1 skipped, 2 xpassed**.
- Adversarial suite and a corpus soundness panel (dual bound vs `minlplib.solu`
  oracle on every in-repo instance, plus the convex-MINLP families the change
  serves) are running; results will be posted in a follow-up comment before merge.

## `expel_zero_artificials` stays OFF

Along the way I found a second real defect: `primal.rs`'s basis extractor drops a
basis slot when a basic artificial has no zero-valued singleton substitute,
emitting `basic_vars.len() < m`, which makes `PreparedDual::prepare` reject the
node basis on shape (measured: 619/620 rejections, 556 pivots/LP on this master).
The existing `expel_zero_artificials` option fixes that — but once the root loop
was warm, flipping it was **exactly neutral** on node count and wall. Sound but
not net-positive is not a graduation (CLAUDE.md §5), so it stays off; the comment
at the call site now records that measurement instead of implying the option is on.

### DIFF
--- a/crates/discopt-core/src/bnb/milp_driver.rs
+++ b/crates/discopt-core/src/bnb/milp_driver.rs
@@ -703,12 +703,74 @@ pub fn solve_milp_lazy_hooked(
             if total_cuts >= opts.root_cuts {
                 break;
             }
+            crate::profile::incr(crate::profile::Ctr::RootCutRounds);
             let root = {
                 let _t = crate::profile::Timer::new(crate::profile::Phase::RootSolve);
-                // Solve the root relaxation directly from the working CSC (T3b5: no
-                // dense matrix). Bit-identical to `solve_lp_root` (gated by
-                // `driver_matches_golden`).
-                solve_lp_root_csc(&csc_w, m_w, n_w, &c_w, &l_w, &u_w, &b_w, &node_simplex)
+                // Re-optimize from the PREVIOUS round's optimal basis instead of
+                // re-deriving the augmented LP from scratch. Appending a cut adds a
+                // row and its surplus slack; `extend_basis` makes that slack basic,
+                // which leaves the basis nonsingular and still dual-feasible (the
+                // slack's cost is 0), so a warm DUAL re-solve is the textbook
+                // re-optimizer — it pivots in the few violated cut rows rather than
+                // running a fresh primal phase-1 over the whole matrix.
+                //
+                // The loop already kept `root_basis` for the post-loop root node; it
+                // simply never fed it back in, so every round paid a cold solve.
+                // Measured on the rsyn0840m OA master at `root_cuts=500,
+                // cut_rounds=15`: 14 cold root solves = 23.1 s of the 24.2 s cut
+                // loop, 16127 phase-1 pivots, `EntryColsWarm = 0`.
+                //
+                // Bound-safe: the LP optimum does not depend on the starting basis,
+                // and the warm solve falls back to the cold `solve_lp_root_csc` on
+                // IterLimit/Numerical exactly as the node path does. Only which
+                // optimal vertex is reached — hence which Gomory cuts are read off
+                // the basis — can differ, the same latitude the post-loop
+                // `root_warm_basis` path already has.
+                // Borrowed, not taken: a round whose LP comes back non-Optimal
+                // breaks out of the loop *without* refreshing `root_basis`, and the
+                // post-loop root node still wants the last good basis to warm-start
+                // from. Taking it here would silently cold-solve that root.
+                match root_basis.as_ref() {
+                    Some(&(ref b, basis_n)) => {
+                        crate::profile::incr(crate::profile::Ctr::RootCutWarmReopt);
+                        let b = if basis_n < n_w {
+                            extend_basis(b.clone(), n_w)
+                        } else {
+                            b.clone()
+                        };
+                        let lp = LpView {
+                            a: &[],
+                            m: m_w,
+                            n: n_w,
+                            c: &c_w,
+                            l: &l_w,
+                            u: &u_w,
+                        };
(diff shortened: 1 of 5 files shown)


jkitchin:
## Verification results

**CI**: all 9 non-skipped checks pass (Rust, Python fast, correctness fast subset, claim-boundary serial, AMP coverage, lint+typecheck, codecov patch/project).

**Local**
- `cargo test -p discopt-core`: 636 passed, 0 failed.
- `pytest -m smoke`: 1104 passed, 1 skipped, 2 xpassed.
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py`: 19 passed.

**Corpus soundness panel (the cert-clean half of the §5 gate)**

Every in-repo MINLPLib instance with a `minlplib.solu` reference optimum, solved
end-to-end at `time_limit=30` on this build, asserting the dual bound never
crosses the reference optimum and no claimed optimum beats it:

```
63 instances   118 executed checks   0 violations
48 optimal   6 feasible (time limit)   rest certified at the root
```

The probe prints its executed-assertion count and exits non-zero on zero checks,
so a silent no-op panel cannot read as a pass (CLAUDE.md §6).

**Convex-MINLP class (the net-positive half)**

Spot-checked on the `rsyn`/`syn` family end-to-end at `time_limit=60`, default
settings. `rsyn0805m` (1.8 s), `rsyn0810m` (3.9 s), `syn30m` (1.6 s) and
`syn20m02m` (6.8 s) now certify optimality at the root against their published
optima. The full `#1066` re-run is still going and will be posted there rather
than here, since the remaining rows are about the **default cut budget**
(`root_cuts=16, cut_rounds=1`), which this PR deliberately does not change — that
is a separate bound-changing default needing its own graduation panel. What this
PR establishes is that a real budget is now *affordable*, which it was not before.

Issue #120, opened 2026-06-14
Nonconvex pure-integer NLP: invalid dual bound certifies a suboptimal point as global optimum (nvs16)


jkitchin:
## Summary

On a **nonconvex, pure-integer NLP**, `Model.solve()` certifies a **suboptimal point as globally optimal** by reporting an **invalid dual bound that lies _above_ the true optimum**. This is a soundness bug (silent wrong answer with `status=optimal`), not just a weak-bound / slow-convergence issue.

Reproducer: MINLPLib `nvs16`.

- discopt: `status=optimal`, `objective = bound = 14.203125`, `nodes = 1`
- True optimum (SCIP and Couenne agree): **0.703125**

Since `14.203125 > 0.703125`, the reported value is not a valid lower bound — a valid lower bound must be `≤` the global optimum. The gap closes at the root on this invalid bound and the local point is certified global.

## Reproduce

```python
import discopt.modeling as dm
m = dm.from_nl("nvs16.nl")  # MINLPLib; 2 integer vars in [0,200], 0 constraints
r = m.solve(time_limit=30)
print(r.status, r.objective, r.bound, r.node_count)
# optimal 14.203125 14.203125 1     <-- WRONG (true optimum 0.703125)
```

`nvs16` is a Beale-like unconstrained box-bounded integer NLP. Objective:

```
min (1.5 + x0*(1 + x1))^2 + (2.25 + x0*(1 + x1^2))^2 + (2.625 + x0*(1 + x1^3))^2
    x0, x1 integer in [0, 200]
```

## Root cause

The convexity machinery is **not** at fault — it classifies the model correctly:

```python
from discopt._jax.convexity import classify_model
classify_model(m)                         # (False, [])  -> nonconvex
classify_model(m, use_certificate=True)   # (False, [])  -> nonconvex
```

`solve()` therefore sets `tree.set_nonconvex(True)` and builds the alphaBB underestimator. The bug is in **lower-bound mode selection** (`solver.py`):

- `mccormick_bounds` defaults to `"auto"`. For a nonconvex model with an objective, `"auto"` resolves to **`"nlp"`** (`solver.py:2097-2106`):

  ```python
  if _mc_mode == "auto":
      if _model_is_convex:
          _mc_mode = "none"
      elif model._objective is not None:
          # "nlp" mode solves a convex relaxation NLP for valid lower bounds.
          _mc_mode = "nlp"
      ...
  ```

[...]

PR #1110, opened 2026-08-22
docs(mecp): MECP/MECI readiness assessment, formulation guidance, and probe suite

jkitchin:
## Summary

Research assessment of **minimum-energy crossing point (MECP) / minimum-energy conical intersection (MECI)** optimization as a target problem class for discopt, plus the practitioner-facing guidance page that makes it usable. Docs and probe scripts only — no solver code touched, no solver behaviour changed.

The headline: **discopt already solves this class to a certified global optimum, and no established MECP method returns a bound at all.** Gradient projection, Lagrange–Newton, the smooth penalty method and branching-plane updating are all local; the state of the art for finding *multiple* crossing points is stochastic exploration (multistate metadynamics). So the capability is already there — what was missing is knowing how to pose the problem.

Because that turns out to matter enormously. The same physics spans ~10⁴ in tractability across algebraically equivalent formulations, and **the formulation quantum chemistry standardized on is the one that fails**:

| Formulation | certified | nodes | wall |
|---|---|---|---|
| diabatic MECI (`min (W₁₁+W₂₂)/2 s.t. W₁₁−W₂₂=0, W₁₂=0`) | ✅ | 0 | 0.03 s |
| spin-crossing MECP (`min W₁ s.t. W₁−W₂=0`) | ✅ | 5 | 0.5 s |
| characteristic-polynomial (`min T/2 s.t. T²−4D ≤ 0`) | ✅ | 7 | 0.85 s |
| adiabatic with `sqrt` | ✅ | 1 | 1.53 s |
| **Levine/Coe/Martínez penalty** (`min (E₁+E₂)/2 + σΔE²/(\|ΔE\|+α)`) | ❌ | 571 | 300 s, **rel. gap 3.4×10³** on 2 variables |

Relaxing a squared numerator over a denominator bounded below by α≈0.025 is near worst-case. Same physics, ~10⁴ difference.

**What the certificate buys, quantified.** On a spin-crossing model whose lower state has two conformer wells (two seam basins, 0.278 apart), from 200 random starting geometries:

| Method | reached the seam | found the **global** basin |
|---|---|---|
| Smooth penalty (σ escalated, L-BFGS-B) | 200/200 | **127/200 (63.5 %)** |
| SLSQP on the constrained form | 199/200 | **78/199 (39.2 %)** |
| **discopt** | — | **certified, every time, 5 nodes, 0.5 s** |

Both local methods reliably reach the seam; neither reliably reaches the bottom of it. Since the MECP energy enters a nonadiabatic-TST rate through a Boltzmann factor, a 0.278 eV basin error is ~4.6×10⁴ in the rate at 300 K — not a slightly worse answer, a rate wrong by four to five orders of magnitude.

**Dimensional ceiling** (2 → 9 internal coordinates, 300 s cap): certified through n=8 in 194 s, growing ~×2.7 per coordinate; n=9 reached a 0.6 % gap when the clock ran out. 3N−6 means triatomics through ~5-atom molecules outright, larger systems only in a reduced active subspace — and the certificate then applies to the subspace, not the full geometry.

**Free Cartesian coordinates cannot be certified**, and this is structural rather than a defect: every translation and rotation of the MECP is also an MECP, so the optimal set is a 6-manifold and no box isolates it. Measured: dual bound −0.109858 against a true optimum of 0.819996 after 180 s, versus a certificate in 5 nodes / 0.96 s in internal coordinates. Gauge fixing restores it (329 nodes, 14.6 s). Internal or gauge-fixed coordinates is a hard requirement, now documented.

**Surrogate routes** (real energies are black boxes, so a certified answer is necessarily about a surrogate): symbolic `dm.exp`/`dm.udf` certifies in 47 nodes / 1.6 s; `solver="direct"` gives a good answer fast with no bound, honestly reported; `discopt.nn` with two tanh 12×12 nets models it correctly but returns a dual bound of −20.6 after 601 s. That last one is the only place genuine performance work would change what is possible.

Contributes to no existing issue — this opens a new area. Two narrow follow-ups are identified in §4 of the report and are **not** addressed here (`sqrt`'s endpoint tangent at zero; `mccormick_bounds='lp'` being inert on the CustomCall/`nn` paths, which cross-references #120).

## Correctness

- [x] Cannot produce a false certificate (or: N/A — no solver-math change) — **N/A, docs and scratchpad probes only.** No file under `python/discopt/` or `crates/` is touched.

Every probe asserts soundness independently: no certified bound may exceed a known feasible seam point. **Zero violations** across 9 scaling instances plus every formulation variant, checked against an analytic oracle (LVC), an exact grid-and-bisect oracle (n=2, ~2900 crossings located), and a sample-and-project oracle (n>2).

- [ ] `pytest -m smoke` → **running at the time of opening this PR; result will be posted as a comment.** Not a gate on this change (no importable code is modified), but stated rather than skipped.
- [ ] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **not run.** N/A — no solver behaviour changed.
- [x] `ruff check` / `ruff format --check` clean → all 8 probe scripts clean
- [x] `jupyter-book build docs/` → **build succeeded, zero warnings.** The first attempt reported one warning, a MyST cross-reference to a heading anchor this book's config does not enable; removed.

## Regression test

- [x] Added a regression test that fails before / passes after (or: N/A — docs/refactor only) — **N/A, docs only.**

The eight probe scripts are the verification artifact instead. Each ends with an executed-check count and **exits non-zero if it checked nothing**, per CLAUDE.md §6 — the measurement-discipline rule that exists because probes that silently measure nothing have been believed before. Counts from the runs behind this PR: exp1 21 checks / 2 failures (both the penalty formulation, expected and reported), exp2 403, exp3 8, exp3b 4, exp3c 6, exp4 8, exp5 7, `mecp_models` 206.

## What is honestly incomplete

Stated rather than buried:

- **Two claims corrected mid-campaign, both recorded in the report.** (1) I predicted the characteristic-polynomial form would defeat local NLP solvers, because its constraint gradient `∇(T²−4D) = 2(E₁−E₂)∇(E₁−E₂)` vanishes identically on its own feasible set, so LICQ fails everywhere. SLSQP does fail on 47 % of starts — but `trust-constr` handles the degeneracy completely (60/60, median error 2.7e−10). The claim was too strong and is retracted in §3.5. (2) My own probe bug: exp3 passed a coordinate *vector* where `dm.custom` requires scalar arguments, so a non-scalar leaf disqualified the MCBox path and the reduced-space route was never exercised — its "no certificate" result was meaningless. `exp3b` retests it correctly (it does engage, 299 nodes, and still does not certify).
- **No symmetry measurement.** discopt has no orbit/symmetry reduction in the solver core (only an advisory `_detect_symmetry` in `llm/reformulation.py`), and real molecules with equivalent atoms have permutationally equivalent crossing points. The models here were deliberately symmetry-broken, so this is unmeasured and should be sized before claiming readiness on real molecules.
- **The discrete structure is untested.** All formulations here are purely continuous. Which *pair* of states crosses, which conformer basin, which multiplicity — these are natural disjunctions, and posing MECP as a GDP rather than N separate NLPs is the obvious next formulation question. Not attempted.
- **Bibliography DOIs were verified against sources, not recalled.** The Yarkony 1993 DOI I first wrote from memory was wrong (`j100119a023`; the real one is `j100119a026`) and was corrected before committing.

## Files

- `docs/dev/mecp-readiness-2026-08-22.md` (581 lines) — the assessment: problem class, formulation map, five measurement sections, gaps with recommended next steps in priority order
- `docs/notebooks/mecp_crossing_points.md` + TOC entry — practitioner guidance; all three code blocks executed and their printed outputs are the actual values (8 assertions; an earlier draft had a hand-sorted variable order and a wrong final digit, both corrected)
- `scratchpad/mecp/` (8 scripts, ~1900 lines) — three model PES families with two independent oracles, and the five experiments. Each surface is written **once** against an injected `exp` (`np.exp` evaluates, `dm.exp` builds the DAG) so the oracle and the solver input cannot drift apart; an earlier version had two implementations and that risk was removed.
- `docs/references.bib` — 11 chemistry entries (the file had 220, none on conical intersections)

https://claude.ai/code/session_01QXHqRuvTdQX1i5sv8GT1Ng

---
_Generated by [Claude Code](https://claude.ai/code/session_01QXHqRuvTdQX1i5sv8GT1Ng)_


jkitchin:
Smoke result, filling the checkbox left open when this PR was opened:

```
pytest python/tests/ -m smoke -q --timeout=600
1087 passed, 19 skipped, 8949 deselected, 2 xpassed, 26 warnings in 370.10s
EXIT=0
```

Green. As noted in the description this isn't a gate on this change — no file under `python/discopt/` or `crates/` is modified — but it was stated as running rather than skipped, so here is the number.

The 26 warnings are pre-existing (unbounded-variable advisories from `discopt/debug/__init__.py` and similar), not introduced here.

---

jkitchin:
@bernalde this is a preliminary investigation on MECP in discopt. I don't think this includes any discrete examples yet. If you have other problems you think would be worth exploring, leave them in a comment!

jkitchin:
## Added: the discrete formulation (`18e80ee`)

The description flagged "the discrete structure is untested" as an honest gap. Closing part of it with a sixth experiment, since the disjunctive framing is what makes this an MINLP/GDP problem rather than an application note.

**What it does.** Experiments 1–5 are all purely continuous: name two surfaces, find their lowest crossing. The question a chemist usually has is one level up — a molecule has several low-lying states, and what matters is the lowest crossing over *any* pair, because that is the one the reaction goes through. That is a disjunction:

```
min_{x,y}  E
s.t.  y_p = 1  =>  W_i(x) = W_j(x)  and  E = W_i(x)     for each pair p=(i,j)
      sum_p y_p = 1,   y_p in {0,1},   x in [x^L, x^U]
```

`scratchpad/mecp/exp6_multistate_gdp.py` poses this as a big-M MINLP over three states (three candidate pairs, two coordinates) and compares it against the obvious baseline — enumerate the pairs, one continuous solve each:

| Route | energy | certified | nodes | wall |
|---|---|---|---|---|
| enumeration (3 separate solves) | 1.270684 | yes | 11 | **2.14 s** |
| disjunctive MINLP (1 solve) | 1.270684 | yes | 51 | **5.70 s** |

Both certify and both agree, as equivalent formulations must. The MINLP selects pair (0,1), and the selected disjunct is genuinely degenerate at the returned point (`|W₀−W₁| = 5.3e-15`) — which doubles as the check that the big-M isn't so loose that it admits a spurious "crossing". 8 executed checks, 0 anomalies.

**The result is negative at this size, and reported that way: enumeration wins by 2.7×.** Three pairs is exactly the regime where it should — three tiny independent problems beat one MINLP carrying binaries and big-M slack for all of them simultaneously. Recorded anyway for two reasons:

- The interesting regime is *many* states, where pairs grow as O(K²) and a single tree can prune a whole pair on a bound instead of solving it. **The crossover was not measured**; this is the datapoint it would start from.
- One of the three pairs here — the two upper states — is **infeasible**, they never cross inside the box, and enumeration spends a whole solve discovering that. In a many-state problem most pairs are of that kind, which is precisely the structure a disjunctive formulation should exploit and enumeration cannot.

**Two shortcuts, both taken for time and both stated in the report rather than buried:** the big-M is a grid maximum × 2 rather than a rigorous interval bound (it is validated *after* the solve via disjunct degeneracy, not derived soundly), and this is hand-rolled big-M instead of discopt's native GDP path (`gdp_method`), which would take the disjunction directly. Either would need fixing before this is more than illustrative.

Report gains §3.7; the guidance page gains a "Several states" section. `jupyter-book build docs/` clean, `ruff` clean.

---

bernalde:
I ran the pair-selection experiment again using DiscOpt's native GDP interface rather than manually expanding the implications.

The distinction matters. When the model uses `m.either_or(...)`, `gdp_method="big-m"` computes a separate M for each constraint from interval bounds. `gdp_method="mbigm"` attempts LP tightening for linear constraint bodies and falls back to interval bounds for nonlinear ones. The experiment in this PR hand-constructs the big-M constraints, so it bypasses those mechanisms.

For the original three-state problem, using the second of two runs on the same machine:

| Formulation | M | Variables / constraints | Nodes | Solve result |
|---|---:|---:|---:|---:|
| Serial enumeration | n/a | 3 separate problems | 11 | certified in 5.15 s |
| Native GDP big-M | 17.16 to 23.60, constraint-specific | 6 / 13 | 51 | certified in 7.66 s |
| Native GDP MBigM | same effective values | 6 / 13 | 51 | certified in 7.91 s |
| Native nonlinear hull | n/a | 15 / 34 | 167 | not certified after 120 s |

All certified formulations found pair (0,1) at E = 1.27068427. The hull run stopped with a feasible incumbent of 1.54854 and a lower bound of 1.00052, so it did not find or certify the known optimum.

Adding one state while leaving the original three surfaces unchanged gives six candidate pairs. Native automatic big-M then took 10.99 s, compared with 12.60 s for serial enumeration. Thus the serial-enumeration versus joint-MINLP crossover occurs by four states in this family.

The automatic M values are sound but can still be loose because interval evaluation does not cancel common terms. For the two original upper surfaces,

`W2 - W1 = 0.75 + 0.22*q0^2`.

With `|q0| <= 2.6`, a sufficient M including the 1% margin is about 2.26, whereas the general interval calculation returned 17.16. MBigM does not improve this case because the unsimplified constraint body is nonlinear, so it falls back to interval arithmetic.

The nonlinear hull formulation did not help here. These disjuncts contain nonconvex nonlinear equalities, so the perspective reformulation is not an ideal convex hull. It also introduces disaggregated variables and expressions of the form `y*f(v/(y+eps))` involving the quartic and exponential surface terms. Those expressions were substantially harder for the current spatial relaxations than the big-M model.

My conclusions are:

1. The native GDP path should be used for this comparison because DiscOpt already computes valid, constraint-specific M values.
2. Native big-M confirms the crossover at four states.
3. Nonlinear hull is not competitive on this instance.
4. Simplifying state-difference expressions before interval evaluation, or otherwise deriving tighter bounds for `Wi-Wj`, appears more promising than nonlinear hull for this problem.

PR #1121, opened 2026-08-23
test(#1050): finish the no-op benchmark suite — delete the interop stubs, guard the class, fix the sense-blind bound check

### LINKED ISSUE #1050: Pre-release benchmark run asserts nothing: 118 stub tests in discopt_benchmarks never execute
## Scope (clarified 2026-08-16)

**This is not a request to run benchmarks in CI.** Benchmarks are run before a
release, or when we want to confirm things are on track — not per-CI, which would be
far too expensive. That policy is right and nothing here asks to change it.

The problem is narrower and survives that policy: **a deliberate pre-release
benchmark run asserts nothing in these two files.** They do not skip because a
dependency is missing or because CI declined to run them — they skip
unconditionally, every time, including on exactly the pre-release run this project
relies on:

```
$ pytest discopt_benchmarks/tests/test_correctness.py \
         discopt_benchmarks/tests/test_interop.py
118 skipped in 0.70s
```

0 passed, 0 failed, 0.70 s. `solve_instance` (`test_correctness.py:39`) raises
`NotImplementedError` unconditionally and all 98 callers catch it and skip with the
false reason "discopt not yet available". So running benchmarks before a release —
the intended safety net — does not exercise them at all, and the suite reports
`321 passed, 118 skipped` either way.

Left open on that basis. Priority is low: the certificate invariant these files
nominally cover is genuinely covered by `python/tests/test_correctness.py` and the
CI correctness jobs, so this is a hollow duplicate, not an open correctness hole.

---

## Summary

`discopt_benchmarks/tests/test_correctness.py` and
`discopt_benchmarks/tests/test_interop.py` contain **118 tests that have never
executed**. Every one skips unconditionally. The suite reports green, and
`RELEASE.md` §2 lists `pytest discopt_benchmarks/tests/ -v` as a release gate, so
this reads as coverage that does not exist.

Measured on `release/v0.8.0` + PR #1049 (`4167982e`), worktree asserted at
`discopt.__version__ == "0.8.0"`:

```
$ pytest discopt_benchmarks/tests/test_correctness.py \
         discopt_benchmarks/tests/test_interop.py
118 skipped in 0.70s
```

118 skipped, **0 passed, 0 failed**, in 0.70 s — nothing solved, nothing asserted.
The full benchmark suite reports `321 passed, 118 skipped`; those 118 are exactly
these two files, so every skip in that suite is a dead stub rather than an
optional-dependency gate.

## Mechanism

`test_correctness.py:39` is a stub that raises unconditionally:

```python
def solve_instance(name: str, **kwargs) -> MockSolution:
    """Stub: replace with actual discopt API."""
    # from discopt import solve, load_problem
    # problem = load_problem(f"instances/{name}.nl")
    # return solve(problem, **kwargs)
    raise NotImplementedError("Replace with actual discopt API")
```

and all 98 of its tests wrap the call in `except NotImplementedError:
pytest.skip("discopt not yet available")`. The skip reason is false — discopt *is*
available; the harness is a placeholder. `test_interop.py`'s 20 tests are the same
shape with their bodies commented out and a bare
`pytest.skip("discopt Rust bindings not yet available")`.

## Why this matters beyond the count

`test_correctness.py` is where the certificate invariant is supposed to be checked:

- `test_bound_validity` — "Lower bound must never exceed the true optimum"
- `test_solution_feasibility` — returned point satisfies all constraints
- `test_optimal_value` — objective matches the known optimum

That is `CLAUDE.md`'s non-negotiable invariant, and this file asserts none of it.

**The invariant itself is not unguarded** — `python/tests/test_correctness.py` is a
real suite (26 tests, no stubs, imports `discopt.modeling`) and the CI "Python
correctness" jobs run it. So this is a *hollow duplicate*, not an open correctness
hole. It is still worth fixing: it inflates the apparent gate, and the two files
are where a reader looks for benchmark-level correctness coverage.

The clearest illustration is `e51a60e3` (2026-06-16), which corrected nine wrong
entries in `KNOWN_OPTIMA` against the authoritative `minlplib.solu` — careful,
correct work on a data table that no executing test consumes.

## Proposed fix

1. Implement `solve_instance` against the real API (`discopt.interfaces.from_nl` +
   `Model.solve()`), drawing instances from
   `~/Dropbox/projects/discopt-minlp-benchmark/minlplib/nl/` with `minlplib.solu`
   as the oracle.
2. Scope the runtime. 98 tests at the current `time_limit=3600` is not a suite
   anyone will run; select a small-instance subset via `problems_small.txt` and
   mark the rest `slow`.
3. Delete the `except NotImplementedError: pytest.skip(...)` arms. A missing API
   should fail, not skip — a skip on a correctness gate is indistinguishable from
   a pass.
4. For `test_interop.py`, either implement against the actual `discopt._rust`
   surface or delete the file if the helpers it references
   (`create_test_bounds`, `create_test_sparse`, `get_all_array_outputs`) were
   never real.
5. Add a guard that fails if either file's skip count equals its collected count,
   so a future regression to a no-op is loud (`CLAUDE.md` §6).

## Not a v0.8.0 blocker

The invariant is covered by `python/tests/test_correctness.py` and CI. Filing this
rather than fixing it inside PR #1049 because implementing 118 real tests plus
instance selection is well beyond a documentation PR — the `CLAUDE.md` "Working on
an issue" §3 split exception.

jkitchin:
Closes #1050.

#1050 listed five items. #1120 landed items 1–3 (a real `solve_instance`, a scoped
runtime, and the deletion of the `except NotImplementedError: pytest.skip(...)` arms).
This finishes items 4 and 5, and fixes the first real defect the now-executing suite
surfaced.

## Item 4 — `test_interop.py` deleted, on measurement rather than preference

The issue offered a choice: implement it against the actual `discopt._rust` surface, or
delete it if its helpers were never real. **0 of 14 exist.** `create_test_bounds`,
`read_bounds`, `create_test_sparse`, `read_sparse`, `get_all_array_outputs`,
`get_shared_buffer`, `check_buffer_modified`, `create_batch`, `BatchConfig`,
`TreeManager`, `benchmark_roundtrip`, `trigger_test_panic`, `evaluate_expression_rust`,
`solve_lp` — none are among the module's 30 exports. The file was a spec for an API
that was never written.

Its one concrete claim about the boundary was also wrong: it asserted CSR
`indices.dtype == np.int32`, while the real entry point takes `i64`
(`crates/discopt-python/src/lp_bindings.rs:612`). Had it ever run it would have failed
against a *correct* implementation. Its header premise ("Rust ↔ JAX Interop") names a
layer that is not on the solve path.

Deleting is only right if nothing is lost, so all 20 tests were mapped to executing
coverage first. `python/tests/test_batch_dispatch.py` (34 tests) is a class-for-class
superset — dtype, zero-copy (the identical `ctypes.data` pointer-identity assertion),
batch shapes, latency (warmed, best-of-5, size-scaled — strictly better than the stub's
flat 100 µs) and invalid-input rejection. `test_simplex_lp.py` covers the LP infeasible
status; `relaxation_harness.assert_containment` covers the sampled McCormick soundness
property. Two tests had no counterpart and are dropped explicitly, not silently:
`test_overhead_fraction` (asserts `result.python_time_fraction`, a field that does not
exist) and `test_evaluation_matches_across_layers` (a Rust-vs-JAX comparison the
architecture retired). The mapped files were run rather than assumed: 34 / 94 / 69 / 59
passed.

## Item 5 — a no-op suite guard, aimed at the class

The issue asked for a guard that fires when a file's skip count equals its collected
count. A literal count check is the wrong shape here: this suite has *legitimate*
all-skipped modules (GDPlib, external solvers, GPU), so it would false-positive on
ordinary developer commands, and a guard that gets turned off is worth less than none.

`discopt_benchmarks/tests/conftest.py` enforces the two independent signatures of the
actual defect instead:

1. **A placeholder skip *reason* fails the session** — "not yet available", "replace
   with actual", `NotImplementedError`. This is the discriminator the count check
   lacks: it separates "the dependency is absent" (honest) from "the harness was never
   written" (the bug), and catches a future copy of the idiom anywhere in the suite.
2. **A module declaring `MUST_EXECUTE = True` that executes nothing fails** — the
   backstop that survives a rewording of the reason. Suppressed for a module narrowed
   by `-k`/`-m`, and inside xdist workers, whose partial view carries no information.

The guard is itself an instrument, so per CLAUDE.md §6 it has tests that prove it
fires. `discopt_benchmarks/tests/test_1050_no_op_suite_guard.py` (6 tests) copies the
real `conftest.py` into a temp dir and runs pytest in a subprocess against synthetic
modules, asserting **both** arms fire and that the guard stays silent on an honest
`importorskip`, on a module where one test ran, and under a narrowing `-k`.
Retrospective check: pointed at the un-deleted `test_interop.py`, the guard exits 1 and
names all 20 tests — it would have caught this issue the day it was introduced.

## The bug the executing suite caught in itself

Making a suite run only pays if you read what it says. The first full post-#1120 run
produced two failures that read like solver soundness violations:

```
INVALID BOUND: procurement1large bound=4.02373510e+04 optimal=3.80217975e+03
INVALID BOUND: procurement1mot   bound=3.59368811e+03 optimal=2.91541658e+02
```

They are not. `procurement1mot` reported `obj=251.75`, ref `291.54`, `bound=3593.69` —
an incumbent below the reference and a bound above it, impossible for a minimization
and consistent for a maximization. `_objective.sense` is `MAXIMIZE` for
`procurement1large`, `procurement1mot`, `procurement2mot`; the other 26 available
instances minimize.

The defect is in `test_bound_validity`, which **#1120 wrote**: it asserted
`sol.bound <= expected + ABS_TOL` unconditionally, under the comment "For
minimization". `.solu` records an optimum without a sense, so the check reads as
plausible while testing the wrong inequality — and it is wrong in both directions at
once, failing valid upper bounds while being structurally unable to catch an invalid
one (an invalid *upper* bound falls **below** the optimum, a condition the assertion
does not express). It now reads the sense from the parsed model and asserts the
matching direction, and a new smoke test asserts the corpus still contains both senses
so neither arm silently stops being exercised.

**Retraction (CLAUDE.md §11):** PR #1120's body stated "No bound exceeded its reference
optimum on any of the 29." That was produced by the sense-blind check and was not a
valid verification for the three maximizing instances. The sense-aware run below is the
verification of record.

## Verification

- `discopt_benchmarks/tests/test_correctness.py -m correctness`: **exit 0**, 100 collected, 13 skipped -- every skip stating a true condition (`status=feasible`/`time_limit`, `No bound reported`, `No feasible solution`). The same command before this commit was **2 failed** (the two above).
- `discopt_benchmarks/tests/`: **exit 0**, 435 tests, 0 failed, 12 skipped, and the guard fired nowhere -- including on `test_correctness.py`, whose 100 tests the root `addopts` deselects by marker, exercising the deselection-suppression arm on a real command
- `discopt_benchmarks/tests/test_1050_no_op_suite_guard.py`: 6 passed
- `pytest -m smoke`: 1104 passed, 1 skipped, 2 xpassed in 163 s
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py`: 19 passed in 144 s
- No Rust changes, so `cargo test -p discopt-core` was not required.

Test-only change: no solver code is touched, so the bound-neutrality regime of
CLAUDE.md §5 does not apply.


jkitchin:
### CI caught the sense probe making the same mistake the bound check made

The first push failed `Python correctness (fast subset)`:

```
FAILED discopt_benchmarks/tests/test_correctness.py::test_the_corpus_contains_both_objective_senses
  AssertionError: no maximize instance available; the max arm is untested
```

The probe was right about the gap and wrong about how to state it. It asserted
that a maximizing instance *exists* — a property of a machine carrying the MINLPLib
snapshot, not of this repository. CI resolves only the 66-file in-repo corpus, and
none of the five `KNOWN_OPTIMA` entries in it maximized. Asserting a local-corpus
property from a test that runs everywhere is the same class of error as the bound
check itself: a claim that reads as general and is not.

Two changes make the arm genuinely covered rather than merely asserted:

1. **The direction decision is now a pure function**, `bound_is_valid(bound,
   optimum, maximize=...)`, unit-tested in both directions with no corpus, no solve
   and no environment dependence. The cases include the two real bounds that
   exposed the bug (`procurement1mot` 3593.688 vs 291.542; `procurement1large`
   40237.351 vs 3802.180 — both sound upper bounds the old check rejected) and the
   case the old check was structurally unable to catch: an upper bound *below* the
   optimum.
2. **`syn05hfsg` joins `KNOWN_OPTIMA`.** It is the only maximizing instance with a
   `.solu` optimum that already ships in `python/tests/data/minlplib_nl/`, so it
   adds no corpus data, and it solves to optimality in 5.9 s (obj 837.7323677,
   bound 837.7324009, ref 837.7324009) — inside both the 30 s budget and CI's 120 s
   per-test timeout. The maximize arm now runs on CI.

The remaining corpus probe asserts only what holds everywhere — that every
available instance yields a readable sense, so `_is_maximize` raises rather than
quietly answering "minimize" if the sense ever moves — and prints the split.

**Re-verified**

- CI-emulated (`DISCOPT_MINLPLIB_NL` pointed at a nonexistent path): 6 instances
  resolve, 1 maximizing; the whole module exits 0; the two `syn05hfsg` tests pass
  in 5.87 s.
- Full local corpus: 109 tests, 0 failed, 12 skipped, exit 0; probe reports 30
  instances read, 4 maximize, 26 minimize.

Issue #1124, opened 2026-08-27
PR 0: define disjunction semantics and lock existing behavior with regression tests


bernalde:
Parent: #1123  
Cross-repository application track: [jkitchin/pounce#776](https://github.com/jkitchin/pounce/issues/776)  
Gate 0 benchmark: [jkitchin/pounce#794](https://github.com/jkitchin/pounce/issues/794)

## Objective

Define the source semantics that every later MCP/MPEC and continuous-logic lowering must preserve, and lock the existing behavior with regression tests before changing the intermediate representation or solver paths.

This is the first implementation step from #1123. It should be independently reviewable and should not change solver defaults or the feasible set of existing models.

## Contract to make explicit

- `OR`: at least one disjunct is selected/active.
- `SELECT_ONE`: exactly one selector is active, while an unselected predicate may also happen to be true. This is the backward-compatible meaning of the current `Model.either_or(...)` API.
- `TRUTH_XOR`: exactly one source predicate is true, requiring full reification or an explicit strict-negation policy.
- User-visible Boolean identities are distinct from existential compiler auxiliaries introduced during lowering.
- A local continuous solve result is distinct from a discrete result carrying a certified bound or optimality gap.

## Scope

- Add an internal disjunction-semantics enum or equivalent typed representation, initially defaulting existing `either_or` calls to `SELECT_ONE`.
- Update API and developer documentation to use the vocabulary above consistently.
- Preserve the existing projected-union behavior for overlapping disjuncts.
- Add convex and nonconvex regression examples showing that OR/select-one versus truth-XOR is a semantic distinction, not a convexity distinction.
- Exercise both Big-M and hull lowering where supported.
- Distinguish named Boolean identities from generated existential auxiliaries in the relevant metadata/tests.
- Record the local-versus-certified terminology that later result objects must use.

## Acceptance criteria

- [ ] Existing `either_or` models retain their current feasible set and default lowering behavior.
- [ ] Overlapping disjuncts are tested through both Big-M and hull paths.
- [ ] At least one convex and one nonconvex example distinguish `SELECT_ONE` from `TRUTH_XOR`.
- [ ] The public/developer documentation no longer uses ambiguous OR/XOR terminology.
- [ ] Tests pin the distinction between user Boolean identities and generated existential auxiliaries.
- [ ] No new MCP/NCP lowering, solver backend, or solver default is introduced.
- [ ] Existing test suites remain green.

## Out of scope

[...]

PR #1128, opened 2026-08-29
feat(#1124): define disjunction semantics and lock select-one behavior

jkitchin:
## Summary

Makes the meaning of a disjunction an explicit source-level contract instead of something inferred from whichever activation a lowering happens to emit, and pins the existing behavior before any later PR changes the IR. This is PR 0 of the MCP/GDP RFC (#1123).

`DisjunctionSemantics` members are defined **as** the `(activation, cardinality)` pair they stand for — a member's `.value` *is* that tuple — and every lowering branches on the pair, never on the member. `SELECT_ONE = (ONE_WAY, EXACTLY_ONE)`, `OR = (ONE_WAY, AT_LEAST_ONE)`, `EXACTLY_ONE_TRUE = (REIFIED, EXACTLY_ONE)`. Two consequences that matter downstream:

- The axis separating `SELECT_ONE` from `EXACTLY_ONE_TRUE` is **reification, not cardinality** (both are exactly-one). That is the axis a continuous/complementarity lowering has to dispatch on, and a flat three-member enum would hide it.
- Adding a combination later (an optional-mode disjunction is `(ONE_WAY, AT_MOST_ONE)`) becomes an added member rather than a redefinition of what the enum means.

One predicate decides support for every lowering. `_SELECT_ONE_ROWS` — the pair emitted by one-way activation under `sum_k y_k == 1` — is what both `Model.add_disjunction` (`if_then` + `exactly(1)`) and every `_relax.gdp_reformulate` reformulation build; both gate through the same `_require_semantics_supported`, so they cannot drift apart on what "supported" means.

Named `EXACTLY_ONE_TRUE`, not `XOR`: over three or more operands XOR is odd parity rather than "exactly one true", `either_or` is not limited to two disjuncts, and `xor` is already taken in this tree by the pairwise Boolean operator (the GAMS `boolxor` opcode). The `xor=` / `truth_xor` spellings are refused with a message naming both meanings, since Pyomo.GDP's `Disjunction(xor=True)` means select-one.

**Identity vs. auxiliary is enforced, not conventional.** `Model._check_name` — the single choke point every user-facing variable/parameter factory funnels through — refuses names inside the reserved `GDP_AUX_PREFIX`, and the GDP pass allocates auxiliary names against the names already present in the model. Without both, `m.boolean("_gdp_aux_disj_d_0_0")` plus `either_or(..., name="d")` produced four variables under three distinct names.

Also fixes a pre-existing defect found while pinning the paths: a nested disjunction under `hull` died on a bare `AttributeError` (the pass reads `con.body` on a nested `_DisjunctiveConstraint`). It now raises `NotImplementedError` naming big-M as the working path; nesting under big-M is untouched.

Terminology, per acceptance criterion 4: `either_or`'s docstring, the LLM prompt and tool description, and the published GDP tutorial (both its introduction and its summary API table, plus a `\veebar`/XOR rendering) all taught "exactly one group of constraints must hold" — the truth reading of select-one behavior. All now state select-one. The `add_disjunction` selector row is renamed `_disj_{name}_xor` → `_disj_{name}_select` (nothing read that name). The contract lives at `docs/disjunction_semantics.md`, published in `_toc.yml` under Advanced Features, and records the indicator trap, the identity-vs-auxiliary distinction, and the local-versus-certified vocabulary later result objects must use.

Contributes to #1124. Semantics naming and the enum shape follow the decisions in https://github.com/jkitchin/discopt/issues/1123#issuecomment-5463289126.

## Correctness

No solver math changes. `either_or` remains `SELECT_ONE`; no solver default, lowering, or feasible set moves. Behavior neutrality was checked directly rather than argued: the select-one and indicator probes produce **byte-identical** output — objectives, solution vectors, and emitted selector rows — before and after the change (`diff` clean on both).

The behavior changes are all in the refusing direction: a disjunction declaring semantics no lowering implements now raises instead of being served by the select-one machinery (that would return a different feasible set than the model declares and then certify it), a nested disjunction under hull raises instead of dying on an `AttributeError`, and a user name in the reserved auxiliary namespace raises instead of silently colliding. The semantics refusal fires through the full `solve()` path, not just a direct `reformulate_gdp` call.

- [x] Cannot produce a false certificate (no solver-math change; the added paths only refuse)

Current head `cd3b1c3f`:

- [x] `python/tests/test_1124_disjunction_semantics.py` → **40 passed**
- [x] GDP/LLM/display selection (`test_gdp`, `test_gdp1_gdp2_bigm_soundness`, `test_gdp_advisor`, `test_issue1043_hull_perspective`, `test_gdpopt_loa`, `test_gdpopt_heuristics_core_units`, `test_nl_complementarity`, `test_llm_modules`, `test_display_cluster`) → **426 passed**, 3 skipped, 2 xpassed
- [x] `ruff check python/` / `ruff format --check` → clean
- [x] `cargo test -p discopt-core` → N/A, Rust untouched

On the previous head `cd64b4a` (unchanged by this round's fix, which is scoped to the `add_disjunction` gate and one docstring path): `pytest -m smoke` → 1104 passed; `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → 19 passed.

`main` is currently red independently of this PR: smoke on the merged head gives 15 failures (7 `test_1060_native_single_tree`, 6 `test_1141_fractional_node_cuts`, 2 `test_issue_1060_open_bounds_marshal`) plus one in `test_mip_nlp`. Verified byte-identical on `origin/main`'s source with this PR's marker asserted **absent**, so none is this PR's and there is no fix to port. CI's own selections are green here regardless.

## Regression test

`python/tests/test_1124_disjunction_semantics.py` — **40 cases**, all passing; the file errors on collection before the change (the API is absent), so fail-before/pass-after holds.

The discriminating witness the acceptance criteria were missing: the pre-existing `test_hull_overlapping_disjuncts` does **not** discriminate, since its `min x` objective is 0 under either reading. Overlap is now pinned with `min (x−1/2)²  s.t.  [x ≤ 1] ∨ [x ≥ 0]`, which is 0 at x = 1/2 under select-one and 0.25 under truth semantics, across **big-m / mbigm / hull / auto** and the `Disjunct`-block path.

Convex (`[x≤1] ∨ [x≥0]`) and nonconvex (`[z²≥1] ∨ [z≥0]`) witnesses show the distinction is semantic and not convexity-driven, both using named indicators — deselecting a mode does not falsify its predicate. That indicator-facing form is the one users actually trip over, and it was not in the issue's original scope.

Two tests are themselves discriminators rather than assertions that pass either way: the namespace test instantiates the exact colliding name across three spellings and every user-facing factory, and `test_add_disjunction_dispatch_is_pair_based_not_member_identity` widens the emitted-pair set so `OR` becomes servable — verified to **fail** against a restored identity gate and pass against the pair gate.

N/A — no bound, cut, or relaxation is touched.


jkitchin:
@bernalde — review please, since this implements PR 0 from your RFC (#1123) and follows the decisions in https://github.com/jkitchin/discopt/issues/1123#issuecomment-5463289126. (GitHub won't let me add you as a formal reviewer — review requests are limited to repo collaborators.)

Three things worth your eye specifically, all departures from #1124 as written:

1. **`EXACTLY_ONE_TRUE`, not `TRUTH_XOR`**, and the enum members are defined *as* `(activation, cardinality)` pairs so lowerings branch on the pair. The practical payoff is that the axis separating select-one from truth semantics is **reification, not cardinality** — which is the axis your PR 3/PR 4 lowerings need to dispatch on, and the one a flat three-member enum hides.

2. **The overlapping-disjunct test you cited doesn't discriminate.** `test_hull_overlapping_disjuncts` is `min x` over `[x≤5] ∨ [x≥3]` asserting obj ≈ 0 — under truth semantics the feasible set is `[0,3) ∪ (5,10]` and the minimum is still 0, so it passes either way. Replaced with a witness that actually separates the two (`min (x−1/2)²`, 0 vs 0.25), across big-m / mbigm / hull / auto.

3. **The convex/nonconvex witnesses use named indicators**, which wasn't in the issue's scope but is the form users actually trip over: with `add_disjunction`, constraining an indicator to 0 means "this mode is not *selected*", not "this predicate is false". Both witnesses show a deselected disjunct's predicate holding at the optimum.

Behavior is unchanged — the select-one and indicator probes produce byte-identical output before and after. Non-default semantics raise rather than lower, which keeps solver defaults untouched per the issue's scope.

---

jkitchin:
Full sweep result, as promised in the PR body.

`pytest python/tests/ -m "not slow"` → **8929 passed, 5 failed**, 160 skipped, 9 xfailed, 2 xpassed (33m 11s).

All 5 failures are **pre-existing and unrelated to this change**, established by running the same 5 against the parent commit's source rather than by inspection:

| test | with this PR | baseline (`HEAD~1`) | cause |
|---|---|---|---|
| `test_1064…::test_round_fix_resolve_is_bounded_by_a_time_budget` | fail (1 of 2 runs) | fail | timing-dependent fixture ("never reaches the MIQP round gate") |
| `test_1064…::test_the_time_budget_is_what_bounds_the_spend` | fail | fail | same |
| `test_75_tape_nlp_evaluator…::test_solve_degrades_to_jax_when_pounce_is_missing` | fail | fail | environment: pounce **is** installed here, so the degradation path can't be exercised |
| `test_convex_nlp_certificate_853…[ipopt]` | fail | fail | environment: cyipopt not installed (`available_backends() == ['pounce']`) |
| `test_relax_compiler_convexity_units…::test_ipopt_backend_active_constraint_gradient` | fail | fail | same |

Method notes, since two of my own measurements were invalid before this one and I'd rather record that than bury it:

- The first baseline attempt used `git stash`, which was a **no-op** on a clean post-commit tree — so it silently re-ran my own code and I nearly reported it as the baseline. The run above instead restores the parent source with `git checkout HEAD~1 -- python/discopt discopt_benchmarks docs` and **asserts the marker absent** (`hasattr(core, 'DisjunctionSemantics') is False`) before running, and asserts it present again after restoring.
- An earlier repeat-run of `test_1064` was invalid for load: I had a background suite holding ~2.4 load average on 4 cores. The runs above were on a quiet machine (load ~1.1–1.35), both arms 327.5s and 327.8s wall.

`test_1064`'s first row is genuinely flaky — 5 failures in one run of this branch's code and 4 in another run of the identical code — so the with-PR/baseline difference in that row is noise, not an improvement from this PR.

Nothing here touches the disjunction paths. The GDP-adjacent suites (628 passed), `-m smoke` (1092 passed), and the adversarial suite (19 passed) are all green, as recorded in the PR body.

---

python/discopt/modeling/core.py
```diff
@@ -338,6 +338,139 @@ class            n    what happens
...
+
+    Notes
+    -----
+    ``SELECT_ONE`` and ``EXACTLY_ONE_TRUE`` are a *semantic* distinction, not a
+    convexity one: with ``S_1 = {x <= 1}`` and ``S_2 = {x >= 0}``, the point
+    ``x = 1/2`` is feasible under ``SELECT_ONE`` and infeasible under
+    ``EXACTLY_ONE_TRUE`` — and the same holds verbatim for nonconvex ``S_k``.
+    """
+
+    SELECT_ONE = "select_one"
```
bernalde: Blocking: The enum contract described immediately above is not what this code exposes: these members remain flat string values, the axes live in a separate lookup table, and `_require_select_one` dispatches on `semantics is DisjunctionSemantics.SELECT_ONE` rather than on `(activation, cardinality)`. That contradicts the explicit #1123 maintainer decision to define each public member internally as the pair and make lowerings branch on the pair, while also locking the wrong `.value` shape into the public API. Please store the axes on each member (preserving string coercion separately), dispatch the lowerings on the pair, and test the actual representation and axis-based dispatch.

python/tests/test_1124_disjunction_semantics.py
```diff
@@ -0,0 +1,313 @@
...
+        names = [v.name for v in ref._variables]
+
+        generated = [n for n in names if n.startswith("_gdp_aux_")]
+        assert len(generated) == 2  # one selector per disjunct
+
+        assert "mode_is_hot" in names
+        assert not any(n.startswith("_") for n in ("mode_is_hot",))
+        assert "mode_is_hot" not in generated
+
+    def test_user_boolean_cannot_collide_with_the_reserved_aux_prefix(self):
```
bernalde: Blocking: This test never creates a user Boolean with the reserved prefix, so it cannot prove the claimed isolation. On the exact reviewed head, adding `m.boolean("_gdp_aux_disj_d_0_0")` before this disjunction produces four variables but only three unique names because the lowering generates `_gdp_aux_disj_d_0_0` again. That violates AC-5 and can make name-based result mapping ambiguous. Please reject/reserve this namespace or allocate collision-proof auxiliary identities with explicit metadata, then make this test instantiate the conflicting user name.

bernalde (changes requested):
The new terminology and overlap coverage are useful, but the implementation still misses three explicit parts of the #1123/#1124 contract.

### Blocking: Public documentation still teaches the rejected truth semantics


The published GDP tutorial still says “exactly one group of constraints must hold” in both its introduction and API table. That is the precise truth-based reading this issue is meant to remove. The new contract page does not repair the public surface because `docs/_config.yml` excludes `design/`, while the changed `docs/comparison.md` links into that excluded path. Please update both notebook cells and publish or link the semantics contract from an included book page.

## Summary

- Blocking issues: 3 (2 inline, 1 review body)
- Nonblocking issues: 0
- Questions: 0
- Tests: On a clean merge of `c8000362da8c0792d460eaa09bc70cd07919368e` into `main@9e19486963012b8a0baec4c7cf808a3fb44f4fc9`, the new test file passed (`29 passed`); the GDP/LLM selection passed its default and slow groups (`242 passed, 14 deselected` plus `14 passed, 242 deselected`, 256 selected cases total). `ruff check` and `git diff --check` passed. Current-head lint/typecheck, Rust, Python fast/correctness/claim-boundary/AMP, and Codecov checks are green; optional slow and coverage jobs were skipped by workflow configuration.
- Coordination: #1128 is the only open upstream PR, and it merges cleanly into the current `upstream/main`.
- Merge-readiness: Requesting changes until the enum representation/dispatch, auxiliary-name isolation, and published documentation match the agreed contract.

python/discopt/modeling/core.py
jkitchin: Agreed, and fixed in `cd64b4a`. You're right that the docstring described a contract the code didn't implement.

A member's `.value` is now the pair itself:

```python
SELECT_ONE = (SelectorActivation.ONE_WAY, SelectorCardinality.EXACTLY_ONE)
OR = (SelectorActivation.ONE_WAY, SelectorCardinality.AT_LEAST_ONE)
EXACTLY_ONE_TRUE = (SelectorActivation.REIFIED, SelectorCardinality.EXACTLY_ONE)
```

so the axes are carried by the member and the side lookup table is gone — it was the thing that could drift out of sync. The lowercase spelling survives as a coercion alias exposed as `.label`; `_coerce_disjunction_semantics` now resolves by member name rather than by value.

Dispatch is keyed on the pair. The GDP pass declares what its rows actually encode and serves a disjunction iff its pair is in that set:

```python
_SUPPORTED_SEMANTICS = frozenset({(SelectorActivation.ONE_WAY, SelectorCardinality.EXACTLY_ONE)})

def _semantics_is_supported(semantics):
    return (semantics.activation, semantics.cardinality) in _SUPPORTED_SEMANTICS
```

The consequence you were after: a future member carrying an already-implemented pair is served without touching the lowering, and one carrying a different pair is refused without touching it either. The refusal message now names the pair the rows encode instead of naming `SELECT_ONE`.

Two tests pin the representation and the dispatch rather than the docstring: `test_member_value_is_the_pair_itself` asserts `member.value == (member.activation, member.cardinality)` for every member, and `test_lowering_dispatch_is_keyed_on_the_pair_not_the_member` asserts `_semantics_is_supported(member)` agrees with the pair lookup for all three.

---

python/tests/test_1124_disjunction_semantics.py
jkitchin: Confirmed and fixed in `cd64b4a`. I reproduced your case exactly on the reviewed head before changing anything:

```
variables: ['x', '_gdp_aux_disj_d_0_0', '_gdp_aux_disj_d_0_0', '_gdp_aux_disj_d_1_1']
count: 4  unique: 3
duplicated names: {'_gdp_aux_disj_d_0_0'}
```

You're also right about the more important part: the test asserted an isolation it never exercised. It checked that *generated* names were unique among themselves, which is true and beside the point — it never created the colliding user name, so it could not have caught this.

Closed from both sides, since either alone leaves a hole:

- **Reservation.** `Model._check_name` — the single choke point every user-facing variable/parameter factory funnels through — refuses any name inside `GDP_AUX_PREFIX`, with a message saying the prefix belongs to compiler auxiliaries. That stops a user taking the namespace.
- **Allocation.** The GDP pass now allocates auxiliary names against the names already present in the model, not from a counter alone. A model can carry reserved-prefix names without any user typing one — a second lowering pass, or an imported model — and the counter does not prevent a clash there.

Tests follow the same split. `test_user_name_in_the_reserved_namespace_is_refused` instantiates the conflicting name across three spellings (including the exact `_gdp_aux_disj_d_0_0` from your repro) against `boolean`/`continuous`/`binary`; `test_the_collision_that_motivated_the_reservation_cannot_recur` runs your reproducer end to end; and `test_generated_names_are_unique_even_against_preexisting_auxiliaries` appends reserved-prefix variables directly to `_variables` — bypassing the user-facing check exactly as an earlier pass would — and asserts the allocator still yields unique names.

---

jkitchin:
@bernalde — all three blocking findings are fixed in `cd64b4a`. Each was reproduced before being changed; none was taken on faith. The two inline threads carry the detail; this covers the review-body finding and the verification.

**Public documentation (review body).** Both parts confirmed. `docs/notebooks/tutorial_gdp.ipynb` taught the rejected reading in two cells — the section-1 introduction and the summary API table — and additionally rendered the disjunction as `\veebar` (⊻, the XOR symbol), which reinforced it. Both cells now state select-one, name the union projection, and link the contract; the symbol is `\vee`. And the contract page genuinely was unreachable: `docs/_config.yml` has `design` in `exclude_patterns`, so it was never published, and my `docs/comparison.md` link pointed into that excluded path — a broken link, not just an unpublished page. Moved to `docs/disjunction_semantics.md`, added to `_toc.yml` under Advanced Features, both links retargeted. `grep -rn "exactly one group" docs/` now returns only the passage that explicitly names it as the *rejected* reading.

The contract page was also updated to match the code rather than restate the old design: §1 now describes the pair-valued members and pair-keyed dispatch, and §4 documents that the namespace separation is enforced at both the reservation and allocation ends, with the collision it prevents spelled out.

**Verification on the merged head.** GDP/LLM/display selection: **586 passed**, 5 skipped, 2 xpassed. The `#1124` file is now **36 cases** (was 29). `ruff check python/` and `ruff format --check` clean.

One thing to flag rather than bury: `main` is currently red independently of this PR. `pytest -m smoke` on the merged head gives 15 failures — 7 in `test_1060_native_single_tree.py`, 6 in `test_1141_fractional_node_cuts.py`, 2 in `test_issue_1060_open_bounds_marshal.py` — plus `test_mip_nlp.py::test_mip_nlp_method_lp_nlp_bb_runs_on_the_native_simplex_backend`. I checked out `origin/main`'s source, asserted the marker for this PR's API was **absent** to prove I was really running the baseline, and got a byte-identical failure list. All of it arrived with the recent #1066/#1141 work; none is this PR's, and there is no fix of mine to port. If CI shows those checks red here, that is why.

Re-requesting your review.

---

python/discopt/modeling/core.py
```diff
@@ -4053,10 +4241,22 @@ def add_disjunction(
         >>> d2.subject_to(x >= 7)
         >>> m.add_disjunction([d1, d2], name="mode_select")
         """
+        resolved = _coerce_disjunction_semantics(semantics)
+        if resolved is not DisjunctionSemantics.SELECT_ONE:
```
bernalde: Blocking: `add_disjunction()` is itself an immediate lowering (as this docstring and its regression test say), but this gate still dispatches on enum-member identity. That leaves the prior pair-dispatch finding partially unresolved and directly contradicts the new `DisjunctionSemantics` contract above that every lowering branches on `(activation, cardinality)`. Please gate this path on the pair it emits—`(ONE_WAY, EXACTLY_ONE)`—or reuse a shared pair-support predicate, and add coverage that pins pair-based dispatch for `add_disjunction()` rather than only `_semantics_is_supported()`.

discopt_benchmarks/benchmarks/gdplib_native.py
```diff
@@ -94,8 +94,10 @@ def build_jobshop():
 def build_ex1_linan_2023():
     """Toy GDP of Liñán & Ricardez-Sandoval (2023): six-hump-camel over a grid.
 
-    Minimize the (nonconvex) six-hump camel function while two xor disjunctions
-    pin ``alpha`` and ``beta`` to grid points. A logical constraint forces the
+    Minimize the (nonconvex) six-hump camel function while two select-one
+    disjunctions pin ``alpha`` and ``beta`` to grid points. The disjuncts are
+    mutually exclusive equalities, so select-one and truth semantics coincide
+    here (see ``docs/design/disjunction-semantics.md``). A logical constraint forces the
```
bernalde: Nonblocking: The contract page moved to `docs/disjunction_semantics.md`, so this changed docstring now points at a path that does not exist. Please update the reference to the published path.

bernalde (changes requested):
The update resolves the auxiliary-namespace collision and the unpublished/incorrect semantics documentation, and the enum members now carry their `(activation, cardinality)` pairs. One immediate lowering still bypasses that pair contract, so the earlier dispatch blocker is only partially resolved.

### Nonblocking: Refresh the regression-test count in the PR description


The PR body still says `python/tests/test_1124_disjunction_semantics.py` contains 29 cases. The exact updated head runs 36 cases; please refresh the current-head verification count so the description matches the code under review.

## Summary

- Blocking issues: 1 (inline)
- Nonblocking issues: 2 (1 inline, 1 review body)
- Questions: 0
- Prior findings: The auxiliary-name isolation and public-documentation findings are addressed. The enum representation is addressed, but one immediate-lowering dispatch site still uses member identity rather than the semantic pair.
- Tests: Built the Python native extension from the exact `cd64b4a294092721923be4e0d5b671584fff9d6a` tree. The new regression file passed (`36 passed`); the GDP/LLM selection passed its default group (`249 passed, 14 deselected`) and slow group (`14 passed, 249 deselected`). `ruff check`, `ruff format --check`, and `git diff --check` passed. Current-head CI completed successfully; optional slow, coverage, and AMP-integration jobs were skipped by workflow configuration.
- Coordination: The PR is up to date and cleanly mergeable against `main@9e19486963012b8a0baec4c7cf808a3fb44f4fc9`. The only other open upstream PR, #1144, has no overlapping changed files.
- Merge-readiness: I would merge this once `add_disjunction()` dispatches on `(activation, cardinality)` as the contract requires.

python/discopt/modeling/core.py
jkitchin: You're right, and it's the sharper version of the original finding: I replaced identity dispatch everywhere the GDP pass touches and left the one lowering that doesn't go through it. Fixed in `cd3b1c3f`.

Rather than write a second pair check beside the first, the predicate now lives once, in `modeling/core.py` next to the enum it interprets:

- `_SELECT_ONE_ROWS` — the pair emitted by one-way activation under `sum_k y_k == 1`, which is what **both** `add_disjunction` (`if_then` + `exactly(1)`) and every `_relax.gdp_reformulate` reformulation build;
- `_semantics_pair` / `_semantics_supported_by` / `_require_semantics_supported` — the shared predicate and refusal, parameterized by the emitting site so the message names it.

`add_disjunction`'s gate is now:

```python
resolved = _coerce_disjunction_semantics(semantics)
_require_semantics_supported(resolved, _SELECT_ONE_ROWS, "add_disjunction()")
```

and the GDP pass's `_SUPPORTED_SEMANTICS` **is** `_SELECT_ONE_ROWS` rather than a second frozenset spelling out the same pair — so the two gates cannot drift apart on what "supported" means, which was the failure mode that produced this finding in the first place.

On coverage: a test asserting `add_disjunction` refuses `OR` passes under either dispatch, so it would not have caught this. The discriminator instead widens the emitted-pair set and checks that `OR` becomes servable — true under pair dispatch, false under an identity check:

```python
monkeypatch.setattr(core, "_SELECT_ONE_ROWS", frozenset({
    (SelectorActivation.ONE_WAY, SelectorCardinality.EXACTLY_ONE),
    (SelectorActivation.ONE_WAY, SelectorCardinality.AT_LEAST_ONE),
}))
m.add_disjunction(ds, name="m", semantics=DisjunctionSemantics.OR)   # no raise
```

I restored the old identity gate and confirmed it **fails** there, then restored the new one and confirmed it passes — so it cannot quietly degrade into a tautology. Alongside it, `test_add_disjunction_acceptance_agrees_with_the_pair_predicate` checks acceptance matches the pair predicate for every member, and `test_add_disjunction_and_the_gdp_pass_share_one_supported_pair_set` pins the single declaration. 40 cases now, from 36.

---

discopt_benchmarks/benchmarks/gdplib_native.py
jkitchin: Caught — that reference was left dangling by the very commit that published the page. Retargeted to `docs/disjunction_semantics.md` in `cd3b1c3f`.

`grep -rn "design/disjunction-semantics" .` across `*.py`, `*.md` and `*.yml` now returns nothing, so this was the last stale pointer to the old path.

---

jkitchin:
@bernalde — all three items from the second round are addressed in `cd3b1c3f`. Details are on the two threads; this covers the review-body item and the verification.

**Nonblocking (review body): stale test count.** Fixed — the PR body now reports the current head's numbers. It is **40** cases, not 36: the pair-dispatch coverage you asked for on `add_disjunction()` added four more on top of the 36 you measured.

**On the blocking find.** You were right, and it was the sharper version of the original: I replaced identity dispatch everywhere the GDP pass reaches and left the one lowering that doesn't go through it. The fix puts the predicate in one place rather than adding a second copy of the check — `_SELECT_ONE_ROWS` plus `_require_semantics_supported` in `modeling/core.py`, used by both `add_disjunction` and the GDP pass, whose `_SUPPORTED_SEMANTICS` now *is* that same frozenset. Two gates spelling out the same pair separately is what let them diverge in the first place, so a shared declaration seemed like the fix that keeps this from recurring rather than just closing today's instance.

Worth flagging about the new test: an `add_disjunction` test that merely asserts `OR` is refused passes under *either* dispatch, so it would not have caught your finding. The discriminator widens the emitted-pair set and asserts `OR` becomes servable. I verified it **fails** against the restored identity gate and passes against the pair gate, so it cannot silently degrade into a tautology — the same failure mode as the AC-5 test you flagged last round.

**Verification on `cd3b1c3f`:** the `#1124` file 40 passed; GDP/LLM/display selection 426 passed, 3 skipped, 2 xpassed; `ruff check python/` and `ruff format --check` clean. This round's diff is confined to the `add_disjunction` gate, the shared predicate, the tests, and one docstring path, so the smoke and adversarial results from `cd64b4a` still stand.

Re-requesting your review.

---

bernalde:
Thanks for flagging these departures. My view on each:

1. I agree with `EXACTLY_ONE_TRUE` rather than `TRUTH_XOR`, and with defining the enum values as `(activation, cardinality)` pairs. For an n-ary disjunction, “XOR” is easy to read as odd parity, while `EXACTLY_ONE_TRUE` states the intended invariant directly. The pair representation also exposes the semantic axes that downstream lowerings need to dispatch on.

2. I agree that the original overlapping-disjunct test did not distinguish select-one from truth semantics, and the replacement is a useful select-one regression. There is one mathematical nuance to retain for future `EXACTLY_ONE_TRUE` work: with the closed predicates `x <= 1` and `x >= 0`, exactly one predicate is true only when `x < 0` or `x > 1`. Therefore `(x - 0.5)^2` has infimum `0.25` under truth semantics, but that value is not attained at `x = 0` or `x = 1`, because both predicates are true at either boundary. A future truth-semantics implementation will need an explicit strict-complement/boundary policy, and related prose should say “infimum” rather than locating an optimum at those boundary points. This does not undermine the current select-one test.

3. I agree with adding the named-indicator witnesses despite their being beyond the narrow issue scope. They test the user-visible distinction that matters here: a deselected mode is not the same thing as a false predicate.

I also completed a resolution pass against head `cd3b1c3f1d6e7ef121bd6ac2ff3cc46d9e5f255d`. Every finding from my posted review rounds has been addressed: the pair-based dispatch now covers both reformulation paths through the shared predicate, the auxiliary-name collision and its regression coverage are fixed, the public documentation is aligned with select-one semantics, and the stale benchmark link and test-count text were corrected. The review threads are resolved, the exact-head CI run is green, and the focused semantics test module passes all cases. I do not see a remaining issue with the merged select-one contract; the strict-boundary point above is a follow-up consideration before non-default truth semantics are implemented.

PR #1130, opened 2026-08-29
ci: cut GitHub Actions runner time (optimized build profile + stop duplicate runs)

jkitchin:
## Summary

A full CI run on a PR costs **56.5 min of runner time** — measured on run [33263673485](https://github.com/jkitchin/discopt/actions/runs/33263673485): 3,391 s across 7 executing jobs, at 10–17 CI runs/day. The three Python lanes are 86% of that. This cuts the per-run cost and stops several categories of run that bought no signal at all.

**The main cause: every Python lane built the Rust extension unoptimized.** They ran a bare `maturin develop` — the `dev` profile, opt-level 0. The Makefile developers use locally builds `--release`, and so does `graduation-gate.yml`, so the CI lanes were the only place an unoptimized solver core was ever exercised. That, not runner speed, is the "~5 min locally and 23 min on a runner" gap the comment on `python-correctness-slow` recorded; that attribution is retracted in place.

Entry experiment (CLAUDE.md §4), interleaved A/B on an idle 4-vCPU/15 GB box (a runner's shape), `cargo test -p discopt-core -- --test-threads=4`, 2 reps each, load-gated (`uptime` 0.14 at start):

| profile | rep 1 | rep 2 | mean | tests passed |
| --- | --- | --- | --- | --- |
| `dev` (what CI built) | 113.57 s | 110.62 s | 112.10 s | 646 |
| `release` (plain) | 12.91 s | 12.70 s | 12.80 s | **645** |
| `release` + the two overrides (**shipped**) | 14.11 s | 14.11 s | 14.11 s | 646 |

**7.9× faster than what CI was building.** Lanes now build `--release` with `CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS` and `CARGO_PROFILE_RELEASE_OVERFLOW_CHECKS` set at the workflow level.

Plain `--release` would have been 10% faster still and was rejected: it drops to **645** tests because it compiles out the `#[cfg(debug_assertions)]` test in `presolve/orchestrator.rs`, and it silences the 20 `debug_assert!` invariant checks (index bounds, length agreement) in the simplex/B&B kernels plus integer overflow checking. Buying CI minutes by disarming a correctness check is the trade CLAUDE.md §1/§3 forbids. Arming them restores the 646th test for 1.3 s. That count is the marker to assert on if this ever needs re-verifying.

The overrides go on the standard `release` profile rather than a custom one in `Cargo.toml`, so artifacts land in `target/release` — the path `Swatinem/rust-cache` is known to cache, and a cache miss would cost more than the profile saves — and **nothing about the shipped wheels changes**. Cargo caches added to the two lanes that lacked one and now compile optimized.

**Runs that bought nothing, now stopped:**

* **Draft PRs** no longer run the matrix. `synchronize` fires on every push to a draft; `ready_for_review` is already in `types:`, so the full matrix runs on the final tree. Deferred, not skipped — and a draft cannot merge.
* **Label events** no longer run the matrix. Adding any label re-ran all 56 min. The lanes that care about a label gate themselves on it and do not depend on the `changes` job, so they still fire.
* **`python-fast` is skipped whenever `python-coverage` runs** (pushes to `main`, PRs labelled `coverage`). `python-coverage` is a strict superset — same marker expression, same `--ignore`, same worker count, plus `test_amp.py` and the optional extras. That duplicate was **917 s (15.3 min) per main push**.
* **The docs book** is only rebuilt when `docs/` or `python/discopt/` changed; a `workflow_dispatch` button keeps manual republishing possible.
* **`amp-integration` had no concurrency group**, so superseded pushes each ran the ~13 min suite to completion. Added, along with one on `graduation-gate`; both exempt `schedule` and `workflow_dispatch` by run id so a cron or manual verdict is never cancelled.
* The docs-only allowlist now covers `.crucible/`, `.claude/` and `*.org`.

Contributes to nothing tracked — opened directly from the runner-time budget. Note for the record: this repository is **public**, so its standard-runner minutes are free and unbilled (`billable.UBUNTU.total_ms = 0` on every job of the run above). The runner burn is real regardless, but if this was opened against a "90% of Actions allotment" alert, that allotment is being consumed somewhere other than this repo.

## Correctness

No solver math, no Python source, and no Rust source is touched — the diff is four workflow files. The one substantive question is whether moving the Python lanes from `dev` to `release` weakens a check; it does not, because both `debug-assertions` and `overflow-checks` are explicitly re-armed, and the 646-vs-645 test count above is the evidence that they are actually on rather than nominally requested.

- [x] Cannot produce a false certificate — N/A, no solver-math change
- [x] No validation, fallback, or safety guard was weakened to make a check pass — the reverse: plain `--release` was rejected *because* it would have disarmed the debug assertions and the `#[cfg(debug_assertions)]` test

The one deliberate reduction in *when* checks run is draft PRs and label events. Neither loses a check on any tree that can merge: `ready_for_review` re-runs the full matrix, and a draft is unmergeable until it fires.

Run in this session:

- [x] `cargo test -p discopt-core` → **646 passed, 0 failed** (6 suites) on `dev`; **646 passed, 0 failed** on `release` with the two overrides; 645 on plain `release` (the rejected variant). Rust source is unchanged — these runs are the A/B measurement, not a regression check.

Not run locally, and why:

- [ ] `pytest -m smoke` — not run in this session; no Python or Rust source changed, and the lanes this PR edits are exactly what runs it. **This PR's own CI is the verification**; the numbers it reports are the point of the change.
- [ ] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — same; `slow` is not on an automatic lane, so unchanged by this PR either way.
- [ ] `ruff check` / `ruff format --check` — no Python files in the diff.

All four workflow files were validated as parsing YAML with the expected top-level keys (`name`, `on`, `env`, `concurrency`, `jobs`). There is no `actionlint` hook configured in `.pre-commit-config.yaml`, so the GitHub expression syntax is verified by this PR's run, not locally.

## Regression test

- [x] N/A — CI configuration only. The behavioral assertion this change rests on is the 646-test count under the new build flags, which is recorded in the `env:` block comment at the top of `ci.yml` as the marker to re-check.

Not a solver performance change — no bound, cut, relaxation, or flag is touched, and no shipped artifact changes. The performance claim is entirely about CI wall time, and its measurement is the A/B table above.

One number is deliberately **not** claimed: the 7.9× is measured on the Rust kernels. The Python lanes' wall clock is a mix of that and numpy/Python, so their realized saving will be reported from this PR's own first run rather than projected here.

### Follow-ups deliberately left out of this PR

* `PYTEST_XDIST_WORKERS: "2"` on a 4-vCPU runner is the obvious next lever on `python-fast`, but OOM risk means a red run costs more than it saves — it needs its own measured change.
* `PYTEST_MEMORY_LIMIT_MB: 32768` is a `prlimit --as` cap above the runner's 16 GB, so it is currently inert. Worth fixing, but it is a safety question rather than a minutes question.
* `release.yml` uses `macos-14` (10× billing) and `windows-latest` (2×). Tag-only, so rare, and trimming it would be a real platform-coverage loss.

https://claude.ai/code/session_016F5N9ey1iLEJCyFMW8Jb4U

---
_Generated by [Claude Code](https://claude.ai/code/session_016F5N9ey1iLEJCyFMW8Jb4U)_

### DIFF
--- a/.github/workflows/amp-integration.yml
+++ b/.github/workflows/amp-integration.yml
@@ -11,6 +11,57 @@ on:
   schedule:
     - cron: "18 8 * * 1"
 
+# Every Python test lane below builds the Rust extension with `maturin develop
+# --release` and these two overrides, instead of the bare `maturin develop` they
+# used to run.
+#
+# `maturin develop` with no profile flag builds `dev` — opt-level 0. The Makefile
+# developers use locally builds `--release`, and so does graduation-gate.yml, so
+# the Python lanes were the only place that ever exercised an unoptimized solver
+# core; that is the real content of the comment on `python-correctness-slow`
+# recording `python-correctness` as "~5 min locally and 23 min on a runner" and
+# attributing the gap to the runner.
+#
+# Measured on an idle 4-vCPU/15 GB box (the runner's shape), interleaved A/B,
+# `cargo test -p discopt-core -- --test-threads=4`, 2 reps each:
+#
+#   dev (what CI built)            113.57 s, 110.62 s   mean 112.10   646 tests
+#   release (plain)                 12.91 s,  12.70 s   mean  12.80   645 tests
+#   release + the two overrides     14.11 s,  14.11 s   mean  14.11   646 tests
+#
+# i.e. 7.9x faster than what CI was building, and the Python lanes are dominated
+# by the same simplex/B&B kernels.
+#
+# The two overrides matter, and the test COUNT is why they are here rather than a
+# bare `--release`: plain release drops to 645 because it compiles out the
+# `#[cfg(debug_assertions)]` test in presolve/orchestrator.rs, and silences the 20
+# `debug_assert!` invariant checks (index bounds, length agreement) in the
+# simplex/B&B kernels plus integer overflow checking. Buying CI minutes by
+# disarming a correctness check is the trade CLAUDE.md §1/§3 forbids. Arming them
+# restores the 646th test and costs 1.3 s — 10% of an already 8x win. That count
+# is the marker to assert on if this ever needs re-verifying.
+#
+# They go on the standard `release` profile (rather than a custom one in
+# Cargo.toml) so the artifacts land in `target/release`, which is the path
+# Swatinem/rust-cache is known to cache — a cache miss would cost more than the
+# profile saves — and so nothing about the shipped wheels changes.
+env:
+  CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS: "true"
+  CARGO_PROFILE_RELEASE_OVERFLOW_CHECKS: "true"
+
+# This workflow had no concurrency group, so N pushes to main in quick
+# succession each ran the full ~13 min Alpine/incidence suite to completion,
+# including the runs whose SHA had already been superseded. ci.yml has cancelled
+# superseded runs this way since it was written; this is the same rule.
+# `schedule` and `workflow_dispatch` get their own group per run id so a manual
+# or weekly run is never cancelled by a push.
+concurrency:
+  group: >-
+    ${{ github.workflow }}-${{
+      (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch')
+        && github.run_id || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   amp-integration:
     name: AMP Alpine/incidence suite
@@ -29,6 +80,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
+      # Same reason as the ci.yml lanes: this suite now builds the optimized
+      # `release` profile, so the cargo cache has to be here or the build cost eats
+      # the speedup.
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: ". -> target"
       - uses: actions/setup-python@v5
         with:
           python-version: "3.12"
@@ -42,7 +99,7 @@ jobs:
           source .venv/bin/activate
           pip install maturin
           pip install jax jaxlib numpy scipy highspy cyipopt pytest pytest-timeout pytest-xdist
-          maturin develop
+          maturin develop --release
       - name: Verify cyipopt is importable
         run: |
           source .venv/bin/activate
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,44 @@ on:
 # below is `workflow_dispatch`-only: run it deliberately, before a release or
 # when touching the certificate path.
 
+# Every Python test lane below builds the Rust extension with `maturin develop
+# --release` and these two overrides, instead of the bare `maturin develop` they
+# used to run.
+#
+# `maturin develop` with no profile flag builds `dev` — opt-level 0. The Makefile
+# developers use locally builds `--release`, and so does graduation-gate.yml, so
+# the Python lanes were the only place that ever exercised an unoptimized solver
+# core; that is the real content of the comment on `python-correctness-slow`
+# recording `python-correctness` as "~5 min locally and 23 min on a runner" and
+# attributing the gap to the runner.
+#
+# Measured on an idle 4-vCPU/15 GB box (the runner's shape), interleaved A/B,
+# `cargo test -p discopt-core -- --test-threads=4`, 2 reps each:
+#
+#   dev (what CI built)            113.57 s, 110.62 s   mean 112.10   646 tests
+#   release (plain)                 12.91 s,  12.70 s   mean  12.80   645 tests
+#   release + the two overrides     14.11 s,  14.11 s   mean  14.11   646 tests
+#
+# i.e. 7.9x faster than what CI was building, and the Python lanes are dominated
+# by the same simplex/B&B kernels.
+#
+# The two overrides matter, and the test COUNT is why they are here rather than a
+# bare `--release`: plain release drops to 645 because it compiles out the
+# `#[cfg(debug_assertions)]` test in presolve/orchestrator.rs, and silences the 20
+# `debug_assert!` invariant checks (index bounds, length agreement) in the
+# simplex/B&B kernels plus integer overflow checking. Buying CI minutes by
+# disarming a correctness check is the trade CLAUDE.md §1/§3 forbids. Arming them
+# restores the 646th test and costs 1.3 s — 10% of an already 8x win. That count
+# is the marker to assert on if this ever needs re-verifying.
+#
+# They go on the standard `release` profile (rather than a custom one in
+# Cargo.toml) so the artifacts land in `target/release`, which is the path
+# Swatinem/rust-cache is known to cache — a cache miss would cost more than the
+# profile saves — and so nothing about the shipped wheels changes.
+env:
+  CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS: "true"
+  CARGO_PROFILE_RELEASE_OVERFLOW_CHECKS: "true"
+
 # Cancel superseded runs on the same PR/branch — saves CI minutes when
 # pushes land in quick succession.
 concurrency:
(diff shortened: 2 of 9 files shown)


jkitchin:
## Measured: the Python-lane number the PR body owed

CI is green on `a86472f`. Comparing [run 33268908560](https://github.com/jkitchin/discopt/actions/runs/33268908560) (this branch, warm cargo cache) against the baseline [run 33263673485](https://github.com/jkitchin/discopt/actions/runs/33263673485), per-job wall clock:

| job | before | after | |
| --- | ---: | ---: | ---: |
| Python claim-boundary (serial) | 11m37s | 5m40s | **−51.2%** |
| Python correctness (fast subset, serial) | 21m30s | 15m31s | **−27.8%** |
| Detect code changes | 0m11s | 0m09s | −18.2% |
| Rust | 4m52s | 4m46s | −2.1% |
| Python fast | 15m17s | 16m02s | **+4.9%** |
| Python AMP coverage | 2m26s | 3m15s | **+33.6%** |
| Lint + typecheck | 0m38s | 0m52s | +36.8% |
| **TOTAL runner time** | **56m31s** | **46m15s** | **−18.2%** (616 s) |

Wall clock for the whole run: 21m47s → 16m17s.

### The two regressions are real, and they have a coherent mechanism

The result is not "everything got faster", and the split is exactly where the physics says it should be. The optimized build buys time proportional to how much of a lane is *solve*, and costs a fixed extra compile:

* The two **serial, solve-dominated** lanes are where the Rust kernels dominate the wall clock — they take −51% and −28%.
* **`python-fast`** is xdist-parallel over mostly-unit tests, so there is less kernel time to recover and the extra compile isn't repaid. +4.9%, consistent across both runs on this branch (cold: +7.5%, warm: +4.9%), so it is signal, not noise.
* **`python-amp-coverage`** is the clearest case: its actual test battery is ~24 s, so build time is essentially the whole job and an optimized build is close to pure cost. +33.6%.

`Lint + typecheck` touches none of this (no Rust, no maturin) — 38 s → 52 s is runner variance on a job with a 3-second floor, not an effect of this PR.

Net is still −18.2% per run, and the trigger-level changes in this PR are worth considerably more than that in aggregate (a draft-PR push and a label event each drop from ~56 min to ~1 min; every push to `main` drops the duplicated 917 s `python-fast` lane).

### What I have deliberately not done

The obvious instinct is to put the two regressing lanes back on `dev`. I don't think that's right as a reflex: it would leave them testing an unoptimized core that nothing ships, which is the thing this PR argues against, and picking per-lane profiles by looking at a table is tuning to instances rather than fixing the class.

The general fix is that **no lane should build the extension at all**: build the wheel once in a dedicated job, publish it as an artifact, and have all five test lanes install it. That removes the regression everywhere instead of hiding it on two lanes, and it also removes the ~2 min of duplicated `uv pip install` each lane currently pays. It is a larger change to five jobs, so I've kept it out of this PR rather than widening it — happy to do it as a follow-up, or here if preferred.

The remaining `cert-neutrality regression guard` check is still running; this comment covers the CI workflow.

---

jkitchin:
## `cert-neutrality regression guard` is failing, and it is not this PR's

[Job 99143779760](https://github.com/jkitchin/discopt/actions/runs/33268908504/job/99143779760) — step *Run the graduation gate CI subset*, all 7 arms `eligible=no soundness=FAIL cert=FAIL`.

**This check was already red on `main`.** The previous execution of this workflow — [run 32706094552](https://github.com/jkitchin/discopt/actions/runs/32706094552), scheduled, `main` @ `48ac0463`, 2026-08-24 — failed in the same job, in the same step. There were **no graduation-gate runs between that one and this PR's**, so nothing landed in between that could have fixed or broken it. The step ran 44m27s there and 44m03s here.

One limit on that evidence, stated rather than glossed: the log API truncates that older run before its verdict table, so I matched on check + step + duration, not on the verdict lines themselves.

**Nothing in this PR can produce it.** The whole diff to `graduation-gate.yml` is a `concurrency:` block — it cannot influence a verdict. The `CARGO_PROFILE_RELEASE_*` overrides this PR adds are workflow-level `env:` in `ci.yml` and `amp-integration.yml` only; `graduation-gate.yml` already ran `maturin develop --release` and its build is untouched. The branch's merge-base is `origin/main` exactly, so this is not a stale branch missing a fix.

**No fix exists to port.** The gate was green on `main` at `1a0df198` ([run 32010001879](https://github.com/jkitchin/discopt/actions/runs/32010001879), attempt 2, Aug 17) and red at `48ac0463` (Aug 24), so it regressed somewhere in that week's merges. #1122 was filed by the reporter for that failure and closed 2026-08-25 — but the cron is Mondays only, so **the gate never actually ran green after that closure**; this PR's run is the first execution since, and it still fails. Worth a look: #1122 is closed on a guard that is still down, which is the exact "dark signal" shape the reporter exists to prevent. Same for #1058 before it.

### A lead on the root cause

In `--ci-subset`, `graduation_gate.py` assigns **both** fields from one boolean:

```python
soundness_ok=cert.neutral,
cert_neutral=cert.neutral,
```

So `soundness=FAIL cert=FAIL` across all 7 arms is **one** failing check reported twice per arm — not seven independent soundness regressions. Every arm also took a uniform ~6m17s, which is the same vendored panel re-solving each time. That points at `run_cert_neutrality` / the cert baseline rather than at the flags, and it means the headline reads far more alarmingly than what is actually broken.

Chasing that is a separate change against a failure that predates this branch, so I have not widened this PR to include it.

### The one re-run, deliberately not spent

The re-run exists to confirm a failure isn't the PR's. The base-branch red already confirms that, and re-running would cost **44 minutes of runner time on a PR whose purpose is cutting runner time** to re-derive something I can already show. If a second observation is wanted anyway, the button is there — I'd rather it be a deliberate call than a reflex.

This check is not required: the PR reports `mergeable_state: unstable`, not `blocked`, and there is no merge conflict. The CI workflow itself is green on `a86472f`.

---

jkitchin:
## Gate fixed — it was measuring drift, not the flags

Pushed in `ad45acc`. Verified end to end on a 4-vCPU box, same commit, real gate, not a proxy:

```
origin/main gate code : root_fixpoint    eligible=no  soundness=FAIL cert=FAIL   exit 1
this branch           : root_fixpoint    eligible=YES soundness=ok  cert=neutral exit 0
                        lu_density_route eligible=YES soundness=ok  cert=neutral
```

### Root cause

The gate compared each flag-ON panel run against the committed `cert-baseline.jsonl` snapshot. That conflates two deltas:

```
(flag ON  vs  flag OFF)        <- what a flag guard is for
(today's main  vs  snapshot)   <- drift, identical for every flag
```

and the second swamped the first. Measured: the **flag-OFF control alone** produces 9 violations against the snapshot, and `root_fixpoint` and `lu_density_route` each produce the *same* 9. All seven arms were failing identically for something no flag did — which is exactly the 7/7 identical `soundness=FAIL` in the logs.

**The drift is not a soundness fault.** Zero objective violations in any comparison, and a bracket of all 47 certified instances against `cert-optima.json` found none disagreeing with the true optimum. It is three instances (`clay0303hfsg`, `nvs05`, `tanksize`) no longer certifying inside their 60 s wall-clock budget, plus node-count movement. A wall-clock budget deciding a `soundness=FAIL` verdict is the "the bar measures the runner" failure this repo already names in `_assert_budget_not_clock`.

### Why not just regenerate the baseline

`gen_cert_baseline.py` writes only the deterministically-certifying subset, so a refresh today would **silently drop** those three instances from the panel — shrinking coverage instead of recording the regression. I checked this before ruling it out.

### What changed

* **CI subset measures a flag-OFF control once and compares every arm against it** — same session, same machine, immune to snapshot staleness and hardware speed. The held-out arm in step (1) of the protocol already shared its OFF control across arms; this is the same pattern applied to the cert panel. 8 panels for 7 arms, so the workflow timeout goes 90 → 120 min.
* **Two absolute checks in `check_neutrality` become reference-relative.** Both are *provably* no-ops on the pre-existing path — all 52 committed rows are `optimal` with a non-null objective, which a new test pins so the two changes can't quietly stop being behaviour-preserving:
  * `status`: violation iff the reference certified and the new run did not, rather than demanding `new.status == "optimal"` outright.
  * `objective`: `nb is None or no is None` also fired for `None -> None` (neither side certified — no certificate for either to be wrong about) and `None -> value` (the flag certified what the reference could not, an improvement). Only a genuinely lost certificate (`value -> None`) flags now.
* **The snapshot comparison is kept and printed** as an explicit drift report, and the gate still hard-fails on the hardware-independent soundness question: any instance it *certified* whose objective disagrees with `cert-optima.json` beyond correctness tolerance. Uncertified incumbents are skipped — an open gap is not a false certificate. That check exits non-zero if it compared nothing, rather than reporting a pass it never measured.

### The judgement call worth flagging

I reclassified *"an instance stopped certifying inside its wall-clock budget"* from fatal to reported. Everything hardware-independent stays fatal and unweakened. I think that is correcting a mislabel rather than weakening a gate — the soundness half was passing the whole time — but it is a call, and easy to reverse if you disagree.

### Tests

12 pass in `test_cert_neutrality_regime.py`. The 3 new behavioural tests fail against `origin/main`'s `cert_neutrality.py` and pass here; 4 controls (lost certificate, lost status, committed-baseline shape) pass in both states, which is what makes them controls.

### Not fixed here

The drift is a **real solver regression** — `nvs14` 129 → 839 nodes, `nvs09` 5 → 31, and the three lost certifications — landing between `1a0df198` (2026-08-17, gate green) and `48ac0463` (2026-08-24, gate red). Filed as #1134 with the measurements and a suggested bisect target. Nothing in this PR fixes it; the point is that the gate should never have been the thing discovering it.

---

jkitchin:
## Green, including the gate

`cert-neutrality regression guard` passed on `c1f6e23` in **49m32s** ([run 33276175904](https://github.com/jkitchin/discopt/actions/runs/33276175904)) — the first time this check has been green since 2026-08-17.

```
root_fixpoint             eligible=YES  soundness=ok  cert=neutral
psd_cost_gate             eligible=YES  soundness=ok  cert=neutral
lift_zero_spanning        eligible=YES  soundness=ok  cert=neutral
lift_loose_products       eligible=YES  soundness=ok  cert=neutral
lu_density_route          eligible=YES  soundness=ok  cert=neutral
node_numerical_dual_bound eligible=YES  soundness=ok  cert=neutral
root_build_deadline       eligible=YES  soundness=ok  cert=neutral
```

It is green because it gated, not because it skipped: all 7 arms ran at ~5m44s each on top of the shared flag-OFF control — 8 panels × 52 instances = 416 solves — inside the raised 120-minute timeout with ~70 min to spare.

**This supersedes my earlier standing-down comment** (#issuecomment-5464412375), which said the failure was not this PR's and no fix existed to port. The first half was right — it was red on `main` at `48ac0463`, before this branch. The second half is now moot: the fix is in this PR.

`Python fast` also went green. It failed once on `ad45acc` because the PR's merge ref caught `main` between #1131 changing `oa.py` and #1133 re-syncing `test_912_wall_budget_inventory.py`'s gate inventory; merging `main` (`1e5a2cc`) in resolved it, and `oa.py` is not in this PR's diff.

Full CI on `c1f6e23`: Rust, Python fast, Python correctness, Python claim-boundary, Python AMP coverage, Lint, Detect changes, codecov — all green. `mergeable_state` clean.

Still open and not addressed here: #1134, the solver perf regression the OFF control surfaced (`nvs14` 129 → 839 nodes, three instances losing certification). That is real work, separate from this PR.

---

PR #1133, opened 2026-08-29
fix(#1066): a closed gap at the wall is a certificate, not "feasible"

jkitchin:
`solve_lp_nlp_bb` decided its exit status from *how the loop ended* and only then,
on two of the six branches, looked at the gap. The branch that fires when the
callback stops the search -- the clock, or a caller's termination hook -- set
`status = "feasible"` unconditionally whenever an incumbent existed, without ever
consulting the bound sitting right next to it.

5be6b0ce fixed the adjacent half of this: it made the driver *stop* when its own
bound certifies the incumbent. This is the other half -- what to report when the
driver did not get to stop on its own. The bound is the same object in both cases.
`master_bound_valid` says the master is a relaxation of the MINLP, and a MIP tree's
dual bound bounds its own optimum whether the tree finished or was cut off
mid-search, so `gap <= gap_tolerance` on that bound is the same certificate a
converged run reports. Reporting it as `feasible` is not conservative, it is wrong:
it throws away a proof the solver holds.

Measured, `squfl015-060` at default settings (60 s, `gap_tolerance=1e-4`, with the
disaggregated perspective epigraph of #1066 supplying the tight bound):

    dual bound  366.6218167383147
    incumbent   366.62181673996474
    relative gap 4.5e-14          reported status: feasible

The gap is ten orders of magnitude inside the tolerance. Without the tight bound
the clause is inert -- on the 104-instance route panel of PR #1131, 0 of the 7
`feasible` rows that carried both a bound and an incumbent had a closed gap -- so
this changes no status that was previously correct.

`termination_reason` is deliberately NOT rewritten to `gap_tolerance` when the new
clause fires. The search really did stop on the clock and the trace should keep
saying so: status answers "is the answer proven", the reason answers "why did the
loop end". Two different questions, two fields.

The decision moved out of the driver into `_lp_nlp_bb_exit_status`, a pure
function, so it can be tested at every combination of its inputs rather than only
at whichever ones a 60-second solve happens to produce. The extraction is
behaviour-preserving except for the one new clause.

Verification: 14 tests in
`python/tests/test_issue_1066_certificate_at_time_limit.py` -- the new clause, its
kill criteria (an open gap, an invalid bound, no bound, no incumbent), the
preserved meaning of every other branch, and an anti-vacuity test (CLAUDE.md §6)
that solves a real convex MINLP through `solve_lp_nlp_bb` and fails if the driver
reached its exit without consulting the helper. ruff, ruff format and mypy clean
on `oa.py`. Rust untouched.

Contributes to #1066.

### DIFF
--- a/python/discopt/solvers/oa.py
+++ b/python/discopt/solvers/oa.py
@@ -45,6 +45,7 @@ from discopt.solvers.mip_nlp_options import (
 
 if TYPE_CHECKING:
     from discopt._relax.nlp_evaluator import NLPEvaluator
+    from discopt.solvers import SolveStatus
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


jkitchin:
Local verification (2026-08-29, this branch):

- `pytest -m smoke` — 1137 passed, 1 skipped, 2 xpassed (510 s)
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — 19 passed (145 s)
- `pytest python/tests/test_issue_1066_certificate_at_time_limit.py` — 14 passed
- `pytest python/tests/test_912_wall_budget_inventory.py` — 5 passed

No Rust touched, so `cargo test -p discopt-core` was not run.

**Note on the second commit.** `origin/main` at 0d4daf8a already fails both
`test_912_wall_budget_inventory.py` assertions — verified on a detached worktree
of main (2 failed, 3 passed). The inventory went stale in 5be6b0ce (merged in
#1131), which rewrote the single-tree callback's wall gate from a `return` form
to an `if ...: return True` form without updating KNOWN. The commit here
re-records that gate in its current spelling and records the exit-status helper's
`hook is not None and elapsed < time_limit` question, both as `contract` (each
reads the caller's own `time_limit` to decide when to stop, never how much work
to do).

Issue #1143, opened 2026-08-31
Convex-MINLP auto-route spends 50-66% of the budget before abstaining, on instances the fallback certifies in seconds


jkitchin:
Split out of jkitchin/discopt#1134, where it turned up as "Cause 2" of the cert-panel drift. #1134 covers the investigation and the baseline tooling; this is the performance defect it found and deliberately did not fix, because changing it is a bound-changing default that owes the CLAUDE.md §5 graduation pair on a machine that can run it.

## What happens

`_convex_minlp_auto_route` sends a convex MINLP to the `oa` master. When the route does not certify, `solve_model` hands the remainder of the budget to the default spatial path, which then solves the model from scratch. The route's cut-off is `_CONVEX_ROUTE_BUDGET_FRACTION = 0.5`, and the #1066 progress guard cannot fire earlier than that by construction (`checkpoint = max(limit * fraction, min(limit, _CONVEX_ROUTE_FALLBACK_FLOOR_S))`, with `fraction` the same 0.5).

So on any instance where the route abstains, the caller pays **at least half the budget** for nothing. On the cert panel that is routinely 3-8x more than the whole solve costs unrouted.

Measured on `main` (`9e19486`), 4 vCPU / idle, 60 s budget, quoting the solver's own `algorithm_route`. "reference" is the committed `cert-baseline.jsonl` row:

| instance | reference | on `main` | what the route did |
| --- | --- | --- | --- |
| `cvxnonsep_nsig30` | 165 nodes, 1.12 s | 165 nodes, 49.6 s | OA did not certify in **39.63 s** of 60; the default path then solved the *same* 165-node tree |
| `clay0303hfsg` | 283 nodes, 15.29 s | no certificate at 60 s | OA `unknown` after **30.64 s**; fallback left 29.36 s |
| `fac2` | 39 nodes, 2.77 s | 39 nodes, 38.9 s | OA `feasible` after **30.13 s**; fallback certified in ~8.8 s |
| `cvxnonsep_psig30` | 89 nodes, 0.41 s | 89 nodes, 8.5 s | routed |
| `alan` | 21 nodes, 0.21 s | 13 nodes, 1.3 s | routed (OA abstained after 1.14 s) |

`fac2` is the clearest shape: the route burns 30.13 s, abstains, and the path that was going to certify anyway then does it in 8.8 s.

[...]

PR #1144, opened 2026-08-31
fix(cert): #1134 root-cause the panel drift (two bisects); stop the reference shrinking silently

### LINKED ISSUE #1134: Cert panel: 3 instances lost certification and 4 regressed on nodes since v0.8.0
Surfaced while fixing the flag-graduation gate (jkitchin/discopt#1130). The gate was comparing flag-ON runs against the committed `cert-baseline.jsonl` snapshot, so this drift was being charged to all 7 flags as `soundness=FAIL`. That PR makes the flag verdict ON-vs-OFF and reports the drift separately — it does **not** fix what the drift is telling us, which is this.

## What was measured

Flag-OFF control run of the vendored 52-instance cert panel on `main` (`9bed6a8`), against the committed baseline. Box: 4 vCPU / 15 GB, Python 3.12, `maturin develop --release`, idle (`uptime` load-gated), 60 s per-instance budgets from `_instance_budgets`. Reproduced across three independent runs.

**No soundness problem.** Zero objective violations in the `bound_changing` regime, and a bracket of all 47 certified instances against `cert-optima.json` found none disagreeing with the true optimum. Everything below is performance.

### Lost certification (was `optimal` in the baseline, now not, within the same 60 s budget)

| instance | baseline | now | nodes |
| --- | --- | --- | --- |
| `clay0303hfsg` | optimal | `time_limit` | 283 → 333–365 |
| `nvs05` | optimal | `feasible` | 175 → 95 |
| `tanksize` | optimal | `time_limit` | 16879 → 15041 |

`cvxnonsep_nsig30` is marginal rather than reliably lost — `optimal` in two runs, `feasible` in a third, i.e. it now sits on its budget cliff.

`nvs17` also times out but is already covered by `_KNOWN_PERF_GATED` for exactly this.

### Node-count regressions

| instance | baseline | now | |
| --- | --- | --- | --- |
| `nvs14` | 129 | 839 | **+550%** |
| `nvs09` | 5 | 31 | **+520%** |
| `tspn05` | 39 | 51 | +31% |
| `nvs02` | 345 | 421 | +22% |
| `clay0303hfsg` | 283 | 365 | +29% |

### For balance — the same run improved a lot of the panel

`st_testgr3` 549 → 0, `nvs12` 231 → 0, `flay03m` 107 → 0, `nvs11` 55 → 0, `m3` 37 → 0, `st_miqp1` 15 → 0, `st_miqp4/5` → 0, `gbd` 3 → 0, `nvs03`/`nvs10`/`nvs15` → 0. Whatever changed is clearly a net win on most of the panel; this issue is about the tail that went the other way.

## Window

The gate was green on `main` at `1a0df198` ([run 32010001879](https://github.com/jkitchin/discopt/actions/runs/32010001879), 2026-08-17) and red at `48ac0463` ([run 32706094552](https://github.com/jkitchin/discopt/actions/runs/32706094552), 2026-08-24), so the regression landed in that week's merges. ~30 merges in that window; `feat/1061-1062-graduate-nlpbb-root-cuts`, `perf/1066-node-throughput` and `perf/warm-root-cut-loop` are the ones whose blast radius most plausibly reaches node counts, but that is a reading of the titles, not a bisect.

The committed baseline itself was last regenerated at the v0.8.0 release (`deba4ce`), so part of the delta is simply a month of accumulated change rather than one commit.

## Why not just regenerate the baseline

`gen_cert_baseline.py` writes only the deterministically-certifying subset (`all_opt and nodes_same and obj_ok and margin_ok`). Refreshing it today would **silently drop** `clay0303hfsg`, `nvs05` and `tanksize` from the panel — shrinking coverage rather than recording the regression. The baseline should be regenerated once these are understood, not before.

## Suggested next step

Bisect `nvs14` (129 → 839 nodes) across the 2026-08-17…08-24 merges — it is the largest, is well inside its budget so it is not confounded by the wall clock, and node count is deterministic, so it makes a clean bisect signal.

jkitchin:
## Summary

Closes #1134. The 7 rows the cert panel reports as regressions are **two independent causes**, both bisected, and the window in that issue could not have been right: `git log --follow` puts the reference's last regeneration at `23a64ab6` (2026-08-11), five days *before* v0.8.0, not at it — `nvs14` measures 839 nodes at both ends of the supposed "good" end of the window.

**Cause 1 — the node counts.** `git bisect --first-parent` on `nvs14`'s node count (deterministic, well inside its budget, so unconfounded by wall clock) over `23a64ab6` → `main`, 104 revisions, 7 steps → **`bd9c4c5c` (#1025)**, the feral 0.15.1 → 0.16.0 bump defaulting `SparseLu::factor` to threshold Markowitz. Confirmed forward on `main` by an A/B whose arms differ only in `LuParams::pivoting`: the `GilbertPeierls` arm restores `nvs02` 421 → **345** and `tspn05` 51 → **39**, exactly the committed values, and takes `nvs14` 839 → 175. Already recorded and accepted by that bump under its regime-(b) acceptance; the reference was simply never refreshed after it.

**Cause 2 — the wall.** `cvxnonsep_nsig30` certifies the *same* 165-node tree with the same objective and dual bound, yet went 1.12 s → 42.5 s. Bisecting on wall **with both arms pinned to `GilbertPeierls`** (so the pivoting rule is not the variable) names **`533c05e1` (#1100)**, which corrected `_convex_minlp_auto_route`'s availability gate. The route abstains at a budget checkpoint and the fallback then restarts, so it costs 50–66% of the budget. Split to **#1143**; not fixed here, because re-tuning it is a bound-changing default owing the CLAUDE.md §5 graduation pair. §0.6.1 dates the full sequence: the route is `d9fec841` (#1059, default-ON 08-19) and `533c05e1` (08-20) extended it to machines without `highspy` — a bisect names one or the other depending on the box, and the cost belongs to the #1059 graduation.

**Verdict on the issue's question — the panel did not regress since v0.8.0.** Re-run on a second machine with the extension rebuilt from the tree under test and asserted at import (§8): **all 52 rows certify `optimal`**, `incorrect_count = 0` against 58 oracles, and of the 11 rows drifting past the 1e-8 byte-reproducibility tolerance, none disagrees with the true optimum and **8 land closer to it than the committed reference does**. The three "lost certifications" were the reporting box's wall (3.46x the reference machine there, 1.53x on the fast box); the node drift is Cause 1 and predates v0.8.0. §0.6.1 carries the two retractions this produced (`nvs09` does reproduce, at 31 nodes; and the "45 of 52 rows" / highspy-free framings were box-scoped).

**What this PR changes** is the tooling that let a stale reference sit undetected and would have hidden the coverage loss on refresh:

- `gen_cert_baseline.py` writes `docs/dev/data/cert-baseline-meta.json` (generating commit, timestamp, budget, host, `baseline_written`, and the full drop record with each dropped row's previous verdict) and **refuses to overwrite the reference when coverage would shrink**, unless `--allow-shrink`. The meta and the full-panel report are written either way, so a refused run still leaves its evidence — and is reported as a *refused* regeneration rather than having its provenance attributed to the reference still on disk.
- `check_cert_neutrality.py` prints that provenance and a host-speed calibration measured only on **unrouted** instances whose `node_count` reproduced exactly, and annotates `status` violations with it.
- `benchmarks.metrics.SolveResult` gains an optional `algorithm_route`, populated by the runner. Its absence is why Cause 2 was invisible in a table of nodes and seconds and cost a second bisect; with it, the one row still blocking a refresh (`clay0303hfsg`, 49 s of 60) is explained in one line — 30.27 s of it is an abstaining OA route — instead of another bisect.
- `docs/dev/certification-gap-plan.md` §0.6.1 records the falsification per §0.4; `crates/discopt-core/Cargo.toml` carries a narrowed retraction per §11.
- `.github/workflows/ci.yml` runs the cert-baseline tooling tests, which no lane exercised — `ci.yml` pulled exactly one file out of `discopt_benchmarks/tests`, so the reference every solver change is judged against had no CI covering the code that generates or reads it.

**No baseline is committed.** On the fast box the generator admits 51 of 52 and the guard refuses on `clay0303hfsg` alone; strip the abstaining route and it finishes in ~18.5 s, inside the margin, with zero coverage loss. Refreshing the reference is therefore an acceptance criterion on **#1143**, and the refusal is the guard working, not something to route around with `--allow-shrink`.

## Correctness

- [x] Cannot produce a false certificate (**N/A — no solver-math change**). The diff touches the benchmark harness, two generator/checker scripts, CI config, docs, and TOML comments. The one field added to a solver-facing dataclass is a reporting string. No relaxation, bound, cut, or reduction is touched; `cargo build -p discopt-core` recompiles only because a comment moved. Independently, the fresh-build panel run reports `incorrect_count = 0` against 58 oracles with every drifting row bracketed against `cert-optima.json`.
- [x] No validation, fallback, or safety guard was weakened to make a check pass. The guard-adjacent additions are strictly one-directional: the shrink guard *adds* a refusal that did not exist, and the host-speed calibration is **reporting-only** — it annotates a `status` violation with the machine ratio and never suppresses one (§0.3), confirmed on a live 52-instance run where all 17 violations stand. Excluding routed rows removes samples from that statistic; it does not change any verdict. `test_cert_neutrality_regime.py` passes unchanged.

- [x] `pytest -m smoke python/tests/` → **1119 passed, 21 skipped, 2 xpassed** (367 s)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **19 passed** (158 s)
- [x] `cargo test -p discopt-core` → **N/A** — no Rust source changed; `Cargo.toml` edit is comments only (`cargo metadata` parses, `cargo build --release -p discopt-core` clean)
- [x] `ruff check` / `ruff format --check python/` → **All checks passed / 957 files already formatted**
- [x] `pytest discopt_benchmarks/tests/ -m "not slow"` → **green**; the two cert modules on the merged head → **33 passed** (21 + 12)

## Regression test

`discopt_benchmarks/tests/test_1134_cert_baseline_shrink_guard.py` — **21 tests**. Fail-before/pass-after confirmed: on the pre-change tree the module **errors at collection** (`coverage_loss`, `build_meta`, `host_speed_ratio`, `provenance_lines`, `meta_describes_the_committed_reference` and `_MIN_CALIBRATION_SAMPLES` do not exist); after, 21 pass.

Covered: `coverage_loss` names exactly the rows the previous reference covered and a new run will not admit; `build_meta` records each drop's reason plus the lost row's previous verdict, distinguishes a never-admitted instance from a coverage loss, carries the generating commit, and records whether the reference was actually written; `provenance_lines` reports a refused run as `UNKNOWN` rather than claiming the reference's provenance, and re-derives that verdict for a meta written before the field existed; `host_speed_ratio` uses only unrouted instances whose node count reproduced exactly (a row whose tree exploded 100x is excluded, and so is a routed row on either side — both asserted), abstains below the sample floor, drops sub-noise baseline walls, and returns 1.0 for a machine against itself; and `algorithm_route` round-trips through `to_dict`/`from_dict` while all 52 rows of the committed baseline still load.

**End-to-end validation on real panels, on two machines.** On the 4-vCPU box `gen_cert_baseline.py --time-limit 60` refused the write and named its losses; `check_cert_neutrality.py` over all 52 rows then exercised the reporting paths (provenance correctly `NONE`, calibration 3.46x on 17 unrouted rows, 17 violations with the three `status` rows annotated and none suppressed). On the 14-core box with a verified-fresh extension the same generator admits 51 of 52 and refuses on one row, and `algorithm_route` explains it. Runner wiring verified with a live solve: `alan` comes back carrying the OA route note, `nvs14` with `None`.

Not a behavior change — the solver path is untouched — but the PR **retracts a published claim**, which CLAUDE.md §11 requires be recorded in writing, so the measurement is here.

The #1025 record asserts that "the certifying panel cannot" see an LU kernel bump, and measures net-positive only on captured relaxation LPs. It cannot see the *benefit* — that stands — but it does see a per-node cost. Measured on `main`, interleaved A/B, 3 reps, arms distinguished by a marker string baked into the extension, restricted to **unrouted** instances whose node count is identical in both arms (so the comparison is per-node cost, not search). Wall in seconds, Markowitz / GilbertPeierls:

```
nvs09  2.78/2.65   st_e29 2.07/2.03   ex1224 2.03/2.00   ex1225 1.88/1.84
gkocis 1.35/1.39   ex1226 1.53/1.35   oaer   1.32/1.32   st_e36 6.83/7.03
nvs22 15.35/12.11  nvs05 >=60.32/56.62  (Markowitz arm censored at the 60 s limit)
```

Median ratio **~1.03**, carried by `nvs22` **+27%** and `nvs05` **≥+7%**. It is two-sided: on the routed `m3` Markowitz is **2.4x faster** (1.53 s vs 3.70 s). The bump is **not** retracted — fill, `numerical` refusals and wall on the LPs that dominate a real solve all improved, and no verdict was ever false. What is retracted is that the panel is blind to it, at that smaller size. Size-routed pivoting (Gilbert–Peierls on small low-fill bases, Markowitz on large) is named as the follow-up lead in `Cargo.toml`, with these numbers as its entry measurement.

**Withdrawn from that evidence:** an earlier pass in this branch read `cvxnonsep_nsig30` as +49% from pivoting. It is auto-routed, so a wall difference there is not a per-factorization measurement; its real regression is Cause 2. Recorded in `Cargo.toml` alongside the retraction.

**Corrected in review:** the host-speed calibration was first published as 3.45x over 21 equal-node rows, which admitted three auto-routed rows carrying 14–44x inflation that is the router, not the box — Cause 2 in this same PR is the proof equal node counts are necessary but not sufficient. `host_speed_ratio` now excludes routed rows on either side, and the figure was re-measured on the same idle box: **3.46x over 17 unrouted rows**. The contaminated value was high by 0.01x and every conclusion drawn from it holds unchanged. §0.6.1 records the correction, the re-measurement, and the fast box's independent 1.53x.

*Measurement hygiene (CLAUDE.md §8/§9):* timing runs were load-gated on an idle box, run A/B interleaved rather than sequentially, and report a per-instance standard deviation over 3 reps. Every arm asserts both `module.__file__` and a marker string unique to the version under test, with the other arm's marker asserted absent. The extension-shadowing hazard the fresh-build re-run uncovered was checked against this container's builds: `_rust.abi3.so` is the only extension in the tree (fresh clone, `*.so` gitignored) and `discopt._rust.__file__` resolves to it, so the bisect attributions are against the tree under test.

---

https://claude.ai/code/session_011x5obMV6ZStxUyEe5WezGK

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,9 +235,11 @@ jobs:
     runs-on: ubuntu-latest
     needs: changes
     # `python-coverage` below runs the SAME pytest invocation as this job — same
-    # `-m` marker expression, same `--ignore`, same worker count — plus
-    # test_amp.py and the optional-extra deps, i.e. a strict superset, with
-    # coverage instrumentation on top. Whenever it runs (every push to main, and
+    # test paths, same `-m` marker expression, same `--ignore`, same worker
+    # count — plus test_amp.py and the optional-extra deps, i.e. a strict
+    # superset, with coverage instrumentation on top. Adding a test path to one
+    # of the two and not the other silently drops it from whichever event runs
+    # the other job, so change them together. Whenever it runs (every push to main, and
     # any PR carrying the `coverage` label) this job was re-running that superset
     # for a second time on the same tree, for ~15 min of runner time that bought
     # no signal. Skip here and let the superset be the signal; on an ordinary PR
@@ -307,7 +309,20 @@ jobs:
           # `python-coverage` job (main pushes + the `coverage` label) instead.
           # That parenthetical used to read "nightly + label-triggered";
           # `python-coverage` has never had a schedule trigger.
+          # The cert-tooling tests join this lane. `cert-baseline.jsonl` is the
+          # §0.2.5 bound-neutrality reference every solver change is judged
+          # against, and until #1134 NOTHING in CI exercised the code that
+          # generates, validates or reads it: ci.yml pulled exactly one file out
+          # of discopt_benchmarks/tests, so both cert modules could only ever be
+          # run by hand. This one is unmarked and lands here; its sibling
+          # `test_cert_neutrality_regime.py` is `pytest.mark.correctness` and
+          # would collect ZERO tests under this lane's `not correctness`, so it
+          # joins `python-correctness` instead. Both are pure functions over the
+          # committed reference — no solve, no .nl file, numpy their only
+          # third-party import — so they cost ~a second and cannot flake on a
+          # runner's wall clock.
           scripts/run_memory_capped_pytest.sh pytest python/tests/ \
+            discopt_benchmarks/tests/test_1134_cert_baseline_shrink_guard.py \
             -q --tb=short --timeout=120 \
             -m "not slow and not correctness and not integration and not amp_benchmark and not requires_cyipopt and not memory_heavy" \
             --ignore=python/tests/test_correctness.py \
@@ -401,8 +416,13 @@ jobs:
           # the consumer of #1116's `deterministic=`) and it lived its whole life
           # skipping. It resolves instances from python/tests/data/minlplib_nl/,
           # which is in the checkout, and skips the ones that are not.
+          # test_cert_neutrality_regime.py joins for the same reason as #1134's
+          # shrink-guard module joins python-fast: it is `correctness`-marked, so
+          # this is the only lane whose marker expression selects it, and until
+          # now nothing in CI ran it at all.
           scripts/run_memory_capped_pytest.sh pytest python/tests/ \
             discopt_benchmarks/tests/test_correctness.py \
+            discopt_benchmarks/tests/test_cert_neutrality_regime.py \
             -q --tb=short --timeout=120 -p no:cacheprovider \
             -m "correctness and not slow and not integration and not amp_benchmark and not requires_cyipopt and not memory_heavy" \
             --ignore=python/tests/test_amp.py \
@@ -660,7 +680,12 @@ jobs:
       - name: Run pytest with coverage
         run: |
           source .venv/bin/activate
+          # Kept a strict superset of `python-fast` (see that job's comment):
+          # the cert-tooling tests are added HERE TOO, or they would silently
+          # stop running on every push to main, which is the only event that
+          # skips python-fast.
           scripts/run_memory_capped_pytest.sh pytest python/tests/ \
+            discopt_benchmarks/tests/test_1134_cert_baseline_shrink_guard.py \
             -q --tb=short --timeout=120 \
             -m "not slow and not correctness and not integration and not amp_benchmark and not requires_cyipopt and not memory_heavy" \
             --ignore=python/tests/test_correctness.py \
--- a/crates/discopt-core/Cargo.toml
+++ b/crates/discopt-core/Cargo.toml
@@ -172,34 +172,43 @@ description = "Core MINLP solver: B&B engine, expression IR, presolve"
 # build; POUNCE 0.10.0 pins feral 0.15.1, so the shared-release alignment noted
 # above returns as soon as POUNCE follows to 0.16.0.
 #
-# RETRACTION (#1134, 2026-08-31, CLAUDE.md §11). The (b) paragraph above asserts
-# that "the certifying panel cannot" see an LU kernel bump's benefit, and so
-# measures net-positive only on the captured relaxation LPs. The panel cannot see
-# the benefit — that part stands — but it does see a COST that this record does
-# not carry, and two of the three certifications #1134 reports lost are it.
-# Measured on `main`, interleaved A/B, 3 reps, the two arms differing only in
+# RETRACTION (#1134, 2026-08-31, CLAUDE.md §11), narrow. The (b) paragraph above
+# asserts that "the certifying panel cannot" see an LU kernel bump, and so measures
+# net-positive only on the captured relaxation LPs. It cannot see the BENEFIT --
+# that part stands -- but it does see a per-node COST that this record does not
+# carry. Measured on `main`, interleaved A/B, 3 reps, arms differing only in
 # `LuParams::pivoting` and distinguished by a marker string baked into the
-# extension:
+# extension, restricted to instances that are NOT auto-routed and whose node count
+# is identical in both arms (so the comparison is per-node cost, not search) --
+# wall in seconds, Markowitz / GilbertPeierls:
 #
-#   cvxnonsep_nsig30   identical 165-node tree, identical objective.
-#                      GilbertPeierls 39.5 s (sd 1.05) vs Markowitz 59.4 s
-#                      (sd 0.80) — +49% wall, against a 60 s budget it used to
-#                      clear in 1.12 s on the reference machine.
-#   nvs05              identical 175-node tree. GilbertPeierls 57.0 s (sd 0.24)
-#                      certifies; Markowitz does not finish inside 60 s.
+#   nvs09  2.78/2.65   st_e29 2.07/2.03   ex1224 2.03/2.00   ex1225 1.88/1.84
+#   gkocis 1.35/1.39   ex1226 1.53/1.35   oaer   1.32/1.32   st_e36 6.83/7.03
+#   nvs22 15.35/12.11  nvs05 >=60.32/56.62  (the Markowitz arm censored at 60 s)
 #
-# Same trees, same answers: this is per-factorization overhead, not search. It is
-# the expected shape — Markowitz chooses its column order DURING factorization, and
-# on the panel's small, low-fill bases that search has nothing to amortize against,
-# while the 0.16.0 win (fill 8.32x -> 1.28x) is measured on large high-fill QPLIB
-# relaxation bases. The bump is NOT retracted: fill, `numerical` refusals and wall
-# on the LPs that dominate a real solve all improved, and no verdict was ever false.
-# What is retracted is the claim that this panel is blind to the change, and with it
-# the assumption that a kernel bump needs no wall measurement at panel scale.
+# Median ratio ~1.03. Two rows carry it: `nvs22` +27%, and `nvs05` >=+7%, which is
+# enough to lose a certification 3/3 reps that the GilbertPeierls arm holds at
+# 56.6 s (sd 1.35) -- one of the three #1134 reports lost. It is two-sided: on the
+# routed `m3` Markowitz is 2.4x FASTER (1.53 s vs 3.70 s). The shape is expected --
+# Markowitz chooses its column order DURING factorization, and on the panel's small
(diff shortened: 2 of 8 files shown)


jkitchin:
Reviewed the three defects pushed onto the branch — all three are real bugs in what I added, and the fixes are right. Recording where they land against what this PR published.

**(1) The meta claiming a refused run's provenance.** Mine, and the worst of the three: the checker would have reported a refused run's commit and host as the reference's, which is a *confidently wrong* answer to the exact question #1134 exists to make answerable. It was reachable by this PR's own end-to-end validation — that run refused the write and left a meta on disk. Driving the write from the recorded `baseline_written` flag rather than re-evaluating the condition is the right shape; the record can't drift from what happened.

**(2) `host_speed_ratio` admitting routed rows.** Also mine, and the sharper miss: Cause 2 in this very PR is the proof that equal node counts are necessary but not sufficient, and I used them as sufficient. **I have re-measured it** — the correction left it as an upper bound pending a panel re-run, and the box was free:

> `check_cert_neutrality.py` over all 52 reference instances at their budgets, same box, idle, corrected filter → **3.46x on 17 unrouted instances**.

So the contaminated figure was high by **0.01x**. The three inflated samples all landed above the median, which is the ~1.5 rank positions the correction predicted, and a median absorbed them. **Every conclusion drawn from 3.45x holds unchanged at 3.46x** — `clay0303hfsg` (15.29 s × 3.46 ≈ 53 s) and `nvs05` (20.53 s × 3.46 ≈ 71 s) both still exceed their 60 s budgets, and no attribution moves. What changes is that the number is now a measurement rather than a bound. Pushed as `c55dbfd9`, which updates §0.6.1 in place; **the PR body above has been corrected to 3.46x / 17 rows**, since a number in a PR body is read as a measurement.

That run also exercises both new reporting paths end to end: the provenance line correctly reports `NONE` (no meta is committed, because the reference was deliberately not regenerated), and `clay0303hfsg`, `nvs05` and `tanksize` each report their `status` violation annotated with the ratio — 17 violations in total, none suppressed by the annotation.

**(3) The `"algorithm_route" not in row` assertion.** Correct catch, and the failure mode is the pointed one: `to_dict` is `asdict`, so the regeneration §0.6.1 *prescribes* would have written the key into all 52 rows and failed my test with nothing wrong. A regression test that breaks on the intended next step is worse than no test there.

Nothing in (1)–(3) touches the solver path, and none of the fixes weakens a guard — (2) removes samples from a reporting-only statistic that never suppresses a violation.

Separately, the CI commit closes a gap worth naming: `cert-baseline.jsonl` is the reference every solver change is judged against and no lane exercised the code that generates or reads it, so my 21-test module would have gated nothing. Routing `test_cert_neutrality_regime.py` to `python-correctness` rather than the fast lane — where `not correctness` would have collected zero and reported success — is the distinction that matters.

Local state on the merged head: 33 tests pass (21 + 12). CI is green on `73c28bd2`; `c55dbfd9` is docs-only.

---

Issue #1148, opened 2026-09-03
mpec: source complementarity residuals and the local-vs-certified result contract (RFC #1123 slice 2)


jkitchin:
## Background

This is slice 2 of the two remaining implementation slices of the MCP/MPEC RFC (#1123),
which is being closed in favor of these focused issues. Read #1123 for the full design
rationale; **this issue is self-contained and can be picked up without it.**

It **depends on #1147** (first-class complementarity relation with durable source
provenance): a source residual cannot be computed after lowering until the source
operands survive a reformulation pass.

### 1. `solve_mpec` returns two different types with two different status vocabularies

Measured on the current `main`, same model (`min (x-1)^2 + (y-1)^2` s.t. `0 <= x ⊥ y >= 0`):

```
method="scholtes" -> discopt.solvers.NLPResult
                     status = SolveStatus.OPTIMAL   (enum)
                     gap_certified: attribute absent
                     x = [9.99999985e-01, 1.66549914e-08]

method="gdp"      -> discopt.modeling.core.SolveResult
                     status = "optimal"             (str)
                     gap_certified = True
```

Switching one keyword changes the returned type, the status type, and whether a
certification field exists at all. The Scholtes arm reports `OPTIMAL` for a local
stationary point of a *regularized* problem — and nothing in the result says the achieved
source complementarity was `1.67e-08` rather than zero, or that `t_min` was `1e-8`.

### 2. There is no local/stationary terminal status anywhere

- `discopt.modeling.core.SolveResult.status` is a plain `str`, documented (`:2183`) as
  `"optimal" | "feasible" | "infeasible" | "time_limit" | "node_limit" |
  "iteration_limit" | "error"`.
- `discopt.solvers.SolveStatus` (`python/discopt/solvers/__init__.py:10`) is
  `OPTIMAL | INFEASIBLE | UNBOUNDED | ITERATION_LIMIT | TIME_LIMIT | CUTOFF | ERROR`.

Neither has a member meaning "local stationary point, no global claim".

### 3. What the release gate does with a status it does not know — measured

The benchmark harness has its own `SolveStatus`
(`discopt_benchmarks/benchmarks/metrics.py:25`) and maps discopt's string status through
`status_map.get(result.status, SolveStatus.UNKNOWN)`
(`discopt_benchmarks/benchmarks/runner.py:342`). So a **new, distinct** status string
maps to `UNKNOWN`, `is_solved` is False, and both `incorrect_count` (`:290`) and
[...]

PR #1149, opened 2026-09-04
feat(mpec): #1147 first-class complementarity relation with durable source provenance

### LINKED ISSUE #1147: mpec: first-class complementarity/MCP relation with durable source provenance (RFC #1123 slice 1)
## Background

This is slice 1 of the two remaining implementation slices of the MCP/MPEC RFC (#1123),
which is being closed in favor of these focused issues. Read #1123 for the full design
rationale; **this issue is self-contained and can be picked up without it.**

Prior art in this line: #231 (closed) added the `Complementarity` object and its
disjunctive lowering, and left "the native complementarity node" as an explicit
follow-up — see the comment in `_add_nl_complementarity`
(`python/discopt/modeling/core.py:6208`). #658 preserved `.nl` complementarity rows into
the GDP path. #1124/#1128/#1146 (PR 0 of the RFC) settled the disjunction-semantics
vocabulary (`DisjunctionSemantics` at `python/discopt/modeling/core.py:378`) without
changing solver defaults. This slice does the same for complementarity: representation
and provenance only, no lowering, default, or result change.

## The defect, measured

`Model.complementarity(...)` records a pair on `Model._complementarities`
(`core.py:3991`) and immediately lowers it. **Every model-rebuilding pass in the tree
then discards the record.** Measured on the current `main`:

```python
import discopt
from discopt import Model
from discopt._relax.gdp_reformulate import reformulate_gdp

m = Model()
x = m.continuous("x", lb=0, ub=10)
y = m.continuous("y", lb=0, ub=10)
m.minimize((x - 1) ** 2 + (y - 1) ** 2)
m.complementarity(x, y, name="pair0")

print(len(m._complementarities))                       # 1
out = reformulate_gdp(m, "big-m")
print(len(getattr(out, "_complementarities", [])))     # 0
print([v.name for v in out._variables])
# ['x', 'y', '_gdp_aux_disj_pair0_0_0', '_gdp_aux_disj_pair0_1_1']
```

Result across the four passes that construct a fresh `Model` (all measured, `before=1`
in each case):

| Pass | Entry point | Pairs after |
|---|---|---|
| GDP lowering (`big-m`, `hull`, `mbigm`) | `_relax/gdp_reformulate.py:96` (rebuild at `:139`) | 0 |
| Integer-product expansion | `_relax/integer_product_reform.py:856` (rebuild at `:867`) | 0 |
| Factorable reformulation | `_relax/factorable_reform.py:factorable_reformulate` (rebuilds at `:1552`, `:1592`) | 0 |
| Binary-multilinear linearization | `_relax/binary_multilinear_reform.py:931` (rebuild at `:1007`) | abstains today — see below |

Each rebuild copies variables, parameters, objective and constraints and simply does not
carry `_complementarities`. What survives lowering is **string-level only**: the pair
name baked into generated identifiers (`_gdp_aux_disj_pair0_0_0`,
`_gdp_pair0_d0_c0_le`).

### Why that is the blocking problem

The next slice has to compute a **source** complementarity residual — e.g.
`r = max_i min(f_i(x), g_i(x))` on the *original* operands — and report it separately
from the lowered NLP's own residual. After lowering, the source operands are not
recoverable except by parsing generated variable names. A residual probe written against
the lowered model would measure the *relaxed row* instead of the *source product*, print
a small number, and be believed. That is precisely the instrument-that-measures-nothing
failure mode CLAUDE.md §6–§7 exist for, and it is the one thing the CCOpt comparison in
#1123 turns on (track source complementarity separately from the solver's own KKT
complementarity).

### Related fragility

`_model_carries_unsupported_state` (`_relax/binary_multilinear_reform.py:913`) makes the
binary-multilinear pass abstain when `_complementarities` is non-empty — a guard keyed on
state that any earlier rebuilding pass silently discards. This is **not** a live wrong
answer: after GDP lowering the structure is materialized into ordinary rows, so the pass
is legitimately applicable there. But the guard's premise evaporates at a pass boundary,
which is not a property worth keeping once provenance is durable.

## Scope

1. **Extend the relation.** `mpec.Complementarity`
   (`python/discopt/mpec.py:57`) is a three-field dataclass (`f`, `g`, `name`). Extend it
   to carry:
   - the source operands and their declared bounds;
   - a source scale (for a meaningful residual tolerance when operands have unrelated
     physical magnitudes);
   - a role (NCP pair / box-MCP variable-residual pair / generated-from-KKT / generated-
     from-disjunct);
   - the parent logical or disjunct identity when the pair was generated rather than
     user-declared;
   - shape/index information for vectorized pairs;
   - lowering state, so repeated solves and clones do not duplicate generated rows.
2. **Add the box-bounded MCP form** alongside the symmetric NCP pair: residual `F_i(z)`
   paired with a bounded variable `z_i in [l_i, u_i]`, with the NCP pair as the
   `l=0, u=+inf` special case. Represent it; do not lower it in this slice.
3. **Carry the pair map through every rebuilding pass**, explicitly. Each of the four
   passes above must translate and forward the relation set.
4. **Key provenance to durable object identities, never to indices.** Presolve/FBBT
   eliminate and renumber variables, so an index-based map is stale exactly when it
   becomes useful. Store the pair object and `Variable` identities; derive backend-facing
   indices at the solver boundary rather than persisting them through it.
5. **Raise when provenance cannot be resolved.** No silent drop, no `None` return, no
   best-effort name parsing.
6. **Keep `Model.complementarity` as a compatibility wrapper** over the richer relation.

## Acceptance criteria

- [ ] A regression that fails on `main` and passes after: a pair declared with
      `Model.complementarity(x, y, name="pair0")` is still resolvable to its **source
      operands** after `reformulate_gdp` under each of `big-m`, `hull`, `mbigm`, after
      `expand_integer_products`, and after `factorable_reformulate`.
- [ ] Resolution is by object identity, and a test asserts it survives a pass that
      renumbers or eliminates variables.
- [ ] Unresolvable provenance raises with a message naming the pair and the pass.
- [ ] Box-MCP pairs round-trip through the IR: lower-active, interior, upper-active,
      one-sided-infinite, and fixed (`l == u`) cases.
- [ ] Vector/array pairs keep elementwise identity through a rebuilding pass.
- [ ] `binary_multilinear_reform`'s abstain guard is re-expressed against durable state
      (or removed with a stated reason) rather than against a list an earlier pass empties.
- [ ] **No behavior change:** solver defaults, lowerings, statuses, and objective values
      are untouched. Per CLAUDE.md's bound-neutral regime, `node_count` and certified
      `objective` are exactly unchanged on a certifying panel.
- [ ] `pytest -m smoke`, `pytest -m slow python/tests/test_adversarial_recent_fixes.py`,
      and the existing complementarity tests
      (`python/tests/test_gdp.py:1588`, `python/tests/test_nl_complementarity.py`) pass.

## Non-goals for this slice

- Any new lowering, NCP function, or solve method.
- Source residual computation and the result contract — that is slice 2.
- Pure objectiveless MCP solves, native MCP backends (PATH), GAMS MCP equation-variable
  matching, general VIs.
- Widening the `.nl` importer's accepted complementarity forms. The doubly-bounded /
  free / equality-pinned refusal in `_add_nl_complementarity` stays a loud refusal here;
  the richer node makes lifting it *possible* later, and that is a separate issue.

## Files

- `python/discopt/mpec.py` — the relation, `complementarity()`, the three lowerings.
- `python/discopt/modeling/core.py` — `Model._complementarities` (`:2767`, `:3991`),
  `from_nl` complementarity reconstruction (`:6185`), `_add_nl_complementarity` (`:6208`).
- `python/discopt/_relax/gdp_reformulate.py`, `integer_product_reform.py`,
  `factorable_reform.py`, `binary_multilinear_reform.py` — the rebuilding passes.

## Downstream

#1148 (source residuals + local-vs-certified result contract) depends on this. Both
together unblock the shared Gate 1 phase-appearance flash in
[jkitchin/pounce#776](https://github.com/jkitchin/pounce/issues/776), under the
applications epic [jkitchin/pounce#780](https://github.com/jkitchin/pounce/issues/780).
POUNCE Gate 0 ([jkitchin/pounce#794](https://github.com/jkitchin/pounce/issues/794)) is
complete and its supported route is `scholtes_then_ncp`.

Background: #1123.

jkitchin:
## Summary

Slice 1 of the MCP/MPEC RFC (#1123): **representation and provenance only** — no new
lowering, no default change, no solver-behaviour change.

Measured on `main`, `Model.complementarity` recorded a pair on `Model._complementarities`
and **every** pass that rebuilds a `Model` discarded it — `before=1 / after=0` for GDP
lowering under `big-m`/`hull`/`mbigm`, for `expand_integer_products`, and for
`factorable_reformulate`. What survived a lowering was string-level only: the pair name
baked into generated identifiers (`_gdp_aux_disj_pair0_0_0`). The next slice (#1148) has to
report a **source** complementarity residual over the *declared* operands; written against a
lowered model it would have measured the relaxed row instead, printed a small number, and
been believed — the instrument-that-measures-nothing failure CLAUDE.md §6–§7 exist for.

What changed:

- **The relation is now the durable object.** `mpec.Complementarity` carries the source
  operands, the bounds the *relation* declares on them, a residual `scale`, a `role`
  (`ComplementarityRole`: NCP pair / box MCP / from-KKT / from-disjunct), the generating
  `parent`'s identity, and shape/index for vectorized relations. It is identity-compared
  and identity-hashed, so it stays a usable provenance key as passes carry it from one
  `Model` to the next (value equality was never available anyway: `Expression.__eq__`
  builds a `Constraint`).
- **The box-bounded MCP form**: `Model.mcp` / `mpec.box_mcp`, residual `F(z)` paired with
  `z in [l, u]`. `l=0, u=+inf` is recorded *as* the NCP pair and lowered, so the two forms
  cannot diverge on the case they share. Which form a relation is, is read off its
  **declared bounds** (`is_symmetric_nonnegative`), never off its `role` — the role is
  provenance, and `Complementarity` is public, so a directly-constructed relation with box
  bounds and a default `role=NCP_PAIR` must not be able to talk its way into a two-branch
  lowering. The general box form is **represented and not lowered**: every
  `mpec.reformulate_*` refuses it, and so does the solver boundary — `solve_model`, so the
  callers that never go through `Model.solve` are refused too.
- **All four rebuilding passes forward the relation set explicitly** via
  `carry_complementarities`, which resolves each relation against the rebuilt model **by
  object identity** and raises `ComplementarityProvenanceError` — naming the relation and the
  pass — rather than dropping one it cannot resolve. The three passes with a defensive
  `except Exception: return model` re-raise that error ahead of it, so a broken provenance
  chain can never degrade to a silent abstain.
- **Flat indices are derived at the solver boundary** (`flat_source_indices`), never
  persisted, and are prefix sums over the **target model's** variable list rather than
  `Variable._index` (the declaring model's position): operands are shared objects, so a model
  holding them in a different order needs its own layout. Element-accurate for a statically
  indexed operand, widening only for an index that cannot be resolved statically.
- **Lowering state is model state** — `Model._lowered_complementarities`, an identity map
  from relation to the method that lowered it into *that* model, read via
  `pair.is_lowered_into(model)` / `pair.lowering_in(model)`. A relation is shared, so neither
  the mark nor the method can live on it.
- **`binary_multilinear_reform`'s abstain guard no longer keys on `_complementarities`.**
  Every earlier pass emptied that list, so the guard's premise evaporated at a pass boundary:
  after GDP lowering the condition is materialized into ordinary rows and the list was empty.
  Forwarding the relation set replaces it, and the pass now fires on the GDP-lowered models
  it was wrongly abstaining from. The SOS/indicator and builder-block clauses are unchanged.
- `bilevel.kkt` labels its generated pairs `FROM_KKT` with `parent` naming the lower-level
  row, and `bilevel/problem.py` records them on the model after lowering so that provenance
  is live rather than inert. Unnamed relations get a model-unique name at the single point
  every lowering funnels through — the fallback used to count within the one-element list
  handed to the lowering, so every unnamed declaration became `compl0`.
- New normative doc `docs/complementarity_semantics.md`, the companion of
  `docs/disjunction_semantics.md` (#1124), added to `docs/_toc.yml`.

Closes #1147.

## Correctness

No solver math changed: the relation set is a Python-side record that no lowering, bound,
cut or relaxation reads. The behaviour changes are all *refusals* in cases that could not
previously arise — a model carrying an unlowered box-MCP relation (a form with no
declaration path before this PR), a relation whose declared bounds are not the symmetric
pair, and a rebuilding pass that cannot resolve a relation's source operands. Nothing was
weakened to make a check pass; the one guard removed
(`binary_multilinear_reform`'s `_complementarities` clause) was keyed on state an earlier
pass emptied, and is replaced by explicit forwarding that raises rather than drops.

Bound-neutral evidence (CLAUDE.md §5, bound-neutral regime) over the 66-instance in-repo
`.nl` corpus (`python/tests/data/minlplib_nl/`) at 10 s/instance, before vs after, **both
arms on the same boot of an idle host**, each asserting the loaded module and a version
marker present/absent (§8) — `scratchpad/i1147/panel_before3.json` / `panel_after3.json`:

- status **identical on 66/66** instances;
- every recorded field (status, objective, bound, `node_count`, `gap_certified`)
  **identical on all 45 instances that converged in both arms**;
- the 7 remaining differences are all unconverged (`time_limit`/`feasible`) rows whose node
  count varies with how much wall clock the fixed budget buys — not a bound change.

> **Retractions (§11).** Two earlier revisions of this description reported panel numbers
> that do not stand. The first ("identical 66/66 … 44 converged, 2 `node_count` differences")
> compared arms measured under *different* machine load — the container restarted between
> them — so their unconverged rows were never comparable. The second (the `*2` arms, "45
> converged, 4 differences") was measured on a boot that has since been replaced. Only the
> `*3` pair above, measured on one boot for this head, stands. The superseded arms and why
> they were discarded are recorded in `scratchpad/i1147/README.md`.

- [x] Cannot produce a false certificate (no solver-math change)

- [x] `pytest -m smoke` → 1119 passed, 20 skipped, 2 xpassed
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → 19 passed
- [x] `ruff check` / `ruff format --check` → clean (960 files); `mypy python/discopt/` → no
      issues in 335 source files

Also green: `test_mpec.py`, `test_mpec_mp1_mp2.py`, `test_nl_complementarity.py`,
`test_gdp.py`, `test_bilevel_phase{1,2,3}.py` — 221 passed together with the provenance file.

An earlier full non-slow run (`-n 4`) reported 8 failures, none in a file this change
touches: 4 were load flakes that pass on a serial re-run, and 4 fail **identically on
`main`** in this container (verified by checking out `main`'s `python/discopt` with the
version marker asserted absent) because they need optional backends it does not provide
(cyipopt, POUNCE). CI is the authority on those.

## Regression test

`python/tests/test_complementarity_provenance.py` — 50 tests. The relation survives each of
`big-m`/`hull`/`mbigm`, `expand_integer_products`, `factorable_reformulate` and
`reformulate_binary_multilinear`; resolution survives a pass that renumbers *and* renames
variables; unresolvable provenance raises naming pair and pass, and is not swallowed by a
pass's defensive handler; vector and broadcast relations keep elementwise identity; box-MCP
round-trips for lower-active, interior, upper-active, one-sided-infinite and fixed (`l == u`)
boxes.

Thirteen of those lock review findings, each verified failing on the head it was reported
against and passing after. From the first review: the `solve_model` bypass, the role-gated
lowering (and the same hole in bound tightening), `deepcopy`/`pickle` of a model carrying a
relation, unnamed-relation name collisions, the inert bilevel KKT provenance,
element-accurate flat indices, `elements()`'s `[self]` contract, and a genuine `ub=2e19`
staying finite. From the second: flat indices following the *target* model's order with two
live models and nothing mutated, per-model lowering methods, the unlowered-source carry that
laundered a lowered mark past the solve guard, and the carry preserving the source's own
method.

Fail-before/pass-after for the original slice: the file cannot import on `main` (the API does
not exist there), and the underlying defect is measured directly by
`scratchpad/i1147/repro_1147.py` — `after=0` on `main` for every pass, `after=1` after this
change.

---

https://claude.ai/code/session_01X83REmDmGJghCMhgfLMizJ


jkitchin:
## Review

CI is green across the board (30 new tests in `test_complementarity_provenance.py`). The core of this PR is right, and it does address the central defect described in #1147: `carry_complementarities` closes the `before=1 / after=0` drop through all four rebuilding passes, and it raises rather than drops. **But I would not merge it as-is** — three HIGH findings hold up under independent reproduction, and two of them are regressions against `main`.

All eight findings below were verified in a worktree at the PR head, asserting `discopt.__file__` and the `carry_complementarities` marker (present on the branch, asserted **absent** on the `main` baseline, per CLAUDE.md §8).

### HIGH — confirmed

**1. The unlowered-relation refusal is on the wrong function.** It lives in `Model.solve` (`python/discopt/modeling/core.py:4754`), not in `solve_model()`. Every caller of `solve_model` — `_relax/differentiable_solve.py:268,310`, `_relax/differentiable.py:518`, `_relax/primal_heuristics.py:2589` — silently drops a declared box MCP.

Reproducer: `z in [-1,1]`, `maximize z`, `m.mcp(z + 1.0, z)`. The declared relation forces `z = -1`.

```
Model.solve  -> REFUSED (NotImplementedError)      <- correct
solve_model  -> status=optimal  obj=0.99999999  z=0.9999999975
   MCP branch: z==u  ==>  requires F<=0 ;  F = +2.0
   CERTIFIED POINT VIOLATES DECLARED RELATION
```

A certified `optimal` on a point that violates the declared relation is a §1 violation with no slack.

**2. `_require_lowerable` branches on `role`, not on declared bounds** (`python/discopt/mpec.py:659`; same hole at `:805` in `tighten_complementarity_bounds`). This contradicts `ComplementarityRole`'s own docstring at `mpec.py:91` and the new `docs/complementarity_semantics.md` §3 — "role is *provenance*, never a lowering switch." `Complementarity` is exported in `mpec.__all__`, so constructing one directly keeps `role=NCP_PAIR` even with a declared box:

```
mpec.Complementarity(z + w, z, "p0", g_bounds=(-1.0, 1.0))
  -> reformulate_gdp lowers it, emitting  z >= 0  against the declared lb=-1
  -> status=optimal  obj=1.0000002   (z=2.19e-07)
```

The PR's own normative doc is the spec the code fails.

**3. `_lowered_into` holds a `weakref`** (`python/discopt/mpec.py:220`), which breaks `deepcopy` and `pickle`. Both are regressions:

```
PR branch:  clone.solve()   -> NotImplementedError (model declares complementarity
                                relation(s) with no lowering)
            pickle.dumps()  -> TypeError: cannot pickle 'weakref.ReferenceType' object

main:       clone.solve()   -> status=optimal obj=1
            pickle.dumps()  -> ok
```

The deepcopy case is the worse one: the clone is *silently* stripped of its lowering record and then loudly refuses, so an ordinary model-copy workflow stops working.

### MEDIUM — confirmed

**4. The name-collision fix went into only one of two entry points** (`python/discopt/modeling/core.py:6366`). Two unnamed `m.complementarity(...)` calls through the public API:

```
constraint names: ['compl0_f_nonneg','compl0_g_nonneg','compl0',
                   'compl0_f_nonneg','compl0_g_nonneg','compl0']
COLLISIONS: {'compl0_f_nonneg': 2, 'compl0_g_nonneg': 2, 'compl0': 2}   (no warning)
```

The fix landed in `_add_nl_complementarity` only — §2, fix the class not the instance.

**5. The bilevel KKT provenance labels are inert.** `python/discopt/bilevel/kkt.py:94-95` sets `role=FROM_KKT` and `parent=...`, but nothing under `bilevel/` ever touches `model._complementarities` (grepped: zero references in the package). `bilevel/problem.py:431` calls `reformulate_gdp(self.model, kkt.comp_pairs)` with the list passed directly. So `carry_complementarities` never sees these pairs — the new provenance is dead code on the only in-tree path that produces it.

**6. `flat_source_indices` over-reports for an indexed operand** (`python/discopt/mpec.py:473`). It extends by `range(off, off + v.size)` for the whole parent variable:

```
x[2] _|_ y   on shape-(3,) x   ->   [0, 1, 2, 3]     (4 columns for a 2-column relation)
```

This is the solver-boundary accessor, so a consumer masking a residual by these indices touches columns the relation never reads.

### LOW — confirmed

**7. `elements()` breaks its own contract** (`python/discopt/mpec.py:287`). The docstring says "A scalar relation yields `[self]`." Actual behaviour is a fresh child on each call — `e1[0] is self` -> `False`, and `e1[0] is e2[0]` -> `False`. The identity-keyed provenance this PR advertises does not hold for scalar relations. (`e1[0].source is self` is `True`, so the back-pointer works; it is the advertised identity key that does not.)

**8. `_UNBOUNDED_SENTINEL = 1e19`** (`python/discopt/mpec.py:75`) silently reinterprets a genuine large finite bound as infinite: `m.mcp(..., ub=2e19)` yields `g_bounds=(0.0, inf)`. It is also a fourth, mutually inconsistent copy of a sentinel — `discopt/constants.py` has `CONSTRAINT_INF = 1e20` / `SENTINEL_THRESHOLD = 1e29`, and `bilevel/problem.py` has `_BIG = 1e19`.

### On the `binary_multilinear_reform` abstain-guard removal

I flagged this separately as the highest-risk item in the diff and read it directly rather than let its absence from the finding list read as clearance. **It is defensible.** The guard keyed on `_complementarities` being non-empty; before this PR that list was emptied by every upstream rebuilding pass, so the guard fired only on models the pass would have handled correctly anyway. With `carry_complementarities` now preserving the list *through* lowering, keeping the guard would make the pass newly abstain on models it previously reformulated. Replacing it with explicit forwarding-that-raises is the right trade.

One residual, worth a comment rather than a change: the pass Fortet-linearizes binary monomials in constraint and objective bodies but not in a relation's `f`/`g` bodies (the new comment at the `carry_complementarities` call site asserts this). A relation whose operands contain a binary product therefore survives un-linearized and emits a nonlinear row if it is lowered later.

### Bottom line

HIGH 1 and HIGH 2 each produce a certified `optimal` on a point that violates the declared relation — merge blockers under CLAUDE.md §1. HIGH 3 is a functional regression on `deepcopy`/`pickle` versus `main`. The MEDIUMs are genuine but not certificate-affecting.

Suggested disposition: hold for 1-3 plus 4 (§2, same class of bug), and let 5-8 ride as follow-ups if the main change should land sooner.

jkitchin:
All eight findings reproduced exactly as reported, and all eight are fixed in 1bd4641. I ran your reproducers before touching anything — `scratchpad/i1147/verify_review.py` is committed and re-checks each one. HIGH 1 in particular came back with a certified `optimal` at `z = 1` where the MCP's `z = u` branch requires `F <= 0` and `F = +2`, exactly as you had it.

**HIGH 1 — refusal on the wrong function.** Moved to `mpec.require_all_relations_lowered(model, context=...)` and called from `solve_model`, the boundary every solve passes through. `Model.solve` still calls it too, so an alternate backend (`direct`/`amp`/`gurobi`/`mip-nlp`) is refused before it dispatches rather than after. Both paths now refuse your reproducer.

**HIGH 2 — `role` gating the lowering.** You're right that the PR's own doc was the spec the code failed. `_require_lowerable` and `tighten_complementarity_bounds` now branch on `Complementarity.is_symmetric_nonnegative` — the declared bounds. Your `Complementarity(z + w, z, "p0", g_bounds=(-1.0, 1.0))` is refused by all three lowerings while keeping `role=NCP_PAIR`, which is the point: the label can be wrong, the bounds are the semantics.

**HIGH 3 — weakrefs.** The mark moved off the relation onto the model: `Model._lowered_complementarities`, an identity set. `deepcopy` and `pickle` both work again, and the clone solves rather than refusing. Model-owned also states the fact more directly — *these are the relations whose rows I carry*.

**MEDIUM 4 — one entry point only.** Fixed as a class: `_ensure_relation_name` assigns a model-unique name at the single point every lowering funnels through (`_scalarize_pairs`), so the `.nl` importer, the fluent API and a hand-built pair all get one. Your two-unnamed-calls case now yields `compl0`/`compl1` with no collisions.

**MEDIUM 5 — inert KKT labels.** Correct, and worth catching: `bilevel/problem.py` now records the pairs on the model after lowering, so `FROM_KKT` and the `parent` row are live provenance the passes carry. Recorded *after* the lowering so `solve` doesn't refuse them.

**MEDIUM 6 — over-reporting.** `flat_source_indices` is element-accurate for a statically resolvable index (`x[2] ⊥ y` → `[2, 3]`). An index it can't resolve statically still widens to the whole variable, deliberately: over-reporting is the safe direction for a masking consumer, and that's now documented at the function and in the doc.

**LOW 7 / LOW 8.** `elements()` returns `[self]` for a scalar relation, so the identity key holds. The sentinel derives from the shared `constants.CONSTRAINT_INF` instead of being a fourth private copy, and `ub=2e19` stays finite.

On the `binary_multilinear_reform` note: agreed, and I've left the residual you describe as a comment at the call site rather than a change — a relation whose operands contain a binary product survives un-linearized. Worth an issue when a lowering can consume it.

**Verification.** 9 new tests, one per finding, each failing on the PR head (a350f09) and passing here — 46 in `test_complementarity_provenance.py`. smoke 1119 passed; adversarial slow 19 passed; ruff and mypy clean.

I also **retract the bound-neutral numbers in the PR body**. Those two arms were measured under different machine load (the container restarted between them), so their unconverged rows were never comparable — 13 rows moved across conditions while the pre-fix arm tracked `main`, which is the host talking, not the code. Re-run with both arms on the same idle host: status and objective identical 66/66, every recorded field identical on all **45** instances that converged in both arms, and the 4 remaining differences are unconverged rows whose node count varies with how much wall clock the fixed 10 s budget buys. Raw arms are in `scratchpad/i1147/panel_before2.json` / `panel_after2.json`; the superseded pair is called out in that directory's README.

---

jkitchin:
## Re-review at `1bd4641b`

All eight findings are addressed. I re-verified each one independently rather than reading the response, in a worktree at the new head with `discopt.__file__` and the `require_all_relations_lowered` marker asserted (CLAUDE.md §8).

| Finding | Result |
| --- | --- |
| HIGH 1 | `solve_model` **and** `Model.solve` both raise `NotImplementedError` on the `max z s.t. mcp(z+1, z), z in [-1,1]` reproducer |
| HIGH 2 | `Complementarity(z+w, z, "p0", g_bounds=(-1,1))` is refused by **all three** lowerings — gdp, sos1 and scholtes (with its real 3-arg signature) — while `role` stays `NCP_PAIR` |
| HIGH 3 | `deepcopy` -> clone solves `optimal`; `pickle` round-trip -> solves `optimal` |
| MEDIUM 4 | two unnamed pairs -> `compl0*` / `compl1*`, zero collisions |
| MEDIUM 5 | `bilevel/problem.py:442` extends `model._complementarities` (confirmed at source level — my first grep matched a stale `.pyc`, which is not evidence) |
| MEDIUM 6 | `x[2] _|_ y` -> `[2, 3]` |
| LOW 7 | `elements()[0] is self` -> `True` |
| LOW 8 | `ub=2e19` -> `g_bounds=(0.0, 2e+19)`, stays finite |

### Checks beyond what the response claimed

**The 9 new tests genuinely bisect.** I built a separate pre-fix worktree at `a350f097`, asserted the fix marker **absent** there, and copied the new test file in: all 9 fail on the pre-fix head, all 46 pass at `1bd4641b`.

**The bounds gate does not over-refuse.** This was the real regression risk in the HIGH 2 fix — gating on declared bounds could start rejecting models that previously lowered fine. It does not: expression operands (`x - 1 _|_ y + 2`), operand variables with a negative `lb`, and a `[0, inf)` box MCP all still lower and solve. Only the genuine two-sided box is refused.

**The lowered-set redesign is really identity-keyed.** `@dataclass(eq=False)` gives identity hash/eq, so with two value-identical relations where only one is lowered, only that one is marked and the sibling is still refused at solve. (My first probe here was wrong — it asserted against `Model.complementarity`, which lowers immediately by design, so both were legitimately marked. Redone with detached relations.)

**The mark survives rebuilding passes.** `carry_complementarities` forwards it via `pair.mark_lowered(dst, pair.lowering)`, and a model with binary products that forces a rebuild solves `optimal` end to end.

**The retracted panel numbers reconcile exactly.** I reparsed the raw arms rather than trusting the summary: 66 instances, status and objective identical 66/66; the 45 rows with `status == optimal` are identical in **every** recorded field; the only 4 differing rows are `feasible` / `time_limit` and differ solely in `bound` and `node_count`. Both arms carry a `marker` field (`false` / `true`), so the §8 load-gate discipline is baked into the artifact itself. Retracting the earlier load-contaminated pair was the right call and is documented in the directory README.

### Caveats

* **Local smoke shows 17 failures that are not this PR's.** All 17 fail identically on `main` in my environment (`test_1060_native_single_tree`, `test_1141_fractional_node_cuts`, `test_issue_1060_open_bounds_marshal`, `test_issue_1066_lp_nlp_bb_progress_budget`) — a stale local Rust build. CI is the authority.
* **CI is not finished** as of this comment: 7 of 9 pass, including Rust and claim-boundary; `Python fast` and `Python correctness (fast subset)` are still pending.
* **MEDIUM 6 is narrowed rather than eliminated, by design.** A statically resolvable index is exact; an index that cannot be resolved statically still widens to the whole variable, now documented as deliberate over-reporting. Defensible for a masking consumer.
* The `binary_multilinear_reform` residual — a relation whose operands contain a binary product survives un-linearized — is left as a call-site comment rather than a change. Reasonable, but nothing tracks it; §3 wants a workaround to ship with an issue and a comment linking it.

### Assessment

The blockers are cleared. My recommendation is to merge once `Python fast` and `Python correctness (fast subset)` go green, ideally after filing the one follow-up for the un-linearized-operand residual.

jkitchin:
@bernalde do you want to take a look at this before it merges?

python/discopt/mpec.py
```diff
@@ -36,39 +36,678 @@
...
+    An index that cannot be resolved statically (a slice, a computed index)
+    falls back to the whole variable. That direction is deliberate: this is a
+    *conservative* answer, never a narrower one than the truth, so a consumer
+    can rely on the set covering everything the relation reads.
+    """
+    resolved = resolve_source_variables(model, pair, context="flat_source_indices")
+    slots = _read_slots(pair)
+    out: list[int] = []
+    for v in resolved:
+        off = model._flat_var_offset(v)
```
bernalde: Blocking: this asks `_flat_var_offset` for an offset using `Variable._index`, but that index belongs to the shared variable object, not to the target model. With a relation over a two-element `x` and scalar `y`, a target model ordered as `[y, x]` resolves the right objects yet this returns `[0, 1, 1]`; the target layout requires `[1, 2, 0]`. The current renumbering test hides this by rewriting `_index` on the shared variables, which also changes every other model holding them. Build a target-local identity-to-prefix-offset map from `model._variables` (or provide an equivalent model API), and add a regression with two simultaneously live models that order differently sized shared variables differently without mutating them. Otherwise the residual work in #1148 can mask the wrong or duplicate solver columns.

python/discopt/mpec.py
```diff
@@ -36,39 +36,678 @@
...
+        The mark lives on the **model** (``Model._lowered_complementarities``,
+        an identity set), not on the relation. It first lived here as a list of
+        ``weakref``s, which made a ``Model`` carrying a relation unpicklable
+        (``TypeError: cannot pickle 'weakref.ReferenceType'``) and made
+        ``copy.deepcopy`` silently drop the mark — so the clone then refused to
+        solve, a regression against a plain model copy on ``main`` (review of
+        #1149, HIGH 3). Model-owned state also says the thing more directly:
+        "these are the relations whose rows I already carry".
+        """
+        self.lowering = method
```
bernalde: Blocking: the lowering method is still global mutable state on the shared relation, despite the lowering-presence mark being model-owned. Lowering one relation with GDP in `m1` and SOS1 in `m2` leaves both models marked but changes `pair.lowering` to `"sos1"`, so `m1`'s provenance no longer describes its disjunctive rows. More seriously, `carry_complementarities` uses any non-`None` `pair.lowering` as proof that `src` was lowered: an unlowered source sharing this relation produces a destination marked as lowered even though neither model carries generated rows, bypassing the solve guard. Store relation-to-method state on each model, and carry a mark only when `pair.is_lowered_into(src)` is true, preserving that source model's method. Please cover both different lowerings in two models and the unlowered-source carry case.

bernalde (changes requested):
I found two blocking provenance-correctness issues on head `1bd4641b0e9b184bb7c662e9e6482c3b376c1dc8`.

### Review summary

- Blocking issues: 2 (2 inline)
- Nonblocking issues: 0
- Questions: 0
- Tests/checks:
  - `/home/bernalde/repos/discopt/.venv/bin/python -m pytest tests/test_complementarity_provenance.py -q -rs -k 'not nl_imported'`: 45 passed, 1 deselected. The deselected `.nl` integration case was not reliable in the local worktree because the active environment loaded its native extension from the primary checkout; the exact-head GitHub Python and Rust lanes are green.
  - `/home/bernalde/repos/discopt/.venv/bin/jupyter-book build docs/`: succeeded, including the new complementarity semantics page.
  - Direct PR-head reproductions confirmed both the target-model index mismatch and cross-model lowering-state leak described inline.
  - Current-head GitHub checks completed successfully for lint/typecheck, Rust, Python fast, the fast correctness subset, claim-boundary, AMP coverage, and both Codecov contexts. Slow-on-demand, coverage, and AMP Alpine/incidence jobs were intentionally skipped.
- Coordination: open PR #1150 overlaps `gdp_reformulate.py` and `solver.py`; its current patches touch different identifiers and I found no incompatibility with this head.
- Merge readiness: I would not merge this until the blocking issues above are addressed.

python/discopt/mpec.py
jkitchin: Confirmed and fixed in 300bacc. Your `[y, x]` case reproduced with exactly the `[0, 1, 1]` you predicted.

`flat_source_indices` now builds a target-local identity→offset map from `model._variables` rather than going through `_flat_var_offset`. I left `_flat_var_offset` alone and documented the distinction at the new helper instead: it is correct for the rebuilding passes, which preserve declaration order, and wrong as a general provenance accessor — that's the trap, since every in-tree caller happened to satisfy its assumption. O(n) per call and deliberately not memoized; a cache keyed on anything but the live list is how the stale answer arrived in the first place.

You were also right that the renumbering test was hiding it, and for the worse reason: it rewrote `_index` on variables other live models hold. That mutation is gone with a note saying why, and `test_flat_indices_follow_the_target_models_order` tests the actual property — two simultaneously live models ordering a two-element `x` and a scalar `y` differently, nothing mutated, asserting `[0, 1, 2]` and `[1, 2, 0]` respectively, that each model's answer is unaffected by querying the other, and that `_index` is untouched afterwards.

---

python/discopt/mpec.py
jkitchin: Both halves confirmed and fixed in 300bacc. The second one is the serious one, and you're right that it defeats the guard added for the previous review's HIGH 1: an unlowered `src` sharing a relation handed `dst` a lowered mark, `unlowered_relations(dst)` came back empty, and `dst.solve()` sailed through with the condition unenforced.

`Model._lowered_complementarities` is now an identity **map** from relation to method. The `lowering` field is gone — not deprecated, removed — replaced by `pair.lowering_in(model)`, and `carry_complementarities` propagates `src`'s own method only when `pair.is_lowered_into(src)`. Making the mark model-owned while leaving the method global was the same mistake one level down; the rule I've written into the doc is that a relation is shared, so what a given model did with it is that model's fact, and you ask the model carrying the rows you're reasoning about.

Three tests, all failing on 1bd4641: GDP into `m1` and SOS1 into `m2` reporting `"gdp"` and `"sos1"` respectively (and asserting no global field exists); the unlowered-source carry, asserting the relation is still carried, the mark is not, and `dst.solve()` raises; plus the positive direction — a model rebuilt from `m1` inherits `"gdp"` even after a later SOS1 lowering elsewhere. That last one caught a mistake in my own first draft: `mpec.reformulate_*` lowers without recording on the model, so the test had to go through the fluent API for the carry to see anything.

---

jkitchin:
Both blocking issues are fixed in 300bacc; replies are on the two threads and I've resolved them.

Both reproduced exactly against `1bd4641` before I touched anything — including the `[0, 1, 1]` you predicted for the `[y, x]` target — and `scratchpad/i1147/verify_review2.py` is committed so the reproductions are re-runnable.

**Blocking 1.** Target-local identity→offset map from `model._variables`, replacing the `_flat_var_offset` call. I left `_flat_var_offset` itself alone and documented the distinction at the new helper: correct for the rebuilding passes (they preserve declaration order), wrong as a general provenance accessor. Your point about the renumbering test mattered independently — it was rewriting `_index` on variables other live models hold, so it was both hiding this and corrupting shared state. Gone, replaced by the two-live-models regression you asked for.

**Blocking 2.** `Model._lowered_complementarities` is now an identity map from relation to method; the `lowering` field is removed in favour of `pair.lowering_in(model)`; and `carry_complementarities` propagates `src`'s own method only when `pair.is_lowered_into(src)`. Making the mark model-owned while leaving the method global was the same mistake one level down. Both of your cases are covered plus the positive direction (a model rebuilt from `m1` inherits `"gdp"` after a later SOS1 lowering elsewhere).

**Verification.** 4 new tests plus the 2 corrected ones, all failing on `1bd4641` and passing here — 50 in `test_complementarity_provenance.py`. smoke 1119 passed; adversarial slow 19 passed; ruff and mypy clean.

The bound-neutral panel was **re-run for this head**, both arms on the same boot of an idle host: status identical 66/66, every recorded field identical on all **45** instances that converged in both arms; the 7 differing rows are all unconverged (`time_limit`/`feasible`) and vary with how much wall clock the fixed 10 s budget buys. The `*2` arms quoted in the PR body are superseded — the container had restarted between them and this head — and the directory README records which pair is current and why the earlier ones were discarded. I'll update the PR body's numbers to the `*3` pair.

Thanks for the `#1150` overlap check on `gdp_reformulate.py` / `solver.py` — my change to `solver.py` is a single guard call at the top of `solve_model` and the one in `gdp_reformulate.py` is the `carry_complementarities` call before `return new_model`, so I'd expect the same clean separation you found.

---

bernalde (approved):
Follow-up review on address commit `300bacc02b60716c7f0e9f0e625fdc70f0ade71d`.

Both blocking provenance findings from the [previous review](https://github.com/jkitchin/discopt/pull/1149#pullrequestreview-5115268644) are resolved:

- `flat_source_indices()` now derives identity-keyed offsets from the target model's variable order, so shared operands are mapped to the correct backend columns without mutating `Variable._index`.
- Lowering methods now live in each model's `_lowered_complementarities` identity map, and `carry_complementarities()` propagates only the source model's own lowering state and method.

The focused tests cover the target-order counterexample, independent per-model lowering methods, an unlowered source that must not launder a lowering mark, and preservation of the source method during carry. This satisfies the relevant provenance and durable-identity requirements in #1147.

### Review summary

- Blocking issues: 0
- Nonblocking issues: 0
- Questions: 0
- Tests/checks:
  - `python -m pytest tests/test_complementarity_provenance.py -q -rs -k 'not nl_imported' -p no:cacheprovider`: 49 passed, 1 deselected. The warnings were limited to the sandbox refusing JAX persistent-cache writes.
  - Exact-head mutation probes restored each rejected implementation: the old offset path failed the target-order regression, and global relation-level lowering state failed the three per-model/carry regressions.
  - Current-head GitHub checks succeeded for lint/typecheck, Rust, Python fast, the fast correctness subset, claim-boundary, AMP coverage, and both Codecov contexts. Slow-on-demand, coverage, and AMP Alpine/incidence jobs were intentionally skipped.
- Coordination: open PR #1150 still overlaps `gdp_reformulate.py` and `solver.py`, but its current patches touch separate identifiers and do not conflict semantically with this change.
- Merge readiness: no blocking findings remain; this is ready to merge from a review standpoint.

jkitchin:
## Correctness review

Careful PR — two prior review rounds are visible in the code, and most of what I probed holds up. One blocking regression, two smaller things.

---

### 🔴 Blocker: `_pending()` ignores *which* method lowered a relation, so a requested Scholtes regularization is silently not emitted

`_pending(model, pairs)` skips any relation with `is_lowered_into(model)` — regardless of the method that lowered it. `mark_lowered` records the method, but `_pending` never reads it. So asking for a *different* lowering after one is already in place emits **nothing**.

```python
minimize (x-1)**2 + (y-1)**2,   x, y in [0, 5]
m.complementarity(x, y, method="gdp", name="p")        # auto-lowers
solve_mpec(m, m._complementarities, method="scholtes")
```

```
main:     x=0.000000  y=1.000000   x*y=1.7e-08   COMPLEMENTARY
this PR:  x=1.000000  y=1.000000   x*y=1.0e+00   *** NOT COMPLEMENTARY ***
```

`reformulate_scholtes` adds **0 rows** (constraints `4 -> 4`; no `p_reg` row emitted). `solve_mpec`'s Scholtes branch is a local NLP over the raw model — it never reads the GDP `_DisjunctiveConstraint` objects — so the declared condition is simply absent, and `_mpec_t` becomes a parameter no row references: the homotopy loop then iterates `max_iter` times over an unchanged problem. That is the silent-no-op-instrument failure (CLAUDE.md §6/§7) this PR exists to prevent, reintroduced one level up. The returned point violates a constraint the user declared.

**Verified fix, one line** — key `_pending` on the method:

```python
def _pending(model, pairs, method):
    return [p for p in pairs if p.lowering_in(model) != method]
```

With it, the pre-lowered flow returns `(0, 1)` again and the MPEC panel below stays bound-neutral. This still satisfies `_pending`'s stated purpose (never lower the *same* relation by the *same* method twice into one model).

Caveat on the complete fix: `mark_lowered` stores one method per `(relation, model)`, so `gdp -> scholtes -> gdp` would re-emit the GDP rows. A *set* of methods per `(relation, model)` closes it fully.

No test covers "a second, different lowering after the first" — all 74 tests across `test_complementarity_provenance.py` / `test_mpec.py` / `test_mpec_mp1_mp2.py` / `test_nl_complementarity.py` pass on this head.

---

### 🟡 The abstain-guard claim in the description is backwards

> "the pass now fires on the GDP-lowered models it was wrongly abstaining from"

Measured: on `main`, after `reformulate_gdp` the `_complementarities` list is already **empty**, so the guard never fired there and the pass already ran — `8 -> 15` constraints, `7 -> 9` variables. On this branch it fires identically.

Removing the guard is *required* to preserve that behaviour once forwarding makes the list non-empty, so the change itself is right. But the rationale is inverted, the in-code docstring in `_model_carries_unsupported_state` says something different again, and this is the one claim in a PR whose correctness argument rests on "no solver-behaviour change" that asserts one.

(Where the guard *did* fire on `main` is a model with **unlowered** relations handed straight to `reformulate_binary_multilinear`. That case now runs the pass instead of abstaining — harmless given the new solve-boundary refusal, but it is the actual behaviour delta, not the one described.)

### 🟡 `ro/affine_policy.py:326` is the one in-tree site that drops a live `Variable`

`m._variables = [v for v in m._variables if v is not y]`, in place. A model carrying a relation over `y` would now raise `ComplementarityProvenanceError` at the *next* rebuilding pass rather than at the removal site. Exotic combination, but it is the failure vector the new hard error introduces, and it is not covered.

---

### ✅ Verified holding

* **Import health** — `discopt`, `discopt.mpec`, and all four patched `_relax` modules each import cleanly in isolation. No cycle from the new module-level `from discopt.mpec import ...`.
* **Operand-walker completeness** — all 11 concrete `Expression` subclasses in the tree (`BinaryOp`, `Constant`, `CustomCall`, `FunctionCall`, `IndexExpression`, `MatMulExpression`, `Parameter`, `SumExpression`, `SumOverExpression`, `UnaryOp`, `Variable`) are in `_operand_children`'s dispatch, so nothing silently under-reports, and a future node type raises. Right shape for a provenance walker.
* **Sentinel arithmetic** — `0.9999 * CONSTRAINT_INF == 9.999e19` exactly, and `Model.continuous`'s default bound is exactly that double, so the `>=` catches it; the LOW-8 case (`ub=2e19`) stays finite. No off-by-epsilon.
* **Rebuild-site coverage** — 5 `_variables = list(model._variables)` sites across 4 modules; all 5 call `carry_complementarities`.
* **Bound-neutrality on the class the 66-instance `.nl` panel does not cover.** MINLPLib carries no complementarity rows, so that panel never exercises MPEC at all. Four models — NCP; NCP + binary multilinear; NCP + integer products + `exp` under `sos1`; vector NCP — give **identical** `status`, `objective`, `bound`, `node_count` and `gap_certified` on `main` and this head, and stay identical with the `_pending` fix applied.
* `ruff check` clean; `ruff format --check` clean across all 90 touched files.
* `bilevel` double-`formulate()` is already guarded by the `RuntimeError` at `problem.py:405`, so the new `_complementarities.extend` cannot duplicate.
* One process note, not a finding: a bare `pytest` inside a worktree imports `discopt` from the **main** tree and fails collection on the new API — the §8 trap. `PYTHONPATH` set to the worktree gives the 74 passes.

---

The design calls all look right to me: identity comparison and identity hashing, per-model lowering state rather than a field on the shared relation, gating the lowerings on declared bounds rather than on `role`, flat indices derived at the boundary and never persisted, and loud refusals in place of approximations. Fix the `_pending` method-keying (plus a regression test for a second, different lowering) and I have nothing else blocking.

Issue #1116, opened 2026-08-23
kriging_peaks-full200 does not reproduce run-to-run: bound moves in the 13th digit and flips the node count, with every structural stage bit-identical


jkitchin:
## Summary

`kriging_peaks-full200` at `max_nodes=300` with **no** `time_limit` does not
reproduce run to run, in the same process, on the same binary, on an idle
machine. The dual bound moves in roughly the 13th significant digit, and that
occasionally flips a discrete decision so the node count changes.

This is pre-existing and unrelated to any flag. It was found while measuring
#1115, and it invalidated one row of that issue's panel, so it is filed on its
own rather than buried there.

## Evidence

Default configuration, `singular_tangent` **off**, three repetitions per run,
one process, `max_nodes=300`, no wall limit.

Repetition set A (reproducibility probe, 3 reps):

```
off   DISTINCT nodes=[301, 303]  separated rows=[0]  distinct bounds=2
```

Repetition set B (attribution probe, 3 reps, every stage instrumented):

```
rep0 nodes=303 bound=-746.5661745094023  lp={solves:582, nonoptimal:4, timelimited:0}
rep1 nodes=303 bound=-746.566174509407   lp={solves:582, nonoptimal:4, timelimited:0}
rep2 nodes=303 bound=-746.566174509407   lp={solves:582, nonoptimal:4, timelimited:0}

_separate_multilinear        rows across reps = [0]      STABLE
_separate_edge_concave       rows across reps = [0]      STABLE
_separate_univariate_square  rows across reps = [0]      STABLE
_separate_convex             rows across reps = [27544]  STABLE
_separate_singular_tangent   rows across reps = [0]      STABLE
LP.solves      across reps = [582]  STABLE
LP.nonoptimal  across reps = [4]    STABLE
LP.timelimited across reps = [0]    STABLE
```

Every structural quantity is bit-identical across reps. The bound is not.

By contrast `kriging_peaks-full050`, `full100`, `eq6_1` and `maxmin` are fully
reproducible across 3 reps in both arms at the same budget, so this is not a
universal property of the solver -- it is reachable at some size/conditioning.

## What is ruled out

* **Separator truncation on the shared `_deadline`.** Zero deadline hits in every
  family in every rep, and identical row counts.
* **The node LP wall limit** (`milp.solve(time_limit=_remaining(), ...)`,
  `mccormick_lp.py:1615`). Zero time-limited returns.
[...]

Issue #1038, opened 2026-08-16
Regression: callback-vetoed point returned as incumbent with status=optimal, gap_certified=True (#748 guard broken)


jkitchin:
## Summary

A point rejected by `incumbent_callback` is returned as the incumbent with
`status="optimal"` and `gap_certified=True`. This is the exact defect #748 was filed
for and that PR #755 (`ecf8466b`, 2026-07-18) fixed; the guard has since stopped
working. The regression test for it —
`test_callbacks.py::TestIssue748MilpMiqpCallbackDispatch::test_reproduction_sketch_never_returns_vetoed_point`
— is currently failing.

It went unnoticed because the test class is marked `@pytest.mark.slow`
`@pytest.mark.integration`, and the `slow` tier is excluded by the default `addopts`
while CI's slow job is on-demand (see #1034).

## Reproduction

```python
import numpy as np, discopt

def cb(ctx, model, sol):
    return float(sol["x"]) < 0.5      # veto anything with x >= 0.5

m = discopt.Model("milp748_sketch")
x = m.binary("x"); y = m.binary("y")
m.minimize(x + 2 * y)
m.subject_to(x + y >= 1)
res = m.solve(incumbent_callback=cb, time_limit=30)
print(res.status, res.objective, res.x, res.gap_certified)
```

Observed:

```
arm                  status            obj   x   y   cert  cb_calls
default              optimal         1.000   1   0   True         3
     callback saw x=1.0 y=0.0 -> keep=False
     callback saw x=1.0 y=0.0 -> keep=False
     callback saw x=0.9999999926061179 y=3.39638054827748e-09 -> keep=False
skip_convex_check    optimal         1.000   1   0   True         3
     (identical)
```

The callback **is** invoked — three times — and returns `False` every time. Its verdict
is then discarded.

Expected, per #748 and the test's docstring: the vetoed point is never returned. Because
rejecting the root leaves nothing to branch on, the honest outcome is an *uncertified*
result (feasibility undetermined) — never the vetoed point, and never a certified-false
`infeasible`, since a callback rejection is a non-rigorous fathom.

Note the feasible alternatives exist and are cheap to find: `x=0,y=1` (obj 2) and
`x=1,y=1` (obj 3). The solver is not stuck; it is ignoring the veto.

## Why this is high severity

`gap_certified=True` on a point the user explicitly rejected is a false certificate in
the sense of CLAUDE.md §1 — the solver's product is its certificate. A caller using
[...]

Issue #1036, opened 2026-08-16
surrogate: hartman_3 misses its 1e-2 convergence tolerance, and the budget rationale behind the whole panel is unsound


jkitchin:
## Symptom

`test_reaches_the_published_optimum_within_a_small_budget[hartman_3-0.01-100]`
fails: relative error **1.1202e-2** against a **1e-2** tolerance. Deterministic
on the default seed.

Found during v0.8.0 release prep. It is a `slow`-tier test, and the slow tier is
on-demand in CI ("Python correctness (slow, on demand)" reports `skipping` on
every PR), so it has most likely never run since it was written.

## Not a soundness issue, and not caused by the RBF residual fix

`solver="surrogate"` is explicitly non-certifying — it never claims a bound, and
`test_never_claims_a_certificate` covers that. This is a *convergence quality*
miss of 12%, not a false bound.

Confirmed independent of the RBF residual-guard fix that landed alongside it: the
failure reproduces byte-identically with that fix stashed.

Also confirmed **not** an incumbent-tracking bug. `_ClosedLoopState._offer`
(`surrogate.py:1244`) keeps the minimum over all feasible evaluations, so the
reported objective is the best point actually evaluated.

## The measurement, and what it says about the test

Budget sweep, `hartman_3`, default seed and config:

| max_evals | objective | rel_err | |
|---|---|---|---|
| 46  | -3.857687 | 1.3e-3 | reached |
| 60  | -3.650674 | 5.5e-2 | short |
| 80  | -3.793509 | 1.8e-2 | short |
| 100 | -3.819510 | 1.1e-2 | short |
| 120 | -3.848625 | 3.7e-3 | reached |
| 150 | -3.859252 | 9.1e-4 | reached |
| 200 | -3.862737 | 1.1e-5 | reached |

The result is **non-monotone in the budget**. That is expected rather than
alarming — `max_evals` feeds the initial design size, so each row is a *different
trajectory*, not a longer prefix of one trajectory.

But it makes the test's stated rationale unsound as written. From its docstring:

> each is set from a measurement rather than from hope — the evaluation at which
> the incumbent first reached the tolerance on the default seed, with headroom:
> branin 38, six_hump_camel 32, ackley_2 48, **hartman_3 46**

"First reached the tolerance at 46, so 100 has headroom" only follows if the
search at budget 100 contains the search at budget 46. It does not. The 46-eval
run does reach 1.3e-3; the 100-eval run is a different search that lands
[...]

PR #1150, opened 2026-09-04
fix(tests): #1039 close out the 18 slow-tier failures

### LINKED ISSUE #1039: slow/correctness tier: 102 failures triaged (18 unaccounted for after #1036/#1037/#1038)
## Summary

Running the `slow` and `correctness` tiers for the v0.8.0 release turned up **102
failures** (1384 passed, 16 skipped, 5 xfailed, 2 xpassed, 68m49s). Neither tier runs in
CI (excluded by the default `addopts`; the slow job is on-demand), so these have
accumulated invisibly — see #1034.

Triage of all 102:

| bucket | count | disposition |
|---|---|---|
| Presolve-box duals reported against the user's model | 79 | #1037 (single root cause) |
| Callback veto ignored, vetoed point certified optimal | 1 | #1038 (highest severity) |
| `hartman_3` surrogate budget miss | 1 | #1036 |
| Load artifacts (re-ran green on a quiet machine) | 3 | not defects — see below |
| Everything else | 18 | **this issue** |

### The 3 load artifacts

The tier ran while three unrelated `pounce` benchmark processes held ~100% CPU each
(load 5.4–8.8). Re-running the 23 non-minlptests failures at load 3.0 turned these
green:

* `test_912_work_budget.py::test_solve_deadline_still_stops_a_work_budgeted_search`
* `test_issue654_deadline_root_setup.py::test_issue654_dod_panel_honors_and_scales_with_time_limit[eg_all_s-7.657752093]`
* `test_sparsity.py::TestPerformance::test_sparse_faster_for_large_sparse`

All three are deadline/throughput assertions — exactly the class CLAUDE.md §9 warns
about. The other 20 reproduced under quiet load and are real.

---

## A. The HDA opt-out is inert — 3 tests, 1 root cause

All three assert that turning a flag **off** restores the legacy loose bound, and all
three get the same tight bound instead:

```
test_issue_517_numerical_dual_bound.py:59
    flag disabled must be the no-bound baseline, got -64473.44239643668
test_issue_671_lp_iterative_refinement.py:114
    flag OFF should be the loose candidate-A floor, got -64473.44239643685
test_relax_row_filter.py:119
    opt-out bound -64473.4 is unexpectedly tight -- the legacy no-filter path
    should give the loose candidate-A floor
```

The identical value across all three says one mechanism: the HDA / candidate-A
row-filter opt-out no longer disables anything.

This is policy-relevant, not cosmetic. The graduation rule in CLAUDE.md §5 requires that
a graduated flag "keep the `=0` opt-out and the legacy path intact". If the opt-out is
inert, the escape hatch promised when these flags graduated does not exist, and any
future A/B measurement using the flag silently compares a path against itself — the same
failure mode as #1035 (`relaxation_arithmetic="superposition"` inert since #632).

A tighter bound is not itself unsound, so this is a contract/measurement defect rather
than a correctness one. But it disarms the instrument used to validate correctness
changes, which is why it leads this list.

## B. `time_limit` overrun — 2 tests

```
test_875_root_setup_budget.py:479
    solve took 61.1s against a 30s time_limit (2.0x); root setup is still unbounded
    solve took 90.0s against a 60s time_limit (1.5x); root setup is still unbounded
```

Reproduced at load 3.0, so not a load artifact. `watercontamination0202` overruns its
budget by 1.5–2.0x, i.e. root setup remains outside the deadline — the thing #875 set
out to bound.

## C. Mechanisms that never fire — 3 tests

```
test_c42_cut_inherit_coldpath.py:213  pool-drop retry never fired
test_c42_cut_inherit_coldpath.py:289  no re-separation trigger fired on the stalling class
test_tx1_adaptive_nlp.py:107          expected the adaptive node-NLP back-off to fire on nvs09
```

Each asserts an adaptive mechanism engages on an instance chosen to provoke it. Same
shape as bucket A: the feature may be inert, or the probe may no longer select for it.
Worth checking whether the trigger conditions still match the instances.

## D. Node-count / gap targets missed — 3 tests

```
test_auto_cut_policy.py:92        assert 111 < (181 / 2)
test_psd_node_reduction.py:56     assert 111 < (181 / 2)
test_cut_recognizer.py:478        assert (3.0024852206298087 is None or ... < 2.0)
```

The first two are byte-identical numbers, so they are one instance measured twice: the
policy delivers 111 nodes where the test demands < 90.5 against a 181-node baseline —
a real reduction (39%), just short of the 50% the test encodes. These are performance
assertions of the kind PR #1033 removed elsewhere; they need either a re-derived
threshold backed by a measurement or removal, not a nudge.

## E. Accuracy / status misses — 4 tests

```
test_gauss_newton_hessian.py:369   assert 'feasible' == 'optimal'
test_gdpopt_loa.py:112             assert 'unknown' in ('optimal', 'feasible')
test_gp_corpus.py:199              assert 1.998683979470214 == 2.0 +- 1.0e-04
test_relax_row_filter.py:149       bchoco07: bound drifted
                                   (1.0000000000002582 -> 1.0000000000002498)
```

`test_gdpopt_loa` returning `unknown` is the most concerning of these. The `bchoco07`
drift is 8.4e-15 relative — a byte-identical assertion tripping on last-bit
nondeterminism, which needs deciding on its own terms (either the path is genuinely
deterministic and something broke it, or the assertion is too strict to be meaningful).

## F. Completeness misses — 2 tests

```
test_incumbent_injection_soundness.py:79
    nvs19 returned -1001.2 in 60s; known optimum is -1098.4
test_issue654_deadline_root_setup.py:282
    sonet23v4 lost its dual bound -- a §8 truncation regression
```

`test_incumbent_injection_soundness` is explicitly documented as "a *completeness*
check ... not a soundness one", and its sibling soundness test
(`test_suboptimal_warm_start_never_false_certifies`) passes — no false certificate here,
the search just does not reach the optimum in budget.

## G. `gap_certified` contract ambiguity — 1 test

```
test_tainted_tree_bound.py:85
    assert 338.2746666667722 <= ((0.0001 * 1100.4000000000003) + 1e-06)
```

Measured on nvs17:

```
time_limit=6s : status=time_limit  objective=-1100.4  bound=-1442.2
                gap=31.06%  gap_certified=True   nodes=19614
time_limit=20s: status=optimal     objective=-1100.4  bound=-1100.4
                gap=9.5e-14  gap_certified=True  nodes=60451
```

**Nothing here is unsound.** The bound −1442.2 is a valid lower bound (≤ the true
optimum −1100.4), the incumbent *is* the true optimum, the 31% gap is honestly
reported, and the status is `time_limit`, not `optimal`.

The failure is a genuine disagreement about what `gap_certified` means:

* Its docstring says "True if the reported optimality **gap** is mathematically
  certified" — under which `True` with a rigorous 31% gap is correct.
* The `__post_init__` guard comments say "the graduation panels count a
  `gap_certified=True` instance as certified" and call premature `True` "a false
  certification (the benchmark gate would miscount it as a solved/certified instance)"
  — under which the test is right and the flag should be `False` here.

Both readings are load-bearing somewhere. This needs an owner decision on the contract
before either the field or the test is touched; changing the test to match current
behavior without settling it would be weakening a validation to make a test pass.

---

## Suggested handling

Buckets A and B are the ones with contracts behind them (a promised opt-out; a promised
deadline). C and D are probe-health questions. G needs a decision, not a code change.

Prerequisite for all of it: #1034. Until these tiers run somewhere on a schedule, any fix
here can regress the same way — #1038 is a fixed correctness guard that broke and stayed
broken because nothing ran its test.

jkitchin:
## Summary

Works #1039 — the 18 slow/correctness-tier failures left after #1037, #1038 and
#1036 took their buckets. All 18 are dispositioned: **11 repaired, 4 pinned as
strict xfails against defects that are real and out of scope for a test-repair
PR, 3 folded into repairs of their neighbours.** The tier is green.

One production-code fix, one new instrument, and a lot of measurement. The
guiding rule throughout: a failing assertion is repaired only when the *test* was
wrong. Where the code was wrong, nothing was loosened.

### Retraction: the first version of this fix was a correctness regression (§11)

**Superseded by the review on this PR, and corrected in `ac42014a`.** This
section originally described widening `_is_linear` with a `SumOverExpression`
case. That was wrong and I am retracting it. `_is_linear` is a shared gate with
~10 consumers, and the hull substitution family (`_hull_linear_substitute` →
`_substitute_vars` → `_collect_variables`) and `_bound_expression` have no case
for the node. Admitting it made hull emit the disjunct body *globally* with its
selector coefficient collapsed to zero, and on

```python
m.either_or([[dm.sum(x[i] - 1 for i in range(3)) <= 0.0], [x[0] >= 8.0]])
m.minimize(-(x[0] + x[1] + x[2]))     # xi in [0, 10], true optimum -30.0
```

both `auto` and `hull` returned `status=optimal, objective=-3.0, bound=-3.0`. A
dual bound **above** the true minimum of a minimization is invalid. Three loud
refusals on `main` had become two silent false certificates — a §1 violation.
Verified on both arms with §8 marker counts before acting on the review.

### What actually ships

The defect is real: `_relax/gdp_reformulate.py` silently dropped its exactly-one
row, so a GDP whose exactly-one constraint is built as `Σ[N terms]` reformulated
without it and LOA returned `unknown`. But `_is_linear` was never the right
question. OA and LOA both gated their linear row block behind **two** predicates
— `_is_linear(c.body)` first, then `_extract_body_coeffs` — and the conservative
one wins. The extractor answers with the row itself, so it cannot promise more
capability than it delivers; it is the correct and only gate.

- `_is_linear` — `SumOverExpression` case **reverted**, with a comment recording
  why admitting it is unsound for the hull consumers.
- `_extract_body_coeffs` — keeps the term-by-term fold. This is the actual fix.
- `gdpopt_loa.py` **and** `oa.py` — redundant pre-gate dropped. `oa.py:1348` had
  the byte-identical shape, so both sites are fixed (§2, the class not the site).

Completing the hull family so those routes *solve* this model instead of
declining it is **#1154**: it is bound-changing (it moves the FBBT structural
mask at `solver.py:3068` and the aux-lift gate at `factorable_reform.py:684`) and
owes the §5 differential panel, so it does not ride along in a test-repair PR.
With the widening reverted, those consumers are byte-identical to `main`.

### Regression test

`python/tests/test_1039_gdp_sumover_rows.py` pins **both** defects. Fail-before /
pass-after, run in the installed tree with §8 marker counts confirming which
sources each arm loaded:

| sources at | marker | result |
|---|---|---|
| `bf079fc1` (the widening) | 4 | **2 failed** — `[auto]`, `[hull]`: bound `-3.0` above the optimum |
| `main` | 2 | **2 failed** — extractor returns `None`; LOA returns `unknown` |
| `HEAD` | 3 | **5 passed** |

The blocker arm asserts the refusal rather than skipping on it (a skip on all
three routes would make it a no-op, §6), and is phrased as "no route may report a
bound above the true optimum" so #1154 strengthens it rather than breaks it.
354 GDP tests pass; 911 pass on the gdp/hull/bigm/oa/benders/gbd/mpec selection.

## The instrument (bucket A)

`MccormickLPRelaxer` now keeps `_row_filter_stats` (invocations / rows_dropped),
surfaced through `solver_stats` as `row_filter/*`. Sweeping all 66 vendored
instances with it: exactly three open the failure-triggered branch — `bchoco07`
(2 inv / 158 rows), `bchoco08` (2/144), `hda` (2/356). Every other instance is 0.

### Retraction: this issue's own bucket-A comment is contradicted (CLAUDE.md §11)

The retraction comment on #1039 measured `filter_invocations=0` in **both** arms
on `hda` and concluded the mechanism was dormant, revising bucket A to "not a
contract defect". With the counter surfaced, the count on `hda` is **2, not 0**,
and the arms are far apart:

```
DISCOPT_RELAX_ROW_FILTER=0 -> bound -13992288065.86   inv=0
DISCOPT_RELAX_ROW_FILTER=1 -> bound      -64509.85    inv=2, 356 rows dropped
```

(2 reps interleaved.) The opt-out is live and load-bearing. What *was* right in
that comment is the diagnosis of the tests: they encoded a stale premise. But the
premise was subtler than "hda no longer false-fails" — each test varied the flag
it names while leaving `DISCOPT_RELAX_ROW_FILTER` at its graduated default-ON
(#671, 2026-07-18), so a **later-graduated mechanism answered for both arms** and
no test isolated its own flag.

With the other rescue flags pinned (3 reps interleaved, sd 0 in every arm):

| flag | OFF | ON | verdict |
|---|---|---|---|
| #671 refinement | -141647.03848991546 | -64509.850876897275 | live |
| #517 dual bound | -141697.43348991615 | -141697.43348991545 | agrees to 12 digits |

So `#671`'s test now asserts the **differential** (ON tighter than OFF, both
sound) instead of a hardcoded floor a third mechanism can move. `#517`'s flag is
still consulted but no longer load-bearing on `hda`; it is disposed per the
C-42/C-43/C-44 precedent — assert inertness *with the measurement*, rather than
demand it fire, which would be asserting a premise known false. Its ON path keeps
coverage in `test_hda_gets_first_finite_dual_bound`. **Re-pointing it needs a
`#517` invocation counter; that is recorded in the docstring as the one piece
left open.**

`bchoco07` also moves out of the "never fires" list — it fires twice, so
asserting zero on it was asserting a false premise — and the real-instance
positive control the retraction comment asked for is added.

Whether the branch opens is itself budget-dependent (a #1116 role-2 effect), so
the firing test is pinned to the measured plateau with the curve in the
docstring: `bchoco07` fires at tl=15/30 but not tl=60. `deterministic=True` does
not remove the dependence, it moves it (0 nodes at tl<=60, fires at tl=120).

## Retraction: my own bucket-G review recommendation was wrong (§11)

My review of #1039 said the test was right and the `gap_certified` docstring
needed correcting. That was wrong, and #1059 had already settled it on
2026-08-19 with a measurement: `gap_certified` means "the reported gap is
*mathematically valid*", not "the gap is closed", so `True` at a rigorous 37.7%
gap on nvs17 is **correct**. The **test** and the **graduation panel** were both
wrong. Both are fixed; `_route_result_is_certified` (solver.py:5760) is the right
predicate for "did this actually finish", and the panel now uses an equivalent.

## Pinned, not fixed — 4 strict xfails

Each preserves every assertion and threshold. `strict=True` means none can pass
by having its goalposts moved, and each fails the suite the moment it is fixed.

Per §3 each now ships with a filed issue tracking the real fix, linked from the
xfail's own `reason`: **#1151** (the super-optimal division objective),
**#1152** (the root-setup overrun and the sonet23v4 contradiction — one contract
hole, two tests), **#1153** (nvs19's budget non-monotonicity).

**1. A super-optimal reported objective (new, and the most severe thing here).**
Bucket E listed `assert 1.998683979470214 == 2.0 +- 1e-4` as an accuracy miss.
Widening the tolerance would have masked a soundness defect. `minimize x/y + y/x`
over a positive box has global minimum **exactly 2** by AM-GM, so 1.9987 is a
value no feasible point attains — returned with `status=optimal`. The returned
point is fine (true objective 2.000002247829649); the reported number is wrong by
-1.318268e-03 against its own incumbent.

The error is the absolute feasibility tolerance amplified by 1/denominator. Only
the quotient misbehaves — affine and **bilinear** objectives (the latter also
auxiliary-backed) match an outside oracle to 4e-16 — so it is specific to the
division reformulation, not to auxiliaries generally. Scaling the box floor
confirms the 1/y law, `|delta| x denominator` flat at ~1.9e-6 (~2x the 1e-6
tolerance, two quotient terms):

```
floor 1e-3  denom 0.00140525  delta -1.318268e-03  product 1.852e-06
floor 1e-2  denom 0.0106986   delta -1.850776e-04  product 1.980e-06
floor 1e-1  denom 8.13524     delta -4.360956e-13  product 3.548e-12
floor 1e+0  denom 479.758     delta -4.440892e-16  product 2.131e-13
```

A trace over the whole solve found **zero Python frames** returning the bad
value, so it is produced in the Rust B&B incumbent path. The general fix is to
recompute the incumbent objective from the original expression at the incumbent
point — one evaluation, and it can only correct the number, since the point is
feasible and its true objective is a valid upper bound. That is a §5
bound-changing solver change, not a test repair.

**2. Root setup still overruns its deadline (bucket B).** Re-measured at load
3.33, so not the §9 load artifact three other failures in that sweep turned out
to be: 30s budget -> 59.7s (2.0x), 60s -> 89.4s (1.5x). #875 took this from 579.3s
(19.3x) with `nodes=0`, so most of the class is fixed and a residual remains. The
1.25x threshold is deliberately **not** relaxed. Not one instance — the same
overrun appeared incidentally on `casctanks` (~321s/120s, 2.7x), `nvs19`
(80.7s/60s, 1.35x) and `sonet23v4` (4.6s/2s, 2.32x).

**3. `sonet23v4` vs bucket B — a contract contradiction, now machine-checked.**
At tl=2.0 the bound is `None`; at tl=8.0 it is -473508.0000005656 and at tl=30.0
-50587.97955518025, both sound against the -22747.5 oracle. The bound is not lost
to a truncation *bug* — the #654 gating correctly declines to **start** a 17s
operation. So this test ("spend the 17s, keep the bound") and the bucket B test
("never exceed 1.25x") assert opposing contracts on one mechanism, and
`sonet23v4` already overruns 2.32x *while declining* the op. Both hold only once
the root-relaxation LP build is made interruptible so a short budget yields a
weaker-but-present bound. That is the §3 hard fix.

**4. `nvs19`: more budget buys a worse answer (bucket F).** The test's "~11s
locally" premise is false — the optimum -1098.4 is never reached at any budget to
480s, measured on a quiet machine (load 2.63):

```
tl    wall     status      nodes    objective   bound
30    30.2s    time_limit   38403   -1098.2     -2076.33
60    80.7s    time_limit    7619   -1001.2     -7401.66
120  110.7s    feasible    100009   -1097.6     -1104.24
240  188.4s    feasible    100001   -1097.6     -1104.21
480  344.1s    feasible    100013   -1097.6     -1103.83
```

tl=30 finds -1098.2; tl=60 finds -1001.2 while exploring **5x fewer nodes** on
twice the wall clock — the #1116 role-2 signature. A performance pathology, not a
budget shortfall; raising the budget cannot fix it. At tl>=120 the search stops
on a ~100,000 **node** cap with wall under the limit. Everything is sound: every
bound is valid, every incumbent feasible and >= the optimum. No false
certificate — which is what the soundness sibling guards, and it passes.

## Repaired, because the test was wrong

* **Bucket C** — three "mechanism never fired" assertions re-pointed. The
  pool-drop retry and re-separation triggers went inert *because the source fix
  removed the condition* (C-42/C-43/C-44 precedent: the right assertion is
  `inherited_nodes >= 1` + `dropped_nodes == 0`, not `>= 1`). TX1's back-off does
  fire, just not on `nvs09` — measured `nvs09` 0/0, `casctanks` 0/0, `bchoco07`
  0/0, `tspn05` 2/2 — so the test is split and re-pointed at `tspn05`, carrying
  its own soundness assertions.
* **Bucket D** — node-count thresholds re-derived from an 18-draw sweep instead
  of nudged. Only 3 draws branch at all; the rest solve in 1-5 nodes. PSD's
  no-harm and optimum-preservation properties hold 18/18; `auto == selected
  family` 8/8. Both tests are now parametrized over the draws that actually
  branch, with a §6 probe-fired guard (`base.node_count > 20`).
* **Bucket E gauss_newton** — asserted `status=='optimal'`; **both** arms return
  `feasible` with `node_count==0` and `bound=None`, and 120 -> 600s changes
  nothing, so the root relaxation structurally yields no dual bound for a
  sum-of-squares-of-exponentials objective. The assertion is now
  soundness-shaped: `feasible` is accepted, but a claim of `optimal` must carry
  `bound <= objective`. Coverage is *strengthened* — it now also checks the
  reported objective is attained at the returned point and that both backends
  recover the generating parameters (2.0, 1.3) against a numpy oracle. Its actual
  subject (the two Hessian backends agree) holds to ~1e-12.
* **Bucket E bchoco07 drift** and **bucket A byte-identity** — the old
  comparisons ran both arms under a bare `time_limit`, which is not a comparison:
  a wall-truncated solve is cut at a machine-speed-dependent point, so the two
  arms are two different *amounts of search*. `beuster` produced **both** outcomes
  in **both** arms, proving the variation is within an arm. Both now run under
  `deterministic=True` on instances that terminate on work, with the precondition
  **asserted** (§6) rather than assumed.

## A note on the issue's prerequisite

#1039 closes with "Prerequisite for all of it: #1034." #1034 was closed by
*declining* to add a schedule — `ci.yml:15` records "NO `schedule:` trigger" — so
that line is stale. It does not block this PR, but the underlying worry is real
and unaddressed: nothing runs these tiers automatically, which is how #1038's
guard broke and stayed broken.

## Verification

* `pytest -m smoke` — **1165 passed**, 1 skipped, 2 xpassed (both pre-existing).
* `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — **19 passed**.
* `cargo test -p discopt-core` — **not run; no Rust was touched** (`git diff
  main...HEAD -- crates/` is empty).
* Per-file, all at `-m "slow or not slow"`: row filter **18 passed**; #517 + #671
  **11 passed**; gp_corpus **21 passed / 1 xfailed**; gauss_newton **32 passed**;
  incumbent-injection + #875 **18 passed / 3 xfailed**; #654 **9 passed / 1
  xfailed**; GDP **322 passed** / 1 skipped / 2 xpassed; C-42 **7 passed**; cut
  policy + PSD **14 passed**; tainted-tree **2 passed**.

Every probe prints an executed-assertion count and exits non-zero at zero; none
swallows an exception; the ones that made timing claims ran interleaved with a
load gate and report a spread. Probes are in `scratchpad/issue1039/`.

## Disposition

Closes #1039 — all 18 failures triaged and the tier is green.

It should **not** be closed silently, though: the 4 strict xfails stand for 4
real defects, and per §3 each ships with a filed issue tracking the real fix,
linked from its xfail `reason`:

| issue | defect | xfail |
|---|---|---|
| #1151 | reported objective BELOW the global minimum on quotients (false certificate) | `test_gp_corpus` |
| #1152 | `time_limit` role-1: root-setup overrun vs deadline-gated bound loss, contradictory as tested | `test_875_root_setup_budget`, `test_issue654_deadline_root_setup` |
| #1153 | incumbent quality non-monotone in `time_limit` (nvs19: 2x budget, 5x fewer nodes, worse answer) | `test_incumbent_injection_soundness` |

**#1154** additionally tracks completing the hull substitution family for
`SumOverExpression` — the bound-changing half of the bucket-E fix that this PR
deliberately does not ship.

Found while verifying, and **not** part of this PR:
`test_oa.py::TestOAEdgeCases::test_infeasible_model` asserts `result.x == {}` and
gets `None`. Confirmed pre-existing by running it against `main`'s sources in
this tree (§8 marker=2). It is outside the 18 this issue triaged, so it is left
alone rather than widening the PR.


jkitchin:
## Correctness review — one blocking regression

Reviewed for correctness. The test-repair work is largely well-argued (the graduation-panel change is a *tightening*, the `tainted_tree` re-point is right, the xfails are all `strict=True` with thresholds preserved). The problem is the one production fix.

---

### 🔴 Blocker: the `_is_linear` widening produces a false `optimal` certificate on the GDP `auto` route

The fix widens `_is_linear` (`gdp_reformulate.py:1448`) to accept `SumOverExpression` and teaches `_extract_body_coeffs` to fold it. But `_is_linear` is a **global gate with ~10 consumers**, and the hull reformulation is one of them (`gdp_reformulate.py:1139`). The hull substitution family did not learn the same node:

| function | handles `SumOverExpression`? |
|---|---|
| `_is_linear` (1448) | ✅ added here |
| `_extract_body_coeffs` (537) | ✅ added here |
| `_hull_linear_substitute` (1322) | ❌ falls through to `_substitute_vars` |
| `_substitute_vars` (1436) | ❌ `# Constant or unknown — pass through` |
| `_collect_variables` (1371) | ❌ returns `{}` → vars never disaggregated |
| `_body_at_zero` (1235) | ❌ raises — **this is what used to save it** |

`grep -c SumOverExpression python/discopt/_relax/gdp_reformulate.py` on this branch = 4 (import, local import, extractor, `_is_linear`).

So a disjunct body written as `dm.sum(...)` now takes the *linear* hull path, and `_hull_linear_substitute` returns it **unchanged** — original variables, no `v_k` disaggregation, no `y_k` scaling on the constant. The row is then enforced unconditionally.

**Measured**, both arms with the version marker asserted present/absent (§8):

```python
m = dm.Model("t")
x = [m.continuous(f"x{i}", lb=0.0, ub=10.0) for i in range(3)]
m.either_or([[dm.sum(x[i] - 1 for i in range(3)) <= 0.0], [x[0] >= 8.0]], name="modes")
m.minimize(-(x[0] + x[1] + x[2]))
# true optimum = -30.0  (mode B: x = 10, 10, 10)
```

```
                     main                              this branch
gdp_method="auto"    RAISED (loud refusal)             status=optimal  obj=-3.0   ← FALSE CERTIFICATE
gdp_method="hull"    HullPerspectiveOriginError        status=optimal  obj=-3.0   ← FALSE CERTIFICATE
gdp_method="big-m"   RAISED                            RAISED (unchanged, safe)
```

The emitted row is `_hull_modes_d0_c0: Σ[3 terms] - (0 * y0) <= 0`, i.e. `x0+x1+x2 <= 3` imposed **globally**, so mode B is cut off and `-3.0` is certified optimal against a true `-30.0`.

`auto` reaches it because the widened predicate also moves the advisor's method choice (`gdp_advisor.py:145`) — the user never asks for hull. This is a §1 violation, and it converts a loud refusal into a silent wrong answer, which §3 specifically prohibits.

**Same exposure in MPEC:** `mpec.py:227` `_gdp_operand` previously lifted a `SumOverExpression` side to a bounded auxiliary; it now returns the raw node straight into `either_or`.

**Verified fix** — 4 cases, ~10 lines: add `SumOverExpression` handling to `_collect_variables`, `_substitute_vars`, `_hull_linear_substitute` and `_body_at_zero`. With that patch applied on this branch, both `auto` and `hull` return `-30.0`. `_bound_expression` still lacks the case so big-M keeps refusing — safe, but worth closing as the fifth.

---

### 🟠 This is a bound-changing change shipped inside a test-repair PR

Beyond hull, the widened predicate silently changes behavior at:

* `solver.py:3060` — the **structural linear-FBBT mask**, whose docstring calls the conservatism deliberate and cites the #27a false-prune;
* `factorable_reform.py:684` — the aux-lift gate for `sqrt`/univariate outers;
* `decomposition/benders/solver.py:114` — linear-Benders routing;
* `_relax/implied_integer.py:97` — this one is fine, `_extract_linear_coefficients_sparse` already handles the node (`problem_classifier.py:480`).

Each is plausibly sound (a sum of affine terms is affine), but all are bound-changing under §5 and none was measured. Either narrow the fix to the LOA row-forwarding predicate, or keep the widening and run the differential panel.

### 🟠 The 4 strict xfails ship with no issues filed

§3: a workaround ships only with an issue tracking the real fix and a comment linking it. The body says "say the word and I will file them" — but one of the four is a **super-optimal reported objective returned with `status=optimal`** (`1.998683979470214` against an AM-GM minimum of exactly 2). That is a certificate defect and the most severe item in the PR; it needs its issue before this merges, not after.

### 🟡 No regression test for the production fix

Workflow rule: *"New behavior requires a regression test that fails before the change and passes after."* 13 test files touched, **zero** covering the GDP/LOA fix. "354 GDP tests pass" is the point — none of them covers this path, which is why the hull regression got through.

### 🟡 Minor

* `ruff format --check` fails on `discopt_benchmarks/scripts/gp_minlp_graduation_panel.py` (extra blank line before `_closed_its_tree`) — CI will be red. `ruff check` is clean.
* `solver.py:15610` only emits `row_filter/*` when the value is `> 0`, so `stats.get("row_filter/invocations", 0) == 0` (`test_relax_row_filter.py:261`) passes identically when the instrument was never wired (e.g. `_mc_lp_relaxer is None`). The positive controls at :308/:359 cover it as a pair, but that negative assertion alone has the §6 "probe never fired" shape — consider asserting the key is *present* and zero.

---

### ✅ What holds up

* The graduation-panel change **strengthens** the bar (`gap_certified` → `gap_certified AND gap <= tol`) and matches the #1059 reading. Not a weakening.
* `test_tainted_tree_bound` is re-pointed at the right invariant and replaces a vacuous `if r.bound is not None` guard with a real assertion — coverage is strictly better.
* All four xfails are `strict=True` with every threshold preserved; the `nvs19`, `sonet23v4` and root-setup measurements are convincing and the §11 retractions are handled properly.
* The `row_filter` counter is correctly placed inside the flag-gated branch and is a real answer to the "compare two whole solves under a wall clock" problem.

jkitchin:
## Review addressed — the blocker was real, and it was mine

Thank you for the catch. I verified the 🔴 finding on both arms before acting on
it, and it reproduces exactly as described. §8 marker counts confirm which code
each arm ran (`grep -c SumOverExpression gdp_reformulate.py`):

|  | main (marker=2) | this branch @ bf079fc1 (marker=4) |
|---|---|---|
| `auto` | RAISED `ValueError` | `status=optimal obj=-3.0 bound=-3.0` |
| `hull` | RAISED `HullPerspectiveOriginError` | `status=optimal obj=-3.0 bound=-3.0` |
| `big-m` | RAISED `ValueError` | RAISED `ValueError` |

`bound = -3.0` is **above** the true minimum `-30.0` of a minimization — an
invalid dual bound. My change converted three loud refusals into two silent
false certificates. That is a §1 violation, and §3 besides: a silent wrong
answer replacing a refusal.

**I am retracting the "one real code fix" framing in the PR body (§11).** That
commit was not a fix; it introduced a correctness regression that the 354
passing GDP tests did not cover.

### The fix I took, and why it is not the one you proposed

Your verified patch (teach `_collect_variables`, `_substitute_vars`,
`_hull_linear_substitute`, `_body_at_zero`) is correct and I confirmed it gives
`-30.0`. I did **not** ship it, because it only becomes reachable by keeping the
`_is_linear` widening — and that is exactly the bound-changing change your 🟠
finding says must not ride along in a test-repair PR. It is filed as **#1154**
with your patch recorded, the five functions listed, and the §5 differential
panel as its gate.

What I shipped instead is that `_is_linear` was never the right question at the
LOA call site. Both OA and LOA gated their linear row block behind **two**
predicates — `_is_linear(c.body)` first, then `_extract_body_coeffs` — and the
conservative one wins. The extractor answers with the row itself, so it cannot
promise more capability than it delivers; it is the correct and only gate.

- `_is_linear`: `SumOverExpression` case **reverted**, with a comment recording
  why admitting it is unsound for the hull consumers.
- `_extract_body_coeffs`: keeps the fold — this is the actual bucket-E fix.
- `gdpopt_loa.py` **and** `oa.py`: redundant pre-gate dropped. `oa.py:1348` had
  the byte-identical shape, so both are fixed (§2).

This makes your 🟠 §5 finding moot rather than answered: the FBBT structural mask
(`solver.py:3068`), the aux-lift gate (`factorable_reform.py:684`) and the
Benders route selection are now byte-identical to `main`. No panel is owed.

### 🟡 Missing regression test — added, and it fails before on *both* defects

`python/tests/test_1039_gdp_sumover_rows.py`. Fail-before/pass-after run in the
installed tree (the worktree attempt was void — §8: it imported the main tree's
`discopt`, which is why I re-ran it this way):

| sources at | marker | result |
|---|---|---|
| `bf079fc1` (the widening) | 4 | **2 failed** — `[auto]`, `[hull]`: bound `-3.0` above the optimum |
| `main` | 2 | **2 failed** — extractor returns `None`; LOA returns `unknown` |
| `HEAD` | 3 | **5 passed** |

The blocker arm **asserts** the refusal rather than skipping on it — a skip on
all three routes would have made it a no-op (§6) — and is written as "no route
may report a bound above the true optimum", so #1154 will *strengthen* it rather
than break it.

### 🟡 `row_filter/*` emitted only when > 0 — you were right, and it was hiding one

Fixed to emit zeros, and the assertion now requires the key to be **present**.
That immediately caught a genuine vacuous pass: **`alan` is routed to the convex
`mip-nlp/oa` algorithm and emits no `solver_stats` at all**, so it had been
"passing" against an empty dict. Dormancy is now a positive observation with two
arms — counter present and zero, or a *named non-default* route with no
McCormick LP relaxer. (`algorithm_route is None` means the default path and is
the first arm, not an excuse for a missing counter.) 18 passed.

### 🟡 `ruff format` on the panel script — confirmed and fixed

Also worth noting *why* it got through: the lint job runs `ruff format --check
python/`, so `discopt_benchmarks/` is unchecked.

### 🟠 Issues filed for all four strict xfails, each linked from its reason

- **#1151** — reported objective BELOW the global minimum on quotient
  expressions. False certificate; error is the absolute tolerance amplified by
  `1/denominator`, so it is unbounded as the denominator shrinks. Filed before
  merge as you asked.
- **#1152** — `time_limit` role-1 contract: root-setup overrun (#875) and
  deadline-gated bound loss (#654) are contradictory *as tested*; covers both
  budget xfails.
- **#1153** — incumbent quality non-monotone in `time_limit` (nvs19: 2x budget,
  5x fewer nodes, worse answer).
- **#1154** — complete the hull substitution family, then admit the node.

### Also found, and NOT mine

`test_oa.py::TestOAEdgeCases::test_infeasible_model` asserts `result.x == {}` and
gets `None`. Confirmed pre-existing by running it against `main`'s sources in
this tree (marker=2). It is outside the 18 this issue triaged, so I have left it
alone rather than widen the PR — say the word if you want it pulled in.

Issue #875, opened 2026-07-26
Root setup runs unbounded before B&B: watercontamination0202 still 19x over time_limit (successor to #863)


jkitchin:
Successor to #863, which is now down to a single dominant cause plus two smaller ones. PR #868 (merged) took `watercontamination0202` from an **unbounded hang** to a **bounded overrun with fully attributed time**; this issue covers what is left.

## Where it stands after #868

| | before #868 | after #868 |
|---|---|---|
| `extract_qp_data` | never returned (killed at 400 s, 1500 s) | **14.8 s** |
| `solve(time_limit=30)` | never returned | **579.3 s (19.3×)** |
| `solve(time_limit=60)` | never returned | **620.8 s (10.4×)** |

Both now return `status=time_limit`, `objective=None`, `best_bound=None`, `nodes=0` — **sound** (nothing is reported, so no bound can cross an incumbent). The failure mode changed from "hangs indefinitely" to "honours nothing but terminates", which is why this is a follow-up rather than a reopen.

## Remaining time, fully attributed (T=30, total 579 s)

| phase | cost | note |
|---|---|---|
| **`_fix_single_var_equalities`** (`_jax/uniform_relax.py:2495`, via `_root_relaxation_lower_bound` → `build_uniform_relaxation`) | **~460 s** | **dominant** — 23 of 29 stack samples |
| `tighten_nonlinear_bounds` ×3 | 80.9 s | no deadline awareness; #868 already halved this by sharing one computation between two consumers, the pass itself still needs the `_until` treatment |
| `_classify_model_convexity` ×2 | 14.7 s | |
| presolve + load + classification | ~12 s | now correctly capped by #868 |

Instance: `~/Dropbox/projects/discopt-minlp-benchmark/minlplib/nl/watercontamination0202.nl` — 106,711 vars / 107,209 rows / **7 binaries**, parses in ~0.9 s.

## The unifying insight

Every remaining item is **pre-B&B root setup with no deadline polling**. That is the same class as the LP-engine loops fixed in #858, and the root-presolve passes fixed in #868 — one layer earlier each time. #868 found that **5 of 10** root-presolve passes ignored `time_limit_ms` entirely, and that `PyModelRepr.fbbt` had **no budget parameter at all** (it ran >10 minutes). The pattern is worth auditing for directly rather than discovering one phase at a time: *which root-setup steps can run unbounded before a single node is created?*

[...]

PR #1155, opened 2026-09-04
fix(solver): #1152 bound the root-setup builds and the fallback reserve

### LINKED ISSUE #1152: time_limit role-1 contract: root-setup overrun and deadline-gated bound loss are contradictory as tested
## Summary

Two slow-tier tests encode **contradictory** contracts about what the solver owes
a `time_limit`, and both currently fail. Neither can be repaired without deciding
the policy, so both ship as strict xfails in PR #1150 with nothing weakened.

## Side A — the deadline is overrun

`watercontamination0202`, measured at load 3.33 (§9 load gate):

| time_limit | wall | ratio |
|---|---|---|
| 30 s | 59.7 s | **2.0x** |
| 60 s | 89.4 s | **1.5x** |

#875 improved this class from 19.3x to 2.0x but did not close it: root setup is
still not fully inside the deadline. The test's threshold is 1.25x and is
**deliberately unchanged** — relaxing it would retire the contract rather than
meet it.

This is a class, not one instance:

| instance | overrun |
|---|---|
| casctanks | 2.7x |
| sonet23v4 | 2.32x |
| watercontamination0202 | 2.0x |
| nvs19 | 1.35x |

## Side B — honouring the deadline destroys the bound

`sonet23v4`'s dual bound comes from a single ~17 s uninterruptible operation. The
#654 deadline gating correctly declines to *start* it when the budget cannot
cover it, and the bound is then `None`:

| time_limit | bound |
|---|---|
| 2.0 s | **None** |
| 8.0 s | -473508.0000005656 |
| 30.0 s | -50587.97955518025 |

`test_issue654_deadline_root_setup.py::test_sonet23v4_bound_survives_the_deadline_gating`
asserts the bound survives; `test_875_root_setup_budget.py::test_watercontamination0202_honours_its_time_limit`
asserts the deadline is honoured. Satisfying either one as written breaks the
other, because the only way to stop overrunning is to decline long
bound-producing operations, which is precisely what loses the bound.

## What has to be decided

Which role-1 guarantee `time_limit` actually makes:

1. **Hard wall.** The solver returns by the deadline, and a truncated run may
   return `bound = None`. Side B's test is then wrong as written and must assert
   "bound is None **or** valid", never a bound.
2. **Soft target.** An uninterruptible bound-producing operation may overrun.
   Side A's 1.25x threshold is then wrong and needs a re-derived, documented
   bound on the overrun (not a nudge to whatever today's number is).
3. **Make the operation interruptible.** The only option that satisfies both
   tests as written, and the most work: the ~17 s root operation would need to
   yield a valid intermediate bound.

Option 3 is the honest fix. 1 and 2 both retire a contract someone is relying on,
so the choice is an owner decision, not a test edit.

## Tests

Both are strict xfails referencing this issue; every threshold is preserved.

Found while closing out #1039 (PR #1150).

jkitchin:
## Summary

Closes #1152.

The issue asked the owner to pick between three readings of the role-1 `time_limit`
contract, because two slow-tier tests read as contradictory. **They are not
contradictory, and no threshold is retired.** They were two symptoms of one defect,
and the issue's own option 3 — "make the operation interruptible" — is what ships.

A root-setup phase polls a deadline between its LP solves but **not inside its
relaxation build**, and clamps its own budget to *all* of the live remainder — so it
both runs past `T` and spends the `_ROOT_FALLBACK_RESERVE_S` slice the last-ditch
bound producer needs. Measured on `casctanks` at `time_limit=5` (in-repo corpus,
`scratchpad/i1152/phase_attrib3.py` + `fb_probe.py`, quiet 4-core Linux box):

| t (s) | phase | note |
|---|---|---|
| ~4.4 | root OBBT enters its round `build_milp_relaxation` | **0.61 s of budget left** |
| 4.4–6.3 | that build | **1.85 s, uninterruptible** |
| 6.3 | the #654 short-circuit | `fallback grant 0.000s of the 1.500s reserve` |
| 6.4 | return | wall **1.29x** *and* `bound=None` |

Side A (the overrun) and side B (the lost bound) are the same missing budget, one
size down from `sonet23v4`/`watercontamination0202`. The decision, recorded in
`docs/dev/1152-time-limit-root-setup-contract-2026-09-04.md`:

> `time_limit` is a **hard wall with an anytime bound**. The solve returns by the
> deadline; the bound it reports is whatever it could *prove* inside the budget —
> weaker under a short budget, never unsound, and never absent merely because the
> operation that would have produced it was long.

### What ships

Reuses the anytime-build mechanism #694 built and #832 graduated, one layer up:

* `_root_setup_build_deadline()` — `time_limit` minus the reserve, threaded as the
  `build_deadline` of the root OBBT round build (new
  `obbt_tighten_root(build_deadline=...)`, applied to the round build *and* the DBBT
  rebuild), the root LP probe, and both root cut pools;
* `_setup_remaining_budget()` — the same subtraction for those phases' own budget
  clamps, so the reserve is actually reserved;
* the native-kernel spec build is **declined** past that deadline rather than
  truncated: its probe-box and real-box builds' row sets must correspond, so a
  prefix of each need not describe the same relaxation.

Same `casctanks` solve after: **5.35 s (1.07x) with bound 1.2584**.

## Correctness

Every truncation this change reaches only ever **weakens**: dropping constraint rows
*enlarges* the relaxation polytope, so OBBT tightens less, the probe's banked bound
is looser (and #928's cut-short objective floor keeps it finite), and the pool
separates fewer cuts. The one place a truncated build is not merely weaker is the
#208 aux cascade — its carried bounds are keyed by column index and a truncated
build has a different lifted layout, so applying one could tighten the *wrong*
column; both the apply and the capture are skipped on `milp._build_truncated`.
`baron-gap-plan.md` §8.1/§8.2 are un-circumvented: this bounds Python relaxation
*builds*, never an LP solve and never the Rust LP.

- [x] `pytest -m smoke` → **1169 passed, 22 skipped, 2 xpassed** (both xpasses pre-existing at the time; see the xfail note below)
- [x] `cargo test -p discopt-core` → **N/A, no Rust touched** (`git diff main...HEAD -- crates/` is empty)
- [x] `ruff check` / `ruff format --check` → clean (CI's Lint + typecheck green on the head)
- [x] The 49 budget / deadline / OBBT / root test files at `-m "slow or not slow"` → **444 passed, 21 skipped, 3 failed**, none of them this change:
  - `test_930_root_probe_bound.py` (2) — **pre-existing**, confirmed by stashing the diff and re-running against `main`'s sources in this tree (marker count 0), and independently re-confirmed in review (branch 2 failed/4 passed, main 2 failed/2 passed — same two IDs). On a 4-core Linux box `hda`'s probe proves −119284.66, not the −64473.44 the test's threshold was derived from on a 14-core Mac.
  - `test_perf_723_obbt_probe_cost.py::test_per_probe_cost_has_a_pivot_heavy_tail` — a wall-distribution shape assertion (`p90 ≥ 1.8·p50`); **passes in isolation** on both machines, and its code path takes no build deadline.
- [x] After merging `main` in: `test_1152` + `test_875` + `test_654` → **23 passed, 9 skipped** (the 9 skips are the big-corpus tests, which are not skipped on the owner's machine — see below).

## Regression test

`python/tests/test_1152_time_limit_root_setup_contract.py` (5 tests): the OBBT
anytime contract as a deterministic unit test (a truncated pass must return a box
*containing* the full pass's, with `n_tightened > 0` on the full pass as the
positive control, per §6), the inertness of a non-binding deadline, the env
opt-out, and the end-to-end contract on `casctanks`.

Fail-before/pass-after was verified on this same tree via the env opt-out, which
reproduces the legacy path exactly:

```
DISCOPT_ROOT_SETUP_BUILD_DEADLINE=0 pytest -k casctanks -m slow
  E   AssertionError: the solve reported no dual bound — the root-setup phases
      spent the fallback reserve again (#1152 side B; legacy arm bound: None)
(default)                            pytest -k casctanks -m slow  →  passed
```

The end-to-end test takes its own premise rather than assuming it: the legacy arm
runs first, and unless *it* shows the defect on this machine at this budget the test
skips with the measurement in the message. Written at 10 s it passed on the legacy
path too and pinned nothing — that was caught and the budget moved to 5 s.

- [x] Flag: `DISCOPT_ROOT_SETUP_BUILD_DEADLINE` — OFF path byte-identical (no build
      deadline computed, no entry gate consulted, `_setup_remaining_budget` returns
      the plain remainder). **Graduated default-ON**, `=0` opts out.
- [x] Differential panel + soundness checks, below.
- [x] Graduation panel evidence attached, **plus the held-out arm on the two
      instances the issue names**.

**In-repo panel** — 66 instances at `time_limit` {5 s, 20 s}, both arms
**interleaved per instance**: 132 pairs, 559 executed comparisons
(`scratchpad/i1152/panel.py` → `panel_T5_T20.jsonl` → `panel_report.py`).

*Cert-clean:* **0** bounds above their reference optimum, **0** bounds crossing
their own run's incumbent, **0** `gap_certified` lost, **0** certified objectives
changed. The reference optimum is `known_optima.toml` where listed, else the best
feasible objective either arm returned; a 120 s incumbent hunt
(`scratchpad/i1152/oracle_probe.py`) supplies it for `casctanks` (9.163 feasible)
and `tanksize` (1.26864 certified). Three instances whose bound moved — `4stufen`,
`beuster`, `bchoco08` — have **no** feasible point at any budget tried and no
`.solu` in that environment; for those the argument is the structural one plus the
check that each moved bound stays below what the 120 s run proves (19055 < 20712,
6352 < 6395, 1.0−2e-12 < 1.0+5e-13). That gap is stated in the doc, not papered over.

*Net-positive:*

| effect | count | instances |
|---|---|---|
| bound RECOVERED (`None` → finite) | 2 | `casctanks` → 1.2584, `bchoco08` → 1.0 |
| bound TIGHTER | 3 | `4stufen` 18770→19055, `beuster` 5942→6352, `nvs09` −50.59→−48.99 (oracle −43.13) |
| bound LOOSER | 1 | `tanksize` @20 s, 1.253535→1.253040 (4e-4 rel; optimum 1.26864) |
| bound LOST | **0** | — |

| arm | mean `wall/T` | max | pairs > 1.25x |
|---|---|---|---|
| OFF | 0.447 | 1.448 | 2 |
| ON | 0.440 | 1.434 | 2 |

Per-pair delta −0.006x (−0.012x at 5 s, where the reserve binds), with the two large
moves exactly on the class: `casctanks` 1.23x → **1.02x**, `beuster` 1.13x → **0.98x**.

**Held-out arm — the two instances #1152 names.** Not in the in-repo corpus and not
reachable from the environment the panel ran in; run on the owner's machine during
review (`fe4d69c`), back to back at load 5.4 — elevated, which only strengthens the
result, since load pushes the overrun test toward failing:

| arm | result | wall |
|---|---|---|
| flag ON (the default) | **3 XPASS** | 101.6 s |
| `DISCOPT_ROOT_SETUP_BUILD_DEADLINE=0` | 3 XFAIL | 158.0 s |
| xfail markers removed | **3 passed** | 100.6 s |

Those three are `watercontamination0202` @30 s and @60 s (#875's 1.25x threshold) and
`sonet23v4` @2 s (#654's bound-survives guard) — i.e. exactly the two contracts the
issue called irreconcilable, now both satisfied by one solve each. The OFF arm is the
load-bearing half: it reproduces the defect as documented, which attributes the pass
to the flag rather than to the machine, and doubles as an end-to-end check that the
graduated flag's §5 opt-out is live.

**One counter-example, recorded rather than averaged away.** `hda` at an 8 s budget
(not one of the panel's two) goes the other way, consistently — 3 interleaved reps
per arm, load 0.58: OFF 9.68 s (sd 0.25, 1.21x, 1 node), ON 10.26 s (sd 0.10, 1.28x,
3 nodes), **bit-identical bound**. Root setup finishing 2.4 s earlier lets the search
start two more nodes and the last straddles the deadline — #966's per-node residual
given more chances to fire, not a setup overrun.

## The strict xfails #1150 added

#1150 merged first, which inverts the note this description used to carry. `main`'s
xfails for #1152 are `strict=True`, so merging `main` into a tree that fixes the
defect turned three of them into `XPASS(strict) → FAILED`; they are slow-tier, so CI
would not have caught it. Removed in `fe4d69c` along with the stale `test_875`
docstring that called the two contracts irreconcilable. **No threshold was touched** —
the 1.25x and every soundness assertion are still the ones #875 and #654 set.

## What this does not fix

`tls2` (1.45x → 1.43x), `tspn12` (1.39x → 1.38x) and `bchoco07` (1.21x, unchanged)
at 5 s are where the arms agree: their time is not in a root-setup relaxation build.
An instance whose *mandatory* root work alone exceeds `time_limit` will still return
late — none of that work is optional and none of it is a build. That residual is
now bounded by the reserve rather than unbounded by it: the fallback runs whenever
any budget survives setup.

Noted in passing, not fixed here: `generality_sweep`'s `root_build_deadline` arm has
been a no-op since #832 graduated that flag (it sets `=1` against a control that no
longer turns it off). For a default-ON flag the control arm is the `=0` one — which
is how the held-out arm above was run.

https://claude.ai/code/session_017e5a1afCTnzu4yti6xVuQZ


jkitchin:
## Review — approved, with one thing fixed on the branch

Read the production diff rather than the summary, and re-ran the claims that
sit in this change's blast radius. It holds up. Notes below, then the one
problem I found and fixed.

### The mechanism is the right one

This does not invent an interruption mechanism — it threads the existing
#694/#832 anytime build into the phases that were missing it. Checked the
truncation point (`uniform_relax.py:3782`): the deadline is polled **between
whole constraints**, never mid-`rep`, and the objective is fully linearized
before the loop. So a truncated build is a strict row-prefix, and the soundness
argument ("dropping rows enlarges the polytope, so every bound read off it is
weaker") is the correct one.

That prefix property is also exactly what makes the #208 aux-cascade skip
necessary rather than cautious: lifted columns are appended *during* the loop,
so a truncated build's column `c` need not be last round's quantity. Applying a
carried bound by index there would be unsound, not merely weak. Both the apply
and the capture side are guarded. Good catch by whoever wrote it.

Declining the native-kernel spec build instead of truncating it is right for the
stated reason — two builds whose row sets must correspond are not independently
truncatable.

### The "OFF is byte-identical" claim, checked

Worth checking because it is not obviously true: the probe and both pool sites
previously used a hand-rolled `time_limit - (perf_counter() - t_start)`, not
`_remaining_budget()`, and the new helper introduces a `max(0.0, ...)` clamp.

It holds. `t_start = _solve_t0` (solver.py:9595), so the expressions agree, and
the clamp difference (negative vs `0.0`) is absorbed downstream by
`max(_probe_remaining, _DEADLINE_NODE_FLOOR_S)` with the floor at `0.1`. Same
value either way.

Also checked the reserve cannot put the setup deadline in the past at solve
start: `_rr_reserve_s = min(3.0, 0.30 * time_limit)`, so the deadline is at
worst `t0 + 0.70 * time_limit`. No degenerate short-budget case.

### The three "pre-existing" failures

Verified rather than taken on trust, since two of them
(`test_930_root_probe_bound.py`) are named for the mechanism this PR changes.
Ran them on both arms in the installed tree with a marker assertion on the
loaded code (§8):

    branch (marker present): 2 failed, 4 passed
    main   (marker absent):  2 failed, 2 passed   <- same two test IDs

Pre-existing, confirmed. `test_perf_723_obbt_probe_cost.py::test_per_probe_cost_has_a_pivot_heavy_tail`
passed here, so that one looks load-dependent rather than a real failure.

---

## The one problem: three strict xfails go XPASS on merge

The PR body says:

> Note for whoever merges #1150 after this: its strict xfails on
> `test_sonet23v4_bound_survives_the_deadline_gating` and
> `test_watercontamination0202_honours_its_time_limit` are for this defect and
> should come off.

**#1150 merged first**, which inverts this. This branch predates it and does not
touch those files, so the merge takes main's version — xfails included — against
a tree that fixes the defect. `strict=True` then turns each into a failure:

    3 failed  ->  [XPASS(strict)] watercontamination0202 @30s, @60s; sonet23v4

They are slow-tier, so CI would not have caught it; it would have landed as a
broken slow tier.

Fixed on the branch (`fe4d69c5`): merged `origin/main` in, removed the three
xfail markers, and refreshed the stale `test_875` docstring that described the
xfail and called the two contracts irreconcilable — #1152 settled that they are
one defect, so that text was actively misleading.

No threshold was touched. The 1.25x and every soundness assertion are still the
ones #875 and #654 set.

### This also supplies the held-out arm the panel was missing

The PR is candid that `sonet23v4` and `watercontamination0202` "were not
reachable from the environment this ran in". They are reachable here, so the
gap is now closed — and on the two instances the issue actually names:

| arm | result | wall |
|---|---|---|
| flag ON (default) | 3 XPASS | 101.6 s |
| `DISCOPT_ROOT_SETUP_BUILD_DEADLINE=0` | 3 XFAIL | 158.0 s |
| merged, xfails removed | **3 passed** | 100.6 s |

Back-to-back, same machine, load 5.4 throughout — elevated, which only makes the
result stronger, since load pushes the overrun test toward failing and it passed
anyway.

The OFF arm is the part that matters: it reproduces the defect exactly as
documented, which attributes the pass to the flag rather than to the machine,
and it doubles as an end-to-end check that the graduated flag's §5 opt-out is
genuinely live (not the #1035 shape).

jkitchin:
Merged as `6d7e0de9`. #1152 closed; checked the other issue numbers this branch
names (#875, #654, #694, #832, #966, #208, #928, #1034) — all were already
closed weeks ago, so nothing was caught by a stray closing keyword, and #1151 /
#1153 / #1154 are still open.

Final state: 9/9 required checks green, `mergeState=CLEAN`; locally 19/19
adversarial and the benchmark smoke tier clean.

One small thing left behind, not worth its own CI cycle but worth knowing before
someone follows the instructions. The flag docstring records the control arm as

> **3 pass** with the flag on its default and **3 xfail** under
> `DISCOPT_ROOT_SETUP_BUILD_DEADLINE=0`

which is exactly what was measured — but it was measured while the strict xfail
markers were still on those tests. `fe4d69c5` removed them, so re-running the
`=0` arm on current `main` reports **3 failed**, not 3 xfail. Same observation,
different word: the tests do not pass with the flag off. Worth folding into the
wording next time this docstring is touched rather than spending a build on it.

Separately, and unrelated to this PR: a local `-m smoke` over both test trees
turned up `test_cert_instrumentation.py::test_reduction_separation_timers_present_and_bounded`
failing on `origin/main` — my regression from #1150, where the `row_filter/`
counters were never registered in that test's schema whitelist. Confirmed it
reproduces with this PR's code absent. Fix in #1156.

Issue #908, opened 2026-07-31
correctness: incumbent verifiers accept infeasible points (row misalignment on vector constraints)


jkitchin:
## Symptom

The incumbent feasibility verifiers walked rows by advancing **one index per `Constraint` object** while the evaluator emits **one row per flat element**. On any model with a vector constraint the two streams desynchronise, and the verifier checks the wrong row. It reports *feasible* for points that violate a constraint outright.

Three further defects in the same code, all in the wrongly-accept direction:
- builder-resident linear rows were never examined at all;
- `Constraint.rhs` was ignored;
- `_incumbent_is_feasible` checked neither variable bounds nor integrality.

Separately, the tolerance form `abs_tol + rel_tol * |residual|` is self-referential — it scales with the residual rather than the row, so on an equality row it collapses to a flat `1e-6` regardless of whether the row's value is 1 or 10,782, and the `rel_tol` term is arithmetically dead.

## Evidence (measured)

- Against a purpose-built corpus, the pre-fix row loop **wrongly ACCEPTS 6 of 7** infeasible points; the fixed verifier rejects 7/7 on all three entry points.
- A point violating row 2 of a size-3 vector constraint **by 5.0** was reported feasible.
- Every bad point in the corpus is asserted in-bounds and integral first, so the row check is provably the only thing that can reject it; a scalar control is rejected by both old and new, so the discriminator measures alignment rather than failing everything.
- `nvs22`'s incumbent was being rejected at a **relative** residual of 8.1e-9 (absolute 1.71e-5 against a row value of 2121.64) — the scale-blindness in the other direction.
- Corpus sweep of the tolerance change: 99 comparisons, one `False → True` (`nvs22`, worst relative 1.52e-8), zero `True → False`.

## Reproduction

```bash
pytest python/tests/test_vector_constraint_corpus.py -v
python -u discopt_benchmarks/scripts/verifier_scale_sweep.py
```

`test_vector_constraint_corpus.py` transcribes the pre-fix row loop verbatim as `_pre55_verify` and runs it against the same cases, so the 6-of-7 wrong-accept result is reproducible directly rather than argued.

**Coverage caveat worth knowing:** every row in the in-repo `.nl` corpus is scalar, so `from_nl` models cannot exercise this class at all. The corpus is built through the modeling API for exactly that reason — a correctness benchmark over the `.nl` corpus alone is blind to this defect.

## Where it is

- `python/discopt/validation/feasibility.py` (the consolidated verifier)
[...]

PR #1157, opened 2026-09-04
fix(correctness): #1151 incumbent row scale must be a term magnitude, not a coefficient

### LINKED ISSUE #1151: gdp/rust: reported objective is BELOW the global minimum on quotient expressions (false certificate)
## Summary

On the geometric-programming corpus, `min x/y + y/x` returns

```
objective = 1.998683979470214   status = optimal
```

The true global minimum is exactly **2** (AM-GM: `t + 1/t >= 2` for `t > 0`).
The reported objective is **below** the global minimum, and the status is
`optimal`, so this is a false certificate — the highest-severity class under
CLAUDE.md §1, not an accuracy complaint.

## The reported value is not the objective at the returned point

The distinguishing measurement: evaluate the objective in plain Python at the
solver's own returned incumbent.

```
returned objective   1.998683979470214
oracle at that point 2.000002247829649
delta               -1.318268e-03
```

So the solver is not merely finding a slightly-infeasible point and reporting
its true value; it is reporting a number that **no point attains**, its own
incumbent included. `objective` and `x` disagree.

## It is specific to division

Running the same comparison across the corpus's expression families, only the
quotient forms mismatch:

| family | max |reported − oracle at returned point| |
|---|---|
| affine | <= 4e-16 |
| bilinear | <= 4e-16 |
| separable square | <= 4e-16 |
| **division** | **1.3e-03** |

## The error scales as 1/denominator, so it is unbounded

`|delta| x denominator` is flat at approximately **1.9e-6** across denominator
floors of 1e-3 and 1e-2. That is the absolute constraint tolerance (1e-6)
amplified by `1/y`: a point that violates the defining equality of a quotient
aux by `eps` moves the quotient's value by `eps/y`. As `y -> 0` the reported
objective can be arbitrarily far below the true optimum, so no fixed tolerance
widening bounds this.

## Where it is

The returned value does not pass through any Python frame — a frame-level trace
over the solve shows zero Python returns of `1.998683979470214`. It comes from
the Rust incumbent path, which accepts a point as feasible under the absolute
tolerance and records the *relaxation's* objective value for it rather than
re-evaluating the true objective at the accepted point.

## Definition of done

The reported `objective` must equal the objective evaluated at the reported `x`,
to within a documented tolerance that does not depend on the magnitude of an
intermediate denominator. The likely correct fix is to re-evaluate the true
objective at an incumbent when it is accepted, and to treat a quotient aux whose
denominator is near its bound as not-yet-feasible rather than tolerance-feasible.

## Test

`python/tests/test_gp_corpus.py::test_bb_reported_objective_is_attained_by_its_own_incumbent`
is a **strict** xfail pinning exactly this, with a plain-Python oracle. It fails
the suite the moment this is fixed, which is the signal to remove the xfail.

Found while closing out #1039 (PR #1150). Referenced from that xfail's reason.

jkitchin:
## Summary

`minimize x/y + y/x` over `[1e-3, 1e3]^2` returned `objective =
1.998683979470214` at `status = optimal`. The global minimum is exactly 2
(AM-GM: `t + 1/t >= 2` for `t > 0`), so the reported number is **below the
global minimum** and is attained by no point — its own incumbent included, whose
true objective is 2.000002247829649. A false certificate, not an accuracy miss.

**Not where the issue expected.** The issue reports that no Python frame returns
the bad value, so it must come from the Rust incumbent path. A frame trace over
the solve does find one: `_tape_nlp_evaluator.evaluate_objective`. The reason it
looks like a Rust value is that the model being evaluated is the *reformulated*
one, where the objective is literally `_fr_aux_0 + _fr_aux_1` — so re-evaluating
"the objective at the point" reproduces the relaxation reading exactly, and the
#789 reported-vs-verified cross-check at `solver.py:1602` is a tautology on this
class. The native spatial kernel (#764) is the engine; with
`DISCOPT_NATIVE_SPATIAL_KERNEL=0` the Python path already returned 2.0.

**Root cause.** `_lift_objective_atoms` lifts `x/y` to an aux `w == x/y` whose
defining equality `_clear_divisions` clears to `w*y - x == 0` and then multiplies
by `1/dmin = 1000` — the emitted row is `1000*(_fr_aux_0*y) - 1000*x == 0` —
precisely so that a fixed ABSOLUTE residual test on the row bounds the error in
`w`. The incumbent verifier's row-scale term divided that scaling straight back
out:

```
row 0: residual=-9.276261e-04  J=[-1000, 997.84, 1.407, 0]
       max|J|*max(1,|x|) = 1.000e+03  -> tol 1.000e-03   ACCEPTED
       max|J|*|x|        = 1.405e+00  -> tol 1.405e-06   rejected
```

At the accepted point (`x = y ~ 1.4e-3`) `max_j |J_ij| * max(1,|x_j|)` reads the
row's scale as **1000** — the coefficient on `x` — rather than **1.4**, the
magnitude `1000*x` actually attains. That licenses a tolerance of 1e-3 for a row
violated by 9.28e-4, and the residual maps to `residual/(1000*y)` of error in
`w`, which is what the objective reads. Hence the issue's 1/denominator law.

**The fix**: the row scale becomes `max_j |J_ij| * |x_j|`, the first-order
magnitude of the row's *j*-th term. Flooring `|x_j|` at 1 does not make it a
bigger term, it makes it not a term magnitude at all, over-reading the scale by
`1/|x_j|` on every sub-unit column. The floor is redundant anyway — the caller
already floors the whole thing at 1 via `max(anchor, scale)` with `anchor >= 1`.

The guarantee this buys, derived in the module docstring for a defining row
`s*(w*D - N) == 0`: `|dw| <= abs_tol * max(1, |w|)` — a bound on the aux value,
keyed on the aux's own magnitude and on nothing about the denominator, which is
what the issue's definition of done asks for. Under the floored form the same
algebra leaves `|dw|` proportional to `1/|D|`, unbounded as `D -> 0`.

Closes #1151

### Blast radius: three sites, one definition

The first review pass found the defective expression written out **by hand in
three places** — this module's incumbent gate, `validation/examiner.py`'s scaled
primal-feasibility check, and `_dual_recovery.py`'s active-set test. Fixing one
left the other two vouching for exactly the point the fixed one rejects.
Measured on the witness before the other two were fixed:

```
verify_point                         -> ok=False  row 0 violated by 9.276e-04 (allowed 1.405e-06)
examiner primal_con_feas (scaled)    -> passed=True     <- certifies what the gate rejects
_dual_recovery near-test  floored scale=1.000e+03 -> near=True; term scale=1.405e+00 -> near=False
```

All three now import **`validation.feasibility.jacobian_row_scales`**, the single
definition — deliberately public, because three hand-written copies is *how* the
fix came to be half-applied. `test_row_scale_has_a_single_definition_shared_by_all_three_consumers`
asserts no module carries the floored form by hand, so the drift regresses, not
just its symptom.

The change only ever moves these gates in the strict direction —
`|x_j| <= max(1,|x_j|)` columnwise, so the scale never grows and the accepted set
only shrinks. The two forms differ only where the column attaining the floored
maximum carries `|x_j| < 1`. A rejection is decline-and-fall-back at every
consumer, never `infeasible`.

The second review pass then caught that extracting the helper unified the formula
but left its **non-finite guard** behind in `_row_scales`. `d/dx log(x)` at
`x = 0` makes `inf * 0` a NaN, and a NaN scale compares False against every
tolerance — a numpy RuntimeWarning and a spurious `[FAIL] primal_con_feas
(scaled)` on a feasible point. The guard now lives *with* the definition
(non-finite row → `0.0` → the caller's floor → the plain absolute tolerance, the
strictest answer). `_row_scales` keeps its stricter **whole-batch** fallback via
`_jacobian_row_scales_checked`: per-row zeroing there would have left co-occurring
rows on their own larger scales, i.e. a relaxation in the accepting direction
introduced by a refactor — the same shape of mistake as the original defect.

**One correction to the first review's finding 2.** It describes the harm on a
quotient aux's *defining* row. That row is an `==`, and `row_select` takes
`is_eq` unconditionally — `near` is never consulted for an equality — so the
scale cannot change the active set there (verified by driving
`recover_multipliers` on the witness: selected under both scales). The defect is
real and the reasoning about surplus free multipliers holds; it bites on scaled
**inequalities**, which `_clear_divisions` also produces. The regression test is
built on the `<=` form for that reason, with the sense asserted as a precondition.

### The collision with #1150, which happened mid-PR

#1150 merged (`fdbbb72`) while this branch was open, adding
`test_bb_reported_objective_is_attained_by_its_own_incumbent` as a **strict
xfail** against exactly this defect. Git merged `test_gp_corpus.py` with no
textual conflict — its xfail at line 216, mine at the end of the file — so
`mergeable_state` read `clean` and CI went red anyway with `F811: Redefinition`.
A clean merge is not a compatible one.

Resolved the way that xfail's own docstring asks for (*"this test XPASSes and
fails the suite, which is the signal to remove the xfail"*): #1150's test keeps
its name, the `xfail` marker is gone, and it is a verified XPASS — with
`feasibility.py` reverted to pre-fix in the installed tree (§8 marker count 0 vs
3) it FAILS. This branch's parametrized sweep is renamed
`..._attained_across_denominator_floors`; it generalizes rather than replaces,
adding the denominator axis and the two certificate-shaped checks the single box
cannot make. Two stale claims in the retired xfail's docstring (the Rust-path
attribution, and "the fix is bound-changing") are corrected in place, with a note
on what that diagnosis got right — the mechanism and the 1/y table, which is what
made the fix findable.

### Retractions (CLAUDE.md §11)

Three claims from earlier revisions of this body, withdrawn:

* *"`test_bb_opt_out_skips_gp_fast_path` currently FAILS on `main`."* True of the
  fork point (`1e75919`), false of current `main`: #1150 moved the objective
  assertion out of that test, so it passes on `main` and **this PR does not
  repair it**. What this PR repairs is the assertion itself.
* *"#1150's `abs=1e-9` threshold was kept over mine."* It was, and then the
  review pointed out it asserts **more than the fix guarantees**: the derived
  bound is `1e-6 * max(1,|obj|)` (~2e-6 for two auxes), and 1e-9 passes only
  because the accepted incumbent's residual sits at float noise rather than at
  the tolerance boundary — timing-dependent on a 5 s solve. It is now the
  guaranteed bound, with "never below 2.0" kept as the sharp assertion. No signal
  lost: the defect reported -1.318e-03, 660x outside the new band.
* The `deterministic=True` method note below cites root-phase overruns as "#1152
  showing up in a third place". **#1152 has since been fixed and merged (#1155).**
  The measurement was accurate when taken and is why the panels run on the real
  clock, but it is not a live report on current `main`.

## Correctness

The change makes three validation gates STRICTER, never looser, so none can admit
a point its previous form rejected. It touches no bound, cut, or relaxation. What
it removes is a false PRIMAL — a reported objective below the true optimum — and
the `status = optimal` that rode on it.

Where the changed path fires, the differential is one-sided (Panel B below): 8
super-optimal or mismatched reports in the OFF arm, 0 in the ON arm. Where it
does not fire (the whole vendored `.nl` corpus) the arms execute identical code,
measured, not assumed.

- [x] Cannot produce a false certificate — strictly tightens primal-acceptance
      gates; removes a false `optimal`, cannot create one
      — this PR *restores* one that a later widening had neutralized

`main` has moved **six times** while this PR was open (#1150, #1155, #1156,
#1159, #1162, #1163), and each merge was re-verified in full before pushing —
rebuilding the Rust extension first where the merge touched `crates/` (#1162,
#1163), so nothing was measured against a stale `.so`. The numbers below are the
**latest** run, on head `f6a0497`; the per-head figures for the earlier merges
are in this PR's comments, each naming its own head.

- [x] `pytest -m smoke` → **1129 passed**, 21 skipped, 2 xpassed (511 s). The
      2 xpassed are pre-existing and unrelated; the skips are optional-dependency
      gaps in this container.
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` →
      **19 passed** (169 s). `nvs22` is in that set — the instance whose #908
      widening this tightening could plausibly have undone. It passes.
- [x] `cargo test -p discopt-core` → **N/A, no Rust touched by this PR**
      (the Rust in the diff arrives via the merges of #1162/#1163).
- [x] `ruff check python/` + `ruff format --check python/` → clean, 964 files, at
      the repo's pinned **0.14.6**; `mypy python/discopt/` → Success, 335 files.
- CI's own selectors, run locally before each push: fast-correctness lane
  **337 passed** / 24 skipped / 4 xfailed; claim-boundary lane **308 passed** /
  33 skipped; the three touched files **72 passed** / 1 skipped.

One failure seen locally and established as **not this PR's**:
`test_rand_lagrangian_dual_is_valid_lower_bound[kelley-4]` asserts bit-equality
between a sequential and a threaded run, both cut by a 15 s `time_limit`. A reach
probe on the failing case counts **zero** calls into all three changed sites, so
the diff cannot move that bound; and on base sources with the §8 marker asserted
at 0 it still flakes **2 failed / 4 passed** in 6 quiet-machine runs. Pre-existing
on `main`; recorded in `scratchpad/issue1151/README.md` and deliberately not
fixed here.

## Regression test

Fail-before / pass-after, three times, each by reverting only the production
file(s) in the installed tree and confirming with §8 marker counts:

| reverted | result |
|---|---|
| `feasibility.py` | `test_gp_corpus.py`: **3 failed**, 23 passed → **26 passed** |
| `examiner.py` + `_dual_recovery.py` (helper kept so the module imports) | `test_vector_constraint_corpus.py`: **3 failed**, 24 passed → **27 passed** |
| `feasibility.py` (second-pass guard) | `test_vector_constraint_corpus.py`: **3 failed**, 27 passed → **30 passed** |

Verifier level (`test_vector_constraint_corpus.py`), no solve involved:

* `test_row_scale_is_a_term_magnitude_not_a_coefficient_magnitude` — pins the
  decision on the exact point the solver returned, asserting its own
  preconditions first (residual 9.276e-4, old scale 1000, new scale 1.405) so it
  cannot silently stop exercising the regime.
* `test_examiner_scaled_feasibility_rejects_what_the_incumbent_gate_rejects`
* `test_dual_recovery_active_set_uses_the_term_magnitude_scale` — on the `<=`
  form, driving `recover_multipliers` rather than re-deriving its arithmetic.
* `test_row_scale_has_a_single_definition_shared_by_all_three_consumers`
* `test_row_scale_is_finite_when_a_derivative_is_unbounded_at_zero` — runs under
  `simplefilter("error", RuntimeWarning)` so a warning fails the test.
* `test_row_scales_keeps_its_whole_batch_fallback`
* `test_examiner_does_not_warn_or_spuriously_fail_on_an_unbounded_derivative`
* `test_term_magnitude_scale_still_admits_the_nvs22_direction` — guards #908's
  widening. Passes in both arms, by design.
* `test_term_magnitude_scale_is_never_looser_than_the_floored_form` — 200 draws
  pinning the direction, with a compared-count assertion.
* `test_jacobian_row_scales_rejects_shape_mismatches` — the shared helper raises
  rather than broadcasting into a silently wrong scale (§7).

End to end (`test_gp_corpus.py`):

* `test_bb_reported_objective_is_attained_by_its_own_incumbent` — #1150's test,
  xfail retired.
* `test_bb_reported_objective_is_attained_across_denominator_floors[floor…]` —
  sweeps the box floor over 1e-3 … 1e0. **Asserts the changed path actually
  fired** (§6): `_row_scales` is entered only for rows already over the flat
  tolerance, and Panel A measured **0** such invocations across all 119 vendored
  `.nl` instances, so a green solve alone measures nothing about #1151. The
  counter is asserted per-floor, because at floors ≥ 1e-1 no row ever becomes
  suspect and the path legitimately does not fire.

Not bound-changing — no relaxation, cut or bound is touched — so there is no
default-off flag and no graduation gate. It is a primal-acceptance change with a
measurable effect, so the evidence is here anyway. Probes are in
`scratchpad/issue1151/` (see its `README.md`); each prints an executed-comparison
count and exits non-zero at zero, none swallows an exception, arms are
interleaved in one process, and the timing run is load-gated with a spread.

**The issue's own table, re-derived** (`probe_family.py`, 16 comparisons):
reported objective vs a plain-Python oracle at the solver's own returned point.

| family | before | after |
|---|---|---|
| affine | 0.000e+00 | 0.000e+00 |
| bilinear | 0.000e+00 | 0.000e+00 |
| separable square | 1.087e-15 | 1.087e-15 |
| **division** | **1.318e-03** | **4.361e-13** |

The `|delta| x denominator ~ 1.9e-6` signature — flat across box floors, i.e.
unbounded as `y -> 0` — is gone; the division column is float noise, and every
other family is unchanged value-for-value.

**Panel A — 119 vendored `.nl` instances**, arms interleaved per instance, 20 s,
`gap_tolerance=1e-4`. The headline is a negative result, reported as such:

```
verify_point calls        : ON 538  OFF 538
_row_scales calls (pass 2): ON 0    OFF 0
rows where the two forms diverged: 0
#1151 oracle violations   : ON 0  OFF 0
dual bound above a reference optimum: 0 / 52 checked
```

`_row_scales` is reached only for rows already over the flat absolute tolerance,
and **no verified incumbent on that corpus has one**. So the changed code never
executed and the two arms ran identical code. That is a narrowness result, and on
its own it is a probe that measured nothing (CLAUDE.md §6) — two further
instruments carry the weight.

**Null control.** Panel A still logged 10 instances whose bound or node count
differed. Since `divergent_rows = 0` those cannot be caused by this change, and
rather than argue that, the same 10 were re-run with **no code difference between
arms**: **8 of 10 still disagree**, with bounds moving comparably (nvs17
-1486.93 vs -1547.90, nvs19 -7236.3 vs -7095.3, nvs24 -174530 vs -174334) and
node counts drifting 1–10 %. Every objective was identical in both runs.
Wall-clock truncation, not the fix.

**Panel B — the class where the path does fire** (quotient objectives, 10 models
x 2 arms, interleaved). Positive control: `_row_scales` fires 8 times and the two
forms diverge on 10 rows, so the instrument demonstrably measures something.

```
case                    arm      status              reported              oracle@x        delta    opt  rows  div
balance_floor0.001       on    feasible                     2                     2    0.000e+00      2     1    2
balance_floor0.001      off     optimal      1.99868397947021      2.00000224782965   -1.318e-03      2     1    0
balance_floor0.01        on    feasible                     2                     2    0.000e+00      2     1    2
balance_floor0.01       off     optimal      1.99985334212268       2.0000384197609   -1.851e-04      2     1    0
ratio3_floor0.001        on    feasible                     3                     3    0.000e+00      3     1    3
ratio3_floor0.001       off     optimal      2.99800969965084      3.00010933723672   -2.100e-03      3     1    0
ratio3_floor0.01         on    feasible                     3                     3    0.000e+00      3     1    3
ratio3_floor0.01        off     optimal      2.99976909822596      3.00002939785867   -2.603e-04      3     1    0
```

**8 super-optimal / mismatched reports, all OFF, zero ON.** The remaining 12 rows
(`posyobj`, `cobb`, and the loose floors) never enter pass 2 and are
byte-identical between arms.

`ratio3` (`x/y + y/z + z/x`, optimum 3) is **not** in the issue — this panel
found it. It is the evidence that the fix addresses a class and not the named
model (CLAUDE.md §2).

**What it costs**, 3 interleaved reps at a 300 s budget, load 0.73 at start:

```
floor 1e-03 off: optimal 1.99868397947    wall mean   0.60s sd 0.02   <- false
floor 1e-03  on: optimal 2.0              wall mean  37.95s sd 0.27
floor 1e-02 off: optimal 1.999853342123   wall mean   0.75s sd 0.02   <- false
floor 1e-02  on: optimal 2.0              wall mean 228.16s sd 0.34
```

Both arms certify; no certificate is lost. OFF's speed is manufactured by its own
too-low incumbent — the tree stops at `bound >= incumbent - gap_tol`, and a
sub-optimal incumbent makes that trivially satisfiable. The trade is a slow true
certificate for a fast false one, which is the only direction §1 allows. This
cost is recorded in the CHANGELOG entry, since it is user-visible.

### One method note, recorded rather than buried

`deterministic=True` was the first choice for the panel and was abandoned with
the measurement: it renders the role-2 sub-budgets inert, and on the pre-merge
tree that let the root phase run **181 s (bchoco07)** and **299 s (bchoco08)**
against a 30 s budget and still return `nodes=0, bound=None`. Those runs carry no
comparable content, which is why the panels run on the real clock. See the
retraction above: #1152 has since been fixed and merged, and this has not been
re-measured against that fix.

https://claude.ai/code/session_018zU7YAPyMQuw26EdzcEKeg


jkitchin:
## CI red on the first push, fixed; and two retractions the base branch forced (CLAUDE.md §11)

`Lint + typecheck` failed on `0a303f8` with **F811: Redefinition of unused
`test_bb_reported_objective_is_attained_by_its_own_incumbent` from line 216**.

This is the collision the PR body flagged, arriving faster than expected: #1150
merged (`fdbbb72`) while this branch was open, and #1152's fix merged behind it
(#1155, `6d7e0de`). Git merged `test_gp_corpus.py` with no textual conflict — the
xfail went in at line 216, my test at the end — and produced two functions with
the same name. `mergeable_state: clean` was true and still red.

Fixed in `4331bea` (merge `f85c326` + resolution), resolved the way that xfail's
own docstring asks for: *"this test XPASSes and fails the suite, which is the
signal to remove the xfail."*

* **#1150's test keeps its name and every assertion and threshold unchanged** —
  including the `abs=1e-9` equality against its plain-Python oracle, which is
  *tighter* than the relative tolerance my version used. Only the `xfail` marker
  is gone. Verified as a real XPASS rather than a moved goalpost: reverting
  `feasibility.py` to its pre-fix content in the installed tree (§8 marker count
  0 vs 3) makes it FAIL, and it passes here on its own 1e-9 threshold.
* My parametrized sweep is renamed `..._attained_across_denominator_floors`. It
  generalizes rather than replaces: the denominator is the axis the defect was
  unbounded along, and it adds the two checks the single box cannot make — the
  reported value is never below the true global minimum, and the dual bound
  never exceeds it.

### Retraction 1 — "`test_bb_opt_out_skips_gp_fast_path` currently FAILS on main"

**Withdrawn.** That was true of the base this branch forked from (`1e75919`) and
is false of current `main`: #1150 moved the objective assertion out of that test
into the new xfail, so it passes on `main` today and this PR does not repair it.
What this PR does repair is the assertion itself, now living in the un-xfailed
test above. The substance is unchanged; the sentence about *which* test was red
was stale the moment #1150 merged.

### Retraction 2 — the `deterministic=True` method note

The PR body ends with a note that `deterministic=True` let the root phase run
181 s (bchoco07) and 299 s (bchoco08) against a 30 s budget, "the #1152
root-setup overrun class showing up in a third place." That measurement was
taken on the pre-merge tree. **#1152 has since been fixed and merged (#1155).**
The measurement was accurate when taken and is why this panel runs on the real
clock, but the note should not be read as a live report on current `main` — I
have not re-measured it against the #1152 fix, and it is not this PR's to claim
either way.

### Merged-tree verification, all re-run on `4331bea`

| check | result |
|---|---|
| `ruff check python/` + `ruff format --check python/` (pinned 0.14.6) | clean, 962 files formatted |
| `mypy python/discopt/` | Success, no issues in 335 source files |
| `pytest -m smoke` | **1122 passed**, 21 skipped, 2 xpassed (533 s) |
| Fast-correctness lane (CI's exact selector) | **337 passed**, 24 skipped, 4 xfailed (723 s) |
| Claim-boundary lane (CI's exact selector) | **308 passed**, 33 skipped (286 s) |
| `test_gp_corpus.py` at `-m "slow or not slow"` | **26 passed** (fail-before: 3 failed, marker 0 vs 3) |

One note on the earlier lint claim: the PR body says two untouched files are
reported unformatted. That was ruff **0.15.8**, which I had installed at the
time. Under the repo's pinned **0.14.6** the whole of `python/` is clean, so
there is nothing pre-existing to explain away. The 975 errors a wider
`ruff check` reports are all in `scratchpad/`, which pre-commit's `files:
^python/` excludes by design.

---

jkitchin:
## Code review — PR #1157

Verified in a clean worktree of `origin/claude/discopt-issue-1151-p5ihyj`, asserting `discopt.validation.feasibility.__file__` pointed at the worktree and that the floored form was absent before measuring.

**The core fix is sound.** Replacing the floored product with the term magnitude is strictly tightening (`|x_j| <= max(1,|x_j|)` columnwise); the caller's `max(anchor, scale)` floor with `anchor >= 1` keeps the tolerance at or above the flat `ABS_TOL`, so no row is held to zero; non-finite inputs still hit the `np.all(np.isfinite(sub))` guard and fall back to the stricter Jacobian-free bound; both consumers (`_native_kernel_verify_point`, `_convex_kernel._incumbent_is_feasible`) treat a rejection as decline-and-fall-back, never as `infeasible`; the docstring's `s*|D| >= 1` premise matches `_clear_divisions`' `if dmin < 1.0: scale /= dmin`; and the `panel_api.py` arm switch patches the module namespace `check_constraints` actually resolves at call time. Tests run on the branch: `test_vector_constraint_corpus.py` 23 passed; `test_gp_corpus.py -m "slow or not slow"` 26 passed (39.6 s); `test_902_native_seed_general_integers.py` + `test_spatial_native_kernel.py` 43 passed. An independent probe reproduced the fixed behaviour end to end (all four floors report objective 2.0 with bound ~1.9998-1.9999).

The main gap is that the fix was not applied everywhere the defective formula lives.

---

### 1. `python/discopt/validation/examiner.py:206` — the same defective floored row scale is left in the certificate examiner, and it still certifies the exact #1151 point

`jac_scale = np.max(np.abs(jac) * np.maximum(1.0, np.abs(x_flat)), axis=1)` is the pre-fix formula verbatim. Measured on the branch, feeding the examiner the PR's own witness point on the PR's own `1000*(w*y) - 1000*x == 0` model:

```
verify_point -> False: row 0 violated by 9.276e-04 (allowed 1.405e-06)
examiner     -> [PASS] primal_con_feas (scaled) (tol=1.0e-06)
```

The user-facing "is my solution feasible" tool still vouches for a point the fixed incumbent gate rejects, on the very class this PR is about. Either apply the same term-magnitude scale here or have the examiner call `feasibility._row_scales`, so the two cannot drift.

### 2. `python/discopt/_dual_recovery.py:172` — same floored form, and its comment now claims something false

The comment explicitly says this is "the same row_scale the examiner uses for scaled primal feasibility" — false in two directions after this PR. Concretely, on a `1/dmin`-scaled row (`s=1000`) whose columns are all sub-unit, `row_scale` reads 1000 instead of ~1.4, so `near = |signed| <= active_tol * row_scale` admits rows whose signed slack is up to 1e-3 into the active set. Extra inactive rows in the least-squares active set add free multipliers, which can shrink the stationarity residual and make a non-KKT point pass `stationarity` — a validation weakened in the accepting direction. Worth fixing in this PR, since the PR's whole argument is that the floored product is not a row scale.

### 3. `python/tests/test_vector_constraint_corpus.py:259` — `NAIVE_WIDENINGS` header documents the removed form as the reference

The header still describes the chosen form as `violation <= abs_tol * max(1, |rhs|, max_j |J_ij| * max(1,|x_j|))`, i.e. the form this PR removes. That block is the module's documented control set for "each naive widening", so the stale comment now presents one of the rejected forms as the baseline. (The parametrized cases themselves use `ABS_TOL * max(1.0, scale)` and still pass; it is the comment that is wrong.)

### 4. `python/tests/test_gp_corpus.py:345` — the new sweep never asserts the changed path fired

It solves and checks values only, with no executed-assertion count and no proof `_row_scales` ran (CLAUDE.md §6). If routing ever changes — a different reformulation, a fast path that avoids the quotient aux, or a `_row_scales` pass-2 that is never reached — the test keeps passing while measuring nothing about #1151. A cheap counter (assert `_row_scales` was invoked at least once during the solve, or assert the suspect-row regime the way the verifier-level test does) keeps it a probe rather than a smoke test.

### 5. `python/tests/test_gp_corpus.py:281` — objective assertion is 1000x tighter than the fix guarantees

`assert result.objective == pytest.approx(oracle, abs=1e-9)`. The derived bound in the new module docstring is `|Δw| <= abs_tol * max(1,|w|)` = 1e-6 per aux, so an incumbent accepted just inside the new tolerance (residual at the boundary rather than at float noise) would report a delta up to ~2e-6 and fail — on a 5 s time-limited solve whose incumbent path is timing-dependent. It passes today because the accepted incumbent's residual is ~0, not because the property is guaranteed. Either pin the point/residual the way the sibling verifier-level test does, or relax to the guaranteed `1e-6*max(1,|obj|)` and keep the "never below 2.0" assertion as the sharp one.

### 6. `python/discopt/validation/feasibility.py:219` — no CHANGELOG entry

Against the repo's consistent practice: every recent `fix(...)` commit on `main` (#1152, #1147, #1036, #1050) touches `CHANGELOG.md`. This one is user-visible twice over — a previously reported `status=optimal` objective below the true optimum is gone, and the same models now take substantially longer to certify (the PR's own measurement: 0.6 s → 37.95 s at floor 1e-3, 0.75 s → 228 s at 1e-2).

---

Non-findings I checked and cleared are listed in the sound-fix paragraph above.

jkitchin:
## Review addressed — all six, in `f66f362` (merged to `ff65000`)

Thank you for finding **1**. That is the important one and it is a real miss: I
fixed the formula where I found it and did not check whether it lived anywhere
else. It lived in two more places, written out by hand, so the fix left the
user-facing feasibility tool certifying exactly the point the fixed gate
rejects. That is a §2 failure — I fixed a site, not the class.

Reproduced before touching either site, `scratchpad/issue1151/probe_examiner.py`
(prints an executed-check count, exits non-zero at zero):

```
verify_point                         -> ok=False  row 0 violated by 9.276e-04 (allowed 1.405e-06)
examiner primal_con_feas (unscaled)  -> passed=False
examiner primal_con_feas (scaled)    -> passed=True     <- certifies what the gate rejects
_dual_recovery near-test  |signed|=9.276e-04  floored scale=1.000e+03 -> near=True
                                              term    scale=1.405e+00 -> near=False
```

and after: `examiner primal_con_feas (scaled) -> passed=False`.

### 1 + 2 — one definition, three consumers

Rather than copy the corrected expression into two more files, the three sites
now import **`validation.feasibility.jacobian_row_scales`**. It is deliberately
public, with the reason in its own docstring: three hand-written copies is *how*
this defect came to be half-fixed, and the next person to change the formula
should not be able to change one of three. It validates its inputs and raises on
a 1-D Jacobian or a column/point mismatch rather than broadcasting into a
silently wrong scale (§7).

`test_row_scale_has_a_single_definition_shared_by_all_three_consumers` asserts
no module carries `np.maximum(1.0, np.abs(x_flat))` by hand any more, so the
drift itself is what regresses, not just its symptom.

**One correction to finding 2's stated mechanism.** You describe the harm on a
quotient aux's defining row. That row is an `==`, and `row_select` takes `is_eq`
**unconditionally** — `near` is never consulted for an equality — so on that row
the scale cannot change the active set. I checked this by driving
`recover_multipliers` on your exact witness: the row is selected under both
scales. The defect is real, and your reasoning about surplus free multipliers
shrinking the stationarity residual holds; it bites on scaled **inequalities**,
which `_clear_divisions` also produces (it multiplies a `<=` through and flips
the sense for a negative denominator). The regression test is therefore built on
the `<=` form, asserts `sense == "<="` as a precondition, and drives
`recover_multipliers` rather than re-deriving its arithmetic in the test.

### 3 — stale `NAIVE_WIDENINGS` header

Fixed. Worth noting how bad it was: the header presented `max(1,|x_j|)` as the
*reference* form in the one block whose whole job is to enumerate rejected
shapes.

### 4 — the sweep measured nothing about #1151

Fixed with a counter on `_row_scales`, asserted `>= 1` at the two tight floors.
Per-floor rather than unconditional, because at floors ≥ 1e-1 the denominators
are large enough that no row ever becomes suspect — the path legitimately does
not fire there, and asserting it would be a wish rather than a precondition.
This is the sharper version of your point: Panel A already measured **0**
`_row_scales` invocations across all 119 vendored `.nl` instances, so "solves
green" really does mean nothing here without the firing proof.

### 5 — `abs=1e-9` asserted more than the fix guarantees

Agreed, and relaxed to `1e-6 * max(1,|obj|)` with "never below 2.0" kept as the
sharp assertion, as you suggested. I had preserved 1e-9 specifically to avoid
weakening #1150's threshold, but you are right that it is a latent flake rather
than a stronger guarantee: it passes because the accepted incumbent's residual
sits at float noise, not because anything forces it to. No signal is lost — the
defect reported -1.318e-03, which is 660x outside the new band.

### 6 — CHANGELOG

Added under Unreleased/Fixed, including both user-visible halves: the false
`optimal` is gone, **and** these models now take substantially longer to
certify (0.60 s → 37.95 s at box floor 1e-3; 0.75 s → 228.16 s at 1e-2), because
the previous certificate was manufactured by the too-low incumbent.

### Verification on this head

Fail-before / pass-after for the three new tests, with **only the two consumer
sites** reverted (keeping the helper so the module still imports): **3 failed →
27 passed**.

| check | result |
|---|---|
| `ruff check python/` + `ruff format --check python/` (pinned 0.14.6) | clean, 962 files |
| `mypy python/discopt/` | Success, 335 source files |
| `pytest -m smoke` | **1126 passed**, 21 skipped, 2 xpassed (512 s) |
| `pytest -m slow test_adversarial_recent_fixes.py` | **19 passed** (172 s) |
| Fast-correctness lane (CI's selector) | **337 passed**, 24 skipped, 4 xfailed (724 s) |
| Claim-boundary lane (CI's selector) | **308 passed**, 33 skipped (283 s) |
| Every test file touching the two new sites | **313 passed**, 8 skipped, 2 xpassed |
| `cargo test -p discopt-core` | N/A — no Rust touched |

`main` moved again under #1156 while this was in flight; merged in `ff65000` (no
conflicts) and re-verified lint/format and the three touched files (69 passed)
on the merge result, so CI is testing what would actually land.

---

jkitchin:
## Code review — PR #1157, second pass (on `ff650006`)

Verified in an isolated worktree at the PR head, asserting `feasibility.py` / `examiner.py` / `_dual_recovery.py` all resolved to the worktree and that the shared helper was present before measuring.

**All six first-pass findings are fixed**, and the two important ones were fixed better than I suggested: rather than patching each site, `jacobian_row_scales` was extracted into `validation/feasibility.py` as the single definition, and both consumers import it.

```
floored-form copies remaining : (none)
examiner + dual recovery both call jacobian_row_scales : OK
term magnitude = 1        floored form = 1000        (J=[[1000,1000]], x=[1e-3,1e-3])
```

On the #1151 witness point the contradiction between the tools is gone:

```
verify_point                       -> ok=False  row 0 violated by 9.276e-04 (allowed 1.405e-06)
examiner primal_con_feas (unscaled)-> passed=False
examiner primal_con_feas (scaled)  -> passed=False
_dual_recovery near-test           -> floored scale 1.000e+03 -> near=True
                                      term    scale 1.405e+00 -> near=False
```

The rest:

- **Finding 3** — the `NAIVE_WIDENINGS` header now names `max_j |J_ij| * |x_j|` and explains that #1151 replaced the factor it used to document (and that the old form is one of the shapes the list exists to reject).
- **Finding 4** — `abs=1e-9` is replaced by the guaranteed `1e-6 * max(1, |oracle|)`, *plus* a new sharp assertion — `result.objective >= 2.0 - tol` — which is the false-certificate property the issue is actually about. Asserting the contract rather than the observation, with the measured deltas recorded in the comment.
- **Finding 5** — the sweep now monkeypatches `feasibility._row_scales` with a call counter and asserts it fired. Asserted **per floor** (`if floor <= 1e-2`) rather than unconditionally, since at loose floors no row becomes suspect and the path legitimately does not run — that is the right call and keeps it a real precondition instead of a wish.
- **Finding 6** — CHANGELOG entry present, and it records both user-visible effects.

Tests on the branch: `test_gp_corpus.py` + `test_vector_constraint_corpus.py` `-m "slow or not slow"` — 53 passed (was 49; +4 new). Validation / examiner / dual-recovery selection — 98 passed.

---

### 1. LOW/MEDIUM — `python/discopt/validation/examiner.py:219` (and the matching site in `_dual_recovery.py`) — the refactor unified the formula but not its non-finite guard

`feasibility._row_scales` protects the computation with `np.all(np.isfinite(sub))` (line 249) and falls back to the stricter Jacobian-free bound. The two new callers call `jacobian_row_scales` and divide by the result with no such guard, and the helper itself does not handle it — its docstring covers the all-terms-vanish case but not this one.

A row with an unbounded derivative at a variable pinned to zero — `d/dx log(x)` at `x = 0` — makes `inf * 0 = nan` inside the helper. Measured on the branch:

```
J = [[inf, 1.0]],  x = [0.0, 2.0]

term magnitude (new) : [nan]   -> row_scale=nan  scaled=nan  examiner passed=False
floored form   (old) : [inf]   -> row_scale=inf  scaled=0.0  examiner passed=True

RuntimeWarning: invalid value encountered in multiply
  feasibility.py:220 in jacobian_row_scales
```

Direction first, since it matters: **both arms are conservative** — the examiner reports a violation and `_dual_recovery`'s `near` test goes `False`, excluding the row from the active set. There is no false certificate here, and the new behaviour is strictly safer than the old one, which handed the row an infinite tolerance and passed it unconditionally.

What it costs is a user-facing regression in the tool this PR just made authoritative: a numpy `RuntimeWarning` on stderr and a spurious `[FAIL] primal_con_feas (scaled)` for a point that may be perfectly feasible, with `scale=nan` printed in the detail line. The clean fix is to put the finite handling inside `jacobian_row_scales` — it is now "the one definition of this quantity", so the guard belongs with it rather than in two of the three callers — or, failing that, give the two new callers the same `isfinite` fallback `_row_scales` already has.

jkitchin:
## Second-pass finding fixed — `4d318fa` (merged to `cdf21e2`)

You're right, and the diagnosis is exact: the guard was a property of the old
*location*, not of the definition, and extracting the formula did not bring it
along. That is the same shape of mistake as the original defect — a refactor
that moves an expression without its invariant — so it belongs fixed the way you
suggest, **inside** `jacobian_row_scales`, not bolted onto two of three callers.

Reproduced first, then fixed:

```
J = [[inf, 1.0]], x = [0.0, 2.0]
term magnitude (extracted) : [nan]  -> RuntimeWarning; examiner [FAIL] scale=nan
floored form   (pre-#1151) : [inf]  -> INFINITE tolerance, passes unconditionally
term magnitude (guarded)   : [0.0]  -> caller's floor, i.e. plain ABS_TOL
```

A non-finite row now returns **0.0** — no usable estimate, so the caller's floor
applies and the row is held to the plain absolute tolerance. Strictest answer,
same direction `_row_scales` already took, and it keeps the improvement over the
pre-#1151 form you noted (which handed such a row an infinite tolerance).

### One thing your suggestion would have cost, so I split the helper instead

Putting per-row zeroing straight into the shared helper would have changed
`_row_scales` behind its back. It has a **whole-batch** fallback: one non-finite
row sends *every* suspect row to the Jacobian-free bound. Per-row zeroing would
leave the co-occurring rows on their own larger scales — looser than before the
helper was extracted, i.e. a relaxation in the accepting direction introduced by
a refactor. Exactly the failure mode this PR exists to fix, and it would have
gone in under the banner of fixing it.

So: `_jacobian_row_scales_checked` returns `(scales, all_finite)` and
`_row_scales` keeps its whole-batch `None`; the public helper returns the
finite-safe per-row form for callers that have no batch to fall back for.
`test_row_scales_keeps_its_whole_batch_fallback` pins the distinction, since it
is the kind of thing a future simplification would quietly collapse.

### Tests

Three, each fail-before / pass-after with only `feasibility.py` reverted
(**3 failed → 30 passed**):

* `test_row_scale_is_finite_when_a_derivative_is_unbounded_at_zero` — asserts the
  pre-#1151 form is non-finite here *as a precondition* (so it cannot stop
  exercising the case), then runs under `simplefilter("error", RuntimeWarning)`
  so a warning fails the test rather than being invisible, and checks that a
  finite row sharing the batch keeps its own scale.
* `test_row_scales_keeps_its_whole_batch_fallback`
* `test_examiner_does_not_warn_or_spuriously_fail_on_an_unbounded_derivative` —
  end to end on a `log(a) + b <= 100` model at `a = 0`, asserting both no
  RuntimeWarning and `passed is True`.

### Verification on this head

| check | result |
|---|---|
| `ruff check python/` + `ruff format --check python/` (pinned 0.14.6) | clean, 963 files |
| `mypy python/discopt/` | Success, 335 source files |
| `pytest -m smoke` | **1129 passed**, 21 skipped, 2 xpassed (512 s) |
| `pytest -m slow test_adversarial_recent_fixes.py` | **19 passed** (169 s) |
| Fast-correctness lane (CI's selector) | **337 passed**, 24 skipped, 4 xfailed (720 s) |
| Claim-boundary lane (CI's selector) | **308 passed**, 33 skipped (283 s) |
| Every file touching the three row-scale sites | **316 passed**, 8 skipped, 2 xpassed |
| `cargo test -p discopt-core` | N/A — no Rust touched |

`main` moved again under #1159 (the #1154 hull work) while this was in flight;
merged in `cdf21e2` with no conflicts and re-verified lint/format and the three
touched files (72 passed) on the merge result, so CI is testing what would land.

---

jkitchin:
## Code review — PR #1157, third pass (on `f6a04971`)

**The one open finding is fixed, and fixed more carefully than I proposed.** Verified in an isolated worktree at the PR head (asserted `feasibility.__file__` in the worktree and `_jacobian_row_scales_checked` present before measuring).

`4d318fa2` puts the non-finite guard back **with the definition** rather than in the callers — the right home, since `jacobian_row_scales` is now "the one definition of this quantity". More to the point, it caught something my finding missed: naively zeroing per row inside `_row_scales` would have **loosened** that function's whole-batch fallback, where one non-finite row sends *every* suspect row to the stricter Jacobian-free bound. Per-row zeroing would have left the co-occurring rows on their own larger scales — a relaxation in the accepting direction, smuggled in by a refactor. Hence the split into `_jacobian_row_scales_checked` returning `(scales, all_finite)` so `_row_scales` keeps its batch behaviour, with the public helper's per-row `0.0` for callers that have no batch to fall back for. Good catch.

Measured on the review's own repro (`J = [[inf, 1.0], [2.0, 3.0]]`, `x = [0.0, 2.0]`):

```
scales            : [0. 6.]          # non-finite row -> 0.0 (strictest); finite row untouched
RuntimeWarnings   : (none)
examiner row_scale: [1. 6.]   scaled=[1.0e-09 1.67e-10]   passed=True
batch fallback    : _row_scales uses the checked variant  OK
```

No NaN, no `RuntimeWarning`, no spurious `[FAIL] primal_con_feas (scaled)` with `scale=nan`, and the finite row in the same batch is unaffected.

### Test results

GP corpus + vector-constraint corpus + validation/examiner/dual-recovery selection, `-m "slow or not slow"`: **183 passed, 1 failed**.

The 1 failure is `test_issue_517_numerical_dual_bound.py::test_hda_flag_disabled_is_sound_but_no_longer_discriminating`, and it is **not this PR's**. Established three ways rather than asserted:

* **Reach probe** — counters on all three call sites this PR changed, during that test: `{'verify_point': 0, 'jacobian_row_scales': 0, '_row_scales': 0}`. Zero calls; the diff cannot move that bound.
* **Repeat runs on this branch** — pass, pass, fail across 3 runs, with `load averages` moving 4.79 → 2.13 over the sequence.
* **The assertion is nondeterministic by construction** — both arms report `status='time_limit'`, `nodes=3`, `objective=None`, so it compares two wall-clock-truncated bounds (`OFF -141697.43 vs ON -13992288065.86`). The test's *soundness* assertions (`bound <= _HDA_OPT + 1e-2`, `status != 'optimal'`) passed in every run; only the differential/inertness assertion trips.

This is the same shape as the `kelley-4` flake already analyzed in `7767cca1`. Filed as #1165 so it stops costing each unrelated PR a round of "is this mine?" — it is not a blocker here.

### CI

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, every non-skipped check SUCCESS — lint + typecheck, Rust, Python fast, Python correctness (fast subset), claim-boundary (serial), AMP coverage, codecov patch and project.

### Summary across the three passes

| pass | findings | outcome |
|---|---|---|
| 1 | 6 | Core fix sound; the same defective formula stood in `examiner.py` and `_dual_recovery.py` |
| 2 | 1 | All six fixed — via one shared `jacobian_row_scales` rather than three patches; the extraction dropped the non-finite guard |
| 3 | 0 | Guard restored with the definition, and the batch-fallback relaxation it would have introduced avoided |

No open findings. Approving from my side.

PR #1158, opened 2026-09-04
feat(mpec): #1148 source complementarity residuals and the local-vs-certified result contract

### LINKED ISSUE #1148: mpec: source complementarity residuals and the local-vs-certified result contract (RFC #1123 slice 2)

jkitchin:
## Summary

Slice 2 of the MCP/MPEC RFC (#1123), on top of #1147's durable source provenance. Three things were conflated before this and are now separate, named quantities.

**A. Source residuals, computed on the source relation.** New `discopt/mpec_report.py` measures a point against the relations' *declared* operands — which #1147 guarantees survive every rebuilding pass — and reports, each carrying the definition string that produced it: complementarity (`min(f, g)` for a nonnegative pair, the MCP normal map for a box; `product`/Fischer–Burmeister selectable), the declared-bound violation per operand, source primal feasibility against the *original* rows and bounds, `max_i y_i(1-y_i)` over binary selectors after checking `0 <= y <= 1`, the continuation parameter and termination reason, and — beside all of it — the *lowered* rows' own residual so the two can be compared.

Tolerances are scaled by the relation's declared scale, and the worst-case residual a Scholtes regularization at `t` **admits** (POUNCE Gate 0, jkitchin/pounce#794) is reported rather than left to imply exact orthogonality — as `Residual.admitted_scale`, **per residual definition**: `sqrt(t)` for `min` and (on the nonnegative pair) `natural_map`, `t` for the product form, `(2 - sqrt(2))·sqrt(t)` for Fischer–Burmeister. It is documented as an upper bound on the residual the relaxation admits, *not* a limit on attainable accuracy: `f·g <= t` also admits `(f, g) = (0, 1)`, whose residual is exactly `0` at every positive `t`. `within_admitted_scale` compares against that bound and nothing else — it does **not** establish feasibility of the relaxed rows, which is what `lowered_row_residual` is for.

**B. A distinct terminal status.** New `discopt/status.py` defines `local_optimal`, `local_limit` and `local_infeasible`. A status, not a `certified=False` flag beside `"optimal"`: the harness scores `status == OPTIMAL` with no gap as a *proved optimum*, so a local stationary point reported as optimal is read by the release gate as a certificate. A stalled continuation reports `local_infeasible`, never `infeasible`.

`local_optimal` means a local **stationary** point and is reserved for a point a subsolver converged to. A usable-but-stalled iterate is still reported — as a warm start, and with its residuals — under the weaker `local_limit`, which is in `LOCAL_STATUSES` and so inherits the same no-dual-bound guard. The continuation trace keeps the same distinction so the result and the trace cannot disagree: `ContinuationStage.certified` says whether the subsolver converged at that stage, `ContinuationTrace.reported_point_certified` whether the stage behind the *reported* point did, and `converged` reads that rather than `any_stage_accepted`.

**C. The contract, enforced rather than only written.** `SolveResult.__post_init__` raises if a local status carries `bound` or `root_bound`, and forces `gap_certified=False` — a local result may become an incumbent only after independent verification (`accept_local_incumbent`, which gates on the source residuals and not on the lowered relaxation), and may never become a bound. `accept_local_incumbent` **recomputes** that verification at the acceptance boundary against the actual candidate and the model's current relations; it accepts no report from the caller, because a report records neither the point nor the model it was taken at. Stationarity is never claimed, since discopt checks no C-/M-/S- conditions. Written up in `docs/complementarity_semantics.md` §7–§8, together with the shared benchmark schema jkitchin/pounce#780 asks for.

**D. One return type.** `solve_mpec` returns a `SolveResult` for `scholtes`, `sos1` and `gdp`; the Scholtes arm carries its continuation trace (per-stage `t`, status, accept/reject reason, certified flag, source residual) rather than discarding it, and reports the achieved source residual independently of the final homotopy parameter. Lowerings record the rows they emit per relation per model (`LoweringRecord`), which is what makes the lowered-vs-source comparison possible. A model left carrying a Scholtes *relaxation* is refused by the solve guard rather than certified: the guard distinguishes "has a lowering" from "has an **exact** lowering".

Benchmark harness: `SolveStatus.LOCAL` plus one shared `DISCOPT_STATUS_MAP` used by the in-process runner, the subprocess worker, the category runner and the GDPLib runner, so `--subprocess` cannot score a solve differently from the in-process path. The shared table is held to exactly the entries the panel runners already had plus the three local statuses, so panel behaviour is byte-identical for every pre-existing status.

**The measurement this slice exists for.** On `max x+y s.t. x==y, 0 <= x ⊥ y >= 0` after a Scholtes homotopy at `t=1e-8`:

| quantity | value |
|---|---|
| max violation of the generated rows (`x>=0`, `y>=0`, `x·y<=t`) | `1.0e-8` |
| source complementarity `min(x, y)` on the declared operands | `1.4e-4` |

Four orders of magnitude apart at the same point, and the returned objective (`-2.8e-4`) is *below* the true optimum (`0`) — which is exactly why that result may never become a bound. That model is the regression probe; a residual that cannot tell the two apart fails it.

Closes #1148.

## Correctness

- [x] Cannot produce a false certificate. The change adds a *reporting* layer and several *refusals*; it removes no check. `Model.solve`'s math is untouched apart from the shared `interval_eval` fix below, whose effect is to **widen** enclosures that previously did not contain the value. The behavioural changes in the certifying direction are all tightenings: a local status can no longer carry a dual bound at all, a model carrying a Scholtes relaxation can no longer be solved as if the relation were encoded, and a local result can no longer be vouched for on a measurement taken somewhere else.
- [x] No validation, fallback, or safety guard was weakened. Guards were *added*: `evaluate_at_point` refuses a non-degenerate interval enclosure rather than reporting its midpoint; `relation_residuals` refuses when handed relations and measuring none; every constraint-list entry kind (`Constraint`, `_DisjunctiveConstraint`, `_SOSConstraint`) has an explicit residual or raises, so a row silently skipped cannot make the maximum read as complete.

On head `abbc7c2`, after all four review rounds plus the closeout wording fixes:

- [x] `pytest -m smoke python/tests/` → **1196 passed, 20 skipped, 2 xpassed**, 455 s
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **19 passed**, 169 s
- [x] `ruff check python/` / `ruff format --check python/` clean; `mypy python/discopt/` → no issues in 337 files
- [x] full `pytest python/tests/` → **8943 passed, 142 skipped, 5 xfailed, 2 xpassed**, 1122 s, alongside the three `cyipopt` failures noted below
- [x] `pytest python/tests/test_mpec_source_residuals.py` → **50 passed**, with **198 executed residual comparisons** counted and asserted non-zero (CLAUDE.md §6)
- [x] `pytest python/tests/test_interval_sum_reduction.py` → **13 passed** (33 comparisons); the #1160 xfail is gone — see below
- [x] `test_mpec.py`, `test_mpec_mp1_mp2.py`, `test_complementarity_provenance.py`, `test_nl_complementarity.py` → **74 passed**

The executed-comparison counter is a `teardown_module` hook, not a test: as a test it was position-dependent, and this suite runs under `pytest-randomly`, so a shuffle that put it first would have made the guard against measuring nothing itself measure nothing.

**Environment gap, not a result.** Four tests fail or hang in this container only for want of `cyipopt`/the Ipopt C library: `test_convex_nlp_certificate_853::test_neglog_interior_stall_not_false_optimal[ipopt]`, `test_issue_1066_master_bound_inversion`, `test_relax_compiler_convexity_units::test_ipopt_backend_active_constraint_gradient` (all three `ImportError: cyipopt is required`), and `test_75_tape_nlp_evaluator::test_solve_degrades_to_jax_when_pounce_is_missing`, which exceeds its 600 s timeout. All four were measured **identically on the unmodified tree at the same head** (the JAX-fallback subprocess was timed at both: terminated at the same 500 s cap in each), and CI is green on that head.

## Regression test

`python/tests/test_mpec_source_residuals.py` (50 tests) plus `test_interval_sum_reduction.py` pin each acceptance criterion of #1148 and each review finding individually. On `main` the first fails at collection, because `discopt.status` and `discopt.mpec_report` do not exist there — so its fail-before is real but trivial for most cases. The substantive fail-before evidence is narrower and worth stating plainly:

- `test_scholtes_source_residual_differs_from_the_lowered_row_residual` measures a quantity that had **no code path** before this change: nothing in the tree computed a source residual (there was no `r_mpcc`), so the assertion could not have been written against `main` at all.
- `test_mpec.py::test_scholtes_reaches_global_optimum` had to be **updated** here: it read `res.x[0]`, which is a `KeyError` now that every arm returns a `SolveResult` whose `x` is a name→value dict. The behaviour change is pinned in both directions.
- `test_local_result_can_never_carry_a_dual_bound` fails on `main` in the strong sense: `SolveResult(status="local_optimal", bound=0.0)` is silently accepted there.
- `test_review_1_sum_row_residual_is_reduced` fails on `main`: the residual reads `0.0` on a row violated by `2.0`.
- The round-4 regressions (`test_review3_1_*`, `test_review3_3_*`, `test_slow_axis_reduction_is_widened_for_a_sequential_accumulation`) each fail on this branch's own parent `f9f4a2b` and pass here — the strongest fail-before available, since they target code this PR wrote.

Where a blanket fix could have passed a test vacuously, the test carries a control: `accept_local_incumbent` returning `None` for everything would satisfy both the round-1 and round-4 assertions, so the controls check that a genuinely complementary point is still accepted (objective 1.0) and that the point a report *was* taken at is still accepted; a converged homotopy must still report `local_optimal`; and the fast-axis interval reduction must stay **tight**, so widening everything cannot pass the slow-axis fixture.

No flag, cut or relaxation is introduced. One **shared component** did change: `interval_eval` now reduces a `SumExpression` (round-1 HIGH below), which alters enclosures on models that use `dm.sum(...)`. The evidence is therefore in two parts — the `.nl` panel, and a Python-side check, because the `.nl` corpus structurally cannot exercise the changed path.

The round-4 widening fix is confined to that same direction by construction: the pairwise path is bit-identical to before, and the strided path uses `max(n - 1, log2(n) + 2)` rather than a bare `n - 1` precisely so it can only ever widen. The two expressions cross at `n = 5`; both are sound, but taking the smaller would have made a soundness fix *narrow* strided enclosures, which is bound-changing for no benefit. `test_the_sequential_bound_never_narrows_an_enclosure` asserts the ordering across the crossover.

### `.nl` panel: `global50`, 50 instances, 60 s cap, A/B against merge-base `1e75919`

*Historical snapshot, measured 2026-09-05 at head `78bd898`. It is retained because it isolates the `SumExpression` reduction against this branch's merge base; the round-4 commit `1347c2b` changes only the strided branch of that reduction, in the widening direction, and — per the executed-node probe below — the panel instances never reach the reduction at all.*

| | base (`1e75919`) | `78bd898` |
|---|---|---|
| solved | 47/50 | 47/50 |
| proved optimal | 47/50 | 47/50 |
| **incorrect_count** | **0** | **0** |
| statuses | 47 optimal, 2 feasible, 1 time_limit | 47 optimal, 2 feasible, 1 time_limit |

Per-instance A/B over **200 compared fields** (status, objective, bound, node_count × 50) shows 4 differences, all on the three rows that do **not** run to completion:

```
casctanks   node_count  base=15         head=11
nvs05       bound       base=5.0715853  head=4.0460447
nvs05       node_count  base=109        head=105
tanksize    node_count  base=16651      head=16649
```

CLAUDE.md's bound-neutral regime requires `node_count` and the certified objective to be *exactly* unchanged, so none of this was written off as noise. Two independent measurements, not an argument:

**1. The changed code is never executed on these instances.** Patching `interval_eval._eval` to count node types during real solves:

```
nvs05        21375 interval evaluations, SumExpression = 0
casctanks    12460 interval evaluations, SumExpression = 0
tanksize      1473 interval evaluations, SumExpression = 0
```

35,308 evaluations, zero of them the changed branch. This also corrects a weaker claim in my own commit log: "0 of 66 `.nl` models contain a `SumExpression`" was measured on *parsed* models and said nothing about nodes a relaxation pass constructs mid-solve. The probe above is what actually closes that gap.

**2. Both arms produce both values.** Interleaved 3 base / 3 HEAD at the same 60 s budget:

```
            BASE reps                                HEAD reps
nvs05       105/4.0460  105/4.0460  113/5.0890       105/4.0460  113/5.0890  113/5.0890
casctanks   15 nodes    15 nodes    15 nodes         15 nodes    15 nodes    15 nodes
```

`nvs05` takes at least **four** distinct (nodes, bound) values within a single arm at this budget — the two above plus the panel's own base sample (109/5.0716) and the probe run's (111/4.317) — and the base arm reproduces the value the panel attributed to HEAD, bit-for-bit. `casctanks` is 15 nodes in all six controlled runs, so the panel's 11 is another sample of the same variance. `tanksize` was measured the same way earlier and is identical off the time boundary: at a 300 s budget, 3 base / 3 HEAD interleaved, every run gives `nodes=16649 obj=1.2686437535708868 bound=1.268544049560946`, with HEAD no slower (base 59.92 s sd 0.52, HEAD 59.59 s sd 0.06).

### Python side: where the change actually bites

`SumExpression` comes from the Python `dm.sum(...)` API; the `.nl` reader never emits one. `python/tests/test_interval_sum_reduction.py` covers that class directly — enclosure containment under cancellation against an exact `fractions.Fraction` reference, on both the fast and the **slow** reduction axis, the accumulation-error widening, the never-narrows ordering, and, on real solves of Python-API `dm.sum` models, that a dual bound never exceeds the true optimum.

### Measurement hygiene (CLAUDE.md §8), because it nearly invalidated this panel

`discopt/__init__.py` calls `pkgutil.extend_path`, and the editable install's `discopt.pth` keeps the main tree on `sys.path`. The baseline worktree arm therefore resolved `discopt.status` — a module that does not exist at `1e75919` — from *this branch*, while `core.py` came from the worktree: a mixed arm that would have been reported as a clean baseline. It was caught by asserting the branch marker **absent** in the baseline arm. A `sitecustomize.py` strips the main tree from `sys.path` before `discopt` is imported, and both arms print `discopt.__file__` plus the `mpec_report` marker on every run:

| arm | `discopt.__path__` | `discopt.status` | `SolveResult.mpec_report` |
|---|---|---|---|
| base (`1e75919`) | worktree only | absent ✓ | absent ✓ |
| `78bd898` | main tree | present ✓ | present ✓ |

Only the `discopt` library differed between arms; the harness and instance files were the same main-tree copies in both.

`main` moved during the PR and is merged in twice (`deaa3c5`, `ae9a69d`), no conflict. The panel A/B is deliberately **not** re-baselined onto the newer `main`: it compares this branch to its own merge base `1e75919`, which is what isolates *this* change. Re-baselining would fold the intervening merges and #1148 together and measure neither.

## Review rounds

Four review passes plus a verification closeout found 24 findings; all are addressed, each with a regression test and, where a blanket fix could have passed vacuously, a control. Nine were soundness issues, and five of those produced a value better than the true optimum:

| round | severity | what it was |
|---|---|---|
| 1 | HIGH | `interval_eval` never reduced `SumExpression`, so the residual read `0.0` on a row violated by `2.0` — and, being on the solve path, could produce invalid FBBT tightenings on any Python-API model using `dm.sum` |
| 1 | HIGH | `accept_local_incumbent` verified against the model *as mutated by the lowering*; `f·g <= t` is a **relaxation**, so it vouched for a point with source residual 1.4e-4 at an objective better than the true optimum |
| 1 | HIGH | the mandatory report destroyed a completed, certified solve when any row was not interval-walkable |
| 1 | MED | a Scholtes relaxation satisfied the solve guard — `model.solve()` afterwards returned `optimal`/`gap_certified=True` at 2.83e-4 against a true optimum of 0 |
| 1 | MED, LOW | exact-zero selector threshold; aggregate row list per pair (O(N²) and wrong attribution); notebook result shape; stage objective sign on maximize; `final_t` off by a schedule step; `converged` true on a failed run |
| 2 | HIGH | the aggregate residual mixed one relation's value with **its own** scale, so a relation 100× over tolerance hid behind another's denominator — a hole in the round-1 HIGH fix with the same shape as the original bug |
| 2 | MED | the new sum reduction dropped the module's outward-rounding invariant, on the path that feeds FBBT |
| 2 | MED | the per-stage residual call was unguarded, so an unwalkable operand aborted the homotopy on iteration 1 |
| 2 | MED | `_result_point` ran as an *argument expression*, evaluated before `_report_for` was entered — so its refusal bypassed the guard added for exactly that class |
| 2 | MED | only `OPTIMAL` stages accepted, discarding the late stalled-but-usable iterates a Scholtes homotopy actually produces |
| 3 | MED | one `nextafter` does not bound an *n*-term accumulation, so the round-2 outward-rounding fix still returned enclosures that did not contain the true sum (measured: 2289/3000, worst shortfall 1.5e-6) |
| 4 | **BLOCKING** | `accept_local_incumbent` gated on a **cached** report, which records neither the point nor the model it was taken at — so a clean report at `(0, 0)` authorized `(1e-4, 1e-4)`, source residual `1e-4`, objective `-2e-4` against a true optimum of `0`. Now recomputed at the boundary; the `report=` keyword is removed rather than left available |
| 4 | **BLOCKING** | the round-3 widening applied a **pairwise** error bound where numpy accumulates **sequentially**: pairwise precision holds only along the unit-stride axis. A C-contiguous `(10002, 2)` array of ones with `1e16` first and `-1e16` last has exact column sums of `10000`, and `sum(v, axis=0)` enclosed `[-67.89, 67.89]` — excluding the truth, on a certified-bound path. The bound is now chosen from the reduction order numpy actually uses, and the magnitude sum is rounded up before scaling |
| 4 | **BLOCKING** | an iteration-limited iterate was promoted to `local_optimal`. With the subsolver allowed zero iterations it does no optimization and returns its own starting point; the wrapper published that as a stationary point, with the generated product row violated by `24`. The merge base preserved `ITERATION_LIMIT`, so this was a regression from the round-1/2 status work. `local_limit` added; `local_optimal` reserved for a converged stage |
| 4 | NON-BLOCKING | `sqrt(t)` was described as a floor on attainable accuracy — it is an **upper bound on the admitted residual** (`(f, g) = (0, 1)` has residual `0` at every `t`) — and was copied into product and Fischer–Burmeister reports whose formulas admit `t` and `(2-sqrt(2))·sqrt(t)`. Renamed to `admitted_residual_scale`/`admitted_scale`/`within_admitted_scale`, derived per definition, wording corrected in the docstring, the semantics doc and the notebook |
| 4 | NON-BLOCKING | the changelog and this PR body still described #1160/#1161 as open. Corrected — see below |
| 4 closeout | NON-BLOCKING | `within_admitted_scale` was described as saying whether a point is "inside what the relaxation allows", which over-claims: it compares a residual against its upper bound and does **not** establish feasibility of the relaxed rows. At `t = 1e-8`, `(f, g) = (1e-4, 1)` sets the flag while `f·g` exceeds `t` by four orders of magnitude (reproduced). Only the converse holds — above the scale, both operands exceed `sqrt(t)`, so the row *is* violated. Doc, property docstring and notebook now say so and point at `lowered_row_residual`; `_stage_reason`'s leftover "which is what 'local_optimal' claims" now describes `local_limit` retaining an iterate without asserting stationarity |

## Follow-ups filed from this PR, and what remains

Of the three follow-ups this PR filed, all three have landed; one item genuinely remains open, and it is the `DISCOPT_STATUS_MAP` one below.

- **#1160** (the `dm.sum(X, axis=k)` false certificate this PR's panel work surfaced) and **#1161** (`_eval_matmul`'s one-ULP accumulation gap) were both filed from here with their measurements and carried as follow-ups. **Both have since landed and are merged into this head** — #1172 (`474cc18`) and #1171 (`dd6644d`). The #1160 strict xfail in `test_interval_sum_reduction.py` was converted to a passing test in `cfd381f`, and both interval reductions now go through the shared `_widen_sum`, so no one-ULP reduction remains in the module. Nothing is outstanding on either.
- `DISCOPT_STATUS_MAP` is held to the entries the panel runners already had, plus the three local statuses. Two real discopt statuses are still unmapped in `runner.py`/`_subprocess_worker.py` and fall through to `UNKNOWN`: `"unbounded"`, and `"error"` — the latter being the status `Model.solve`'s false-primal guard sets, which `classify_outcome` then tallies as `OUTCOME_UNKNOWN` rather than `OUTCOME_ERROR`. That is the wrong direction for a soundness signal and worth fixing, but widening the map moves rows between outcome buckets on every existing panel report, and folding that into this PR would have muddied the baseline diff it is supposed to leave untouched. The reasoning is recorded in a comment on the table.
- **#1182** carried the deferred simplex/CNF lowering from @bernalde's design note on #1148, with the four requirements he stated. **It is done**: PR #1184 implements the Wehbeh–Kerrigan Theorem-1 lowering as an opt-in `gdp_method="simplex"` and merged into `main` at `a7c1d95`; #1182 was closed by it. Nothing is deferred from this PR on that account. One cross-reference worth keeping straight: #1184 records that requirement 2 of #1182 (reuse `discopt.status`) had no surface there, because its lowering is exact and introduces no local-only result — that requirement becomes live only if a local-NLP arm is ever built on it, and the vocabulary this PR adds is what it would use.

https://claude.ai/code/session_01X83REmDmGJghCMhgfLMizJ


jkitchin:
## Code review — PR #1158

Reviewed `origin/main...origin/claude/discopt-issue-1147-bme983`. Findings 1–6 were reproduced in a worktree at `331dde83`, marker-checked via `mpec_report.__file__` so the branch code was the code that ran.

Three of these are blocking: the residual instrument reads clean on infeasible points for a very common expression shape, and two paths produce a "verified"/"certified" objective strictly better than the true optimum.

---

### 1. HIGH — `python/discopt/mpec_report.py:428` — `evaluate_at_point` returns a 0 residual for any row containing `sum(...)`

`evaluate_interval` does **not** reduce `SumExpression`: `interval_eval._eval_impl` returns `_eval(expr.operand, ...)` for `dm.sum(...)`, i.e. the *elementwise* operand interval. The result is degenerate, so the new `_POINT_EVAL_REL_WIDTH` / non-finite guards never fire and the residual is silently wrong.

```
model: x in [0,10]^2, row "budget": sum(x) <= 10 ; point x = (6,6)
mr._primal_feasibility(...).value -> 0.0     # true violation is 2.0
```

`sum` is ubiquitous in real models, so `source_primal_feasibility` reads 0.0 on genuinely infeasible points — the "instrument that measured nothing and was believed" class this module's docstring claims to prevent. The same root cause makes a legitimately-scalar relation `sum(x) ⊥ y` raise `expected a scalar operand` in `_scalar`.

### 2. HIGH — `python/discopt/mpec_report.py:1003` — `accept_local_incumbent` verifies against the lowered relaxation

It calls `verify_point(model, x)` on the model **as mutated by the lowering that produced the point**. On the Scholtes arm the model no longer contains the complementarity condition — only `f>=0, g>=0, f·g <= t` — so verification is against the relaxation, not the source relation, and the source complementarity residual is never consulted. Reproduced on the PR's own regression model (`max x+y s.t. x==y, 0<=x⊥y>=0`):

```
status local_optimal   obj 2.828e-04
source_complementarity 1.414e-04
source_satisfied       False
accept_local_incumbent -> 2.828e-04     # vouched, for a maximize whose true optimum is 0
```

It returns a "verified" objective better than the true global optimum for a point its own report marks unsatisfied. Fed to a global solve as an incumbent/cutoff this cuts off the optimum. It must gate on the source report (`source_satisfied` / `complementarity.scaled_value`) and verify against the pre-lowering snapshot.

### 3. HIGH — `python/discopt/mpec.py:1233` — the mandatory report destroys a completed solve

`_report_for` is called unconditionally from `solve_mpec` with no opt-out, and `evaluate_at_point` raises on any atom `interval_eval` does not support (variable exponent `u**v`, unsupported unary/binary ops) — including rows with nothing to do with the relations, since `_primal_feasibility` walks every source row:

```
m: u,v in [1,3], row "pow": u**v <= 8, plus a complementarity pair
solve_mpec(m, [complementarity(x,y)], method="gdp", time_limit=20)
-> ValueError: point evaluation of ((u ** v) - 8) is not finite (-inf .. inf)
```

Before this PR that call returned `model.solve()`'s result. The refusal is the right call for the *residual* — but it must not throw away the solve. Record it in `notes` / leave `mpec_report=None`, or give `solve_mpec` a way to skip the report.

### 4. MEDIUM — `python/discopt/mpec.py:1391` (+ `1051`) — a Scholtes relaxation marked as lowered defeats the #1147 guard

`solve_mpec(method="scholtes")` records the pair on `model._complementarities` and marks it lowered with method `"scholtes"`. `require_all_relations_lowered` — whose whole purpose is to catch "a declared relation solved as if absent" — accepts any recorded lowering, but Scholtes is a *relaxation*, not an encoding:

```
solve_mpec(m, [complementarity(x,y)], method="scholtes")   # mutates m
m.solve(time_limit=30)
-> status='optimal'  objective=2.828e-04  gap_certified=True   (true optimum 0, maximize)
```

A certified-optimal result strictly better than the true optimum, straight past the guard. Either the `scholtes` mark should not satisfy `require_all_relations_lowered`, or `solve_mpec` should not leave a relaxed relation marked lowered on the user's model.

### 5. MEDIUM — `python/discopt/mpec_report.py:981` — `_integrality_residual` uses an exact-zero threshold

`if out_of_box > 0.0:` on floating-point solver output. A binary returned as `-1e-15` (routine LP/MILP roundoff) makes the residual `+inf` and `source_satisfied` `False`. Reproduced: `_integrality_residual` with `b = -1e-15` returns `inf`. Compare against `_INTEGRALITY_TOL` (already defined in the module, 1e-5).

### 6. MEDIUM — `python/discopt/mpec.py:1027` — `_RowRecorder.mark` assigns every pair the aggregate row list

`rows = list(self._model._constraints[self._start:])` is the aggregate of *all* rows emitted for *all* pairs in the call, and the same list is assigned to every pair. So `Complementarity.rows_in(model)` — documented as "the rows **this relation** generated in `model`" — returns other relations' rows, and `generated_rows()` then extends that list once per pair, producing `3N²` rows for `N` scalar relations. For an `.nl` MCP or a vector relation with hundreds of elements the lowered-row residual walk goes quadratic (N=1000 → ~3M row evaluations instead of 3000). Record per-pair slices.

### 7. MEDIUM — `docs/notebooks/complementarity_mpec.ipynb:190` — Scholtes cell uses the pre-PR result shape

```python
print("x, y =", np.round(res.x, 6))
print("x*y =", float(res.x[0] * res.x[1]))
```

`solve_mpec` now returns a `SolveResult` whose `x` is a `{name: value}` dict, so this raises (`np.round` on a dict / `KeyError: 0`). `test_mpec.py` was updated for exactly this change; the notebook — the user-facing doc for this API — was not, and its stored outputs are now stale. `execute_notebooks: "off"`, so the docs build won't catch it.

### 8. LOW — `python/discopt/mpec.py:1541`/`1566` — stage objectives sign-disagree with the result on a maximize

`ContinuationStage.objective` records `stage_result.objective` raw, while the returned `SolveResult.objective` is negated back to model units for a `MAXIMIZE` model (~line 1600). On a maximize MPEC, `stages[-1].objective` and `result.objective` differ by a sign — in the trace schema this PR proposes to share across repos.

### 9. LOW — `python/discopt/mpec.py:1578` — `final_t` is one schedule step past the last `t` solved

When no stage was accepted and the loop exits by `max_iter`, `final_t` is `float(tv)` — but `tv` was already advanced by `tv = max(tv * sigma, t_min)` at the end of the last iteration. Separately, `ContinuationTrace.converged` (`termination_reason == "t_min_reached"`) can be `True` on a `local_infeasible` result where nothing converged.

### 10. LOW — `python/discopt/modeling/core.py:2365` — `mpec_report` docstring contradicts the behavior

The docstring says "``None`` for a solve with no complementarity relations", but `_report_for` returns `None` only when the *point* is `None`. `solve_mpec` on a model with zero relations produces a `SourceResidualReport` with `n_scalar_relations=0` and all residuals `0.0` — the all-zeros "clean pass" that `relation_residuals` deliberately refuses to emit when `pairs` is non-empty. Either return `None` for zero relations or fix the doc.

---

jkitchin:
## Code review — PR #1158, second pass (on `1bb92a05`)

Re-reviewed `1e759198...origin/claude/discopt-issue-1147-bme983` after the fix commit.

**All 10 findings from the first pass are genuinely addressed**, each with a regression test and, where a blanket fix could have passed vacuously, a control. The `SumExpression` root fix in particular is the right call: `dag_compiler.py:344` lowers the node to `jnp.sum(a[0], axis=axis)`, so it really is a reduction and the old pass-through returned an enclosure that does not contain the value.

Second pass turned up five new issues, three of them inside the fixes themselves.

---

### 1. HIGH — `python/discopt/mpec_report.py:745` — the aggregate complementarity residual mixes one relation's value with its own scale, so `source_satisfied` can pass with a badly violated relation

`worst_c = max(rows, key=lambda r: r.complementarity.value)` picks by **raw** value; line 759 then sets `scale=worst_c.complementarity.scale`, and `source_satisfied` (line 374) tests `self.complementarity.scaled_value <= 1e-6`. But `effective_scale` is genuinely per-relation — a declared `scale=`, else the largest finite declared-bound magnitude, so a box MCP on `z ∈ [0, 1e3]` gets 1e3 while an NCP pair gets 1.0. Concretely:

```
relation A (box MCP, scale 1e3):  residual 1e-3   -> scaled 1e-6
relation B (NCP pair, scale 1.0): residual 1e-4   -> scaled 1e-4

worst_c picks A (1e-3 > 1e-4)
aggregate scaled_value = 1e-3 / 1e3 = 1e-6 <= tol  ->  source_satisfied = True
```

B's scaled residual is a hundred times the tolerance and the report reads clean. `accept_local_incumbent` gates only on this aggregate, so this is a hole in HIGH 2's fix with the same shape as the original bug. The aggregate should be the max over `scaled_value` (or `source_satisfied` should test every row's `scaled_value`) — which is what this module's own docstring says the scaled number is for. No test covers mixed scales; `test_scale_makes_the_tolerance_meaningful` uses a single relation.

### 2. MEDIUM — `python/discopt/_relax/convexity/interval_eval.py:176` — the new reduction drops the module's outward-rounding invariant, on a path that feeds FBBT

`interval.py`'s docstring states soundness is achieved by "outward rounding after every arithmetic operation", and the other reduction in this same file, `_eval_matmul` (line 337), ends with `Interval(np.nextafter(lo, -np.inf), np.nextafter(hi, np.inf))`. The new code returns `Interval(np.sum(inner.lo, axis=...), np.sum(inner.hi, axis=...))` with neither `_round_down` nor `_round_up`. Summing *n* terms accumulates up to ~*n* ULP of error in each endpoint, so the returned enclosure can be strictly narrower than the true image.

This matters because `evaluate_interval` is not confined to the residual instrument — it is on the solve path (`_relax/nonlinear_bound_tightening.py:2916`, `_relax/uniform_relax.py:1247`, `_relax/oa_relax.py:178`, `_relax/convexity/g_convex_inject.py`, `solver.py:2650`), where a too-narrow enclosure is exactly the FBBT failure mode CLAUDE.md's `INF`-sentinel note describes: a tightening that cuts the optimum out of the box. The `nextafter` pair fixes it.

**Related scoping note.** The commit argues "0 of 66 in-repo `.nl` models contain a `SumExpression` node ... so no `.nl` panel result can move". That is right and it rules out the `.nl` panel — but `SumExpression` comes from the Python `dm.sum(...)` API, so **Python-API models are precisely the exposed class**, and the argument says nothing about them. Two consequences worth recording in this PR: (a) the pre-fix bug could produce **invalid bounds on real solves** of any Python-API model using `dm.sum`, not merely a wrong residual number — that is a pre-existing correctness issue deserving its own issue and a CHANGELOG line; and (b) under CLAUDE.md §5 this is a bound-changing change whose differential evidence should come from a Python-side panel, since the `.nl` panel cannot exercise it.

### 3. MEDIUM — `python/discopt/mpec.py:1657` — the per-stage residual call is unguarded, so a refusal still aborts the Scholtes solve

`_report_for` (~line 1310) deliberately catches `Exception`, warns, and returns `None`, because "the refusal must not take the solve with it". But inside the homotopy loop, `stage_res = max_source_complementarity(model, report_pairs, point_from_flat(model, x_cur), ...)` runs bare. That path is `max_source_complementarity → relation_residuals → _scalar → evaluate_at_point`, which raises `ValueError` for any operand the interval evaluator cannot walk. So a relation containing such an atom makes `solve_mpec(method="scholtes")` raise on the first iteration instead of returning a `SolveResult` — strictly worse than pre-PR behaviour, and inconsistent with `test_review_3_a_failed_report_does_not_destroy_the_solve`, which covers only the `gdp` arm.

### 4. MEDIUM — `python/discopt/mpec.py:1517` — `_result_point` is evaluated outside `_report_for`'s try/except

`_result_point` raises `ValueError("solve result does not carry variable ...")` by design when `result.x` is missing one of `model._variables`. Because the call is an **argument expression**, it is evaluated before `_report_for` is entered, so that function's `except Exception` never sees it. On the `sos1`/`gdp` arm this turns a finished `Model.solve` — status `optimal`, certified bound in hand — into an exception out of `solve_mpec`: precisely the HIGH-3 class this commit claims to have closed. Same at line 1427 in the public `source_residuals()` helper. Move the `_result_point` call inside the guarded block (or into `_report_for`).

### 5. MEDIUM — `python/discopt/mpec.py:1644` — accepting only `OPTIMAL` stages discards good iterates, on exactly the ill-conditioning Scholtes produces

`nlp_ipopt._IPOPT_STATUS_MAP` maps Ipopt code 3 (`Search_Direction_Becomes_Too_Small`) to `ITERATION_LIMIT`, and this repo's own comment there says code 3 means the IPM stalled "at/near a local solution it cannot certify to tolerance". Scholtes subproblems become increasingly degenerate as `t → 0` (MFCQ fails in the limit), so late stages hitting code 3 is the expected case. Previously the loop returned the last result with a point regardless of status; now those stages only warm-start. Two consequences:

- if every stage is non-`OPTIMAL`, `solve_mpec` returns `local_infeasible` with `x=None`, `objective=None`, discarding a usable point;
- more commonly, `t=1, 1e-1` converge and `t=1e-2 … 1e-8` return code 3, so the reported point is the `t=1e-1` one with source complementarity ~1e-1 instead of the ~1e-4 point the old code returned — while `ContinuationTrace.converged` still reads `True` and `final_t` is 1e-1.

Consider accepting `ITERATION_LIMIT` stages that carry a point (recording the distinction in `reason`), or at minimum making `converged` false when `best_t > t_min`.

---

### Checked and cleared

- `result_io.py:60` — `serialize_result` uses an explicit field list and does not carry `mpec_report`, so the "shared benchmark schema" (`SourceResidualReport.as_dict()`) is dropped by `write_json` / the daemon round-trip. Safe, but the schema is not actually wired to any serializer yet.
- The shared `DISCOPT_STATUS_MAP` refactor (`metrics.py`, `runner.py`, `_subprocess_worker.py`, `category_runner.py`, `gdplib_runner.py`) is behaviour-preserving: each runner's pre-existing map reproduces exactly in the merged one; `SolveStatus.LOCAL` is skipped by `is_solved` and `proved_optimal_count`, and `dashboard._STATUS_TO_OUTCOME` / `reporting.py` fall through to `OUTCOME_UNKNOWN` — fail-closed as documented.
- `_model_units` is correct: `_tape_nlp_evaluator` negates for `MAXIMIZE` (lines 217/272), so the back-conversion holds. The PR body's `-2.8e-4` is consistent with the notebook's `minimize(-(x+y))`, not a sign bug.
- The other `SumExpression` pass-throughs (`milp_relaxation.py:1389/1478`, `least_squares.py:226`, `obbt.py:501`) are structural pattern-matchers — term collectors, sum-of-squares detectors, coefficient extractors — where descending into the operand is legitimate because sum is linear. `interval_eval` was the odd one out precisely because it returns a value enclosure. No sibling bug there.

jkitchin:
@bernalde this is converging to done. You may be interested to take a look at it as it is connected to the work you proposed.

jkitchin:
Second review pass addressed in `8829b13`; CI green on that head. All five reproduced before fixing, each with a regression test and a control where a blanket fix could have passed vacuously. Counted comparisons 145 → 165, plus a new 13-comparison interval suite. PR body updated with the current numbers.

Three of the five were inside the round-1 fixes, and one deserves calling out: **HIGH 1 was a hole in my own HIGH-2 fix with the same shape as the original bug.** The aggregate residual ranked by *raw* value and then reported that relation's scale, so a box MCP violated by 1e-3 (scaled 1e-6, fine) outranked an NCP pair violated by 1e-4 (scaled 1e-4, 100× over) and `source_satisfied` came back True — with `accept_local_incumbent` gating on exactly that. It now ranks by scaled value and quantifies over every relation, not the single aggregate row.

**On the panel differences.** The re-run A/B against `1e75919` shows 4 differences, including an `nvs05` bound move (5.0716 → 4.0460). None of them is this PR. Two measurements rather than an argument:

1. Patching `interval_eval._eval` to count node types during real solves: **35,308 interval evaluations across `nvs05`, `casctanks` and `tanksize`, zero of them `SumExpression`** — the changed branch is never executed on these instances.
2. Interleaved 3 base / 3 HEAD at the same 60 s budget: `nvs05` takes at least four distinct (nodes, bound) values *within a single arm*, and the **base arm reproduces the value the panel attributed to HEAD, bit-for-bit** (105 / 4.046044722519642). `casctanks` is 15 nodes in all six controlled runs, so the panel's 11 is another sample.

That also corrects a weaker claim in my earlier commit log: "0 of 66 `.nl` models contain a `SumExpression`" was measured on *parsed* models and said nothing about nodes a relaxation pass constructs mid-solve. The instrumented probe is what actually closes that gap, and I have replaced the claim in the PR body.

**One thing found while building the Python-side panel the review asked for, filed separately as #1160 and not fixed here:** a constraint on `dm.sum(X, axis=k)` is solved as the axis-*collapsed* model and certified there — true optimum −4, certified −2 with a bound of −2, while discopt's own `verify_point` confirms the −4 point feasible. Measured identically on `1e75919` and on this branch, so neither caused nor fixed by this PR. The NLP evaluator is correct (2 rows, feasible at the true optimum), so it is the relaxation path; the issue names the four `SumExpression` call sites to check. It is carried here as an `xfail(strict=True)` so it flips to a failure the moment #1160 lands rather than rotting.

---

jkitchin:
## Code review — PR #1158, third pass (on `8829b135`)

Verified in an isolated worktree at `8829b135`, asserting `mpec.py` / `mpec_report.py` / `interval_eval.py` all resolved to the worktree and checking a marker unique to each fix (12 executed assertions, printed).

**Four of the five second-pass findings are properly fixed. The fifth is fixed in direction but not in magnitude**, and I can now quantify the shortfall.

### Fixed

| finding | status |
|---|---|
| HIGH 1 — aggregate mixed one relation's value with its own scale | Fixed twice over: `worst_c` ranks by `complementarity.scaled_value`, **and** `source_satisfied` quantifies over every relation rather than trusting the aggregate. Either alone would have closed it. |
| MEDIUM 3 — unguarded per-stage residual call | Fixed. Wrapped, records `residual_failures`, sets `stage_res = None` ("not measured") rather than swallowing — consistent with `_report_for` and with §7. |
| MEDIUM 4 — `_result_point` evaluated outside the guard | Fixed. `_report_for` takes `point_fn` and resolves it *inside* the `try`; both call sites pass `lambda: _result_point(model, result)`. |
| MEDIUM 5 — only `OPTIMAL` stages accepted | Fixed. `_USABLE_STAGE_STATUSES = (OPTIMAL, ITERATION_LIMIT)` with `INFEASIBLE`/`ERROR` deliberately excluded, and `certified_stage` kept distinct from `usable` so the trace does not flatten the difference either way. |

Tests on the branch: `test_mpec_source_residuals.py` + `test_interval_sum_reduction.py` + `test_mpec.py` — 66 passed, 1 xfailed. `pytest -m smoke` — **1216 passed, 1 skipped, 1 xfailed, 2 xpassed** (exit 0). Nothing regressed.

---

### 1. MEDIUM — `python/discopt/_relax/convexity/interval_eval.py:185` — one `nextafter` does not bound the error of an *n*-term sum

The fix applies a single `np.nextafter` to each endpoint, matching `_eval_matmul`'s convention, and the comment says this is "outward rounding, like every other arithmetic result in this module". For a binary operation that is true. For a reduction over *n* terms it is not: floating-point accumulation costs far more than one ULP, so the returned enclosure can still be strictly narrower than the true image — the same too-narrow-enclosure hazard the fix was written to close.

Measured on the branch against an exact `Fraction` reference, 3000 random sums (`n ∈ [4, 600]`, heavy cancellation, magnitudes ~1e8):

```
enclosure MISSED the truth : 1940 / 3000
worst shortfall (abs)      : 1.83e-06

example n=535
  float sum          = 358390632.63480103
  exact sum          = 358390632.6347991
  returned enclosure = (358390632.634801, 358390632.6348011)   <- exact is BELOW lo
```

A correct bound widens by the accumulation error rather than by one ULP — for numpy's pairwise summation, `(log2(n)+1)·eps·Σ|x_i|`. Note `_relax/convexity/eigenvalue.py:56` already takes the rigorous route for its own accumulation ("pushing every partial one ULP toward +∞ — O(n²) scalar `np.nextafter`"), so the repo has the pattern.

To be clear on scope and attribution: this is **strictly better than before the fix**, and the practical FBBT exposure is small (for bounds ~1e3 and n~100 the shortfall is ~1e-11). `_eval_matmul` at line 321 has the identical gap on its dot products and **predates this PR** — worth its own issue rather than loading it onto this one. What is in scope here is that the new comment asserts an invariant the new code does not establish.

---

### Method note

My first two probes of this reported "0 misses" because they used `np.longdouble` as the exact reference — which is float64 on arm64, so the comparison was `fl == fl`, a probe that measured nothing (CLAUDE.md §6, on myself). The numbers above are from the `Fraction` rerun, which is exact.

jkitchin:
Third-pass finding fixed in `78bd898`. You're right, and it reproduces: one `nextafter` is the correct convention for a binary op but does not bound an *n*-term accumulation, and the comment I added asserted an invariant the code did not establish.

Reproduced independently against an exact `Fraction` reference (3000 sums, n ∈ [4, 600], cancellation at ~1e8):

```
one-ULP widening   : enclosure MISSED the truth on 2289 / 3000, worst shortfall 1.523e-06
accumulation bound : 0 / 3000
```

The widening is now `(log2(n) + 2)·eps·Σ|x_i|` — the pairwise-summation forward-error bound, `+2` rather than `+1` to cover the rounding of the bound's own computation plus slack — followed by the existing `nextafter`. As you noted, `eigenvalue.py:56` already takes the rigorous route, so this is the repo's pattern.

The regression test asserts both directions at once, because they pull against each other: the enclosure must contain the exact sum under cancellation, **and** a degenerate point-box evaluation must stay tight enough for `mpec_report.evaluate_at_point` to still accept it (checked at n = 10, 1000, 20000, where the `log2(n)` term is largest). It uses `Fraction`, not `np.longdouble` — thank you for the method note; that failure mode would have produced a green test that measured nothing, and I'd have shipped it.

**`_eval_matmul`**: confirmed, same gap, and I've taken your scoping. Measured 190/400 random `(1,k) @ (k,)` products not containing the truth, worst shortfall 8.8e-07. Filed as #1161 with the measurement and the exact patch rather than folded in here, since widening matmul enclosures is a bound-affecting change on the solve path that deserves its own differential evidence. Reachability is the same structural blind spot as the sum case: **0 of 66** in-repo `.nl` models contain a `MatMulExpression`, so no `.nl` panel covers that class either — worth knowing when #1161 is scheduled.

Also on this head: `main` had moved to `27aeb38` (picking up #1152, which changes root-setup deadline behaviour on `casctanks`/`tanksize` — the instances whose variance this PR's panel documents), so main is merged in as `03a21c2`. No conflict. The panel A/B is deliberately **not** re-baselined onto `27aeb38`: it compares this branch against its own merge base `1e75919`, which is what isolates this change; re-baselining would fold #1152 and #1148 together.

Post-merge: targeted suites 125 passed / 1 xfailed, `pytest -m smoke` 1170 passed, ruff and mypy clean. Smoke is re-running on `78bd898` now.

---

jkitchin:
Updated for the two PRs that landed on `main` since this branch was last pushed. Still holding this PR for @bernalde — no merge intended.

**1. The `#1160` xfail is flipped.** This PR carried

```
@pytest.mark.xfail(strict=True, reason="jkitchin/discopt#1160: ... strict=True so
this flips to a failure the moment #1160 is fixed.")
def test_axis_reduced_constraint_bound_is_sound():
```

#1172 fixed #1160, so that xfail now XPASSes and `strict=True` turns it into a CI failure — which is exactly what it was written to do. It is now the plain assertion it was always meant to become, routed through `_checked` so it counts toward the file's executed-comparison total (CLAUDE.md §6) rather than sitting outside it.

Verified two-sided: the assertion **fails** against a tree without #1160 and **passes** here. The first attempt at that check was a CLAUDE.md §8 failure worth recording — I ran it from this worktree, which resolves `discopt` through an editable install pointing at the *main* checkout, and that checkout was sitting on a branch predating #1160. It reported a failure that had nothing to do with this branch. Every measurement since asserts `discopt.__file__`, `discopt._rust.__file__` and the `scalarize.sum_is_full_reduction` marker before running.

**2. One real conflict `git` did not flag.** #1171 landed `_accumulation_factor` / `_widen_sum` in `interval_eval.py`, written there explicitly so this branch's `SumExpression` reduction could share them "so the two reductions cannot drift" — its docstring already names the axis case as the second caller. Git merged both sides textually and left the branch carrying its own inline copy of the same arithmetic, which is precisely the drift the shared helper exists to prevent. The reduction now goes through the helper and the copy is gone.

Not a no-op in one direction: the inline copy had **no non-finite guard**, so an infinite `|x_i|` computed `inf - inf = nan` and silently destroyed the enclosure instead of widening it to infinity. Sharing the helper fixes that. `_widen_sum`'s `axis` annotation widens from `Optional[int]` to `Union[int, tuple[int, ...], None]`, since `SumExpression.axis` may be a tuple; `np.sum` already accepted one.

**Verification** (worktree code asserted loaded, as above):

- `pytest -m smoke python/tests` → **1233 passed, 1 skipped, 2 xpassed** (134 s)
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **19 passed** (145 s)
- `cargo test -p discopt-core` → **exit 0**
- `test_interval_sum_reduction` + `test_interval_matmul_reduction` + `test_convexity_interval` + `test_interval_ad_reverse` + `test_interval_node_bound` → **127 passed**
- `ruff check` / `ruff format --check` / `mypy` clean on the changed module

Nothing about the substance under review changed — this is merge upkeep plus the de-duplication #1171 asked for.

python/discopt/mpec_report.py
```diff
@@ -0,0 +1,1133 @@
...
+
+    if x_flat is None:
+        x_flat = _flat_from_result(model, result)
+    if x_flat is None:
+        return None
+    x_flat = np.asarray(x_flat, dtype=np.float64)
+
+    if report is None:
+        report = getattr(result, "mpec_report", None)
+    if report is None and getattr(model, "_complementarities", None):
```
bernalde: Blocking: Revalidate the candidate point instead of trusting a cached source report. `x_flat` can override `result.x`, but a non-None `result.mpec_report` prevents recomputing any source relations. On `min -x-y`, `x == y`, `0 <= x,y <= 1`, `x ⊥ y`, with Scholtes rows at `t=1e-8`, a valid report at `(0,0)` followed by `accept_local_incumbent(model, result, x_flat=[1e-4,1e-4])` returns `-0.0002`. A fresh report rejects that point with source residual `0.0001`; the true optimum is `0`. This can seed an invalid incumbent cutoff even though the generated rows pass. Recompute source feasibility for the actual candidate and all current source relations at this acceptance boundary; a cached diagnostic must not authorize a different point/model. Add the stale-report regression and retain the valid `(0,0)` control.

python/discopt/_relax/convexity/interval_eval.py
```diff
@@ -161,7 +180,49 @@ def _eval_impl(expr: Expression, model: Model, box: dict, cache: dict) -> Interv
...
+        # produced ``inf - inf = nan`` and silently destroyed the enclosure.
+        #
+        # Widening is always the SAFE direction: an enclosure may be loose and
+        # remain sound, but must never be narrow.
+        inner = _eval(expr.operand, model, box, cache)
+        lo_arr = np.asarray(inner.lo, dtype=np.float64)
+        hi_arr = np.asarray(inner.hi, dtype=np.float64)
+        n_terms = _reduced_count(lo_arr, expr.axis)
+        lo = _widen_sum(np.sum(lo_arr, axis=expr.axis), lo_arr, n_terms, expr.axis, -1.0)
+        hi = _widen_sum(np.sum(hi_arr, axis=expr.axis), hi_arr, n_terms, expr.axis, +1.0)
```
bernalde, line 224: Blocking: The new axis reduction applies a pairwise-summation error bound where NumPy can accumulate sequentially. For a C-contiguous `(10002, 2)` float64 array filled with ones, set the first row to `1e16` and the last to `-1e16`. Each exact column sum is `10000`, but `evaluate_interval(dm.sum(v, axis=0), model, {v: Interval(a,a)})` returns approximately `[-67.8924, 67.8924]` in each column, excluding the truth. The contiguous fast-axis transpose control encloses it. [NumPy's sum documentation](https://numpy.org/doc/stable/reference/generated/numpy.sum.html#notes) states that pairwise precision with an explicit axis depends on the memory axis. Use an error enclosure valid for the actual reduction order, including rounding of its magnitude sum, and add the slow-axis cancellation fixture. The base's unreduced SumExpression was already wrong; this finding is that its replacement still fails the claimed enclosure invariant on a path used for certified bounds.

python/discopt/mpec.py
```diff
@@ -1156,22 +1635,209 @@ def solve_mpec(
...
+        )
+
+    from discopt.solver import _unpack_solution
+
+    # ``best_obj`` is already in model units: every stage objective is converted
+    # as it is recorded, so the trace and the result cannot disagree.
+    objective = best_obj
+
+    return SolveResult(
+        status=_status.LOCAL_OPTIMAL,
```
bernalde: Blocking: An iteration-limited point is unconditionally promoted to `local_optimal`, which the new contract defines as a local stationary point. With the committed distance model, `x0=[5,5]`, outer `max_iter=1`, and `nlp_options={'max_iter': 0}`, the backend does no optimization and returns `ITERATION_LIMIT`, objective `32`, and `(5,5)`. This wrapper returns `local_optimal` although the generated product row is violated by `24`. The same call at merge base `1b80dac2` preserves `SolveStatus.ITERATION_LIMIT`. Retain useful stalled iterates and warm starts, but return a limit/nonconverged status with no bound or certification when stationarity was not established. Reserve `local_optimal` for a qualifying converged stage, and make `ContinuationTrace.converged` distinguish acceptance of an iterate from convergence. Add this zero-iteration regression; the source-string assertion in `test_review2_5_a_stalled_but_usable_stage_is_taken_forward` does not test that distinction.

docs/complementarity_semantics.md
```diff
@@ -172,3 +172,159 @@ every lowering funnels through (`_scalarize_pairs`) — not at a call site. The
...
+Which definition is used is chosen from the relation's **declared bounds**, never
+from its `role` — the same rule §3 states for the lowerings, and for the same
+reason.
+
+### Scale, and the accuracy floor
+
+A uniform tolerance is meaningless when the two operands carry unrelated physical
+magnitudes (a multiplier in 1e-6 against a flow in 1e3), so a residual is reported
+both raw and divided by the relation's `effective_scale` (§3). A Scholtes
+regularization at `t` cannot attain a complementarity better than `sqrt(t)` —
```
bernalde: Nonblocking: `sqrt(t)` is not a lower limit on attainable complementarity. For nonnegative operands satisfying `f*g <= t`, it is an upper bound on the admitted `min(f,g)` residual; `(f,g)=(0,1)` still has zero residual at every positive `t`. The current wording says better accuracy is impossible, and `relation_residuals` also copies this number into product/Fischer–Burmeister reports despite their different formulas. Describe it as the worst-case residual scale admitted by the relaxation, separate empirical solver tolerance effects, and make any reported reference scale depend on the selected residual definition. Update the `Residual.floor`/`at_floor` documentation and notebook wording consistently.

CHANGELOG.md
```diff
@@ -10,8 +10,89 @@ The release procedure that produces these entries is documented in
...
+  an enclosure that did **not** contain the true sum on 2289 of them, worst
+  shortfall 1.5e-6; with the accumulation bound, 0 of 3000.
+
+  **Who was exposed:** models built through the Python modeling API that use
+  `dm.sum(...)`. The `.nl` reader emits no `SumExpression` at all (measured: 0 of
+  the 66 files in `python/tests/data/minlplib_nl/`), so no `.nl` instance, panel
+  or gate could exercise this — which is why it survived. Regression coverage is
+  `python/tests/test_interval_sum_reduction.py`, including a differential check
+  that a dual bound never exceeds the true optimum on Python-API `dm.sum` models.
+  The sibling reduction in `_eval_matmul` has the same one-ULP gap over its
```
bernalde: Nonblocking: This new changelog entry still says `_eval_matmul` has the one-ULP defect and tracks it as future #1161 work, although #1171 is merged into this head and both reductions now use `_widen_sum`. The PR body's current-head narrative likewise still describes #1160 as unfixed and its regression as a strict xfail; #1172 landed and `cfd381f6` converted it to a passing test. The latest coordination comment correctly explains both updates. Bring the changelog and PR body's current-state claims into agreement with that comment, retaining old test/panel results only as explicitly identified historical snapshots.

bernalde (changes requested):
Three blocking issues remain at `f9f4a2bc`: incumbent acceptance can trust residuals from a different point, the new axis-sum enclosure can exclude the exact sum, and a stage that performs no optimization can still produce `local_optimal`. The inline comments give runnable counterexamples and the required corrections. The earlier fixes for #1160 and #1161 are present; the axis finding exercises a remaining slow-memory-axis case.

I added the Wehbeh–Kerrigan paper and concrete follow-up requirements to [#1148](https://github.com/jkitchin/discopt/issues/1148#issuecomment-5554603766), cross-referencing the existing #1123 design discussion. Exact simplex/CNF projection remains a deferred lowering; it is not an added acceptance criterion for this residual/reporting PR.

Validation used Python 3.12.13, NumPy 2.5.1, a native extension rebuilt from this head, and CI-matched POUNCE 0.11.0 at `a82c7355`. Source and native import paths were checked explicitly.

- Six focused suites: **128 passed** at the PR head and **128 passed** on its clean merge with `main` at `e006864e`.
- Adversarial recent-fixes suite: **19 passed**. All **9 notebook code cells** executed successfully.
- Independent counterexample/control probes: **6 assertions executed**, reproducing all three blocking findings. The zero-iteration call preserves `ITERATION_LIMIT` at merge base `1b80dac2`.
- Smoke suite: **1,231 passed, 2 failed, 8 skipped, 2 XPASS**. Both failures were subprocesses importing the primary checkout's older installation; both also failed on the base branch and **both passed** when rerun with corrected `PYTHONPATH`. No other smoke failures occurred.
- `git diff --check` passed. The current-head GitHub checks that ran all succeeded; optional lanes are intentionally skipped. The full local test suite and NL benchmark panel were not rerun.

Findings: **3 Blocking**, **2 Nonblocking**, **0 Questions** — 5 inline comments, 0 body-only findings, 5 total.

I would not merge this until the blocking issues above are addressed.

python/discopt/mpec_report.py
jkitchin: Fixed in `1347c2b`. Reproduced first: your construction returns `-0.0002` on the old code with the carried `(0,0)` report clean and a fresh report at `(1e-4, 1e-4)` reading `1.000e-04`.

`accept_local_incumbent` now recomputes the source-residual report at the acceptance boundary against the actual candidate and `model._complementarities` as they stand, and the `report=` keyword is removed rather than left as a way to supply one — no caller passed it, and keeping it would have preserved exactly the hazard.

You framed this as "a cached diagnostic must not authorize a different point/model", and the model half turned out to matter independently: binding a report to the *point* alone would still let a report taken before a relation was declared vouch for a model that now carries it. Recomputing closes both, so there are two regressions rather than one — `test_review3_1_a_carried_report_cannot_vouch_for_a_different_point` and `test_review3_1_a_stale_report_cannot_vouch_across_a_model_change`. Both fail on the parent commit and pass here; the valid `(0,0)` control is retained inside the first (the point the report *was* taken at is still accepted, so the refusal is a property of the measurement and not a blanket rejection).

---

python/discopt/_relax/convexity/interval_eval.py
jkitchin: Fixed in `1347c2b`. Your counterexample reproduced to the digit before the fix — `[-67.89236235297292, 67.89236235297292]` against an exact column sum of `10000` — and the transposed fast-axis control enclosed it, exactly as you described.

`_accumulation_factor` now takes the reduction order rather than assuming pairwise. `_reduction_is_pairwise` inspects the array actually being reduced: `axis=None` is pairwise only if the buffer is C- or F-contiguous, an integer axis only if `strides[axis] == itemsize`, and a tuple axis never (numpy does not specify the order, so it takes the sequential bound). It answers False unless pairwise is provable — over-widening is loose and sound, under-widening is not.

Two details beyond the finding as stated:

- **Rounding of the magnitude sum**, which you called out: `sum |x_i|` is rounded up one ULP before scaling. That single step does not by itself bound an n-term accumulation, so the docstring gives the actual argument rather than implying it does — the summands are nonnegative so there is no cancellation, its relative error is at most `gamma_n` with `n*u < 1e-9` for any n this evaluator sees, and the factor-of-two headroom over the Higham bound absorbs that many times over. The randomized slow-axis fixture at n ∈ [500, 8000] is what backs it empirically.
- **The strided factor is `max(n - 1, log2(n) + 2)`, not a bare `n - 1`.** The two cross at n = 5, so for a handful of terms the sequential bound is the *smaller* of the two; both are sound, but taking it would have made a soundness fix *narrow* strided enclosures — bound-changing, for no benefit at n ≤ 5. As written the pairwise path is bit-identical to before and the strided path only ever widens, with `test_the_sequential_bound_never_narrows_an_enclosure` asserting that across the crossover.

Regressions in `python/tests/test_interval_sum_reduction.py`: your exact `(10002, 2)` fixture with the fast-axis control kept beside it (and an assertion that the control stays tight, so the fix cannot pass by widening everything), plus a randomized slow-axis differential against a `fractions.Fraction` reference. Both fail on the parent commit.

---

python/discopt/mpec.py
jkitchin: Fixed in `1347c2b`. Your call reproduced: `status='local_optimal'`, objective `-18.0` at the starting point, with the generated product row violated by 24.

Taking your two instructions separately, since they pull in opposite directions and the earlier MEDIUM 5 fix depends on the first:

- **Retain the iterate.** It still is: the point, `result.x`, its residuals and the trace all come back unchanged, and the stalled-stage acceptance you asked for in review 2 is untouched.
- **Reserve `local_optimal` for a qualifying converged stage.** Added `discopt.status.LOCAL_LIMIT`, and the Scholtes arm returns it whenever the stage behind the *reported* point did not converge. New status rather than reusing `ITERATION_LIMIT` so it stays inside `LOCAL_STATUSES` and keeps the `SolveResult.__post_init__` no-dual-bound chokepoint; the benchmark harness maps it to `SolveStatus.LOCAL` explicitly, and the existing harness test iterates `LOCAL_STATUSES` so it picked the new member up on its own.

`ContinuationTrace.converged` now distinguishes the two events as you asked: `ContinuationStage.certified` records whether the subsolver converged at that stage, `reported_point_certified` whether the stage behind the reported point did, and `converged` reads the latter instead of `any_stage_accepted` (a zero-iteration run has `any_stage_accepted=True`, which is precisely why that was the wrong discriminator). `any_stage_certified` is exposed as a derived property and both new fields serialize.

Regression `test_review3_3_a_zero_iteration_subsolver_does_not_report_local_optimal`, plus a control asserting a healthy homotopy still reports `local_optimal` — measured, its reported point comes from a converged stage (all 10 stages OPTIMAL on the distance model), so the demotion is driven by evidence rather than blanket. Both fail on the parent commit. You were right that the source-string assertion in `test_review2_5_a_stalled_but_usable_stage_is_taken_forward` tested nothing here; these do not read source.

---

docs/complementarity_semantics.md
jkitchin: Fixed in `1347c2b`. You are right on both halves, and I took the naming with it — leaving a field literally called `accuracy_floor` while documenting "this is not an accuracy floor" would have been a half-fix. None of this schema is on `main`, so the rename costs no compatibility.

- `complementarity_accuracy_floor(t)` → `admitted_residual_scale(t, kind)`, with a companion `admitted_residual_scale_definition(kind)`.
- `Residual.floor` / `at_floor` → `admitted_scale` / `within_admitted_scale`, plus `admitted_scale_definition` beside them so the formula travels with the number, per this module's own rule.
- `ContinuationTrace.accuracy_floor` → `admitted_residual_scale`, documented as the `min`-form summary, with the per-relation numbers carrying their own.

**Per definition**, on `{f, g >= 0, f·g <= t}`: `sqrt(t)` for `min` and (on the nonnegative pair) `natural_map`; `t` for the product form, since the row bounds that quantity directly; `(2 - sqrt(2))·sqrt(t)` for Fischer–Burmeister, whose maximum is at `f = g = sqrt(t)`. `auto` returns `None` — a mixed or unresolved kind names no formula, and borrowing one is the bug. `relation_residuals` now derives each row's scale from that row's resolved kind, and the aggregate takes the worst row's, so the value and the scale it is read against always come from the same formula. At `t = 1e-8` that is the difference between judging a product residual against `1e-4` and against `1e-8`.

**Wording.** Documented as the worst-case residual the relaxation admits, explicitly *not* a limit on attainable accuracy, with your `(0, 1)` case as the reason in the docstring, the semantics doc and the notebook. The empirical half is separated out: whether the subsolver could have done better is a question about the subsolver, and the answer lives in the trace's per-stage statuses, not in this number.

Two tests: one checking all four formulas and that `auto` derives nothing, one asserting an exactly complementary point is *strictly below* the admitted scale — the claim your comment corrects, made executable.

---

CHANGELOG.md
jkitchin: Fixed in `1347c2b`. The `_eval_matmul` paragraph now reads in the past tense and says #1171 landed and is merged into this head, so both reductions go through the shared `_widen_sum` and no one-ULP reduction remains in the module — confirmed against the file rather than the issue state. I also corrected the accumulation-bound sentence in the same entry, which still described the factor as unconditionally `(log2(n) + 2)·eps`; that is now conditional on the reduction order, per your blocking 2.

The PR body's current-head narrative is next — I am rewriting it in full rather than patching the stale sentences, since the local-status contract changed under blocking 3 too, and the #1160/xfail and old panel paragraphs will be marked as historical snapshots with their dates rather than left reading as current state.

---

bernalde:
@jkitchin, the three blocking findings in [my earlier review](https://github.com/jkitchin/discopt/pull/1158#pullrequestreview-5122926860) are addressed at `3560f986`: incumbent acceptance recomputes source feasibility, slow-axis sums receive a conservative accumulation bound, and unproved/stalled iterates retain `local_limit`. The corresponding defect-restoration checks confirmed that the regressions detect the original failures.

Please finish these wording updates:

1. **`docs/complementarity_semantics.md`, “Scale, and the residual the relaxation admits”:** `within_admitted_scale` only compares a residual against its relaxation-derived upper bound; it does **not** establish feasibility of the relaxed rows. For example, at `t = 1e-8`, `(f, g) = (1e-4, 1)` has `min(f, g) = sqrt(t)` and the flag is true, but `f*g = 1e-4 > t`. Please clarify both “inside what the relaxation allows” and the later “edge of the admitted set” statement. Suggested wording: “This comparison alone does not establish feasibility of the relaxed rows.” Keep the matching API explanation consistent.
2. **`python/discopt/mpec.py::_stage_reason`:** replace the stalled-stage explanation ending in “which is what 'local_optimal' claims” with wording that explains that `local_limit` retains an iterate without claiming stationarity.
3. **PR description:** update the deferred-work entry for #1182 to acknowledge that it is completed by merged PR #1184, which implements the paper's simplex/CNF lowering.

Validation: the current PR head has green CI. A clean merge of `3560f986` with current `main` (`cfc0e5c0`) passed **162 focused tests**, including the new #1182 simplex/CNF suite, using the POUNCE revision from this PR's CI and a freshly built native extension.

Once these wording updates are in and CI remains green on the resulting head, I consider this PR ready to merge. No further full review is needed unless the follow-up changes solver behavior.

jkitchin:
All three wording updates are in at `abbc7c2`. Wording only — no solver behavior changed, and the three blocking fixes are untouched.

**1. `within_admitted_scale`.** Your counterexample reproduced exactly before the edit: at `t = 1e-8` the `min`-form admitted scale is `1e-4`, so `(f, g) = (1e-4, 1)` returns `True` while `f·g = 1e-4` exceeds `t` by four orders of magnitude. The doc now carries your sentence verbatim — "This comparison alone does not establish feasibility of the relaxed rows" — and both flagged statements are fixed: the "inside what the relaxation allows" phrasing and the later "edge of the admitted set" one, which now reads "no larger than the worst case the relaxation admits" and says explicitly that this is also a different statement from "the relaxed rows hold at this point". Both point at `lowered_row_residual` as the number that *is* about feasibility.

I also stated the one implication that does survive, since leaving the flag looking useless would be its own inaccuracy: for a nonnegative pair, a `min` residual **above** `sqrt(t)` puts both operands above `sqrt(t)` and therefore forces `f·g > t`, so `False` does mean the relaxed row is violated. It is the converse that fails.

The matching API explanation is updated the same way — `Residual.within_admitted_scale`'s docstring previously said "being outside the admitted scale means the point does not even satisfy the relaxed condition, while being inside says nothing about how close to the true condition the solver got", which covered the solver direction but not the feasibility one. It now carries both directions and your counterexample. The notebook gains the same clarification and prints `lowered_row_residual` beside the flag so the two are not conflated there either. Regression: `test_within_admitted_scale_does_not_imply_the_relaxed_rows_hold`, pinning the flag-true-but-infeasible case, the converse that does hold, and the exactly-complementary case that satisfies both.

**2. `_stage_reason`.** The trailing "taken as a local iterate, which is what 'local_optimal' claims" was left over from before the demotion and now contradicted the code. It reads: *"subsolver stalled (…) at a point it could not certify to tolerance; kept as an iterate — a usable warm start carrying its residuals — under 'local_limit', which asserts no stationarity."*

**3. PR description.** The #1182 entry now records that PR #1184 implements the Wehbeh–Kerrigan Theorem-1 lowering and merged at `a7c1d95`, with #1182 closed by it — verified against the merged PR rather than the issue state. I kept one cross-reference from #1184: it notes requirement 2 of #1182 (reuse `discopt.status`) had no surface there because that lowering is exact and introduces no local-only result, so that requirement only becomes live if a local-NLP arm is built on it — and the vocabulary this PR adds is what it would use. I also retitled that section, since with #1160, #1161 and #1182 all landed, only the `DISCOPT_STATUS_MAP` item is genuinely still open, and refreshed the head/test numbers.

Verification on `abbc7c2`: `pytest -m smoke` **1196 passed, 20 skipped, 2 xpassed**; adversarial slow suite **19 passed**; mpec/interval/provenance suites **127 passed** plus the new regression (**198 executed residual comparisons**); `ruff` and `mypy` clean on `python/`; all notebook code cells execute. Watching CI on this head.

---

PR #1159, opened 2026-09-04
fix(gdp): #1154 complete the hull substitution family for SumOverExpression

### LINKED ISSUE #1154: gdp: complete the hull substitution family for SumOverExpression, then admit it in _is_linear
## Summary

`_is_linear` in `python/discopt/_relax/gdp_reformulate.py` reports `False` for
`SumOverExpression` (the indexed summation `Σ[t0, t1, ...]`), even though such a
node **is** linear when every term is. The under-report is deliberate and is
documented in the code, but it is a capability gap worth closing properly.

## Why it is currently a deliberate False

PR #1150 tried admitting the node and it produced a **false optimality
certificate**. `_is_linear` is a shared gate with ~10 consumers, and several read
a `True` as "and I can therefore take this apart". These have no case for the
node:

| function | handles `SumOverExpression`? |
|---|---|
| `_hull_linear_substitute` | no — falls through to `_substitute_vars` |
| `_substitute_vars` | no — "Constant or unknown, pass through" |
| `_collect_variables` | no — returns `{}`, so vars are never disaggregated |
| `_body_at_zero` | no — raises (this is what refused loudly before) |
| `_bound_expression` | no — raises (big-M) |

With `_is_linear` widened, hull emitted

```
_hull_modes_d0_c0: Σ[3 terms] - (0 * y0) <= 0
```

— the disjunct body imposed **globally**, its selector coefficient collapsed to
zero and the disaggregated variables never created. On

```python
m.either_or([[dm.sum(x[i] - 1 for i in range(3)) <= 0.0], [x[0] >= 8.0]])
m.minimize(-(x[0] + x[1] + x[2]))     # xi in [0, 10], true optimum -30.0
```

both `auto` and `hull` returned `status=optimal, objective=-3.0, bound=-3.0`. A
dual bound of -3.0 is **above** the true minimum of -30.0, i.e. invalid. Three
loud refusals had become two silent wrong answers.

PR #1150 therefore reverted the widening. The `#1039` defect it was trying to fix
(the `exactly-one` row being dropped from the OA/LOA master) was fixed instead by
removing the redundant `_is_linear` pre-gate at those two call sites and relying
on `_extract_body_coeffs`, which answers with the row itself and so cannot
promise more capability than it delivers.

## What is left to do here

Teach the hull substitution family and the bounding path the node, then admit it
in `_is_linear`:

- `_collect_variables` — recurse into `.terms`
- `_substitute_vars` — rebuild the node with substituted terms
- `_hull_linear_substitute` — fold term by term, scaling constants by the selector
- `_body_at_zero` — sum the terms at zero
- `_bound_expression` — interval-sum the terms (closes the big-M route too)

With the first four applied on the #1150 branch, `auto` and `hull` both returned
**-30.0** on the repro above. That patch is verified but was **not** shipped in
#1150, because it is bound-changing.

## Why it needs its own PR and a panel

Admitting the node in `_is_linear` moves four other consumers:

- `solver.py:3068` — the structural linear-FBBT mask, whose docstring cites the
  #27a false-prune. Widening what FBBT treats as linear is bound-changing, and
  unsound if the linearity claim is ever wrong.
- `factorable_reform.py:684` — the aux-lift gate.
- `decomposition/benders/solver.py:114` — Benders route selection.
- `_relax/implied_integer.py:97` — reviewed as safe.

Under CLAUDE.md §5 this is the bound-changing regime: differential bound test,
feasible-point sampling, behind a flag, default-off until a corpus-wide
differential panel passes cert-clean **and** net-positive.

## Definition of done

1. The five functions above handle `SumOverExpression`.
2. `_is_linear` admits it.
3. `python/tests/test_1039_gdp_sumover_rows.py::test_sumover_disjunct_body_never_reports_an_invalid_bound`
   — already in the tree — **strengthens** rather than breaks: it currently
   accepts a loud refusal, and once the routes solve the model it asserts
   `-30.0`. It also already forbids any bound above the true optimum, so it
   catches a regression to the #1150 behaviour.
4. The §5 differential panel, cert-clean and net-positive.

Found while closing out #1039 (PR #1150).

jkitchin:
## Summary

Closes #1154.

`dm.sum(f(i) for i in S)` builds a `SumOverExpression`, an n-ary node holding its
already-expanded term list in `.terms`. Six walkers in
`_relax/gdp_reformulate.py` had no case for it, so a disjunct body written that
way was under-reported as variable-free (`_collect_variables`), nonlinear
(`_is_linear`), unbounded (`_bound_expression`) and un-evaluable at the origin
(`_body_at_zero`). Before this PR the issue's repro is three loud refusals:

```python
x = [m.continuous(f"x{i}", lb=0.0, ub=10.0) for i in range(3)]
m.either_or([[dm.sum(x[i] - 1 for i in range(3)) <= 0.0], [x[0] >= 8.0]])
m.minimize(-(x[0] + x[1] + x[2]))          # true optimum -30.0
```

| sources | `auto` | `big-m` | `hull` |
|---|---|---|---|
| `main` | refuses (no finite big-M) | refuses (no finite big-M) | refuses (`g(0)`) |
| HEAD, `DISCOPT_GDP_SUMOVER=0` | refuses (no finite big-M) | refuses (no finite big-M) | refuses (**coverage**, names `x1`/`x2`) |
| HEAD, default | `optimal -30.0 / -30.0` | `optimal -30.0 / -30.0` | `optimal -30.0 / -30.0` |

Each walker gets a case under **one rule**: `Σ[t1, …, tn]` is handled exactly as
the left-folded chain `t1 + … + tn` — the desugaring the modeling layer could
equally have produced. Nothing about the reformulation mathematics is new; only
the node type is newly recognised. That equivalence is machine-checked per walker
in the new test file, and it is also what the panel measures end to end.

Behind `DISCOPT_GDP_SUMOVER` / `SolverTuning.gdp_sumover`, **graduated
default-ON** by the §5 panel below; `=0` is a live opt-out that restores the
pre-#1154 walkers byte for byte.

### The part that is deliberately NOT behind the flag

`_assert_hull_saw_every_variable`, an independent-walker cross-check run
unconditionally in `_reformulate_disjunction_hull`. It is a refusal, not a
capability.

Everything downstream of `_collect_variables` — the disaggregated variables, the
aggregation rows, the bound-linking rows, the substitution map — is keyed on it.
A variable it misses does not merely go under-disaggregated: it survives into the
reformulated row **un-substituted**, so the disjunct body is imposed *globally*.
That is the row PR #1150 shipped and had to revert:

```
_hull_modes_d0_c0: Σ[3 terms] - (0 * y0) <= 0
```

— which certified `bound = -3.0` on a model whose true minimum is `-30.0`. The
guard is built on `discopt.modeling.core._iter_model_leaves`, a walker this
module does **not** own (a cross-check against the same walker would be a no-op
per CLAUDE.md §6: if the collector cannot see a node, neither could the probe).

**It checks per constraint body, and that is load-bearing** — see the review
section below; the first version compared disjunction-wide unions and missed a
6-line instance of the very class it exists to catch. It costs no model that
reformulates today: on the nonlinear route `_body_at_zero` already raises for
every node type that can trip it, and on the linear route the only way in is an
`IndexExpression` over one of them, which is the defect class itself.

## Correctness

The mathematical claim is airtight: a sum of linear terms is linear, and the
interval sum of interval enclosures is a valid enclosure. Both are implemented as
the *same left fold* the equivalent `BinaryOp` chain already gets, so the change
adds no new rounding exposure beyond what a user writing `t1 + t2 + t3` has
today. The change *removes* a latent under-report rather than adding a claim: a
`_collect_variables` that misses variables makes `presolve/separability` split
coupled blocks, and moving to the correct (larger) variable set can only merge
blocks — the conservative direction.

The `_is_linear` consumers that could over-promise —  `gdpopt_loa.py`, `gbd.py`,
`decomposition/_linear.py`, and `decomposition/benders/solver.py::_model_is_linear`
(that last one added after review; its routing does change arm with the flag, but
`extract_linear` re-gates on `_extract_body_coeffs`, and the flag ON strictly
improves it — a `dm.sum` recourse model that raised on the OFF arm solves on the
ON arm) — each re-gate on `_extract_body_coeffs` and fall back on its `None`, so
a `True` there cannot deliver less than it promises.

Measured, not asserted: **0 invalid bounds over 648 solves** on the firing class,
with every incumbent feasibility-verified in numpy against the *original*
disjunction and every dual bound gated against the best verified-feasible
objective rather than against any solver's answer; **0 bound and 0 primal
violations** over the 52 oracle-scored corpus instances.

All numbers from the current head `0cd93a4` (`main` @ `2d78251` merged in),
unless a flag-arm column says otherwise.

- [x] `pytest -m smoke` → **1119 passed, 20 skipped, 2 xpassed** (the 2 xpassed are pre-existing — PR #1150 reports the same pair)
- [x] `cargo test -p discopt-core` → **N/A — no Rust touched** (`git diff origin/main...HEAD -- crates/` is empty)
- [x] `ruff check` / `ruff format --check` clean over `python/` (963 files); `mypy` clean on both changed modules
- [x] 31-file GDP / OA / Benders / GBD / MPEC selection, `-m "slow or not slow"` → **739 passed, 12 skipped, 2 xpassed, 1 failed**

Run in **both** flag arms with identical results — the byte-identity evidence for
the `=0` opt-out:

| selection | flag OFF | flag ON |
|---|---|---|
| 11 GDP files | 311 passed, 12 skipped, 2 xpassed | 311 passed, 12 skipped, 2 xpassed |
| 19 OA / Benders / GBD / MPEC / hull files | 420 passed, 1 failed | 420 passed, 1 failed |
| `pytest -m smoke` | 1119 passed, 20 skipped, 2 xpassed | 1119 passed, 20 skipped, 2 xpassed |
| adversarial suite | 19 passed | 19 passed |

The one failure throughout is `test_oa.py::TestOAEdgeCases::test_infeasible_model`
(`assert None == {}`), re-confirmed pre-existing against `origin/main`
(`2d78251`) in a `git worktree` with a marker count of 0 for this branch's code.
PR #1150 found and documented the same one and also left it alone.

## Regression test

`python/tests/test_1154_gdp_sumover_hull.py`, **24 passed**. Fail-before run in a
`git worktree` at `origin/main` with §8 marker counts confirming which sources
each arm loaded (`SumOverExpression` count 2 vs 11; `HullVariableCoverageError`
absent vs present):

| sources | result |
|---|---|
| `origin/main` (the new error's import shimmed so the assertions actually run) | **17 failed, 4 passed** |
| HEAD | **24 passed** |

The 4 that pass on `main` are exactly the tests that assert the OFF arm *is*
`main`. The file pins the desugaring equivalence walker by walker, the
end-to-end `-30.0` on all three routes, feasibility of the returned point against
the original disjunction, the coverage guard firing per body and by variable
name, and — so a silent revert of the graduation fails here — that the shipped
default is ON and the `=0` opt-out still works.

- [x] Flag: `DISCOPT_GDP_SUMOVER` / `SolverTuning.gdp_sumover` (the `=0` path is byte-identical to pre-#1154; verified as the OFF arm of the table above)
- [x] Differential-bound test + feasible-point sampling
- [x] **Graduation panel evidence attached — cert-clean AND net-positive.** Graduated default-ON.

Full tables: `docs/dev/issue-1154-gdp-sumover-panel-2026-09-04.md`. Probes in
`scratchpad/issue1154/`; every one prints an executed-assertion count and exits
non-zero at zero (§6), none swallows an exception (§7), and the timing-shaped
claim runs interleaved with a load gate and a spread (§9).

**Corpus arm — inert by construction, and cert-clean.**

* The flag *cannot* fire on the `.nl` corpus: `SumOverExpression` is created only
  by the Python modeling API's `dm.sum`, and the reader never emits one —
  measured at **0 occurrences in 33 376 DAG nodes** across all 66 vendored
  instances, 0 load failures.
* A/B differential anyway (interleaved, `deterministic=True`, `time_limit=10`,
  exact `status`/`objective`/`bound`/`node_count`): **63/66 byte-identical**.
* The 3 that differ reproduce their *entire* difference **within a single arm**:
  3 reps at a fixed flag value on a quiet machine (load 0.04) disagree with
  themselves in **5 of 6** arms. `tls2` with the flag OFF produced both
  `feasible / 2.099999990166769 / 229 nodes` and `time_limit / 2.003761171751075
  / 205 nodes` — the exact pair the A/B flagged. `deterministic=True` neutralizes
  role-2 budgets but leaves the role-1 `time_limit` and the phase-entry gates
  wall-dependent (#1116).
* Oracle-scored (sense read from the model — 5 vendored instances are MAXIMIZE):
  **52 scored, 0 bound violations, 0 primal violations**, 14 unscored and
  reported as unscored.

**Firing class — 108 generated GDP models × 3 routes × 2 arms = 648 solves.** The
family varies term count, disjunct count, sense, coefficient pattern and
linear/nonlinear body; the two arms are the same model with the body as `Σ[...]`
vs the folded chain. The node is confirmed to *survive* the reformulation into
the relaxer (1 surviving node on both `big-m` and `hull`), so the arm comparison
is a real statement about the downstream layers.

```
cases_scored=108
invalid_bounds=0
infeasible_incumbents=2   ← both in the pre-existing hull/CHAIN arm; see below
refusals=0

certification rate (optimal / answered):
  auto   chain: 108/108     auto   sumover: 108/108
  big-m  chain: 108/108     big-m  sumover: 108/108
  hull   chain:  81/108     hull   sumover:  99/108

reaches the best verified-feasible incumbent:   chain 199   sumover 236
```

`auto` and `big-m` are **bit-identical** between the arms on all 108 cases. Every
difference is on `hull` with a nonlinear body, and there the `Σ` form is the
better arm. Repetition control on that one timing-shaped claim (the 54 nonlinear
hull cases, arms interleaved within each case, load 0.71 at start):

| arm | rep 0 | rep 1 | rep 2 | mean | sd |
|---|---|---|---|---|---|
| chain | 29/54 | 29/54 | 29/54 | 29.00 | **0.00** |
| `Σ[...]` | 46/54 | 46/54 | 46/54 | 46.00 | **0.00** |

**Retraction — v1 of this panel was wrong (CLAUDE.md §11).**
`scratchpad/issue1154/panel_capability.py` reported 1 bound violation, 2
certification regressions and 5 objective mismatches. **All 8 are withdrawn.** v1
scored the `Σ` arm against *the chain arm's own result on the same route* — a
solver output used as an oracle — and on nonlinear `hull` that arm routinely
stops at `feasible` short of the optimum `auto` and `big-m` both certify. v1 was
flagging the `Σ` arm for being closer to the truth:

| case (hull, nl=1) | `auto`/`big-m` certify | chain hull | Σ hull |
|---|---|---|---|
| n=3 d=2 `<=` | −8.630462164 | `feasible` −8.603386 | `optimal` −8.630462 |
| n=5 d=3 `==` | −9.116077840 | `feasible` −8.367016 | `feasible` −9.116078 |

v2 replaces the oracle with a gate that cannot have that failure mode: the best
**feasibility-verified** objective across all six (route, arm) pairs is a valid
upper bound on the true minimum, so any bound above it is invalid — no solver
assumed correct. The file is kept unmodified for the retraction record.

**The 2 infeasible incumbents are pre-existing and within tolerance.** Both are
in the `hull`/**chain** arm — the `BinaryOp` path `main` already has — and both
reproduce **bit-for-bit in a worktree at `origin/main`**. Residuals 8.29e−05 on
rhs 3.0 (2.8e−05 relative) and 1.50e−05 on rhs 4.0 (3.8e−06 relative), i.e.
inside the declared `rel = 1e-4` feasibility tolerance; v2's checker used a flat
absolute `1e-5`, stricter than the contract. That is the FSG ε-perspective's
documented O(ε) residual, not a new defect — and note the direction: the strict
checker *excluded* those points from the reference upper bound, which makes the
`invalid_bounds` gate stricter, so `invalid_bounds = 0` holds a fortiori.

### The one judgment call, stated plainly

§5's net-positive bar is written for a mechanism that fires across the `.nl`
corpus. This one **cannot** fire there at all (measured, above). So net-positive
is measured on the class where it does fire, and the corpus arm is used for what
it can establish — that nothing existing moves. This is strictly stronger than
the `DISCOPT_CUT_INHERIT` case the rule was written against (cert-clean but
neutral-or-harmful *where it fired*). If you read §5 as requiring
corpus-measured benefit, the flip back is one word in `solver_tuning.py`
(`default=True` → `default=False`); nothing else in the change depends on the
default, and `test_the_shipped_default_is_on` is the single test to update.

## Review response — commit `0cd93a4`

Both findings confirmed by reproducing them first, then fixed.

**1 (HIGH) — the coverage guard compared disjunction-wide unions.** A per-body
collector miss was invisible whenever the missed variable was collected from any
*other* constraint in the same disjunction — the common case, since the arms of a
disjunction usually constrain the same variables. So the guard shipped as the
general fix for this class and did not catch a 6-line instance of it:

```python
X = m.continuous("X", shape=(3,), lb=0.0, ub=10.0)
A = np.array([[1.0, 1.0, 1.0]])
m.either_or([[(A @ X)[0] - 3.0 <= 0.0], [X[0] >= 8.0]])
m.minimize(-(X[0] + X[1] + X[2]))       # true minimum -30.0
```

`(A @ X)[0]` is an `IndexExpression` over a `MatMulExpression`: `_is_linear`
answers `True` for any `IndexExpression` without inspecting its base, so the body
takes the *linear* route, while `_collect_variables` returns `{}` for it. `X`
reached `all_vars` from the other arm, the union check passed, and hull returned
`status=optimal, objective=-3.0, bound=-3.0`. Reproduced before fixing, and the
reverted-#1150 `SumExpression` row too.

The check is now **per constraint body**, and runs before `all_vars` is built so
it refuses before any work. To be exact about attribution: that `-3.0`
certificate **predates this PR and reproduces on `main`** — what was new was the
guard missing it. The per-body form converts a pre-existing silent false
certificate into a loud refusal, and it makes this PR's own argument for leaving
`SumExpression` unhandled true, which on the union form it was not.

`_collect_variables` is deliberately **not** widened to handle
`MatMulExpression`/`SumExpression` — that is a capability change owing its own §5
panel; the refusal is the conservative answer meanwhile.

**2 (LOW) — the NaN test exercised nothing.** It paired `[-inf, 0]` with
`[0, +inf]`, which folds to `-inf + 0` and `0 + inf`: no NaN either way, so it
passed identically against the unguarded fold (measured: both `(-inf, inf)`).
Re-pointed at `Σ[log(x), exp(y)]` with `x ⊆ (-inf, 0]`, where `_bound_expression`
returns `(-inf, -inf)` — unguarded fold `(-inf, nan)`, guarded `(-inf, inf)`. The
test now asserts that divergence as a **control** before asserting the result, so
it fails loudly if the input ever stops exercising the branch.

Fail-before/pass-after in a worktree at the previous head `a58d7c2` (marker
`PER CONSTRAINT BODY` absent there): **2 failed, 22 passed** → **24 passed**.

### Reconciliation with #1150 (merged) — commit `a58d7c2`

`git merge-tree` reported **no textual conflict**, which was the dangerous case:
#1150 added a comment in `_is_linear` stating the node is "deliberately NOT
admitted here", and two lines below it this branch admits it. The comment is
rewritten to record what #1150 actually established — an **ordering** lesson that
still holds: widening `_is_linear` *alone* leaves `all_vars` empty and produces
the false `-3.0`. What ships is the same widening with the family complete plus
the guard. #1150's two flag-independent invariants are kept and now point at the
panel that graduated the flag.

`test_1039_gdp_sumover_rows.py`'s docstring said this work "is tracked by #1154";
it has landed, and #1150 wrote that test to *strengthen* rather than break when
it did — it now takes its answer branch, all three routes returning `-30.0`.
**Every assertion in that file is untouched.** This head is also the first run of
the two changes together: with #1150's `_extract_body_coeffs` fold and this
branch's `_is_linear` case, the OA/LOA linear-row path takes a `Σ[...]` body as a
real LP row for the first time.

### Not shipped here, on purpose

* **`SumExpression`** (the axis reduction `dm.sum(X, axis=0)`) and
  **`MatMulExpression`** stay unhandled in the walkers. Different nodes, array
  semantics, and the constant-scaling rule in `_hull_linear_substitute` is not
  obviously the same for them. The per-body guard turns the silent-miss hazard
  into a refusal, which is what makes leaving them out safe.
* **The hull FSG ε-residual** — pre-existing, within the declared `rel = 1e-4`
  feasibility tolerance, reproduces on `main`, out of scope here.

https://claude.ai/code/session_01U9EmG5ZXztwQ38RhDdY7i3


jkitchin:
## Code review — PR #1159

I read every hunk, then ran targeted probes against the branch in an isolated worktree. Each probe asserted it had loaded the branch's `gdp_reformulate.py`, asserted the `HullVariableCoverageError` marker present/absent per arm, and printed an executed-assertion count.

### What checked out clean

The six-walker widening is correct as far as I could exercise it:

- `_is_linear` stays conservative on every non-affine `Σ` shape I tried (bilinear-by-index, bilinear-by-scalar, squares, `exp`, `abs`, `z[i]/z[0]`) and admits only genuinely affine ones, including nested `Σ` and `/const`.
- 160-case randomized differential of `big-m`/`hull` × `Σ`-form/chain-form against a brute-force LP enumeration of the disjuncts: 0 wrong objectives, 0 bounds above the true optimum.
- 100-case nonlinear (FSG perspective) hull differential with incumbents re-verified in numpy against the *original* disjunction: 0 violations, 0 cross-arm bound violations.
- The newly-reachable `factorable_reform` aux-lift path (`sqrt(dm.sum(...))` with a nonlinear argument, which `_collect_variables` now sees ≥2 vars in): the `Σ` arm reproduces the folded-chain arm's bound **exactly** on 12/12 random instances.
- `SolverTuning(gdp_sumover=False)` really does reach the walkers (GDP lowering runs inside `_scoped_tuning`), so the typed opt-out is live, not just the env var.
- Benders routing (`decomposition/benders/solver.py::_model_is_linear`, which the PR's correctness section does *not* list among the re-gating consumers) changes arm with the flag, but `extract_linear` re-gates on `_extract_body_coeffs`, and the flag ON strictly improves it — a `dm.sum` recourse model that raised on the OFF arm solves on the ON arm. No regression.
- `_bound_expression`'s explicit infinity handling cannot produce NaN in any reachable combination, and where it diverges from the chain it diverges conservatively.
- The guard does not newly break anything: every node type it can trip on (`SumExpression`, `MatMulExpression`, `CustomCall`, `Parameter`) is already refused by `_body_at_zero` on both `main` and HEAD (verified side by side).

---

### 1. HIGH — `python/discopt/_relax/gdp_reformulate.py:1355` — the coverage guard checks the disjunction-wide union, so a per-body collector miss stays silent

`_assert_hull_saw_every_variable` compares the independent walker's leaves against `all_vars`, which is the **union of collected names over the whole disjunction**, not against what the collector found in *that body*. So a per-body collector miss is invisible whenever the missed variable happens to be collected from any other constraint in the same disjunction — which is the common case, since disjunctions typically pin the same variables in the other arm.

This is not theoretical. `_is_linear` returns `True` unconditionally for `IndexExpression` without inspecting its base, while `_collect_variables` returns `{}` for `IndexExpression(base=MatMulExpression)`. On the branch:

```python
m = dm.Model("ix")
X = m.continuous("X", shape=(3,), lb=0.0, ub=10.0)
A = np.array([[1.0, 1.0, 1.0]])
m.either_or([[(A @ X)[0] - 3.0 <= 0.0], [X[0] >= 8.0]])
m.minimize(-(X[0] + X[1] + X[2]))
m.solve(gdp_method="hull")     # -> status=optimal, objective=-3.0, bound=-3.0
```

The true minimum is `-30.0` (second arm: `X0>=8`, `X1=X2=10`). The guard passes silently — `X` is in `all_vars` via `X[0] >= 8.0` — the body goes down the *linear* route, `_hull_linear_substitute` returns the `IndexExpression` unchanged, and the arm's row is imposed globally. A dual bound above the optimum, i.e. exactly the false certificate the guard's docstring says it exists to prevent.

I also confirmed the union-based check misses the literal `SumExpression` shape, by monkeypatching `_is_linear` to admit it: hull emitted

```
_hull_anon_d0_c0 | ((sum(X[3]) - (3*y0)) - (0*y0)) - (0*y0) <= 0
```

— the reverted-#1150 row, verbatim — with the guard silent.

To be clear about attribution: the false certificate in the repro above pre-dates this PR and reproduces on `main`. What is new is that this PR ships a guard advertised as the general fix for this class, and it does not catch a 6-line instance of it. That also undercuts the PR's own "Not shipped here, on purpose" argument for leaving `SumExpression` unhandled — "a `False` only forgoes a capability, and the new coverage guard turns the silent-miss hazard into a refusal" does not hold while the guard compares against the union.

Fix: do the check **per constraint body** — `set(_iter_model_leaves(con.body)) - {Parameter} ⊆ set(_collect_variables(con.body))` for each `con`, rather than against the disjunction-wide union. That version fires on both repros above.

### 2. LOW — `python/tests/test_1154_gdp_sumover_hull.py:166` — the NaN test never exercises the branch it guards

`test_bound_expression_does_not_produce_nan_on_mixed_infinities` rests on the premise that plain float addition "would give `-inf + inf = nan`". But for its own input (`lo_free` in `[-inf, 0]`, `hi_free` in `[0, inf]`) the term intervals are `(-inf, 0)` and `(0, inf)`, so the folded chain computes `-inf + 0` and `0 + inf` — no NaN either way. The test passes identically against the unguarded fold, so it does not exercise the explicit-infinity branch (CLAUDE.md §6) and would stay green if that branch were deleted.

A case that actually diverges needs a term whose *upper* bound is `-inf` alongside one whose upper bound is `+inf` — e.g. `Σ[log(x), exp(y)]` with `x ⊆ (-inf, 0]` (`_bound_expression` returns `(-inf, -inf)` there) and `y` unbounded above. Worth swapping in, or dropping the NaN claim from both the test and the code comment at `gdp_reformulate.py:661`.

jkitchin:
Both findings confirmed and fixed in `0cd93a4`. I reproduced each before touching anything, and re-ran the repro against the fix.

## 1 (HIGH) — you're right, and the union check was the whole weakness

Reproduced verbatim (`scratchpad/issue1154/probe_review_f1.py`, §8 markers asserted):

```
body node: BinaryOp
  _is_linear         -> True
  _collect_variables -> []
hull -> status=optimal obj=-3.0 bound=-3.0
  CONFIRMED: dual bound -3.0 is ABOVE the true minimum -30.0
```

The guard is now **per constraint body**, and runs *before* `all_vars` is built so it refuses before doing any work. Both shapes you found now raise:

```
HullVariableCoverageError: GDP hull reformulation cannot see every variable in the
body of disjunct 0, constraint 0: ['X'] are reachable from that body's expression
DAG but were not collected from it, ...
```

— the `(A @ X)[0]` repro, and the literal `SumExpression` row under your `_is_linear` monkeypatch (`probe_review_f1b.py`). Fail-before/pass-after in a worktree at `a58d7c2`, marker `PER CONSTRAINT BODY` absent there: **2 failed, 22 passed** → **24 passed** at HEAD. Both repros are pinned as tests, each with a §6 premise guard so they can't pass vacuously if `_is_linear`/`_collect_variables` ever change their answers on that body.

Your attribution point is the one I'd have gotten wrong on my own, so to restate it in the record: the `-3.0` certificate **predates this PR and reproduces on `main`** — what was new was shipping a guard advertised as the general fix that missed a 6-line instance. And you're right that this undercut my own "leaving `SumExpression` unhandled is safe" argument: that argument was circular on the union form. Per body it holds, and the guard now converts a pre-existing silent false certificate into a loud refusal.

I did **not** widen `_collect_variables` to handle `MatMulExpression`/`SumExpression` — that is a capability change owing its own §5 panel, and the refusal is the conservative answer in the meantime.

## 2 (LOW) — measured, and you're right

`probe_review_f2.py`:

```
as written  -> guarded (-inf, inf)   unguarded (-inf, inf)     # branch not exercised
proposed    -> guarded (-inf, inf)   unguarded (-inf, nan)     # branch load-bearing
```

Re-pointed at `Σ[log(x), exp(y)]` with `x ⊆ (-inf, 0]`. Rather than just swapping the input, the test now **asserts the divergence as a control** (`assert np.isnan(plain_hi)` against a plain fold) before asserting the result, so it fails loudly if the input ever stops exercising the branch instead of going quiet again. The comment at `gdp_reformulate.py` keeps the NaN claim, since it is now the measured one.

## On the Benders consumer

Thank you — `decomposition/benders/solver.py::_model_is_linear` was missing from my correctness section, and your finding (routing changes arm, `extract_linear` re-gates on `_extract_body_coeffs`, flag ON strictly improves) matches what the code does. I've added it to the PR body's list rather than leaving the section incomplete.

## Verification of the fix

| check | result |
|---|---|
| 31-file GDP / OA / Benders / GBD / MPEC selection, `-m "slow or not slow"` | **739 passed** (was 737; +2 new), 12 skipped, 2 xpassed, 1 failed |
| `pytest -m smoke` | 1119 passed, 20 skipped, 2 xpassed |
| `ruff check` / `format --check` over `python/`, `mypy` | clean |

The per-body form refuses nothing that reformulated before — the selection count moved only by the two tests I added. The 1 failure is the same pre-existing `test_oa.py::TestOAEdgeCases::test_infeasible_model`.

---

PR #1164, opened 2026-09-05
perf(milp): root cut cleanup + profile reset guard; retract the cut-budget claim

jkitchin:
## What this changes

Two things in the MILP engine's root phase, plus the measurement that says the
feature they were built for does **not** ship on.

**1. Root cut cleanup (`MilpOptions::root_cut_prune`).** Every cut separated at
the root previously stayed in `csc_w` for the rest of the solve, so each one rode
in *every* node LP. This drops the cuts that are slack at the final root LP
before the tree starts.

Soundness: a slack cut has a zero dual multiplier at the root optimum, so the
root bound and its KKT certificate survive its removal exactly. Deeper nodes may
get a *weaker* bound than they would with the full set — never an invalid one,
since what remains is a subset of globally valid cuts. Bound-CHANGING under
CLAUDE.md §5, never bound-invalidating.

The non-obvious part: `separate_gomory_cols` returns coefficients over **all**
columns, including the surplus columns of earlier cuts. So (a) a cut may only be
dropped once nothing kept references its surplus column — hence the reverse
dependency-closure sweep — and (b) `augment_cols_with_cuts` cannot be reused for
the rebuild, because it only scans `j in 0..n_base` and would silently install a
*different* row. `rebuild_csc_with_cuts` exists for exactly that, and
`the_rebuild_places_a_cut_coefficient_on_an_earlier_cuts_slack` pins the
difference so a future simplification back to the batch appender fails loudly.

**2. `profile::begin_solve()`.** Every MILP solve called `init_from_env();
reset();` on entry. Under `cargo test`'s thread pool a sibling test's solve
zeroed an armed probe's counters mid-measurement, so it read 0 and looked like
dead code — the CLAUDE.md §6 failure mode. `begin_solve` yields to
`TEST_OVERRIDE` on both halves; this is the reset half of the hazard the earlier
`TEST_OVERRIDE` change fixed for the arm half. Regression test included.

## Why the root cut budget does NOT graduate — retraction

I previously reported the raised root cut budget as net-positive and said it met
CLAUDE.md §5's bar: 26/30 solved at 93.81 s vs base 25/30 at 103.77 s. **I am
retracting that** (CLAUDE.md §11). That panel was 30 *synthetic lattice families*
— exactly the §2 trap of measuring a class by proxy.

On 38 real MIPLIB 2017 "easy" instances (each one validated by having HiGHS
reproduce the published optimum from the same converted arrays discopt receives;
fidelity failures: 0), TL = 20 s, arms interleaved per instance, 190/190 runs, no
certification abort:

| arm | solved | wall | med(solved) | nodes | kept/gen |
|---|---|---|---|---|---|
| base_16x1 | **18**/38 | 444.93 s | 0.139 s | 8,312,044 | (not counted) |
| cut200_noprune | 11/38 | 573.62 s | 0.235 s | 1,964,690 | (not counted) |
| cut200_prune | 11/38 | 529.55 s | 0.160 s | 1,327,506 | 2544/5720 |
| cut500_prune | 11/38 | 541.52 s | 0.196 s | 694,144 | 3709/12671 |
| highs (reference) | 36/38 | 76.77 s | 0.802 s | 15,073 | n/a |

> **Correction (2026-09-05).** Those two cells previously read `0/0`. That was an
> **instrument artifact and I am retracting it** (CLAUDE.md §11). `RootCutsGenerated`
> and `RootCutsKept` increment only inside the cut-cleanup block
> (`milp_driver.rs:1226-1229`), which is gated by `prune_base` (`:934-939`) —
> `Some(..)` only when `root_cut_prune && root_cuts > 0 && cut_rounds > 1`. At
> `cut_rounds = 1` they never fire, so `0/0` meant *not counted*, not *no cuts*.
> The base arm does generate and keep cuts: `b-ball`'s root bound is −1.705882 at
> `root_cuts=16` against −1.818182 at `root_cuts=0`, and `enlight_hard` goes
> 21 → 23. This is precisely the CLAUDE.md §6 failure mode — a counter that
> measured nothing and was read as a measurement. It does not change any
> conclusion in this section (the solved/wall/node columns are independent of it),
> but the cells were being read as evidence that the base arm cuts nothing.

The budget **loses 7 of base's 18 solves and adds ~85 s of wall.** So
`DISCOPT_MILP_ROOT_CUTS` stays default-OFF and this PR does not propose flipping
it. Nothing here changes any default: `root_cut_prune` only takes effect inside
the cut path, which is itself off by default.

What the cleanup buys, stated carefully: in **aggregate** over the cut path it
beats not pruning — 529.55 s / 1.33 M nodes vs 573.62 s / 1.96 M nodes at the
same 11 solves. I earlier wrote that as "strictly better than not pruning";
**that phrasing is wrong and I retract it** (CLAUDE.md §11). Per instance the
prune is often much *worse*. On `sp150x300d` at a 200-cut budget, removing the
prune reaches the **identical** root bound (60) at 22.3 iterations/node instead
of 2079.6 and 0.38 s instead of 13.64 s; `fiber` and `blend2` show the same sign.

That per-instance split is the thread that led to the root cause, which is not
about pruning at all: `separate_gomory_cols` returns coefficients over the
slack-extended column space, so a cut puts a second nonzero into a base slack
column. That column stops being a singleton, `solve_lp_cols_warm`'s basis
reconstruction can no longer substitute for it, the returned basis is short,
`PreparedDual::prepare` refuses it on shape, and every node cold-solves — which
returns another short basis, so the state is self-sustaining. The prune looked
like a win in aggregate because its dependency-closure sweep keeps a cut
precisely *because* something references its slack, so it retains the
cross-referencing cuts and discards the clean ones, concentrating the pathology
onto fewer instances rather than causing it.

The fix is to rewrite each cut into structural space before it is appended. That
is a **separate PR** — #1167, which stacks on this one and is now open. This PR's
cleanup stands on its own soundness argument and stays default-inert.

**Post-fix update (2026-09-05).** With #1167's structural-space rewrite in the
tree, the same panel re-run at matched `gap_tol = 1e-4` inverts the result above:
`cut500_prune` reaches **21/38 at 4,693,240 nodes** against base's **19/38 at
8,853,388** — cert-clean *and* net-positive, 190/190 runs, zero certification
aborts. The budget's net-negativity measured here was caused entirely by the
warm-start breakage, exactly as this section's root-cause paragraph predicted.
Graduating the budget is therefore back on the table, but in its own PR with its
own panel — nothing in *this* PR changes a default.

Two source reviews (against `ref/HiGHS/` and `ref/scipoptsuite-10.0.2/`) read
that regression the same way and independently: it is a cut **lifecycle**
failure, not a cut **quality** failure. Neither reference solver carries root
cuts forever — HiGHS ages basic cut rows out at `mip_lp_age_limit`
(`HighsLpRelaxation.cpp:595-642`) and drops basic cut rows after the root
(`:522-539`); SCIP has no root round budget at all (`maxroundsroot = -1`,
`set.c:471`) and stops on stalling instead, with dynamic rows removed after
`rowagelimit` LPs nonbasic (`set.c:263`).

SCIP in fact does precisely what this PR adds, by default:
`SCIP_DEFAULT_LP_CLEANUPROWSROOT = TRUE` (`set.c:268`) removes new **basic** rows
after the root LP — the same rows this cleanup drops, on the same argument. That
is prior art for the soundness claim above. `CLEANUPROWS = TRUE` (`:267`) extends
it to every LP; that half is Stage 1 of
`docs/dev/milp-competitiveness-plan.md`, added here, and is not in this PR.

### Methodology note carried forward

That panel gave discopt `gap_tol = 1e-6` against HiGHS's default
`mip_rel_gap = 1e-4` — a 100× tighter stopping rule. It does not change the
conclusion above (all four discopt arms ran at the same tolerance), but it does
inflate the discopt-vs-HiGHS column: at matched 1e-4, `neos-1425699` goes from
640,669 nodes unsolved to **29 nodes / 0.00 s**. Later panels run matched.

## Also in this PR

`docs/dev/milp-competitiveness-plan.md` — the evidence-grounded roadmap that
replaces "raise the cut budget", including the primal/dual decomposition of
base's 20 failures, a HiGHS self-ablation showing presolve ≤2 instances and
symmetry 0 on this panel, and the argument for why the prior CUT-1/CUTS-1 NO-GO
(`certification-gap-plan.md`) does not automatically transfer from the MINLP
class to pure MILP — stated as a precondition to test, not an assumption.

It also carries the plan's first measurement, run before this PR opened. Stage 0
is "flip the switches that already exist and ship off"; the HiGHS review expected
`node_propagation=true` to be worth 3-6 instances. Measured over the same 38
instances at matched `gap_tol=1e-4`, TL = 20 s, 114/114 runs, zero certification
aborts:

| arm | solved | wall | med(solved) | nodes |
|---|---|---|---|---|
| base | 19/38 | 413.62 s | 0.101 s | 8,969,820 |
| nodeprop | 19/38 | 395.90 s | 0.081 s | 7,838,764 |
| highs | 38/38 | 66.35 s | 0.959 s | 15,417 |

**It gains zero instances** — that prediction is falsified and recorded as such.
It is still worth −12.6 % nodes and −20 % median time, with 5-11× wins on
`flugpl`, `enlight8`, `gt2` and `bppc8-02`, so it graduates ON in a separate,
one-line PR with this panel behind it. Not flipped here. (Machine load rose 2.85
→ 13.84 mid-run, so the wall column is soft; node counts are load-independent and
carry the conclusion, and arms rotated per instance.)

The useful part is what it rules *out*. Sixteen of the 38 instances HiGHS closes
in **one node** — root LP plus root cuts plus root heuristics both producing the
optimum and proving it, where discopt spends 350 k-870 k nodes.

### Correction to an earlier retraction in this body

This section previously retracted the phrase "the gap is a **root-strength**
problem" and replaced it with a primal/dual/proof split claiming "8 of 17 failures
already hold a dual bound within ~5 % and still cannot close." **That replacement
was wrong and is itself retracted** (CLAUDE.md §11). It classified failures with an
arbitrary 5 % threshold that has nothing to do with the solver's actual stopping
rule, and I published it without checking it against that rule.

Re-derived against the real tolerance (`gap_tol = 1e-4`), over the 17 instances
still unsolved at the strongest arm of the successor panel:

| class | count | meaning |
|---|---|---|
| dual gap already within `gap_tol` | **0** | no pruning/mechanics failure exists |
| PURE-BOUND (incumbent optimal, only bound missing) | 3 | `b-ball` 4.5e-4, `istra` 9.0e-3, `iskar` 1.0e-2 |
| BOTH SHORT (bound and incumbent) | 14 | dual 4.6e-3 … 1.0; primal 22 % … no feasible point |

**The bound is short on 17 of 17. The incumbent is additionally short on 14 of 17.**

The specific claim I made about `b-ball` — "dual gap 0.045 %, optimal incumbent in
hand, 200,839 nodes vs HiGHS's 1, therefore bounding/pruning mechanics" — is false.
Its bound is `-1.500672636` against an optimum of `-1.5`, a relative dual gap of
`4.484e-4` against a `1e-4` tolerance. The solver has *not* converged; it is
behaving correctly and is short by a factor of ~4.5 in bound. Nothing about
pruning is implicated, on that instance or any other on this panel.

So the original "root-strength" framing was substantially right and my first
retraction over-corrected. The accurate statement is narrower than either: this is
a **bound-strength** problem — root or in-tree — on every failing instance, with a
primal-heuristic problem layered on top of 14 of them. Bound work is necessary
everywhere; primal work is necessary on 14 and sufficient on none, since no
instance closes without the bound arriving too.

## Verification

- `cargo test -p discopt-core` — 677 passed, 0 failed.
- New tests: `the_root_cut_cleanup_drops_slack_cuts_without_moving_the_optimum`,
  `the_rebuild_places_a_cut_coefficient_on_an_earlier_cuts_slack`, and the
  `profile::begin_solve` reset-hazard regression test. Each fails before the
  corresponding change.
- Full CI on this PR is green: Rust, Python fast, Python correctness (fast
  subset), Python AMP coverage, claim-boundary, lint + typecheck, codecov
  patch/project.
- `pytest -m smoke` (1165 passed, 1 skipped, 2 xpassed) and `pytest -m slow
  python/tests/test_adversarial_recent_fixes.py` (19 passed) were run locally on
  the **successor branch** `fix/milp-cuts-structural-space`, which contains this
  PR's commit plus the structural-space fix — so they cover this change but are
  not an isolated measurement of it. This PR's own isolated evidence is the CI
  run above.
- Corpus gate: the 38-instance MIPLIB panel above, 190/190 runs, no false bound
  and no wrong optimal on any arm.

Contributes to the MILP competitiveness work; closes nothing.

### DIFF
--- a/crates/discopt-core/examples/milp_bench.rs
+++ b/crates/discopt-core/examples/milp_bench.rs
@@ -162,6 +162,8 @@ fn opts(n_struct: usize, integer_cols: Vec<usize>, tl: f64) -> MilpOptions {
         initial_incumbent: None,
         node_hook_rounds: 0,
         node_hook_cut_cap: 0,
+        root_cut_time_s: None,
+        root_cut_prune: true,
         simplex: SimplexOptions {
             tol: 1e-9,
             max_iter: 100_000,
--- a/crates/discopt-core/examples/par_bench.rs
+++ b/crates/discopt-core/examples/par_bench.rs
@@ -126,6 +126,8 @@ fn opts(inst: &Instance) -> MilpOptions {
         initial_incumbent: None,
         node_hook_rounds: 0,
         node_hook_cut_cap: 0,
+        root_cut_time_s: None,
+        root_cut_prune: true,
         simplex: SimplexOptions {
             tol: 1e-9,
             max_iter: 100_000,
--- a/crates/discopt-core/src/bnb/milp_driver.rs
+++ b/crates/discopt-core/src/bnb/milp_driver.rs
@@ -447,6 +447,47 @@ pub struct MilpOptions {
     /// the budget is spent the hook stops being called and the search runs on the
     /// relaxation it has.
     pub node_hook_cut_cap: usize,
+    /// Wall-clock ceiling for the ROOT CUT LOOP alone, in seconds.
+    ///
+    /// `None` (the value at every existing call site) leaves the loop bounded
+    /// only by `root_cuts` / `cut_rounds` / tailing-off, exactly as before.
+    ///
+    /// A value makes a many-round configuration safe to ask for. Raising
+    /// `cut_rounds` is how a weak root relaxation gets closed — on a lot-sizing
+    /// model the default single pass leaves the root bound at a fraction of the
+    /// optimum — but the separation phase is worth several seconds only if it
+    /// still leaves time to branch. This is the knob that says how much of the
+    /// budget the root may spend, so the round cap can be set by what the model
+    /// needs rather than by what a worst case can afford.
+    ///
+    /// Checked at the top of each round, so a round already in flight always
+    /// completes and the loop never abandons a partially augmented matrix. It
+    /// therefore bounds *entry* into rounds, not the duration of one round; a
+    /// single very slow round can still overrun it, bounded in turn by
+    /// `time_limit_s` through the LP's own deadline.
+    ///
+    /// This is not the ceiling on the whole solve -- that stays `time_limit_s`,
+    /// which the loop also honors (see the guard at the top of the round).
+    pub root_cut_time_s: Option<f64>,
+
+    /// Drop root cuts that are slack at the final root LP before the tree starts.
+    ///
+    /// Every root cut rides in every node LP for the rest of the solve. A cut that
+    /// is not tight at the root optimum has a zero dual multiplier there, so
+    /// removing it leaves the root bound and its certificate untouched while
+    /// taking a row out of every node — the measured cost of NOT doing this is
+    /// severe (30-instance MILP panel, `root_cuts=500, cut_rounds=50`: node counts
+    /// fell 2.4x but total wall rose from 103 s to 145 s, and `cflp_s101` went
+    /// from 0.04 s to 5.85 s at an unchanged node count).
+    ///
+    /// Deeper nodes may get a weaker bound than they would with the full set —
+    /// this is bound-CHANGING under CLAUDE.md §5, never bound-INVALIDATING, since
+    /// what remains is a subset of globally valid cuts. Kept as a switch so the
+    /// two can be measured against each other in one interleaved process.
+    ///
+    /// Inert unless `cut_rounds > 1`: with a single round every cut was separated
+    /// off the one root solve and is violated there, so nothing is ever slack.
+    pub root_cut_prune: bool,
     /// LP solver options.
     pub simplex: SimplexOptions,
 }
@@ -610,8 +651,7 @@ pub fn solve_milp_node_hooked(
     } else {
         node
     };
-    crate::profile::init_from_env();
-    crate::profile::reset();
+    crate::profile::begin_solve();
     let ns = opts.n_struct;
     let is_int = {
         let mut v = vec![false; ns];
@@ -876,13 +916,54 @@ pub fn solve_milp_node_hooked(
     // (`n_w`) it was computed at. Reused to warm-start the root B&B node so it does
     // not re-derive the augmented LP from a cold slack basis. See `root_warm_basis`.
     let mut root_basis: Option<(Basis, usize)> = None;
+    // Snapshot of the matrix as it stood before any root cut, kept so the cut
+    // cleanup after the loop can rebuild from it with a subset of the cuts. Only
+    // taken when more than one round can run: with `cut_rounds == 1` every cut is
+    // separated off the single root solve and is violated there by construction,
+    // so the cleanup below provably keeps all of them and the clone would be pure
+    // cost. (`cut_rounds == 1` is the shipped default.)
+    let prune_base: Option<(SparseCols, usize, usize)> =
+        if opts.root_cut_prune && opts.root_cuts > 0 && opts.cut_rounds > 1 {
+            Some((csc_w.clone(), m_w, n_w))
+        } else {
+            None
+        };
+    let mut all_root_cuts: Vec<GomoryCut> = Vec::new();
+    let mut last_root_x: Option<Vec<f64>> = None;
     if opts.root_cuts > 0 {
         let _t = crate::profile::Timer::new(crate::profile::Phase::RootCutLoop);
         let mut total_cuts = 0usize;
         let mut prev_obj = f64::NEG_INFINITY;
+        // Wall bound for the separation phase: the earlier of the caller's own
+        // deadline and the (optional) root-cut sub-budget. Checked at the top of
+        // each round -- a round in flight always finishes, so the loop never
+        // leaves a half-augmented matrix behind.
+        //
+        // `root_cut_time_s` is the substantive half: it is the only thing that
+        // stops a high `cut_rounds` from spending the whole budget separating.
+        //
+        // The `deadline` half is a cheap early exit, NOT a fix for an overrun.
+        // The loop already stopped at the caller's deadline indirectly: each
+        // round's root LP carries `node_simplex.deadline`, and a deadline-aborted
+        // LP comes back non-`Optimal`, which breaks the loop below. Testing the
+        // clock here just skips mounting a round whose LP would abort anyway,
+        // along with the separation work that would follow it.
+        let root_cut_deadline: Option<std::time::Instant> = match (deadline, opts.root_cut_time_s) {
+            (Some(dl), Some(rc)) => {
+                Some(dl.min(t_start + std::time::Duration::from_secs_f64(rc.max(0.0))))
+            }
+            (Some(dl), None) => Some(dl),
+            (None, Some(rc)) => Some(t_start + std::time::Duration::from_secs_f64(rc.max(0.0))),
+            (None, None) => None,
+        };
         for _round in 0..opts.cut_rounds {
             if total_cuts >= opts.root_cuts {
                 break;
+            }
+            if let Some(dl) = root_cut_deadline {
+                if std::time::Instant::now() >= dl {
+                    break;
+                }
             }
             crate::profile::incr(crate::profile::Ctr::RootCutRounds);
             let root = {
@@ -962,6 +1043,9 @@ pub fn solve_milp_node_hooked(
             // if more cuts are appended below it is extended to the new size after
             // the loop. The clone is cheap next to the solve it came from.
             root_basis = Some((root.basis.clone(), n_w));
+            if prune_base.is_some() {
+                last_root_x = Some(root.x.clone());
+            }
             // Tailing off: stop once added cuts barely move the bound.
             if root.obj <= prev_obj + 1e-7 * (1.0 + prev_obj.abs()) && prev_obj > f64::NEG_INFINITY
             {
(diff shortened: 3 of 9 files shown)


jkitchin:
## Code review — PR #1164

Scope: `git diff origin/main...origin/perf/milp-root-cut-budget` — root cut cleanup + `rebuild_csc_with_cuts`, `root_cut_time_s`, `profile::begin_solve`, the PyO3 signature changes, and `_milp_root_cut_budget` in `solver.py`.

**The prune's soundness argument holds.** The dependency-closure invariant (`dep < ci`) is genuinely guaranteed — a cut's `coeffs` never span more than `n_base + start_of_its_round` columns; `rebuild_csc_with_cuts` produces row-sorted CSC columns matching `augment_csc_with_cuts`'s `b`/`c`/`l`/`u`/`is_int` conventions exactly; and the truncate/re-push of `b_w`/`c_w`/`l_w`/`u_w`/`is_int_full` restores the base state exactly, since nothing between the snapshot and the loop mutates them. Dropped cuts can only weaken a bound, never invalidate one. The findings are elsewhere.

---

### 1. HIGH — `crates/discopt-python/src/lp_bindings.rs:1232` — `root_cut_prune=true` as the binding default turns a bound-changing prune ON for the shipped MILP path

The PR argues the prune is inert because "the cut path is itself off by default". **It is not.** `python/discopt/solvers/milp_simplex.py` has `_MILP_CUT_BUDGET_DEFAULT = True` (graduated 2026-08-29), and its `solve_milp` escalation calls `_drive(cut_kwargs=dict(_STRONG_CUT_PROFILE))` with `root_cuts=200, cut_rounds=10`. `_drive` (line 909) never passes `root_cut_prune`, so it picks up the new binding default `true`, and the driver's gate `root_cut_prune && root_cuts > 0 && cut_rounds > 1` is satisfied.

So any MILP/OA master the cheap probe fails to close — the `rsyn08*` class #1066 exists for — now runs with the prune active: different cut set, different dual bound, different node counts than `main`, with **no graduation panel and no opt-out**. The PR's own measurements say the prune is often much worse per instance (`sp150x300d`: 0.38 s → 13.64 s at the same root bound), so this is a likely wall-clock regression on exactly that class, and it contradicts the PR body's "nothing here changes any default".

Same default at lines 1104 and 1391. Fix: default `root_cut_prune=false` in all three `#[pyo3(signature = ...)]` blocks and opt in explicitly from the flagged path.

### 2. MEDIUM — `crates/discopt-core/src/bnb/milp_driver.rs:1118` — `all_root_cuts` retains a dense length-`n` `Vec<f64>` per cut for the whole solve

`GomoryCut.coeffs` is dense (`lp/gomory.rs:55`). Before this PR each round's `new_cuts` was dropped after `augment_csc_with_cuts`, so peak cut storage was O(cuts_per_round × n). Now every root cut is retained in `all_root_cuts`, declared at function scope (line 931) and alive through the entire B&B, not just the cleanup.

At the `_STRONG_CUT_PROFILE` budget that finding 1 makes reachable by default (200 cuts), on a model with `n = 100k` columns (structural + one slack per row), that is 200 × 100k × 8 = **160 MB** of pure overhead; at the proposed `root_cuts=500`, 400 MB. Storing the sparse `(col, val)` pairs, or dropping `all_root_cuts` right after the rebuild, removes it.

### 3. MEDIUM — `crates/discopt-core/src/profile.rs:657` — `TEST_OVERRIDE` converts an under-count flake into an over-count flake

With `TEST_OVERRIDE` set, a concurrent test's `begin_solve` returns early, so `ENABLED` stays `true` for that sibling solve — which then increments the *same* global counters. `begin_solve_does_not_clear_counters_under_a_test_override` holds `TEST_LOCK`, but the clobbering test does not (that is the stated premise of the fix), so `assert_eq!(counter(Ctr::RootCutRounds), 7)` (line 668) and `== 5` (line 681) fail whenever any concurrent `solve_milp` in the same test binary runs a root-cut round in that window — `Ctr::RootCutRounds` is incremented unconditionally at `milp_driver.rs:968`. Same hazard class the fix targets, sign flipped. Use a counter no driver path touches, or snapshot a delta rather than an absolute.

### 4. LOW — `crates/discopt-core/src/profile.rs:392` — `0/0` means "not counted", and the doc invites the misreading the PR already had to retract

`RootCutsGenerated`/`RootCutsKept` both live at `milp_driver.rs:1172-1176`, inside `if let (Some(prune_base), Some(x))` — they never fire when `root_cut_prune` is off or `cut_rounds == 1`. The doc says "`Kept == Generated` means the cleanup found nothing to remove", which `0 == 0` satisfies, so a reader concludes "no cuts were generated". That is verbatim the instrument artifact retracted in the PR body ("Correction (2026-09-05)") — the code and the doc still produce it. Either count generation unconditionally, or say that `0/0` means the cleanup did not run.

### 5. LOW — `python/discopt/solver.py:21509` — `root_cut_time_s` is a wall clock that runs even under `deterministic`

`_milp_root_cut_budget` returns `max(0.5, 0.5 * float(engine_budget))`, and `engine_budget` is `_milp_budget`, which under `_tuning().deterministic` is `float(time_limit)` precisely so no machine-speed clock decides how much tree is explored (see the `#1116` comment at ~21667). With `DISCOPT_MILP_ROOT_CUTS=1`, how many of the 50 cut rounds execute — hence the root bound, hence the reported dual bound — becomes machine-speed dependent again in deterministic mode. Default-off, so low, but the flag will not be graduatable in this shape. Suggest `None` for `root_cut_time_s` when `_tuning().deterministic`.

### 6. LOW — `crates/discopt-core/src/bnb/milp_driver.rs:1201` — the comment's safety claim is false in release builds

`coeffs[n_base + newidx[j - n_base]]` with `newidx[..] == usize::MAX` does **not** panic in release (integer overflow checks are off by default): `n_base + usize::MAX` wraps to `n_base - 1` and silently writes the cut coefficient onto a *structural* column, producing an unjustified — possibly optimum-cutting — row. The invariant that makes this unreachable does hold today, but the guard is documented as the backstop and it is not one. Use `n_base.checked_add(newidx[...]).expect(...)`, or `assert_ne!(newidx[dep], usize::MAX)`.

### 7. LOW — `crates/discopt-core/src/bnb/milp_driver.rs:1222` — `pool_sigs` is not remapped when the rebuild renumbers columns

`cut_signature` keys on absolute column indices (`milp_driver.rs:3052`). After the rebuild, column `n_base + k` refers to a different cut's slack than when the root signatures were recorded, and the signatures of dropped cuts remain in the set. A node-separated cut can collide with a signature that no longer describes it and be silently discarded by `dedup_new_cuts`. Safe — dedup only rejects, never installs a wrong row — but it drops valid cuts for no reason. Clear or remap `pool_sigs` alongside the rebuild.

### 8. LOW — `crates/discopt-core/examples/par_bench.rs:130` — the parallel benchmark now measures a different configuration than `main`

`par_bench.rs` runs at `cut_rounds: 2` (line 114), so the mechanically added `root_cut_prune: true` activates the prune in the harness used for parallel-throughput numbers. Any before/after comparison against a `main`-built `par_bench` is now cross-configuration. `false` is the configuration-preserving value here (as it is in the two `dive_schedule_*` tests, where `root_cuts: 0` makes it moot either way).

---

Finding 1 is the blocker: it is a silent default change on a graduated path, and it is the one thing the PR body explicitly claims does not happen.

Issue #1153, opened 2026-09-04
incumbent quality is non-monotone in time_limit: more budget explores 5x fewer nodes on nvs19


jkitchin:
## Summary

On `nvs19` the search never reaches the known optimum at any budget tested up to
480 s, and — the part that matters — **more budget yields a worse incumbent**.

Known optimum: **-1098.4**.

| time_limit | incumbent | nodes |
|---|---|---|
| 30 s | -1098.2 | 38,403 |
| 60 s | **-1001.2** | **7,619** |
| >= 120 s | (stops on a ~100,000 node cap) | |

Doubling the budget from 30 s to 60 s made the answer *worse* and explored **5x
fewer nodes**. That is the signature of #1116's role-1 / role-2 budget coupling:
`time_limit` is role-1 ("when do we stop?"), but sub-budgets are carved from it
as role-2 ("how much work do we do per node?"). A larger `time_limit` inflates
the role-2 sub-budgets, so the search spends more per node and covers less tree.

## Why this is filed separately from a soundness bug

There is no false certificate here. The reported incumbents are feasible, the
status is not `optimal`, and the soundness sibling
(`test_suboptimal_warm_start_never_false_certifies`) passes. This is a
**completeness** miss plus a **monotonicity** defect.

The monotonicity defect is the real finding and is the more general one: a user
who doubles `time_limit` and gets a worse answer has no way to reason about the
solver. Node throughput falling 5x as the wall budget doubles is measurable
independently of whether `nvs19` is ever solved.

## Definition of done

1. Incumbent quality is **non-decreasing** in `time_limit` on a panel of
   instances — the general property, not `nvs19` specifically (§2).
2. The role-2 sub-budgets stop scaling with role-1 in a way that reduces node
   throughput as the wall budget grows. #1116 has the analysis; this is a
   concrete instance of the harm it predicts.
3. Only then, whether `nvs19` reaches -1098.4 in budget.

Item 1 is the gate. An implementation that solves `nvs19` without restoring
monotonicity has fixed an instance, not the class.

## Test

`python/tests/test_incumbent_injection_soundness.py::test_feasible_integer_node_objective_is_recorded`
is a strict xfail referencing this issue.

Found while closing out #1039 (PR #1150). Related: #1116.

jkitchin:
## Root cause found — and this issue's stated diagnosis is falsified

PR #1192 has the fix. Recording the falsification here (CLAUDE.md §4) because
[...]

PR #1166, opened 2026-09-05
fix(budgets): #1153 diagnose the budget/throughput coupling; rule out the wall-budget fix family

jkitchin:
## Summary

Contributes to #1153. **The issue is not closed by this PR, and the reason is a
finding rather than an excuse — read §6.6c and §6.7 of the plan doc first.**

#1153 asks for incumbent quality to be non-decreasing in `time_limit`, gated on a
panel. Measuring that gate first turned up three things that reframe the issue:

1. **The gate already passes on the in-repo corpus.** 198 comparisons over 66
   instances at 5/10/20/40 s produced **one** violation (`clay0303hfsg`); all four
   17-instance decision panels since produced **zero** in both arms.
2. **The `nvs19` reproduction the issue is built on does not reproduce here.** It
   returns `-1098.2` stably at 15/30/60/120 s.
3. **The *throughput* half does reproduce, with zero spread, and is diagnosed to
   a single call.** A whole family — `bchoco*`, `heatexch_gen*`, `beuster`,
   `4stufen`, `contvar`, `hda` — explores exactly **3 nodes at every budget from
   5 s to 240 s**. `heatexch_gen2`'s layer profile locates it: `root_time` is
   3.5 s of a 5.1 s wall at 5 s and *the entire* 10.1 s at 10 s. The NLP probe
   names the consumer: at 5 s the feasibility pump is never admitted; at 10 s it
   is, and its stage spends **6.4 s of the 10 s budget** returning **no
   incumbent**.

**The headline result is a family-level negative.** Six mechanisms were built and
measured; five were falsified or inert and deleted, and panel D showed *why none
of them could have worked* — see below. Both flags ship default OFF.

Durable value: the diagnosis, a 38-site inventory of budget carves with a
**static ratchet** that fails the build on a new uncapped one, the monotonicity
panel harness, and every negative result written up so the next attempt starts
from measurement rather than from scratch.

## Correctness

- [x] Cannot produce a false certificate — every gated call is a primal
  *heuristic*, so refusing or truncating one changes which incumbent is found and
  when, never the dual bound or the certificate (§0.3 heuristic-policy). Every
  capped carve is optional *tightening*. All four panels: **no certification
  regressions**, no bound above its reference.
- [x] No validation, fallback, or safety guard was weakened.
  `DISCOPT_ROOT_BUDGET_GATE=0` still disables the whole entry gate independently.

All figures from the **current head**, re-run after the last code change:

- [x] `pytest -m smoke` → **1119 passed, 15 skipped, 2 xpassed** (7m17s)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **19 passed** (2m47s)
- [x] `ruff check` / `ruff format --check` clean; `mypy` clean on changed modules

`python/tests/test_1153_budget_monotonicity.py` → **9 passed, 1 skipped** (6m02s).
The skip is the behavioural control declining to judge, which is correct here: the
cap form does not produce the throughput change that test looks for, so the
control does not reproduce the collapse and the treatment is unfalsifiable on this
run. It asserts whenever the control *does* reproduce it.

## Regression test

- [x] `python/tests/test_1153_budget_monotonicity.py`:
  - unit contracts of `saturate_role2` / `heuristic_entry_share`;
  - a **static ratchet**: a carve of a role-1 budget is either provably bounded
    above by a constant or recorded in `KNOWN` with a category and a **pinned
    count**. Verified firing on three injected arms with a saturated negative
    control staying silent — and it caught this PR's own new carve;
  - a floor on the saturated call-site count, counted by AST;
  - the behavioural gate — arms **interleaved per repetition**, per-cell median,
    sense read from `ObjectiveSense`, parametrised over **both** arms so the
    shipped default path is gated too, **skipping** rather than failing where the
    machine does not show the defect.

- [x] Default-OFF flags: `DISCOPT_BUDGET_SATURATION`, `DISCOPT_HEUR_ENTRY_SHARE`
  (both flag-OFF paths byte-identical)
- [x] No bound rose above its reference; no `gap_certified=True` instance regressed
- [x] **Not graduating — default-off only.**

### Measurement 1 — `DISCOPT_BUDGET_SATURATION` (falsified as the cause)

17 open instances, ladder 30/90/240 s. Monotonicity **34 comparisons, 0
violations in both arms**; incumbent **0/0/17**; **3 dual bounds weaker**, 0
stronger. Cert-clean, not net-positive. Also mis-scoped: inert below a 150 s
budget, so it could never have touched the 30 s vs 60 s inversion the issue
reports.

### Measurement 2 — `DISCOPT_HEUR_ENTRY_SHARE`, four panels

17 open instances, ladder 5/10/20/40 s, both arms interleaved in one process,
136 solves each.

| | A: shared gate | B: finder-scoped refuse | C: + mean estimator | **D: cap the stage** |
|---|---|---|---|---|
| monotonicity violations | 0/0 | 0/0 | 0/0 | **0/0** |
| incumbent better/worse/same | 1/3/64 | 2/2/64 | 1/2/65 | **3/2/63** |
| node count up/down | 17/4 | **20/0** | 18/0 | 5/1 |
| cert regressions | none | none | none | **none** |

**The A→B step** confirmed review's finding that the divisor was landing in a gate
shared by 13 heuristic entries; scoping it to the two feasibility-pump entries
removed a regression and eliminated every node-count decrease.

**The B→D step is the important one.** Capping the stage instead of refusing it
recovers exactly what refusing destroyed — `nvs05` goes from *losing* its
incumbent to finding one at 5 s and improving it at 10 s (`1269.70 → 1107.89`),
and `tspn12`'s regression disappears — but throughput falls from 20 up to 5.

**That symmetry rules out the whole family.** The throughput gain and the
incumbent losses have the *same* cause: not running the pump. Refusing buys nodes
and costs incumbents; capping keeps incumbents and returns the nodes; every
intermediate setting trades along that one line, because a wall-budget policy only
decides *how much pump* runs while cost and value both scale with it. Panels A, B,
C and D, plus §6.4 and §6.5, are all instances of that same family.

### Falsifications recorded (CLAUDE.md §4, §11)

- **§6.1** saturating carves — killed by its own graduation panel.
- **§6.4** bounding the pump's clock — measured inert, then found to be an
  **artifact**: a finder stage is *two* NLP consumers and that change bounded only
  the pump, leaving the more expensive multistart seed on the global deadline.
- **§6.5** success weighting on the finder role — inert on all nine cells: the
  harm is done by the *first* pump, which the rule always admits.
- **§6.6b** the mean estimator — falsified by a **mechanism probe, not its panel**:
  on both regressing instances the first finder decision fires with zero NLPs
  observed, so `max == mean == the 2.0 s default` and the estimators are
  numerically identical at the only decision that matters.
- **§6.6c** the stage cap — sound and incumbent-positive, but returns the
  throughput; the family verdict above.

### Methodology notes worth carrying forward

- **A container restart moved the baseline**, not just the treatment.
- **Per-repetition output separates signal from noise** (`heatexch_gen2`@20 s is
  bimodal under the legacy arm; a last-repetition summary would have logged a
  regression that does not exist).
- **A panel attributes effect; only a probe attributes mechanism** (§6.6b).
- **Verify a cap actually binds before panelling it.** §6.4 skipped that and
  measured an artifact; the first stage cap here was Zeno's bound — per-call
  grants fell 3.0 s → 1.75 s while total stage wall did not move (7.08 → 7.47 s).

## What remains — one direction, not two

Budget policy cannot fix this, per the family verdict. **The pump must become
cheaper for the same value, not smaller.** That is #912's `WorkBudget`:
deterministic operations rather than seconds, so the pump's cost is knowable in
advance instead of estimated at a 2.0 s default that is 3× wrong, and identical on
every machine. It is the only remaining lever that changes the cost/value ratio
rather than sliding along it.

Two things need a human call, neither reachable from here: whether `nvs19` still
reproduces on the reporter's machine, and whether #1153 should be re-scoped to the
throughput property that does reproduce.

https://claude.ai/code/session_01XEU9mwRUdRbMD1tcFFEePN


python/tests/test_1153_budget_monotonicity.py
```diff
@@ -0,0 +1,492 @@
...
+        "split",  # divides the helper's own budget across its candidate variables
+    ),
+    (
+        "solver.py",
+        "_route_budget = max(",
+        "split",  # the auto-routed sub-solver's own role-1 budget
+    ),
+    (
+        "solver.py",
+        "deadline=min(",
```
jkitchin, line 380: **The ratchet does not fire on a new uncapped carve that lands on this line text.**

`KNOWN` is keyed on `(path, stripped first source line)` and `_scan()` returns a *set*, so a new unbounded carve whose widened expression begins with a recorded generic prefix — `deadline=min(` here, and likewise `_route_budget = max(` and `float(time_limit)` — is absorbed by the existing entry and reported as clean. `test_recorded_carves_still_exist` also stays green, because the recorded line is still present.

Verified against a copy of the head `python/discopt` tree, injecting a carve into `solve_model` three ways and running `_scan()` unchanged:

```
A plain new carve                      -> REPORTED  [('solver.py', '_new_stage_budget = time_limit * 0.42')]
B collides with a KNOWN line text      -> SILENT    []      # first line is `deadline=min(`
C role-1 var under a new name          -> SILENT    []      # `_budget_left * 0.42`
```

Arm B is the case the ratchet exists to prevent. Cheapest close: have `_scan()` return a count per key and pin the count in `KNOWN`, so a second site landing on a recorded line goes 1 -> 2 and fails.

Arm C is inherent to the 13-name `_ROLE1_NAMES` whitelist and is a docstring question rather than a bug — but the module docstring's "every sub-budget in the package computed by multiplying or dividing a role-1 budget" claims more coverage than the whitelist delivers.

---

python/discopt/solver.py
```diff
@@ -11764,6 +11801,40 @@ def _root_heur_nlp_entry_ok(_ev=None) -> bool:
...
+        These stages were handed ``_deadline`` — the whole SOLVE deadline — so
+        what they cost was set by the caller's ``time_limit`` rather than by the
+        stage. That is #1116's role-1 clock doing a role-2 job, and #1153
+        measured the harm: on ``heatexch_gen2`` (3 reps, zero spread) a 5 s budget
+        never admits the feasibility pump and explores 7 nodes, while a 10 s
+        budget admits it, its two sub-NLPs spend 6.4 s of the 10 s, they return
+        NO incumbent, and the tree explores 3. Doubling the budget bought one more
+        heuristic that ate 64 % of it and produced nothing.
+
+        Bounding the stage's own clock to a share of what is left fixes that by
```
jkitchin: **This docstring still carries the claim §6.4 falsified** (CLAUDE.md §11).

It says the share-sized clock "fixes that **by construction** … the stage can never take more than its share", but §6.4 measured exactly why it cannot: the pump's sub-NLPs overrun their own `max_wall_time` (3.27 s against a 3.0 s grant) and the pump polls its deadline only between rounds, so the deadline does not bind on round one — ON equalled OFF on 15 of 16 cells.

The comment in `_root_heur_nlp_entry_ok` ~30 lines above records that falsification correctly; these two docstrings in the same file now contradict each other, and a reader landing on this one gets the retracted version.

The second claim needs the same treatment: "strictly better than refusing the stage — measured on `tspn12` … refusing it loses that incumbent, while capping its clock does not" describes the stage-deadline-only arm, which is no longer a reachable configuration. One flag turns on the clock cap *and* the entry rule, and §6.3 measured `tspn12` losing `262.647 -> 282.244` under that combination.

Keeping the cap is defensible (the 20 s incumbent, and it costs nothing), but coupling an effective mechanism and a measured-inert one to a single flag means the next panel cannot separate them again.

---

python/discopt/solver_tuning.py
```diff
@@ -1173,6 +1173,80 @@ class SolverTuning:
...
+    arms). Doubling the budget bought one extra heuristic that consumed 64 % of it
+    and produced nothing, and the tree paid for it.
+
+    The entry gate is what admits that. It refuses a heuristic only when the time
+    left cannot absorb *one whole* solve of the largest size seen so far — so a
+    heuristic may consume up to 100 % of the remaining budget and still be
+    admitted. Crossing that threshold therefore costs more than the budget
+    increment that unlocked it, which is precisely a non-monotone step.
+
+    The repository already applies the right doctrine one role over: an
```
jkitchin, line 1236: **The flag tightens 13 entry sites, not the finder role — and this is probably why three incumbents were lost.**

This paragraph contrasts the finder role with `_improver_allowed` and says the flag "keeps the exemption and bounds it". But the divisor goes into `_root_heur_nlp_entry_ok`, which is the shared gate for *every* root heuristic NLP entry — 13 call sites, several of them improver-role:

- `solver.py:14132` sits directly after `_improver_allowed(_HEUR_COST["enumerate"])`
- `solver.py:14279` (`integer_box_search`) is guarded by `tree.incumbent() is not None`
- `solver.py:14369` is node-diving on a frequency schedule

So under the flag an improver is double-gated: its own success-weighted, node-proportional contingent **and** a 4x-stricter entry test. Soundness is unaffected (all 13 sites are primal heuristics, so refusing is always sound), but the blast radius is much wider than the docstring describes.

This is a cheap, well-posed next experiment rather than only a doc fix: apply the divisor at the two feasibility-pump sites alone and re-run the §6.6 decision panel. Three lost incumbents against one gained is the shape you would expect from over-broad application, and if the finder-only arm keeps the throughput gain without `nvs05`/`tspn12`, the flag's graduation case changes.

Related, and worth stating in §6.7 as the structural reason it cannot graduate as written: `_mean_heur_nlp_cost()` returns `_heur_nlp_cost["default"] = 2.0` until a heuristic NLP has been observed, so at the *first* call this gate is simply "more than 8 s left", on any model, independent of the pump's real cost. That is the entire effect on the 5-10 s rungs where #1153's harm lives, and it is the same 3x-wrong default the "what remains" section already names.

---

jkitchin (commented):
## Review: sound, and the "do not graduate" call is right

Read the full 21-file diff, the plan doc, and re-ran the ratchet scanner standalone against the head tree (`9fdc5c0`). The behavioural tests were **not** run here — no built extension in this container — so those are reviewed by reading only.

Three things worth fixing before merge, one of which I verified breaks a guarantee the PR advertises. All three are inline.

### What I verified

- **Ratchet is self-consistent at head**: 10 scan hits, all in `KNOWN`, zero stale; `saturate_role2` call sites = 6, matching the pinned count.
- **Soundness holds.** All 13 `_root_heur_nlp_entry_ok` sites are primal heuristics; every saturated carve is optional tightening. Flag-OFF is genuinely identity — `heuristic_entry_share() -> 1.0` makes the divisor a no-op and `_heur_stage_deadline()` returns `_deadline` unchanged.
- **Saturation caps never bind below a site's own floor** at the current fracs (smallest is 0.1 -> 15 s against a 2.0 s floor).

### Findings

1. **The ratchet does not fire on a new uncapped carve in two realistic cases** (inline on the test file). `KNOWN` keys on `(path, stripped first source line)` and `_scan()` returns a set, so a new carve landing on a generic recorded prefix (`deadline=min(`) is silently accepted. Demonstrated by injection against a copy of the head package.
2. **`_heur_stage_deadline`'s docstring still asserts what §6.4 falsified** (inline on `solver.py`), and its `tspn12` "strictly better" claim describes an arm that is no longer separately reachable. CLAUDE.md §11.
3. **The flag tightens all 13 root-heuristic entry sites, not the finder role** (inline on `solver_tuning.py`) — improvers are double-gated against their own contingent. Most likely explanation for the three lost incumbents, and a cheap next experiment.

### Smaller items

- `test_incumbent_quality_is_monotone_in_time_limit` hardcodes minimisation (`obj_b > obj_a` = violation) and runs **only with the flag ON**. So the shipped default has no monotonicity test at all, and a max-sense instance added to the panel would silently invert the gate. Read the sense from the model; run the ladder in both arms.
- No `CHANGELOG.md` `[Unreleased]` entry, which CONTRIBUTING §8 requires — two new env flags is exactly what that section is for.
- `saturate_role2(seconds, frac)` takes `frac` from the caller and nothing relates it to the site's own floor; a future `frac=0.01` would cap at 1.5 s and silently undercut a 2 s floor. A one-line note or assert keeps the next caller honest.

### On the issue

This is a diagnosis PR and it is honest about that, but nothing here changes the default path, so #1153's reported behaviour is unchanged for a user. `Contributes to #1153` with the issue left open is the right framing — the durable value is the diagnosis, the ratchet, and the three recorded falsifications, and those are worth merging on their own.

CI at review time: nothing red; `Python fast` and `Python correctness (fast subset)` still in progress.

---

python/tests/test_1153_budget_monotonicity.py
jkitchin: Confirmed and fixed. I reproduced arm B against the head tree before changing anything, and the set-keyed scan was silent exactly as you said:

```
arm B ('deadline=min(' collision): SILENT — the ratchet did NOT fire
```

That is the ratchet failing at the one job it exists to do, so thank you — it would have shipped as a tripwire that reports "clean" forever.

Fixed your way: `_scan()` now returns `dict[key, count]` and `KNOWN` carries a pinned count per entry (all 1 today). The assertion reports any key whose observed count exceeds its recorded one, so a second site landing on a recorded line goes 1 → 2 and fails. Re-verified all three arms on the fixed scanner:

```
A plain new carve                   -> REPORTED [('solver.py', '_new_stage_budget = time_limit * 0.42')]
B collides with a KNOWN line text   -> REPORTED [('solver.py', 'deadline=min(')]
baseline (unmodified tree)          -> clean
```

On arm C — agreed it is inherent, and the docstring was overclaiming. It now states the limit rather than implying coverage it does not have: "*a role-1 budget*" means a name in the 13-name `_ROLE1_NAMES` whitelist, a carve off a role-1 value held under another name is invisible, and no amount of AST work fixes that without whole-package dataflow. It is a tripwire on the spellings this codebase actually uses, not a proof of absence.

---

python/discopt/solver.py
jkitchin: You are right on both counts, and the resolution is stronger than a doc fix: **I removed the clock cap entirely.**

Re-checking its one claimed benefit before deciding, that benefit does not survive either. The "incumbent at 20 s the legacy arm missed" came from the stage-deadline-only run, a single repetition — and the later three-arm run showed `heatexch_gen2` at 20 s is *bimodal under the legacy arm itself* (`None` in rep 0, `808843.77` in rep 1). So the gain I cited was noise in the arm I was crediting, which makes it the same per-repetition trap I flagged elsewhere in this PR and then fell into myself one section earlier. With no measured benefit and a measured-inert mechanism, keeping it was carrying a dead path (§3), so `_heur_stage_deadline` and both `deadline=` call-site changes are gone.

That also resolves your coupling objection at the root rather than documenting around it: the flag now turns on exactly one mechanism, so the next panel measures one thing.

The surviving comment in `_root_heur_nlp_entry_ok` keeps the falsification on record — including the correction that the 20 s gain was bimodality, not the cap.

---

python/discopt/solver_tuning.py
jkitchin: This is the most consequential of the three and you are right — I put the divisor in the shared gate and then described it as if it were finder-scoped. Thirteen entries, several improver-role and already answering to `_improver_allowed`'s contingent, were double-gated.

Fixed rather than just documented: `_root_heur_nlp_entry_ok` takes `finder: bool = False`, the share applies only when it is set, and only the two feasibility-pump entries pass `finder=True`. Every other site is back to the legacy rule.

Your hypothesis that this explains the lost incumbents is well-posed and testable, so I am **re-running the §6.6 decision panel finder-scoped** (17 open instances, 5/10/20/40 s, both arms interleaved in one process) rather than asserting it. `nvs05`'s loss at 10 s is exactly the shape over-broad application would produce. If the finder-only arm keeps the throughput gain without `nvs05`/`tspn12`, the graduation case does change and I will say so; if it does not, that is also worth knowing, because it means the pump entry itself is the cost. I will post the numbers either way.

The `_mean_heur_nlp_cost()` point is the sharpest thing in this review and I have promoted it from "what remains" to a stated structural limit on the flag itself. At the *first* call the default 2.0 s makes this gate "is more than `2.0 / share` seconds left?" — on any model, independent of the pump's real 6.4 s cost. That first call is the entire effect on the 5–10 s rungs where the harm lives, so the rule is acting on a 3x-wrong estimate precisely where it matters most. That is now in the flag docstring as the reason it cannot graduate as written, not buried in the plan doc.

---

jkitchin:
## Code review — PR #1166

Reviewed `origin/main...origin/claude/discopt-issue-1153-bxf1v2` in full, ran the new unit tests (7 passed), and empirically probed the new ratchet.

No soundness issues — both flags are genuinely inert when off, and every `_role2_saturate` site can only shrink a grant. The theme of the findings is that the PR's **enforcement machinery does not hold up to its own doctrine**: the two most important findings are self-referential.

---

### 1. MEDIUM — `python/discopt/solver.py:1835` (`_heur_stage_deadline`) — the PR introduces the defect it forbids

The new helper computes `_share * (_deadline - _now)`: an unsaturated role-2 carve expressed as a fraction of the role-1 remaining budget, with no ceiling. That is precisely the pattern this PR exists to eliminate and adds a build ratchet against.

At `time_limit=3600` it hands one feasibility pump a **900 s** grant. By the PR's own doctrine (`ROLE2_SATURATION_S=150`, share 0.25) it should cap at 37.5 s.

### 2. MEDIUM — `python/tests/test_1153_budget_monotonicity.py:145` (`_ROLE1_NAMES`) — the ratchet meant to catch that is blind to it

The scan only fires on a `Mult`/`Div` `BinOp` containing an `ast.Name` from a 12-entry hardcoded allowlist, so it misses the three most common carve spellings in this codebase. Verified empirically by planting a probe module in `python/discopt/` — `_scan()` returned **zero** hits for all of:

```python
0.25 * (_deadline - _now)            # exactly what _heur_stage_deadline does
0.5  * _remaining_budget()           # a Call, not a Name
0.3  * (deadline - time.perf_counter())
```

The ratchet's live `found` set is currently identical to `KNOWN` (10/10), so it adds no coverage beyond the spellings already listed. The PR body's claim that "a new uncapped carve fails the build" is not true in general — and finding 1 is the proof.

### 3. MEDIUM — `python/discopt/solver.py:11789` — the admission rule divides a running **max**, not a mean

`_mean_heur_nlp_cost()` returns the running max — its own comment says "The *max* (not mean) is deliberate", it records documented ~15 s overruns, and it never decays. Dividing that by `_heuristic_entry_share()` makes the admission rule `remaining > 4 × worst-case-ever`, not "4× typical".

Concretely: one expensive heuristic NLP observed early (heatexch_gen3-class, ~15 s) causes **every** remaining root heuristic to be refused for the rest of the solve at any `time_limit` under 60 s. That is the mechanism behind the PR's own measured `nvs05@10s` regression where the instance loses its incumbent outright. A percentile or mean estimator, or applying the share only to the first finder, would not have that cliff.

### 4. MEDIUM — `scratchpad/i1153/three_arm.py:21,23` — the committed harness cannot run

It references two symbols that do not exist in the tree it ships with (the weighted rule was deleted per §6.5):

```
hasattr(discopt.solver, '_finder_entry_share_for')  -> False
hasattr(solver_tuning, 'finder_entry_share')        -> False
```

The script dies at import: `AssertionError` on line 21, then `AttributeError` on line 23. The PR body cites the three-arm probe as "what made the comparison valid" and as a methodology artifact to carry forward — it cannot be re-run as committed.

### 5. MEDIUM (flaky test) — `python/tests/test_1153_budget_monotonicity.py:412-436`

`test_finder_entry_share_stops_the_throughput_collapse` runs the OFF pair and then the ON pair **sequentially, one repetition each** — not interleaved, no spread. The PR's own methodology notes record that a baseline shift invalidated a cross-run comparison, and that `heatexch_gen2` is bimodal at these budgets (measured `[7, 3]`, sd 2.0). A single unlucky OFF/ON draw produces `on_large < on_small` with a collapsing control and reddens the build with no defect present. CLAUDE.md §9 asks for interleaved arms and a spread; this shipped test has neither.

### 6. LOW — `python/tests/test_1153_budget_monotonicity.py:364`

`assert len(sites) == 6` fails in **both** directions, including the desired one: adding a seventh legitimately-saturated carve — exactly what the ratchet in `test_no_unsaturated_role2_carve` instructs an author to do — breaks the build, while the failure message only explains the lowering case. The line-based grep also counts any prose occurrence of `` `saturate_role2(` `` in a package docstring that does not start with `#` / `def ` / `from ` / `import ` / `*`.

### 7. LOW — `python/tests/test_1153_budget_monotonicity.py:476`

`obj_b > obj_a + tol` hardcodes minimization sense, and nothing in the test consults `ObjectiveSense`. Adding a maximize instance to `_COLLAPSE_PANEL` silently inverts the check: real quality regressions read as passes and improvements read as violations. Related: this test runs only under `_with_entry_share(True)`, so the **shipped default path** (flag OFF) has no monotonicity gate at all.

### 8. LOW — `python/discopt/_relax/lp_spatial_bb.py:509`

`saturate_role2(max(0.0, time_limit - elapsed) / 3.0, 1.0/3.0)` violates the helper's stated contract — `frac` is documented as "the fraction of `time_limit` it was carved from", but the value passed is a third of the *remaining* time, so the cap it derives (50 s) is unrelated to how much of `time_limit` has already been spent. The effective grant `min(remaining/3, 50)` is still safe; the issue is that the ceiling is looser than the surrounding prose asserts, so a future reader tuning `ROLE2_SATURATION_S` will reason about this site wrongly.

### 9. LOW — `scratchpad/i1153/nlp_probe.py:31`

`S.solve_nlp_pounce = wrapped` sets a name that does not exist on `discopt.solver` (`solve_nlp_pounce` is a function-local import alias at `solver.py:16056`). It is a silent no-op that reads as a second interception seam. The probe happens to work anyway because `NP.solve_nlp` is resolved at call time by both the local import and `nlp_backend._try_pounce()` — but the dead line invites the belief that the direct `solver.py` path was covered when it was not (CLAUDE.md §6).

---

### Checked and found sound

- No circular import from the new `lp_spatial_bb` → `solver_tuning` edge (`solver_tuning` imports only stdlib).
- Both flags are genuinely inert when off: `saturate_role2` and `heuristic_entry_share` return the legacy value byte-identically, and `_heur_stage_deadline` short-circuits to `_deadline` at share 1.0.
- All six `_role2_saturate` sites can only shrink a grant, the surrounding `min(..., max(remaining, floor))` structure is preserved at each, and the `_deadline` / `perf_counter` clock sources match.
- `enter_scope(None)` inherits, so the tests' `SolverTuning(...)` scopes really do reach `solve_model` — this could easily have been a silent no-op; it isn't.
- The `.gitignore` `scratchpad/**/*.json` addition does not affect the 36 already-tracked scratchpad JSON files.

The framing in the PR body — that #1153 stays open with the reason stated as a finding rather than an excuse — is the right call on the evidence presented.

jkitchin:
Thanks — this is a genuinely sharp review, and the framing is the part that landed hardest: *the PR's enforcement machinery does not hold up to its own doctrine*, with the two most important findings being self-referential. That's fair and I've fixed it rather than argued with it. Pushed in `e2b6a90`; every finding was reproduced against the head tree before I changed anything.

**F1 / F2 — the self-referential pair.** F1 was already resolved by the preceding commit (the previous review round had me delete `_heur_stage_deadline` for being measured inert), so the offending carve is gone. **But F2 is the serious half and was still live**: the ratchet would not have caught it. Reproduced with your probe-module method:

```
0.25 * (_deadline - _now)               -> NONE   (a Sub of two unlisted names)
0.5  * _remaining_budget()              -> NONE   (a Call, not a Name)
0.3  * (deadline - time.perf_counter()) -> NONE
```

`_carries_role1()` now recognises a whitelisted name, a **call** returning a remaining budget, or a **deadline-minus-clock subtraction**. All three arms report after the fix, and the real tree still yields exactly the same 10 unbounded sites — so coverage grew without new debt. Your point that `found` was identical to `KNOWN` (10/10) and therefore added no coverage was the right diagnosis.

**F4** — you're right that a probe cited as a falsification's evidence must be runnable. `three_arm.py` died at import on two symbols deleted with the weighted rule. It now reconstructs that rule locally and injects both arms through `solver._heuristic_entry_share`; verified end to end.

**F5** — accepted without reservation, and it stings: the test ran OFF then ON sequentially, one repetition, no spread, on an instance the PR itself records as bimodal. That is the exact §9 failure the PR criticises elsewhere. Now interleaved within each repetition, 2 reps, compared on the per-cell median.

**F6** — fixed as a floor rather than equality, and counted by AST instead of a line grep. **F7** — now reads `ObjectiveSense` off the model, and the monotonicity test is parametrised over **both** arms; your observation that the shipped default path had no gate at all was the more important half. **F8** — recorded at the site. **F9** — dead line removed.

**F3 is deliberately not fixed yet, and I want to be explicit about why.** You're right that `_mean_heur_nlp_cost()` returns a running max that never decays, so the rule reads *"remaining > 4 × worst-case-ever"* and one early expensive NLP refuses every later heuristic for the rest of the solve. You propose it as the mechanism behind the measured `nvs05@10s` incumbent loss — and the finder-scoped panel running right now tests precisely that, because the previous round's fix (share applies only at `finder=True`) is the competing explanation for the same regression. Changing the estimator now would invalidate that run and leave *both* hypotheses unmeasured. Data first, then the change — I'll post the numbers and act on them either way.

Two of your "checked and found sound" items were things I had assumed rather than verified (`enter_scope(None)` inheritance actually reaching `solve_model`, and the `.gitignore` addition not orphaning the 36 tracked JSON files). Good to have them checked.

---

jkitchin:
Correction to my previous comment: the commit carrying those fixes is **`fee05ae`**, not `e2b6a90`. I wrote that hash from memory instead of reading it, and it does not exist on this branch. The content of the reply stands; only the reference was wrong.

---

python/discopt/solver_tuning.py
jkitchin: Panel results, as promised — **your hypothesis was right, and measurably so.**

17 open instances, ladder 5/10/20/40 s, both arms interleaved in one process, 136 solves per panel. A = share applied in the shared gate (what you reviewed); B = share applied only where the caller passes `finder=True`.

| | A: shared gate | B: finder-scoped |
|---|---|---|
| monotonicity violations | 0 / 0 (both arms) | 0 / 0 (both arms) |
| incumbent better / worse / same | 1 / **3** / 64 | **2 / 2** / 64 |
| node count flat higher / lower | 17 / **4** | **20 / 0** |
| certification regressions | none | none |

Scoping removes `nvs05`@20 s's regression (`1107.89 -> 1269.7`, gone), adds a gain (`heatexch_gen2`@40 s, none -> `808844`), and **eliminates every node-count decrease** — throughput now improves strictly wherever it moves at all, 20 up and 0 down. So "three lost incumbents is the shape of over-broad application" was the correct read.

Two survive, both from the pump itself: `nvs05`@10 s (`1269.7` -> none) and `tspn12`@10 s (`262.647` -> `282.244`).

That makes your `_mean_heur_nlp_cost()` point the standing suspect, so I have now implemented it (`40c78d0`). Losing an incumbent *while gaining nodes* is precisely the signature of "remaining > 4× worst-case-ever": the pump is refused, the tree runs longer, and nothing finds the point. The share now divides a running **mean** (`_typical_heur_nlp_cost`); the legacy overrun guard keeps the max, because it answers a different question — *could another solve of the worst size still fit?* — where the max is the right statistic. Flag off is unchanged: share 1.0 selects the max, so the legacy rule is byte-identical.

Panel C is running to test whether that recovers `nvs05`/`tspn12` without giving back the 20/0 throughput gain. I'll post it either way; if it does, the graduation case is materially different from what this PR currently claims and I'll say so explicitly rather than quietly re-scoring it.

---

jkitchin:
Follow-up on `40c78d0`: one correction I owe on my own evidence, and one new finding in the estimator swap.

## Correction — my arm-B reproduction was a flawed demonstration (CLAUDE.md §11)

In [the ratchet thread](https://github.com/jkitchin/discopt/pull/1166#discussion_r3940840503) I gave this as the collision case:

```python
deadline=min(
    time_limit * 0.42,
    9e9,
)
```

That snippet is bounded **by its own `9e9` constant** — `_bounded()` sees a `min` with a constant argument and returns `True`, so it never reached the `KNOWN` comparison at all. Its silence did not demonstrate key collision, and you reproduced my snippet and recorded that run in the thread as the verification. The evidence on record is wrong.

**The finding itself holds**, and the fix is validated — I re-ran it with a genuinely unbounded carve whose first source line is still `deadline=min(`, against both heads:

```python
deadline=min(
    time_limit * 0.42,
    _remaining,          # unbounded on both sides -> a real uncapped carve
)
```

```
reviewed head 9fdc5c0 (set-keyed)    -> SILENT     key_state={('solver.py', 'deadline=min(')}
new head 40c78d0 (count-keyed)       -> REPORTED   key_state={('solver.py', 'deadline=min('): 2}
```

So the collision hole was real on the head I reviewed, and the count-per-key fix closes it — for the right reason this time. Worth re-running your arm B with the corrected snippet if you want the thread's record to stand on its own.

While re-running the arms against `40c78d0` I also confirmed the widened `_carries_role1` fires on both spellings the second review named (`0.25 * (_deadline - _now)` and `0.5 * _remaining_budget()` are now REPORTED), the baseline tree is clean, and the unlisted-name arm (`_budget_left * 0.42`) is still silent — which matches the limit the docstring now states honestly.

## New finding — `solver.py:11826`: the estimator change is a swap, not a floor

```python
_cost = _typical_heur_nlp_cost() if _share < 1.0 else _mean_heur_nlp_cost()
```

Whenever `max > 4 × mean`, `mean / 0.25 < max`, so on the finder path the gate now admits a pump with **less** budget remaining than the worst solve already observed. That relaxes the guard the block comment directly above `_heur_nlp_cost` exists for:

> a heuristic NLP can OVERRUN its own `max_wall_time` clamp by ~10 s on the expensive-Hessian class, so once one long solve is observed we must not launch another unless that much budget remains.

One 15 s `heatexch_gen3`-class solve among several ~1 s ones gives `mean ≈ 3`, so the finder gate drops from a 15 s floor to 12 s — precisely the case the max was recorded for.

This is sound (finder entries are primal heuristics, so both admitting and refusing only change which incumbent is found and when), and the flag-OFF path is genuinely untouched — `_share == 1.0` selects the max, byte-identically. But it is a guard relaxation on the ON path, and "flag off is unchanged" does not cover it; §1 is about not weakening a guard on the way to a result.

Expressing it as a floor keeps both properties:

```python
_cost = max(_typical_heur_nlp_cost() / _share, _mean_heur_nlp_cost())
```

That still removes the 4×-worst-case compounding my earlier note described (the result is never above `max / _share`), while never admitting a pump below the observed overrun floor. If panel C's recovery of `nvs05`/`tspn12` turns out to depend on the swap rather than on the mean, that is itself the interesting measurement — it would mean the recovery comes from dropping the overrun guard, not from the better estimator, and those should not be graduated together.

Unrelated and still open from the first review: no `CHANGELOG.md` `[Unreleased]` entry (CONTRIBUTING §8) — `grep 1153 CHANGELOG.md` at head returns nothing, and the PR adds two new env flags.

---

jkitchin:
## Re-review — PR #1166 (on `fee05ae2`)

Verified the fixes in an isolated worktree at the PR head (asserted the test module's `__file__` in the worktree and `_carries_role1` present before measuring).

**F2 — the serious half — is genuinely fixed.** Re-ran my probe-module method against the widened `_carries_role1`, with a negative control added so a scanner that flags everything cannot pass:

```
sub_of_two_names       -> REPORTED
call_remaining         -> REPORTED
deadline_minus_clock   -> REPORTED
NEGATIVE (saturated)   -> silent (good)

real-tree unbounded sites: 10
```

All three previously-missed spellings now report, the saturated control stays silent, and the real tree still yields exactly 10 sites — coverage grew without new debt, as you said.

**Also verified:** F1 — `_heur_stage_deadline` is gone from `python/discopt/` (0 occurrences). F4 — `three_arm.py` parses and `solver._heuristic_entry_share` exists, with the deleted `_finder_entry_share_for` correctly absent. F6 — now a floor (`len(sites) >= 6`) counted by AST. F7 — `ObjectiveSense` is read in the test. F9 — the dead `solve_nlp_pounce` line is removed. F5 — interleaved within each repetition, compared on the per-cell median.

`pytest python/tests/test_1153_budget_monotonicity.py -m "slow or not slow"` — **9 passed, 1 skipped** (346 s).

**On F3, your deferral is the right call and I want to say so explicitly.** Changing the estimator now would invalidate the finder-scoped panel that is already running and leave both competing explanations for `nvs05@10s` unmeasured. That is CLAUDE.md §4 applied correctly — the entry experiment runs before the implementation, and my finding is a hypothesis until that panel discriminates it. Two things worth pinning when the numbers land, so the result is interpretable either way:

- The two hypotheses are not mutually exclusive. The finder-scoping fix and the running-max estimator could each contribute; a panel that improves `nvs05@10s` does not by itself clear the estimator, because the max-never-decays cliff only bites when an expensive NLP is observed *early*. Worth recording how many instances in the panel actually hit that ordering.
- Whatever the panel shows, `_mean_heur_nlp_cost` returning a max while being named `mean` and consumed as one is worth a rename or a comment at the division site, independent of whether the estimator changes.

No new findings. The remaining item is F3, held open deliberately with a measurement in flight — not an outstanding defect.

jkitchin:
Thanks for re-verifying in an isolated worktree — and specifically for adding the **negative control** to the F2 probe. Checking that a saturated carve stays silent is the half I omitted: my own re-verification only showed the three arms now reporting, which a scanner that flags everything would also pass. That is the same "prove the probe fired" discipline the PR argues for, applied to the probe I wrote to enforce it.

Two updates.

**Your naming point is now fixed.** `_mean_heur_nlp_cost` → **`_worst_heur_nlp_cost`**, because you are right that it is causal rather than cosmetic: a function returning a max, named "mean", and consumed as one at a division site is *how* the share came to divide a worst-case-ever by a fraction and read as "4× typical". The docstring now says the max is correct *there* — it answers "could another solve of the worst size still fit?", not "what does one usually cost?" — and records the rename.

**On your first point, which is the sharper one: you are right that the panels do not automatically discriminate.** One thing does help — the two changes are separated by construction, because panel B was finder-scoped with the *max* estimator and panel C is finder-scoped with the *mean*, so B→A isolates scoping and C→B isolates the estimator. But that still only attributes the *effect*; it does not establish the *mechanism*, which is your actual point. If no panel instance ever observes an expensive NLP before the finder entry decision, the max-cliff never bites there and any improvement is something else wearing its clothes.

So I have written `scratchpad/i1153/ordering_probe.py` to record, at every finder-role entry decision, the observed cost state and what each estimator would have decided — the ordering question directly, rather than inferred from the outcome. I will run it on `nvs05`@10 s and `tspn12`@10 s once panel C is off the CPU (running it now would contend with the measurement it is meant to interpret), and report how many instances actually hit that ordering alongside panel C's numbers.

If it turns out none do, I will say so plainly and treat panel C's result as evidence for scoping rather than for the estimator — which would leave the estimator change as a sound-but-unmotivated edit, and I would rather know that than ship it on a coincidence.

Panel C is at 100/136 solves.

---

python/discopt/solver_tuning.py
jkitchin: Panel C and the ordering probe are in, and **your caution was exactly right — the estimator fix is falsified and I have reverted it** (`954f12e`).

**Panel C** (finder-scoped + mean) against **panel B** (finder-scoped + max), same 17 instances, same ladder, arms interleaved:

| | B: max | C: mean |
|---|---|---|
| monotonicity violations | 0 / 0 | 0 / 0 |
| incumbent better / worse / same | **2 / 2** / 64 | 1 / 2 / 65 |
| node count higher / lower | **20 / 0** | 18 / 0 |

No recovery: `nvs05`@10 s and `tspn12`@10 s still regress, and C *loses* B's `heatexch_gen2`@40 s gain. Within noise — but certainly not what the hypothesis predicted.

**The ordering probe is the decisive measurement, and it is your point made concrete:**

```
nvs05  @10 s  first finder decision at index 0, NLPs observed before it: 0
tspn12 @10 s  first finder decision at index 0, NLPs observed before it: 0
```

On **both** regressing instances, in **both** arms, the finder decision fires before any heuristic NLP has been observed. At that moment `max == mean == the 2.0 s default` — the two estimators are **numerically identical** at the only decision that matters. The estimator cannot be the mechanism here, whatever the panel had shown. Had I only looked at the panel I would have read C's 1/2/65 as mild evidence against the change; the probe shows the comparison was empty from the start.

So it is reverted, per §4 — it shipped on a hypothesis and the hypothesis is now measured false. **The rename survives**, which was always the causally interesting half of your finding.

**And the second half of your F3 comment turns out to be the actual finding.** You wrote that at the first call the gate is "simply *more than 8 s left*, on any model, independent of the pump's real cost". That is now not an aside but the measured mechanism: with nothing observed, the rule is `remaining > 2.0 / share` regardless of the model, against a pump that costs 6.4 s. The lever is the **default seed**, not the statistic — which makes it a work-budget question (#912's `WorkBudget` would make the cost knowable in advance rather than guessed) rather than an estimator one.

That is now the top open item in §6.7, backed by a direct measurement instead of an argument. I am leaving this thread open since that item is genuinely unresolved.

---

jkitchin:
## Panel D — and a family-level negative result

Reworked the mechanism from **refusing** the finder stage to **bounding** it, on the reasoning that a productive pump usually succeeds early, so capping its clock should keep the win that refusing throws away. It does — and it costs the throughput.

| | B: refuse entry | D: cap the stage |
|---|---|---|
| monotonicity violations | 0 / 0 | 0 / 0 |
| incumbent better / worse / same | 2 / 2 / 64 | **3 / 2** / 63 |
| node count up / down | **20 / 0** | 5 / 1 |
| certification regressions | none | none |

Capping recovers precisely what refusing destroyed — `nvs05` goes from *losing* its incumbent to finding one at 5 s and improving it at 10 s (`1269.70 → 1107.89`), and `tspn12`'s regression disappears — but throughput drops from 20 instance-rungs up to 5.

**That symmetry is the result, and it is bigger than either arm.** The throughput gain and the incumbent losses have the *same* cause: not running the pump. Refusing buys nodes and costs incumbents; capping keeps incumbents and returns the nodes; every intermediate setting trades along that one line, because a wall-budget policy only decides *how much pump* runs while cost and value both scale with it. **This rules out the entire "gate or bound the pump by wall budget" family** — which is the family §6.1, §6.4, §6.5 and §6.6 were all drawn from — rather than just this instance of it.

So "what remains" collapses from two speculative directions to one supported one: **the pump must become cheaper for the same value, not smaller.** That is #912's `WorkBudget` — deterministic operations instead of seconds, so the cost is knowable in advance rather than estimated at a 2.0 s default that is 3× wrong, and identical on every machine. It is the only remaining lever that changes the cost/value ratio instead of sliding along it.

### Two defects in my own earlier work that this panel exposed

- **§6.4's "capping the clock is inert" was an artifact.** A finder stage is *two* NLP consumers — a multistart seed producing the rounding point (3.27 s) and the pump (3.11 s) — and that change bounded only the pump, leaving the more expensive seed on the global deadline. It capped the cheaper half and measured what you would expect.
- **The first stage cap was Zeno's bound.** Recomputing `now + share × (deadline − now)` per call site gives each call a share of whatever is left *at that moment*, so the total is never bounded. Measured on `heatexch_gen2` at 10 s: per-call grants fell 3.0 s → 1.75 s while total stage wall did **not** move (7.08 s → 7.47 s) — the stage just ran more, cheaper rounds.

Both were caught by verifying the cap actually *bound* before panelling, which is the step §6.4 skipped. I also pushed `7751561` with a failing unit test — my own ratchet correctly flagged the new inline carve and I committed before reading the output; fixed in `80761e0`. And the first version of that fix passed the ratchet **for the wrong reason**, by renaming the watched variable out of the allowlist — review's arm C, with me as the author. `outer_deadline` is now in `_ROLE1_NAMES` and the carve is recorded as a `split`.

**Verification on the current head (`51e785e` / CI green on `80761e0`):** smoke **1119 passed, 15 skipped, 2 xpassed** (7m17s); adversarial **19 passed** (2m47s); ruff and mypy clean.

Neither arm graduates — D still loses `tanksize`@5 s outright and degrades `tspn10`@5 s — so the flag ships **OFF**, in its cap form, which is the more principled mechanism and the one with a net-positive incumbent column.

---

jkitchin:
## Retraction of this PR's claim 2, and #1153's disposition (CLAUDE.md §11)

#1153 was closed by #1192, which found and fixed a root cause this PR's
summary explicitly ruled out. Two claims in the body above are now known to
be wrong and I am retracting them here, where they were published.

### Retracted claim 2 — "the nvs19 reproduction does not reproduce here"

The body says nvs19 "returns `-1098.2` stably at 15/30/60/120 s." It does not.
Re-measured on this machine on `origin/main`, 3 reps interleaved, objective
sd 0.0000:

| budget | objective | bound | wall |
|---|---|---|---|
| 15 s | -1098.2 | -3710 | 15.0 s |
| 30 s | -1098.2 | -1847 | 30.0 s |
| **60 s** | **-1001.2** | **-7133** | **75.6 s** |
| 120 s | -1097.6 | -1104 | 109.0 s |

The 60 s cell is a worse incumbent, a bound five times looser, and a **75.6 s
wall against a 60 s request**. The earlier "does not reproduce" measurement was
taken in a container whose own methodology notes record that a restart moved the
baseline; it should not have been reported as a negative result.

### Retracted claim 3, as it applies to nvs19 — the throughput-coupling mechanism

The body attributes the symptom to role-1/role-2 budget coupling: a larger
`time_limit` inflating role-2 sub-budgets so the search "spends more per node
and covers less tree." On nvs19 throughput does not fall. The native kernel
explored **100,000 nodes at 60 s versus 47,886 at 30 s** — more, not fewer.
The "7,619 nodes" in the issue's table was never the kernel's count.

### The actual root cause

`_try_native_spatial_kernel` (`solver.py:1550`) accepted only `optimal` and
`time_limit` exits. A `node_limit` exit — the kernel stopping on `max_nodes`
holding a feasible incumbent and a rigorous bound — was discarded wholesale,
and the Python fallback re-ran the search from scratch with the wall budget
already spent. Which exit the kernel takes is budget-dependent: at 30 s the
clock binds first (`time_limit`, accepted); at 60 s the larger grant reaches
`max_nodes=100_000` first (`node_limit`, discarded). That is why more budget
produced a worse answer. The one-line accept-list widening in #1192 fixes it:

| budget | before | after |
|---|---|---|
| 60 s | -1001.2, bound -7133, 75.6 s | **-1098.2, bound -1472, 57.6 s** |
| 120 s | -1097.6, bound -1104, 109.0 s | **-1098.2, bound -1472, 58.1 s** |

Notably `solver.py`'s own #764 note had already written the behaviour down as a
benchmark-panel nuisance without recognising it as a live defect.

### What I am not retracting

The three budget-carve findings stand on their own evidence and are unaffected:
the 38-site inventory and its static ratchet, the `bchoco*`/`heatexch_gen*`
3-nodes-at-every-budget family with the feasibility-pump stage attribution, and
panel D's family-level negative on the wall-budget fix family. Those were the
durable output and they remain correct.

### Disposition

This PR does not close #1153 and #1153 is now closed, so it should not be merged
as-is under the repo's contributes-to rule. Its two flags are default-OFF and its
mechanism half is superseded. My recommendation is to close it and, if the ratchet
and the monotonicity harness are wanted, land them as a separate scoped PR that
closes an issue of its own — but that is the owner's call, not mine, so I am
leaving the PR open.

---

PR #1167, opened 2026-09-05
fix(milp): rewrite root cuts into structural space before appending

jkitchin:
## The defect

`separate_gomory_cols` (`crates/discopt-core/src/lp/gomory.rs:215`) loops
`for j in 0..n` over the **full** working width, so every separated cut carries
coefficients on every column — base slack columns and earlier cuts' surplus
columns included. That is not a cut-quality problem; it structurally breaks node
warm starts, and the chain is mechanical:

1. A coefficient on column `ns + r` puts a second nonzero into that column, so it
   is no longer a singleton.
2. `solve_lp_cols_warm`'s basis reconstruction
   (`lp/simplex/primal.rs:2599-2640`) builds `slack_for_row` by requiring
   `rows.len() == 1`. With no singleton substitute, the returned basis is **short**.
3. `PreparedDual::prepare` (`lp/simplex/dual.rs:521,532`) refuses on
   `basic_vars.len() != m`, increments `DualPrepRejectShape`, and the node
   **cold-solves**.
4. The cold solve returns another short basis — **self-sustaining**.
5. Second consequence: `gomory.rs:236` bails out on a short basis
   (`if basis.basic_vars.len() != m { return cuts; }`), so a short basis also
   **silently disables GMI separation entirely**.

So the cut path was paying for cuts it then could not warm-start against, and
past some depth was not separating at all.

## The fix

Rewrite every cut over structural columns only, *before* anything downstream sees
it, using the equality identities the LP rows already supply:

- base row `r`: `s_r = (b_r − A_r·x) / alpha_r`
- appended cut `i` (stored as `c_i·x − s_i = rhs_i`, `s_i ≥ 0`): `s_i = c_i·x − rhs_i`

Both are **exact** — identities, not relaxations. Doing the rewrite before
*selection* (rather than at append time) means selection scores, dedup signatures
and the cleanup's tightness test all see one consistent space, and no cut row ever
references another row's slack.

This is the design HiGHS uses: `HighsTransformedLp.cpp:447,469-482` expands slack
coefficients back into structural ones before the cut is generated, so HiGHS cuts
never reference slack columns at all.

Two follow-on steps genuinely weaken, both bounded and both mirroring HiGHS:

- **Small-coefficient removal** — threshold `100 * tol * max(maxabs, 1e-3)`
  (`HighsCutGeneration.cpp:783`).
- **Shortening** when nnz *grows* past `100 + 0.15·ns` (`:982-1012`; HiGHS
  shortens rather than drops). The cap is `max_len.max(nnz_before)`, so it bounds
  *inflation from substitution*, not intrinsic length — a 300-item knapsack
  separates ~300-nonzero GMI cuts by nature and must not be penalised for it.

For a `≥` cut, dropping term `k` subtracts that term's **maximum** over the box
(`a > 0 ⇒ rhs -= a·u[k]`), so the result is implied by the original. HiGHS's rule
looks inverted only because its cuts are in `≤` form.

**Why the root box is the right box.** Both weakenings use the root `l`/`u`, and
that is what makes them valid tree-wide: every node's box is a subset, so a
relaxation sound on the root box is sound on all of them. This holds because those
bounds come from presolve/FBBT, which are feasibility-based and globally valid —
reduced-cost fixing (`milp_driver.rs:2256`) is objective-based and runs *inside*
the node loop, strictly after. Moving any objective-based tightening ahead of the
root cut loop would invalidate this argument and needs re-arguing here.

## A false-bound hole found in review, not by a test

`BaseRows`'s CSR spans structural columns only. Column-singleton-ness of `ns + r`
does **not** imply row `r` contains only one slack: if some column `ns + q` is not
a singleton and reaches into row `r`, then `slack_coef[q]` is correctly `None`
while `slack_coef[r]` still looks fine. The rewrite would then silently drop that
second slack term and emit a cut **not implied by the LP** — a false-bound class of
bug. `BaseRows::build` now carries a per-row slack tally and refuses the rewrite
unless the row has exactly one.

Measured inert on this panel — 7,803 cuts across all 38 instances,
`SubstDropNoSlack` = 0 — but refused rather than assumed, with
`a_base_row_with_two_slacks_refuses_the_rewrite` pinning it.

## Measurements

38 MIPLIB 2017 "easy" instances, matched `gap_tol = 1e-4`, arms interleaved per
instance, **190/190 runs, zero certification aborts, no false bound on any arm**.

Both target properties now hold at once, which neither earlier design achieved:

| | before | after |
|---|---|---|
| warm-start acceptance | 0.3–1.1 % on cut arms | 100 % on 22/24 arms, 94.8 % worst |
| `DualPrepRejectShape` | 588–822 per arm | 0 on 23/24 arms |
| `fiber` 200×8 | 2986.5 it/node | 99.5 |
| `sp150x300d` 200×8 | 2079.6 it/node | 10.7 |
| `mik-250-20-75-1` 200×8 | 515.3 it/node | 5.6 |
| `fiber` 200×8 root bound | 388985 | 381984 |
| `blend2` 200×8 root bound | 7.14624 | **7.16601** (better) |

Panel totals:

| arm | solved | nodes | kept/gen |
|---|---|---|---|
| base_16x1 | 19/38 | 8,853,388 | 0/0 |
| cut200_noprune | 19/38 | 3,325,608 | 0/0 |
| cut200_prune | 20/38 | 6,459,226 | 1427/4241 |
| cut500_prune | **21**/38 | 4,693,240 | 1432/7673 |
| highs (reference) | 38/38 | 15,417 | 0/0 |

**The sign flips.** Before this change the root cut budget was net-*negative* —
`cut200_prune` 11/38 against base 18/38, losing 7 instances. It is now
net-positive: +1 and +2 over base at 47 % fewer nodes. That was the blocker, and
it is the reason `cut500_prune` is now a CLAUDE.md §5 graduation candidate, which
it was not before. **This PR does not flip that default** — graduation is a
separate change with its own panel.

Wall-clock is deliberately not quoted: the machine carried a load average near 68
during the follow-up work, which invalidates timing (CLAUDE.md §9). Node counts
and counters are load-independent and carry the conclusions above.

## Instrumentation

One rejection counter covered three distinct failures, and `gsvm2rl3` was
mis-diagnosed off it three times running — first as a dynamism-cap problem, then
as the unbounded-coefficient branch, before measurement showed it is the ratio
test acting on cuts the unbounded branch had let through. Split into
`SubstDropUnbounded`, `SubstDropNonFinite` and `SubstDropDynamism`.

The experiment that split it is recorded as **falsified** in
`docs/dev/milp-competitiveness-plan.md`: relaxing both numerical gates recovers 99
cuts on `gsvm2rl3` (root gap 85.16 % → 76.11 %) and improves root gap on 7 of 38
instances, but at a fixed 5000-node budget it is 3 better / 3 worse, 14 solved
either way, 132,724 vs 133,680 nodes — **cert-clean but neutral, so it does not
ship** (the `DISCOPT_CUT_INHERIT` rule). Reverted, with the measurement left in a
comment at the refusal site so it is not re-run.

## Tests

Six new tests in `mod root_cut_budget_tests`, each failing before the change:

- `the_slack_substitution_preserves_the_cut_exactly` — 1600-point identity probe
  with an executed-assertion count.
- `the_slack_substitution_actually_rewrites_the_coefficients` — anti-vacuity.
- `the_slack_substitution_expands_an_earlier_cuts_slack`.
- `the_cut_shortening_only_weakens_and_keeps_the_violation` — 1296-point
  implication probe; the fixture starts as a single slack reference so
  substitution *inflates* it and the length cap actually binds.
- `the_small_coefficient_cleanup_only_relaxes_the_cut` — 1331-point implication
  probe.
- `a_base_row_with_two_slacks_refuses_the_rewrite`.

## Verification

- `cargo test -p discopt-core` — **690 passed, 0 failed**.
- `pytest -m smoke` — **1165 passed, 1 skipped, 2 xpassed**.
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — **19 passed**.
- `cargo clippy -p discopt-core` and `cargo fmt --check` — clean.
- Corpus gate: the 38-instance panel above, 190/190 runs, no false bound and no
  wrong `optimal` on any arm.

## Note on stacking

This branch is built on `perf/milp-root-cut-budget` (#1164), so the diff against
`main` includes that PR's commit. Merge #1164 first.

`crates/discopt-core/src/bnb/convex_kernel.rs:652,1948` has the same slack-space
defect on the MINLP path. Out of scope here and left open deliberately.

Contributes to the MILP competitiveness work; closes nothing.

### DIFF
--- a/crates/discopt-core/examples/milp_bench.rs
+++ b/crates/discopt-core/examples/milp_bench.rs
@@ -162,6 +162,8 @@ fn opts(n_struct: usize, integer_cols: Vec<usize>, tl: f64) -> MilpOptions {
         initial_incumbent: None,
         node_hook_rounds: 0,
         node_hook_cut_cap: 0,
+        root_cut_time_s: None,
+        root_cut_prune: true,
         simplex: SimplexOptions {
             tol: 1e-9,
             max_iter: 100_000,
--- a/crates/discopt-core/examples/par_bench.rs
+++ b/crates/discopt-core/examples/par_bench.rs
@@ -126,6 +126,8 @@ fn opts(inst: &Instance) -> MilpOptions {
         initial_incumbent: None,
         node_hook_rounds: 0,
         node_hook_cut_cap: 0,
+        root_cut_time_s: None,
+        root_cut_prune: true,
         simplex: SimplexOptions {
             tol: 1e-9,
             max_iter: 100_000,
--- a/crates/discopt-core/src/bnb/milp_driver.rs
+++ b/crates/discopt-core/src/bnb/milp_driver.rs
@@ -447,6 +447,47 @@ pub struct MilpOptions {
     /// the budget is spent the hook stops being called and the search runs on the
     /// relaxation it has.
     pub node_hook_cut_cap: usize,
+    /// Wall-clock ceiling for the ROOT CUT LOOP alone, in seconds.
+    ///
+    /// `None` (the value at every existing call site) leaves the loop bounded
+    /// only by `root_cuts` / `cut_rounds` / tailing-off, exactly as before.
+    ///
+    /// A value makes a many-round configuration safe to ask for. Raising
+    /// `cut_rounds` is how a weak root relaxation gets closed — on a lot-sizing
+    /// model the default single pass leaves the root bound at a fraction of the
+    /// optimum — but the separation phase is worth several seconds only if it
+    /// still leaves time to branch. This is the knob that says how much of the
+    /// budget the root may spend, so the round cap can be set by what the model
+    /// needs rather than by what a worst case can afford.
+    ///
+    /// Checked at the top of each round, so a round already in flight always
+    /// completes and the loop never abandons a partially augmented matrix. It
+    /// therefore bounds *entry* into rounds, not the duration of one round; a
+    /// single very slow round can still overrun it, bounded in turn by
+    /// `time_limit_s` through the LP's own deadline.
+    ///
+    /// This is not the ceiling on the whole solve -- that stays `time_limit_s`,
+    /// which the loop also honors (see the guard at the top of the round).
+    pub root_cut_time_s: Option<f64>,
+
+    /// Drop root cuts that are slack at the final root LP before the tree starts.
+    ///
+    /// Every root cut rides in every node LP for the rest of the solve. A cut that
+    /// is not tight at the root optimum has a zero dual multiplier there, so
+    /// removing it leaves the root bound and its certificate untouched while
+    /// taking a row out of every node — the measured cost of NOT doing this is
+    /// severe (30-instance MILP panel, `root_cuts=500, cut_rounds=50`: node counts
+    /// fell 2.4x but total wall rose from 103 s to 145 s, and `cflp_s101` went
+    /// from 0.04 s to 5.85 s at an unchanged node count).
+    ///
+    /// Deeper nodes may get a weaker bound than they would with the full set —
+    /// this is bound-CHANGING under CLAUDE.md §5, never bound-INVALIDATING, since
+    /// what remains is a subset of globally valid cuts. Kept as a switch so the
+    /// two can be measured against each other in one interleaved process.
+    ///
+    /// Inert unless `cut_rounds > 1`: with a single round every cut was separated
+    /// off the one root solve and is violated there, so nothing is ever slack.
+    pub root_cut_prune: bool,
     /// LP solver options.
     pub simplex: SimplexOptions,
 }
@@ -610,8 +651,7 @@ pub fn solve_milp_node_hooked(
     } else {
         node
     };
-    crate::profile::init_from_env();
-    crate::profile::reset();
+    crate::profile::begin_solve();
     let ns = opts.n_struct;
     let is_int = {
         let mut v = vec![false; ns];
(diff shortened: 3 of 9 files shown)


jkitchin:
## Code review — PR #1167

This stacks on #1164, so I reviewed both commits: the root-cut slack→structural rewrite plus the root-cut prune / `root_cut_time_s` / profile-reset guard.

**The substitution algebra is correct in both directions** — base row `s_r = (b_r − A_r·x)/α`, cut row `s_i = c_i·x − rhs_i` — verified against the tests' worked examples. The two weakening steps subtract the term's *maximum* over the root box, which is the right sign for `≥` cuts. The `slacks_in_row` tally genuinely closes the "column singleton ≠ one slack per row" hole the PR describes. Index bookkeeping for `structural_cuts[i] ↔ column n_base0 + i` matches the append order; short coefficient vectors are handled safely by `select_cuts`, `cut_signature`, and `augment_cols_with_cuts`; and `l_w`/`u_w[0..ns]` really are the presolve root box at the substitution site. Full `cargo test -p discopt-core --lib` passed 6/6 runs (683 tests).

---

### 1. `crates/discopt-core/src/profile.rs:52` — the new `TEST_OVERRIDE` guard makes a measuring test flaky

`begin_solve`'s `TEST_OVERRIDE` early-return drops the per-solve `reset()` but leaves `ENABLED` globally armed, so sibling tests' solves accumulate into the counters a measuring test reads. **Reproduced ~1/25** on `cargo test --lib root_cut`: "27 cut rounds but only 25 warm re-optimizes" when the measured solve has `cut_rounds = 12`. The PR's own new `root_cut_budget_tests` are the racing siblings — they run solves without `profile::test_guard()`. `origin/main` was clean over 25 runs of the same filter.

### 2. `crates/discopt-core/src/bnb/milp_driver.rs:1254` — the `usize::MAX` sentinel is not the backstop the comment claims

The comment asserts a `usize::MAX` sentinel index "would panic rather than corrupt". In release (`overflow-checks` off) `n_base + usize::MAX` wraps to `n_base − 1` — an in-bounds write that silently corrupts a cut coefficient. The keep-sweep at :1217 skips `dep >= ci` while this rebuild indexes it unconditionally, so the two sites disagree about what is reachable. (Same defect as finding 6 on #1164; it should be fixed once, in whichever lands first.)

### 3. `crates/discopt-core/src/bnb/milp_driver.rs:3333` — the `SubstDropDynamism` gate cannot fire at the shipped tolerance

At `tol = 1e-9`, the preceding small-coefficient cleanup guarantees `amax/amin < 1/(100·tol) = 1e7` — exactly the cap. Measured with a probe over 246 cuts spanning 40 decades and six `maxabs` scales: **worst ratio 1e6, counter 0**. The doc comment's "tiny-`maxabs` corner" rationale is closed by the same algebra. So the counter split that the plan doc calls "the shippable part of this experiment" is inert in the shipped arm — either the gate needs a tolerance at which it can fire, or the claim about what it measures should be withdrawn.

### 4. `python/discopt/solver.py:21509` — `root_cut_time_s` reintroduces the #1116 role-2 clock

Set unconditionally, including under `_tuning().deterministic`, which is exactly what the comment 20 lines above forbids: the number of cut rounds — hence the dual bound — becomes machine-speed dependent. Default-off behind `DISCOPT_MILP_ROOT_CUTS`, so low severity, but it blocks graduation in this shape. (Also raised on #1164.)

### 5. `crates/discopt-core/src/bnb/milp_driver.rs:948` — `BaseRows::build` runs before the guard that holds its only consumer

It executes three lines before the `if opts.root_cuts > 0` guard, so a cut-free solve pays a full CSR copy of the matrix (~16 bytes/nnz) and three O(nnz) passes for nothing. Move it inside the guard.

---

Findings 2 and 4 are shared with #1164 — worth fixing once in the base PR rather than twice. Finding 1 is the one that will bite CI.

Issue #698, opened 2026-07-17
MILP fast path silently discards most of the user's time_limit on an uncertified result


jkitchin:
## Summary

`_solve_milp_simplex` bounds the Rust MILP engine to

```python
_milp_budget = max(0.5, min(0.5 * _remaining, _SIMPLEX_MILP_BUDGET_CAP_S))   # cap = 10.0
```

That slice is a **stall** bound (#291: defer fast to the robust fallback when the engine wedges on a pathological reformulation). But on an uncertified `feasible` the function **returns** rather than falling back — so the remainder of the user's `time_limit` is not handed to any engine. It is simply discarded.

Measured (`graphpart_3pm-0334-0334`), stock:

| `time_limit` | wall used | budget discarded | bound |
|---|---|---|---|
| 5 s | 2.67 s | 47% | -54 |
| 10 s | 5.17 s | 48% | -50 |
| 20 s | 10.18 s | **49%** | -41 |
| 40 s | 10.23 s | **75%** | -41 |

At the **default `time_limit=3600`** the engine gets 10 s and returns uncertified.

## Why it is worth fixing

The engine is **monotone in its budget** — more time strictly helps (same instance, run 1 only):

| run-1 budget | bound | nodes |
|---|---|---|
| 2.5 s | -54.00 | 7 |
| 5 s | -50.00 | 63 |
| 10 s | -40.50 | 2047 |
| 20 s | -38.25 | 6311 |

So the discarded budget is real search that would have tightened the bound — and an engine that returned an incumbent is demonstrably *not* wedged, which is the only thing the cap exists to protect against.

## Evidence that spending it pays (TL = 60 s, A/B)

| instance | stock | with re-entry |
|---|---|---|
| `edgecross10-090` | feasible, obj 1411, bnd 1336 | **optimal, 1387** |
| `graphpart_3pm-0334-0334` | feasible, obj -32, bnd -40 | **optimal, -36** (= reference optimum) |
| `edgecross10-040` | obj 173, bnd 48 | obj **148**, bnd **118.5** |
| `graphpart_2pm-0077-0777` | obj -37, bnd -42.5 | obj **-40** (= reference optimum) |

2 newly certified, 5/6 tighter dual bounds, 0 bounds above reference. Wall 147 s → 335 s — i.e. it spends the budget the caller granted.

## Note: the first slice must stay bounded

[...]

PR #1168, opened 2026-09-05
fix(lp): A0'.1 complete a short basis on zero reduced cost, not zero value

jkitchin:
## What this fixes

The primal simplex's basis-export pass completes a *short* basis (one holding
fewer than `m` basic variables, because artificials were expelled) by swapping a
nonbasic row singleton into the empty slot. It required that substitute to sit at
**value zero**. That tests the wrong quantity.

Swapping a nonbasic row singleton `s` into a basic artificial's slot replaces
`B`'s column `±e_r` with `a·e_r` — a scalar multiple, still nonsingular — and
leaves `x` bit-identical *whatever* `s`'s value, because a basic variable may hold
any value and `s` simply keeps the one it has. The primal argument never needed
the value to be zero.

What the swap *can* break is the exported `dual`: a basic column requires
`c_j − yᵀA_j = 0`. So the admissible criterion is the **reduced cost**, and the
value test misses every dual-degenerate case — a slack nonbasic at a nonzero
bound whose reduced cost still vanishes.

## Why it matters (the triple blockage)

A short basis silently disables three mechanisms at once, and **none of them
records a counter**:

| consumer | behavior on a short basis |
|---|---|
| `separate_gomory_cols` | returns empty at `basic_vars.len() != m` (`lp/gomory.rs:233`) — no GMI cuts |
| `PreparedDual::prepare` | refuses the warm start on shape — every node cold-solves |
| dual export | already unusable |

`neos-2624317-amur` exported 338 basic variables for `m = 342`: 50 rows had their
slack nonbasic at a nonzero value with no zero-valued alternative. It therefore
got **no cuts and no warm starts on any node**, reported as `0/0` cuts, which
reads as "nothing to cut".

### Counter deltas, `neos-2624317-amur`, identical settings

`max_nodes=200, gap_tol=1e-4, root_cuts=200, cut_rounds=8, cut_select=True, time_limit_s=30`

| counter | before | after |
|---|---|---|
| `DualPrepRejectShape` | 300 | **0** |
| `DualPrepAccept` | 80 | **191** |
| `RootCutsGenerated` | 0 | **102** |
| `RootCutsKept` | 0 | **25** |

### 38-instance node-limited differential panel (the §5 gate)

`max_nodes=5000, gap_tol=1e-4, root_cuts=500, cut_rounds=8, gmi_cuts=True, cut_select=False, root_cut_prune=True`; the harness aborts on any bound above its `minlplib`/MIPLIB reference optimum.

**Cert-clean.** 76 bound-vs-optimum comparisons executed, zero violations. Solved
count unchanged at 14/38; no `optimal` instance regressed to a limit.

**Bound-neutral in the tree.** Total nodes 133,680 → 133,662 (−0.01 %); 36 of 38
instances are bit-identical in node count *and* iteration count. This is the
expected shape: a warm start does not change an LP's optimum, so it cannot change
the tree.

**Net-positive in LP work.** Total simplex iterations **6,492,081 → 4,057,317
(−37.5 %)**, concentrated in exactly the four instances whose basis was short:

| instance | iters before | iters after | Δ |
|---|---|---|---|
| `neos-3610040-iskar` | 1,898,737 | 29,372 | **−98.5 %** |
| `neos-2624317-amur` | 2,548,348 | 2,018,975 | −20.8 % |
| `neos-911970` | 524,406 | 492,061 | −6.2 % |
| `neos-3118745-obra` | 242,881 | 239,200 | −1.5 % |

The other 34 instances changed by zero iterations, which also rules out run-to-run
nondeterminism as the source of the four deltas.

**Mechanism confirmed on the after side.** A counter probe on the three
newly-measured instances (`max_nodes=300`, same cut settings) shows
`DualPrepRejectShape` absent entirely with `DualPrepAccept` at 240 / 290 / 676,
and `RootCutsGenerated` at 127 / 57 / 31 where the shape rejection previously
suppressed both.

Wall-clock is deliberately **not** claimed: machine load has been 30–70 all day
(CLAUDE.md §9). Iterations and nodes are load-independent.

## Scope notes

- Pass 2 only fills rows pass 1 left empty, so the rule is **strictly monotone**:
  no basis that completes today can be shortened by this change.
- Pass 1's own rule is deliberately left alone. It admits a zero-*valued*
  candidate without consulting its reduced cost, which can hand back a basis whose
  duals differ from the exported `y`. That is pre-existing and benign in effect —
  `gomory` recomputes the vertex and `PreparedDual` re-checks and declines — so the
  cost is a refused warm start, never a wrong bound. Tightening it would be
  strictly *narrower* and could shorten a basis that completes today.
- `convex_kernel.rs:652,1948` carries the same slack-space defect on the MINLP
  path. Out of scope here; it needs the MINLP corpus, not the MILP panel.

## Testing

- `cargo test -p discopt-core`: **686 passed, 0 failed** (683 before, plus the
  three new unit tests — one per commit).
- New regression test `a_nonzero_valued_row_singleton_substitutes_when_its_reduced_cost_vanishes`
  pins all three arms of the rule on a hand-built 3×4 LP: a nonzero-valued singleton
  with vanishing reduced cost is **accepted**, one with reduced cost 5 is **refused**,
  and — via the empty-dual path — the pre-change rule **rejected** the case this adds.
  It fails before the change and passes after.
- `pytest -m smoke`: **1165 passed, 1 skipped, 2 xpassed** in 137 s.
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py`: **19 passed** in 147 s.
- The differential panel above.

## Second commit: the instrument that hid this

`separate_gomory_cols` had **five uncounted early returns**, which is why the
defect survived weeks of panels: "separated no cuts" and "never looked at a single
tableau row" were the same observation. Adds `SepGomoryCalls` (the denominator)
plus `SepGomoryShortBasis`, `SepGomorySingular`, `SepGomoryFtranFail`,
`SepGomoryBtranFail`. The pathology that started this reads as
`SepGomoryShortBasis == SepGomoryCalls`.

Its test asserts counter *deltas* around each call and pins both arms — a short
basis increments the refusal, a complete basis does not — so it cannot pass on a
counter that fires unconditionally, and the asserted-nonzero call counter is what
keeps the test itself from going vacuous (CLAUDE.md §6).

## Third commit: `DISCOPT_PROFILE=0` armed profiling

Same family, found while using the instrument above. `init_from_env` armed on
`var_os(...).is_some()`, so the one value a caller exports to *suppress* the
instrument turned it on — against the convention every other discopt env flag
follows. The instrument costs an `Instant::now()` per phase and an atomic per
counter, so a timing panel that defensively set `DISCOPT_PROFILE=0` was timing the
instrument too (CLAUDE.md §9). Caught doing precisely that to a root-cut
graduation panel, which dumped a phase table it had asked not to collect — the
only reason it surfaced.

Now off when unset or set to `""`/`0`/`false`/`no`; `DISCOPT_PROFILE=1` and every
in-repo caller are unchanged. Its test drives the real predicate against the real
variable over an 11-case table (asserting a re-typed copy of the rule would pass
even if the shipped rule were deleted), holds `TEST_LOCK`, restores the prior
value, and asserts the table was walked.

Contributes to the MILP parity plan (`docs/dev/milp-competitiveness-plan.md`,
Track A0′.1).

**Base note:** this began stacked on #1167, but `ci.yml` only triggers for PRs
based on `main`, so it was getting *no checks at all*. Retargeted to `main`; the
diff therefore also shows #1167's commits until that PR merges, at which point it
narrows to the three commits described here.

### DIFF
--- a/crates/discopt-core/examples/milp_bench.rs
+++ b/crates/discopt-core/examples/milp_bench.rs
@@ -162,6 +162,8 @@ fn opts(n_struct: usize, integer_cols: Vec<usize>, tl: f64) -> MilpOptions {
         initial_incumbent: None,
         node_hook_rounds: 0,
         node_hook_cut_cap: 0,
+        root_cut_time_s: None,
+        root_cut_prune: true,
         simplex: SimplexOptions {
             tol: 1e-9,
             max_iter: 100_000,
--- a/crates/discopt-core/examples/par_bench.rs
+++ b/crates/discopt-core/examples/par_bench.rs
@@ -126,6 +126,8 @@ fn opts(inst: &Instance) -> MilpOptions {
         initial_incumbent: None,
         node_hook_rounds: 0,
         node_hook_cut_cap: 0,
+        root_cut_time_s: None,
+        root_cut_prune: true,
         simplex: SimplexOptions {
             tol: 1e-9,
             max_iter: 100_000,
--- a/crates/discopt-core/src/bnb/milp_driver.rs
+++ b/crates/discopt-core/src/bnb/milp_driver.rs
@@ -447,6 +447,47 @@ pub struct MilpOptions {
     /// the budget is spent the hook stops being called and the search runs on the
     /// relaxation it has.
     pub node_hook_cut_cap: usize,
+    /// Wall-clock ceiling for the ROOT CUT LOOP alone, in seconds.
+    ///
+    /// `None` (the value at every existing call site) leaves the loop bounded
+    /// only by `root_cuts` / `cut_rounds` / tailing-off, exactly as before.
+    ///
+    /// A value makes a many-round configuration safe to ask for. Raising
+    /// `cut_rounds` is how a weak root relaxation gets closed — on a lot-sizing
+    /// model the default single pass leaves the root bound at a fraction of the
+    /// optimum — but the separation phase is worth several seconds only if it
+    /// still leaves time to branch. This is the knob that says how much of the
+    /// budget the root may spend, so the round cap can be set by what the model
+    /// needs rather than by what a worst case can afford.
+    ///
+    /// Checked at the top of each round, so a round already in flight always
+    /// completes and the loop never abandons a partially augmented matrix. It
+    /// therefore bounds *entry* into rounds, not the duration of one round; a
+    /// single very slow round can still overrun it, bounded in turn by
+    /// `time_limit_s` through the LP's own deadline.
+    ///
+    /// This is not the ceiling on the whole solve -- that stays `time_limit_s`,
+    /// which the loop also honors (see the guard at the top of the round).
+    pub root_cut_time_s: Option<f64>,
+
+    /// Drop root cuts that are slack at the final root LP before the tree starts.
+    ///
+    /// Every root cut rides in every node LP for the rest of the solve. A cut that
+    /// is not tight at the root optimum has a zero dual multiplier there, so
+    /// removing it leaves the root bound and its certificate untouched while
+    /// taking a row out of every node — the measured cost of NOT doing this is
+    /// severe (30-instance MILP panel, `root_cuts=500, cut_rounds=50`: node counts
+    /// fell 2.4x but total wall rose from 103 s to 145 s, and `cflp_s101` went
+    /// from 0.04 s to 5.85 s at an unchanged node count).
+    ///
+    /// Deeper nodes may get a weaker bound than they would with the full set —
+    /// this is bound-CHANGING under CLAUDE.md §5, never bound-INVALIDATING, since
+    /// what remains is a subset of globally valid cuts. Kept as a switch so the
+    /// two can be measured against each other in one interleaved process.
+    ///
+    /// Inert unless `cut_rounds > 1`: with a single round every cut was separated
+    /// off the one root solve and is violated there, so nothing is ever slack.
+    pub root_cut_prune: bool,
     /// LP solver options.
     pub simplex: SimplexOptions,
 }
@@ -610,8 +651,7 @@ pub fn solve_milp_node_hooked(
     } else {
         node
     };
-    crate::profile::init_from_env();
-    crate::profile::reset();
+    crate::profile::begin_solve();
     let ns = opts.n_struct;
     let is_int = {
         let mut v = vec![false; ns];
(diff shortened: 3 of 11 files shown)


jkitchin:
## Code review — PR #1168

Scope: the full `origin/main...fix/milp-short-basis-export` diff, which — as the PR body notes — still carries #1167's commits (`milp_driver.rs` root-cut rewrite/cleanup, `lp_bindings.rs` new options) alongside this PR's three (primal basis export, GMI counters, `DISCOPT_PROFILE` parsing).

**The headline change's soundness argument holds.** For a row `r` whose artificial is basic, the exported `y` satisfies `y_r = 0` (the artificial's cost is 0 in the `B^T y = c_B` solve), so the new test collapses to `|c_s| ≤ tol`, the swapped basis `B` keeps column `a·e_r`, and reconstruction from the basis returns `x_s` exactly. `PreparedDual::prepare` independently re-derives `y` and re-checks dual feasibility, and `separate_gomory_cols` re-derives the vertex, so the widened rule cannot by itself manufacture a bound.

---

### 1. HIGH — `crates/discopt-python/src/lp_bindings.rs:1232` (also `:1104`, `:1391`) — `root_cut_prune` defaults to `true` on a graduated path

The new parameter defaults to `true` in all three pyo3 entry points, and no Python caller passes it (`milp_simplex._drive` at `milp_simplex.py:909` and the lazy path at `:1319` do not). The driver's own docstring calls this "bound-CHANGING under CLAUDE.md §5", and it is inert only while `cut_rounds == 1`. But `milp_simplex.solve_milp`'s #1066 escalation is default-ON (`_MILP_CUT_BUDGET_DEFAULT = True`) and runs `_STRONG_CUT_PROFILE` with `cut_rounds = 10` — so root-cut pruning is live **today** on every escalated MILP/OA-master solve, a graduated path whose #1066 numbers (`rsyn0820m`/`0830m`/`0840m`) were taken *without* pruning. There is also no opt-out reachable from Python.

The plan doc's own rule ("every bound-changing item behind a default-off flag with the §5 double bar") and its measured table (which only exercised prune under `DISCOPT_MILP_ROOT_CUTS`) both say this should default to `false`. **This is the same finding I raised on #1164** — it should be fixed once, in whichever of the stack lands first.

### 2. MEDIUM — `crates/discopt-core/src/lp/simplex/primal.rs:1712` — pass 2 accepts a row singleton with no pivot-magnitude floor

Pass 2 takes the *first* row singleton in index order whose reduced cost vanishes, with no floor on `vals[0]`:

```rust
if (self.c[j] - dual[rows[0]] * vals[0]).abs() <= self.tol
```

For a zero-cost column the test is `|c_j| ≤ tol` regardless of `|a|`, so a singleton with `a = 1e-12` is admitted and `B`'s slot column becomes `1e-12·e_r`. That basis is numerically singular; `separate_gomory_cols` then factorizes it and reads `w = e_iᵀB⁻¹` with entries amplified by ~1e12, and the only thing between that and an invalid GMI cut — a false bound — is the `max_dynamism`/`MAX_ABS_COEFF` post-filter.

Contrast `expel_zero_basic_artificials` two functions down, which deliberately requires `|ρᵀA_q| > PIVOT_MIN = 1e-7` and picks the **largest** pivot. Pass 2 should pick `argmax |a|` among admissible candidates and reject below a floor. (Pass 1 shares the gap, but pass 2 materially widens the candidate set — the PR's own measurement shows 50 extra rows completing on one instance.)

### 3. MEDIUM — `crates/discopt-core/src/profile.rs:647` — `test_guard()` leaks `TEST_OVERRIDE` on panic, cascading into "this path never executed"

`test_guard()` returns only the `TEST_LOCK` guard; it does not restore `TEST_OVERRIDE`/`ENABLED` on drop. `set_enabled(true)` raises a **process-global** `TEST_OVERRIDE` that makes `init_from_env` a no-op and makes `enabled()` return the (false) thread-local on every other thread.

If any test panics between `set_enabled(true)` and its trailing `set_enabled(false)` — e.g. the new `a_short_basis_refusal_is_counted_not_silent` failing at `assert_eq!(cuts.len(), 1, …)` (`gomory.rs:432`), which sits *before* its `set_enabled(false)` — the override leaks for the rest of the test binary, and every subsequent counter test reads 0 and fails as "this path never executed". That is precisely the §6 misdiagnosis this PR exists to prevent, turned into a cascade. Make `test_guard` return an RAII struct that restores both flags in `Drop`.

### 4. MEDIUM — `python/discopt/solver.py:21796` — `_cut_opts` is dropped on the #698 re-entry call

`_cut_opts` is applied only to the first `solve_milp_py` call (`:21706`). The re-entry call at `:21796` — which receives the *remaining* budget, typically the larger slice — omits `**_cut_opts`, so with `DISCOPT_MILP_ROOT_CUTS=1` run 2 silently reverts to the binding defaults `root_cuts=16, cut_rounds=1, cut_select=False` and no `root_cut_time_s`. Any flag-ON/flag-OFF panel run through this path measures a **mixture of two configurations** and attributes the result to the flag. (Also: `root_cut_time_s` is derived from `_milp_budget`, run 1's slice, so even if passed it would be the wrong ceiling for run 2.)

### 5. MEDIUM — `crates/discopt-core/src/lp/simplex/primal.rs:1647` — the doc block now documents the wrong function

`select_row_substitutes` was inserted *between* `expel_zero_basic_artificials`'s existing doc comment (1647–1660) and its `fn` (now 1719). So the block "Drive every ZERO-valued basic artificial out of the final (phase-2 optimal) basis via degenerate pivots … `Err` only on a factorization breakdown, which the caller ignores" is now the leading doc for `select_row_substitutes` — a `&self` function returning `Vec<i64>` that performs no pivots and cannot return `Err`. `expel_zero_basic_artificials` is left undocumented, and the argument justifying its `PIVOT_MIN`/`ART_TOL` behaviour is no longer anywhere near it.

### 6. LOW — `crates/discopt-core/src/lp/simplex/primal.rs:1700` — pass 2 uses a phase-1 Farkas ray as if it were a dual

`dual` is not always a phase-2 dual: on `LpStatus::Infeasible` it is the phase-1 Farkas ray (`c₁ = 1` on basic artificials). Pass 2 nevertheless computes `self.c[j] - dual[r]*a`, mixing real objective costs with phase-1 multipliers — not a reduced cost in any space. On an infeasible LP a row with a basic artificial has `y_r = ±1`, so the rule admits or refuses substitutes on an arithmetic coincidence (`|c_j ∓ a| ≤ tol`), on a path none of the docstring's reasoning covers. Gate pass 2 on `status == Optimal`.

### 7. LOW — `crates/discopt-core/src/bnb/milp_driver.rs:1254` — the `usize::MAX` comment is false in release, and the branch is now dead

`coeffs[n_base + newidx[j - n_base]] = a;` is commented "indexing a `usize::MAX` here would panic rather than corrupt". In release (how the solver ships) `n_base + usize::MAX` wraps to `n_base - 1` — a valid index into `coeffs` — so a missed dependency writes a bogus coefficient onto a real column and installs an invalid cut silently. The branch is unreachable today only because `substitute_slacks_to_structural` truncates every cut to `ns` columns, i.e. the whole `j >= n_base` dependency-closure path is now dead. Either assert `newidx[dep] != usize::MAX`, or delete the dead closure now that cuts are structural. (Also raised on #1164 and #1167.)

### 8. LOW — `crates/discopt-core/src/profile.rs:750` — `set_var` during a running test pool is a data race

`discopt_profile_zero_means_off_not_on` calls `std::env::set_var` while `cargo test`'s thread pool is running. `TEST_LOCK` does not protect this: sibling threads run solves that call `init_from_env` → `std::env::var("DISCOPT_PROFILE")` without holding the lock, and concurrent `setenv`/`getenv` is a data race (glibc can and does crash on it). Note this PR *removed* the one pre-existing instance of this pattern from `milp_driver.rs:4987` and then reintroduced it here. Test the predicate against an injected value instead.

### 9. LOW — `crates/discopt-python/src/lp_bindings.rs:82` — `parse_budget_secs` hardcodes the wrong parameter name

It is now also used for `root_cut_time_s` (`:1600`), but its error text is fixed: `root_cut_time_s=-1` or `NaN` raises `ValueError: time_limit_s must be non-negative and not NaN`, naming a parameter the caller did not pass. Take the parameter name as an argument.

---

Findings 1 and 7 are shared with #1164/#1167 across the stack; fix them once at the base. Finding 2 is the one specific to this PR that I would want closed before merge — it is the only path here that can reach an invalid cut, even if the post-filter usually catches it.

PR #1170, opened 2026-09-05
perf(milp): A0 graduate the root cut budget to default-ON

jkitchin:
## What this does

`_milp_root_cut_budget` returned `None` unless `DISCOPT_MILP_ROOT_CUTS=1`, so a
plain `Model.solve()` on a pure MILP got the binding's historical
`root_cuts=16, cut_rounds=1` single pass. **It now returns the budget by
default.** `DISCOPT_MILP_ROOT_CUTS=0` restores the legacy single pass — that is
the arm the panel A/Bs, and CLAUDE.md §5 requires it stay intact and reachable.

This is Track A0 of `docs/dev/milp-competitiveness-plan.md`, plus the A0.1 wiring
step it depends on.

## The graduation panel (CLAUDE.md §5)

38 MIPLIB instances, `gap_tol=1e-4`, 20 s each, both arms interleaved **within**
every replicate, 2 replicates, and the whole panel run **twice** under different
machine load.

| arm | solved | total wall | med(solved) | nodes | dual gap on the 17 open |
|---|---|---|---|---|---|
| off (`root_cuts=16`, 1 round) | 18-19/38 | 424.9 s ±4.2 | 0.158 s | 9,409,627 | mean 0.2028, med 0.1052 |
| on (500 / 50 / select) | **21**/38 | 399.8 s ±1.0 | 0.135 s | **5,353,150** | mean **0.1494**, med **0.0265** |

**Cert-clean.** 304 bound-vs-reference-optimum comparisons across the two runs,
**zero violations**, and no certification regression: `fiber`, `gt2` and
`neos-3611689-kaihu` go `feasible → optimal`, and nothing goes the other way.

**Net-positive.** +3 instances, −43 % nodes, and the dual bound improves exactly
on the plan's §1b pure-*dual* failure family — in the companion node-budget panel
the four `mik-250-20-75-*` go from ~0.14 to ~0.037 at a matched node budget.

## Two things I am not dressing up

**The wall column is not the basis of this verdict.** This machine never idled
below load ~5 — two `myst start` dev servers hold ~100 % CPU each — and a foreign
`pytest` arrived mid-run on the first attempt, peaking at load 58. Per CLAUDE.md
§9 the wall numbers are corroboration only. The verdict rests on solved count,
node count and dual bound, which are load-independent, and which **both runs
reproduced to four significant figures** (0.2028 → 0.1494 in each). Two runs
under different load agreeing exactly is stronger than one quiet run would have
been, but it is not the quiet run the plan asked for, and I am labelling it as
such. `gradpanel.py` now samples load per solve, and `gradscore.py` refuses to
issue a timing verdict when it sees contention, so next time this is recorded
rather than inferred afterwards.

**Cuts buy the hard instances by taxing the easy ones.** ON is *slower* on 11 of
the 18 instances both arms solve — `neos-3611447-jijia` 7.5 → 13.9 s, `enlight8`
5.4 → 10.5 s, `22433` 0.31 → 2.51 s — against 7 faster (`bppc8-02`
3.79 → 1.44 s). Net-positive at 38 instances, but that per-instance tax is
precisely what the plan's **A1** (stall-based termination, cut aging, a cut pool)
exists to remove. A1 should be scored on removing this tax *without* giving back
the +3.

## An A0.1 hazard the panel surfaced

The plan banked `cut500_prune` = 500 cuts / 8 rounds / **no** selection. What
`_milp_root_cut_budget` actually ships is **different**: 500 / 50 / `cut_select=True`
with `root_cut_time_s = max(0.5, 0.5·engine_budget)`. Flipping the default as the
plan described would have shipped a configuration no panel had ever run — the
wiring step silently changes the thing A0 measured.

So I measured all three arms (38 instances, `max_nodes=2000`, rotated per
instance — a node budget because the machine was at load ~21 and nodes are
load-independent where wall is not). The shipped configuration is the **better**
of the two: mean gap 0.1883 vs 0.1919, 19 vs 12 best-arm wins, and 1.62 M vs
1.94 M simplex iterations. `_milp_root_cut_budget` therefore stays as written,
and the new tests **pin that exact configuration** rather than merely asserting
"some cuts are on".

## Verification

- `pytest python/tests/test_milp_engine_routing.py` — **28 passed**. Two of the
  three new tests were confirmed **failing** against the default-off gate before
  the change (`2 failed, 26 passed`).
- `pytest -m smoke` — **1165 passed**, 1 skipped, 2 xpassed. This now exercises
  the new default throughout.
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — **19 passed**.
- `ruff check python/` and `ruff format --check python/` — clean.
- No Rust changed, so no `cargo test` is required for this diff (the Rust on this
  branch is #1167/#1168's, already tested there).

**Base note:** this is stacked on #1168, but `ci.yml` only triggers for PRs based
on `main`, so it targets `main` to get checks. The diff therefore also shows
#1164/#1167/#1168's commits until those merge, at which point it reduces to the
three files above.

### DIFF
--- a/crates/discopt-core/examples/milp_bench.rs
+++ b/crates/discopt-core/examples/milp_bench.rs
@@ -162,6 +162,8 @@ fn opts(n_struct: usize, integer_cols: Vec<usize>, tl: f64) -> MilpOptions {
         initial_incumbent: None,
         node_hook_rounds: 0,
         node_hook_cut_cap: 0,
+        root_cut_time_s: None,
+        root_cut_prune: true,
         simplex: SimplexOptions {
             tol: 1e-9,
             max_iter: 100_000,
--- a/crates/discopt-core/examples/par_bench.rs
+++ b/crates/discopt-core/examples/par_bench.rs
@@ -126,6 +126,8 @@ fn opts(inst: &Instance) -> MilpOptions {
         initial_incumbent: None,
         node_hook_rounds: 0,
         node_hook_cut_cap: 0,
+        root_cut_time_s: None,
+        root_cut_prune: true,
         simplex: SimplexOptions {
             tol: 1e-9,
             max_iter: 100_000,
--- a/crates/discopt-core/src/bnb/milp_driver.rs
+++ b/crates/discopt-core/src/bnb/milp_driver.rs
@@ -447,6 +447,47 @@ pub struct MilpOptions {
     /// the budget is spent the hook stops being called and the search runs on the
     /// relaxation it has.
     pub node_hook_cut_cap: usize,
+    /// Wall-clock ceiling for the ROOT CUT LOOP alone, in seconds.
+    ///
+    /// `None` (the value at every existing call site) leaves the loop bounded
+    /// only by `root_cuts` / `cut_rounds` / tailing-off, exactly as before.
+    ///
+    /// A value makes a many-round configuration safe to ask for. Raising
+    /// `cut_rounds` is how a weak root relaxation gets closed — on a lot-sizing
+    /// model the default single pass leaves the root bound at a fraction of the
+    /// optimum — but the separation phase is worth several seconds only if it
+    /// still leaves time to branch. This is the knob that says how much of the
+    /// budget the root may spend, so the round cap can be set by what the model
+    /// needs rather than by what a worst case can afford.
+    ///
+    /// Checked at the top of each round, so a round already in flight always
+    /// completes and the loop never abandons a partially augmented matrix. It
+    /// therefore bounds *entry* into rounds, not the duration of one round; a
+    /// single very slow round can still overrun it, bounded in turn by
+    /// `time_limit_s` through the LP's own deadline.
+    ///
+    /// This is not the ceiling on the whole solve -- that stays `time_limit_s`,
+    /// which the loop also honors (see the guard at the top of the round).
+    pub root_cut_time_s: Option<f64>,
+
+    /// Drop root cuts that are slack at the final root LP before the tree starts.
+    ///
+    /// Every root cut rides in every node LP for the rest of the solve. A cut that
+    /// is not tight at the root optimum has a zero dual multiplier there, so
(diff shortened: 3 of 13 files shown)


jkitchin:
### CI: `test_hook_sound_and_helpful_on_a_branching_tree` — fixed in cf6e1de1

The correctness lane failed at

```
assert off.node_count > 1, "instance no longer branches; pick a harder one"
E   assert 1 > 1
```

A0 going default-ON let the monolithic Rust MILP engine close that 11-binary
covering instance at the root, so the hook-off arm stopped branching.

Taking the assertion's own suggestion (pick a harder instance) would have hidden
the real defect: **the two arms were never on the same solver.** `solver.py`'s
router deliberately declines to reroute a `lagrangian_bound=True` solve to the
Rust engine (it has no per-node hook), so `off` ran the Rust engine while `on`
ran the Python `_solve_milp_bb`. `on.node_count <= off.node_count` was comparing
the hook against an unrelated engine rather than against itself-minus-the-hook,
and it had been passing on luck.

Probe over 5 instance sizes x both `DISCOPT_MILP_ROOT_CUTS` settings (10/10
comparisons executed and printed):

| arm pairing | result |
|---|---|
| cross-engine (status quo) | n=16 `on=31` vs `off=15`; n=17 `on=15` vs `off=3` — the "can only prune more" claim is false across engines, and **no** instance in the family satisfies it under both cut settings |
| same-engine (`DISCOPT_MILP_ENGINE=0`) | every configuration passes; off `13/11/25/101/89` nodes vs on `1/1/1/31/15` — the hook's real benefit, previously hidden |

The instance is unchanged; only the engine is pinned, via `monkeypatch` so the
setting cannot leak. No assertion was weakened.

Verified locally: `pytest python/tests/test_lagrangian_bb_hook.py` 17 passed (all
markers), `test_milp_engine_routing.py` 28 passed, ruff clean.

Issue #1183, opened 2026-09-05
MILP performance: large branch-and-bound tree on a big-M GDP (about 28 to 46x the nodes of SCIP and HiGHS)


devin-griff:
On a small mixed-integer model, discopt proves the optimum but explores far more branch-and-bound nodes than SCIP or HiGHS and is roughly 15x slower in wall time. The model is a plant-layout generalized disjunctive program reformulated to an MILP with `gdp.bigm` (a 4-way no-overlap disjunction per pair of blocks). All three solvers return the same optimum.

**Node counts** on one instance (8 blocks, 202 variables, 112 binary, 291 constraints), all proving objective 224:

| Solver | Nodes | Wall |
|---|---|---|
| HiGHS 1.14.0 | 109 | 0.6 s |
| SCIP 10.0 | 179 | 0.6 s |
| discopt 0.8.1.dev0 | 4973 | ~9 s |

discopt explores about 28x the nodes of SCIP and 46x of HiGHS. The per-node cost is competitive (about 1.8 ms per node), so the gap is the size of the search tree, not the cost of any single relaxation. A tree that large on a tiny instance points to a weak relaxation and little bound tightening on the big-M structure, which SCIP and HiGHS close with cutting planes and presolve in around 100 to 180 nodes.

This is relevant beyond pure MILP: cut generation and bound tightening also govern the MILP master solves in outer approximation and the relaxations in spatial branch and bound, so the same weakness would compound in discopt's MINLP paths on structured integer models.

**Environment:** discopt 0.8.1.dev0, Pyomo 6.10.1, Python 3.12.3. SCIP via PySCIPOpt 6.1.0 (SCIP 10.0), HiGHS via highspy 1.14.0, both at defaults on the same model exported to MPS.

**Reproducer** (requires only discopt and Pyomo, prints objective 224.0 and nodes 4973):

```python
"""Self-contained reproducer: one small structured MILP (a plant-layout GDP
reformulated with gdp.bigm), solved with the discopt Pyomo plugin. Prints the
branch-and-bound node count. Requires only discopt and Pyomo.
"""
import discopt.pyomo  # registers the 'discopt' Pyomo solver
import pyomo.environ as pyo

# One instance: 1 rack, 5 units, 2 zero-length tie-in headers (8 blocks).
L0 = {1: 20, 2: 6, 3: 4, 4: 6, 5: 2, 6: 2, 7: 0, 8: 0}
[...]

Issue #1173, opened 2026-09-05
perf(milp): node_propagation is plumbed but dead -- no default solve runs domain propagation


jkitchin:
## The defect

Per-node domain propagation is implemented in `milp_driver.rs` (`if ctx.opts.node_propagation` at `:2142`) and plumbed all the way out through `crates/discopt-python/src/lp_bindings.rs`, but all three `#[pyo3(signature = ...)]` blocks carry `node_propagation=false` (`:1101`, `:1238`, `:1406`) and **no Python call site passes the keyword**. So the pass exists, is tested in Rust, and has never run on a single default solve.

This is the same shape as the dead-flag class the development philosophy §3 rules out: it is not a knob a user chose to leave off, it is a mechanism the default path cannot reach.

## Why it matters for the parity work

`docs/dev/milp-competitiveness-plan.md` Stage 0 lists this as a candidate flip, with the note that `fiber` was expected to regress. That prediction was made before A0 graduated the root cut budget default-ON, so the measurement had to be redone against current `main` rather than reused.

## Definition of done

The flag graduates to default-ON, or the measurement kills it and it is removed rather than left as a dead default-off knob. Either outcome needs the CLAUDE.md §5 double bar measured on the 38-instance MIPLIB panel with both arms on the shipped cut configuration, and a regression test that fails with the old default.

PR #1185, opened 2026-09-05
docs(milp): A8/B1/A9 -- the #1183 gap is the search, not the relaxation (+2 retractions)

jkitchin:
## Why this exists

PR #1176 merged at `81960cf3`, before the last five commits on this branch were
pushed. Those five are the whole of the #1183 investigation. This re-opens the
same branch against the new `main`; there is no overlap with what #1176 landed.

## What it adds to `docs/dev/milp-competitiveness-plan.md`

**A8** — degeneracy does not explain A4's bimodality; the explanation is dropped
rather than kept as a story that no longer matches the data.

**B1 / A9 — the #1183 investigation.** Issue #1183 reports a plant-layout GDP,
big-M reformulated, where discopt explores ~28-46x the nodes of SCIP and HiGHS.
It is 291 rows x 202 cols, so hypotheses die in seconds. The reporter's diagnosis
is *weak relaxation and missing bound tightening*. Measured on the same MPS, read
by HiGHS for both solvers:

| | root bound | gap closed |
|---|---|---|
| pure LP relaxation | 71.0 | — |
| discopt root | 169.4 | 64.3% |
| HiGHS root | 198.4 | 83.3% |

Six bound-side levers were then run as pre-registered experiments with kill
criteria. All six failed:

| lever | result |
|---|---|
| single-row implied bounds | derived M tighter on 100% of 180 triples, median +108; all cuts valid; root LP 71.0 -> 71.5. Falsified. |
| full probing (fix binary, FBBT over all rows) | 13 binaries fixed, 5 global tightenings, 470 cuts, root LP 71.0 -> 90.0. Nodes 2667 -> 2737: **worse**. |
| discopt's own root FBBT | tightens 86 bounds, moves the root LP by +0.00 |
| seeding the proven optimum | 1.26x. Falsified. |
| MIR / aggregation c-MIR | exist in `lp/mir.rs`, `lp/aggregation.rs`, not wired into the MILP driver; the root cut loop is already exhausted at 169.4 (24 Gomory calls against a 500x50 budget) |
| every relevant engine switch | best single config 1.98x |

Ablating HiGHS's own mechanisms gives the decisive row: **with its cut pool
disabled HiGHS proves optimality from a root bound of 71 in 516 nodes, while
discopt from a root bound of 169.4 takes 2667.** A crippled HiGHS starting from a
far worse root finishes in a fifth of the nodes. The gap is the **search**, not
the relaxation — which redirects the plan's remaining effort.

## Two retractions carried in the same commits (CLAUDE.md #11)

**A #8 failure, retracted in a banner at the head of A9.** A9a-A9e and all of B1
were first measured in the `wt-probe` worktree, which predates #1173 and so still
had `node_propagation=false`. Every `discopt.__file__` assertion passed — the file
*was* the code under test — but the build's **defaults** were not the shipped
ones, so every baseline was inflated. #8 asks for a marker unique to the version
under test; a path is not that marker when the question is "what does shipped
discopt do". Corrected numbers throughout.

**A baseline mislabel, retracted in A9e.** Two scripts in the same session
reported different "defaults" on the identical build (1543 vs 2667). The
difference is `cut_select`, which `_milp_root_cut_budget`
(`python/discopt/solver.py:21569-21637`) sets True and one script omitted.
Measured on main, same MPS, all proving 224:

```
binding defaults (root_cuts=16, cut_rounds=1)   10765
shipped budget (500/50, cut_select, time cap)    2667
shipped budget with cut_select=False             1543
that plus SB 64/5000                             1109
```

So `cut_select` *costs* 1.7x on this instance. That is the documented shape of
its own graduation ("cuts buy the hard instances by taxing the easy ones"), not a
new defect, and one instance is no basis for touching a corpus-graduated default
(CLAUDE.md #2). The fix is to name the configuration rather than write
"defaults".

Also withdrawn: a `71 -> 163.6` jump attributed to HiGHS's first cut pass (those
log lines are throttled by `mip_min_logging_interval`, default 5 s), and the
restart named as a leading candidate (measured at 166 vs 157 nodes, i.e. 1.06x).

## Verification

Documentation only — no source, no test, no default changed by this PR. Every
number above was re-measured on a clean `main` build with
`assert "/wt-" not in discopt.__file__`, and each probe prints an
executed-assertion count and exits non-zero when it is zero (CLAUDE.md #6). The
HiGHS ablation status-checks every option before using it, since
`setOptionValue` returns an error status and raises nothing for a name it does
not know — six of nine ablation rows were silent no-ops before that guard went
in, and that is recorded in the section too.

Findings posted to #1183: https://github.com/jkitchin/discopt/issues/1183#issuecomment-5555166766

Leaves #1183 open — it is now a localized search-side target rather than a vague
relaxation complaint.


jkitchin:
Added **A10** — the corpus-scale follow-up to A9e, which is the reason this PR's story is now complete rather than open-ended.

A9e left the strong-branching budget as the one discopt-side lever still carrying a measured gain (2667 → 1109 on #1183). Per CLAUDE.md §2 that is a probe, not a result, so it went to the 38-instance MIPLIB panel with a kill criterion written before the run: geomean node reduction over the all-arms-solved set, ≥1.5× survives, <1.25× falsified.

The panel was extended from one dimension to two *before it produced a number*. Its arms varied `sb_node_budget` alone; A9e had by then shown the budget alone is non-monotonic while the gain needs the candidate dimension and saturates at 64. A budget-only sweep would have reported a wash and falsified the wrong hypothesis. The budget-alone arm was kept so that dimension is still measured rather than assumed.

| arm | cands | budget | solved | geomean nodes vs shipped |
|---|---|---|---|---|
| shipped | 6 | 48 | **21/38** | — |
| n50k | 6 | 50 000 | **21/38** | 1.22× |
| c64n5000 | 64 | 5 000 | 20/38 | 1.43× |
| c64n50k | 64 | 50 000 | 20/38 | 1.42× |

**WEAK by the criterion, and it fails the §5 net-positive bar outright.** Cert-clean — 152 bound-vs-oracle comparisons and 82 solved-incumbent-in-band checks executed, zero violations — but not helpful:

- It **loses an instance**: `gt2` goes from optimal in 0.06 s / 2643 nodes to unproven after 20.4 s / 828 901 nodes.
- Total wall over the comparable set is **worse** (29.1 s → 31.1 s) *despite* 1.43× fewer nodes. The per-node SB cost eats the entire node gain.
- Violently bimodal: `fiber` 20.3×, `22433` 5.5×, `dcmulti` 3.0× against `kaihu` 0.59×, `gen` 0.73×.

The `gt2` mechanism is the part worth keeping. In every arm the dual bound is **exactly the reference optimum, 21166.0**; the wide-SB arms fail purely on the **primal** side, incumbent stuck at 72966 while the bound already sits on the answer. Wide strong branching reorders the search so the dive never reaches the good solution.

So the fixed-node-prefix **shape** is wrong, not its size — too small under-informs the pseudocosts, too large starves the primal search, and no constant sits between those two failures, which is what "non-monotonic and bimodal" means. HiGHS does not pick a constant: it budgets SB in LP *iterations*, recomputed at every branching decision (`HighsSearch.cpp:1272-1290`). That is a plausible next design, but nothing here is built on the 1.43× — it would enter as its own hypothesis with its own entry experiment.

Net effect on the plan's ledger for this class: relaxation levers falsified (A9a–A9c), SB budget falsified as a default (A10). What remains untested is node selection, conflict analysis and the restart-and-resolve loop — exactly the mechanisms A9d showed HiGHS keeping when its cut pool is taken away.

jkitchin:
## Code review — PR #1185

Documentation-only (+747 to `docs/dev/milp-competitiveness-plan.md`). I verified every source reference against the branch's own tree — `milp_driver.rs:406/690/717/1434/1786/1793`, `pool.rs:10-18`, `tree_manager.rs:719-724`, all six `branching.rs` `best_estimate` sites, `solver.py:21569`, `lp_bindings.rs:1101`, and the MIR/aggregation non-wiring claim. **Those all check out.** I then re-derived every ratio, percentage and sum in the new tables.

The failures are concentrated in the retraction bookkeeping: several numbers the PR explicitly retracts are still load-bearing elsewhere in the same section.

---

### 1. HIGH — `:1893` — the §8 correction banner repeats the exact mislabel A9e retracts

The banner's "on main (correct)" column gives `#1183 default nodes = 1543`. A9e, 40 lines below at `:2298-2311`, retracts precisely this: 1543 is **not** the default, it is the shipped root-cut budget **with `cut_select=False`**; the configuration a user actually gets is 2667. The banner is the artifact whose entire purpose is to fix contaminated baselines, and it hands the reader a wrong one — anyone computing discopt-vs-HiGHS from it gets `1543/157 = 9.8×` instead of the shipped `17×`.

### 2. HIGH — `:2044` — A10's opening ratio mixes two configurations

> "A9e left the SB budget as the one open discopt-side lever with a measured gain (2667 → 1109 on #1183)."

A9e's own table measures `1543 → 1109 = **1.39×**`. 2667 is the `cut_select=True` shipped arm; 1109 is a `cut_select=False` arm. Pairing them credits the SB budget with `cut_select`'s separately-measured 1.7× cost, inflating the lever from 1.39× to 2.4×. This is the number that justified spending the 38-instance A10 panel, and it is the same cross-configuration error the A9e retraction exists to prevent.

### 3. MEDIUM-HIGH — `:1942` — the "six levers" summary carries the retracted probing number

It lists "full probing (**1.07×**)" two paragraphs after A9c (`:1934`) corrects that arm to **0.56× — actively worse** on main, and after the §8 banner explicitly names 1.07× as the contaminated wt-probe value. The list's framing ("levers that ... do not move discopt's tree") also mischaracterizes a lever that moved the tree *backwards* by 1.8×.

### 4. MEDIUM-HIGH — `:1785` — B1's headline baseline is uncorrected and irreconcilable with the banner

The row `discopt nodes (engine, defaults) | 2199` is a wt-probe number. The §8 banner — which states it covers "A9a–A9e **and all of B1**" — reports the wt-probe default as **2371**. So three different values (2199, 2371, 2667) are each presented as "the default" for the same instance and MPS. Compounding it, B1 sits *before* the retraction banner in the document, so a reader meets the contaminated table with no warning, and every derived figure in B1 (the four 1.43×/1.63× lever ratios, "8.6× behind HiGHS") rests on 2199.

### 5. MEDIUM — `:1965` — arithmetic in the "decisive row" does not follow from its own numbers

> "discopt solves ... in 2667 as shipped. A solver with a *much stronger root* taking **4.6×** more nodes."

`2667/516 = 5.17`, not 4.6. 4.6 is `2371/516` — the retracted wt-probe baseline, left in place when 2667 was substituted into the sentence above it. The A9e conclusion at `:2330-2337` gets this right at "three to five times", so the two statements disagree.

### 6. MEDIUM — `:2404` — A13's unsolved-population count is off by two

> "25/44 solved vs 23/44 with **none lost** ... among the **21** instances neither arm closed, the dual bound improved on 6."

If none were lost, the OFF arm's 23 are a subset of the ON arm's 25, so instances neither arm closed = `44 − 25 = 19`. 21 is `44 − 23`, which double-counts `istra` and `iskar` — named in the very next sentence as going from timed-out to certified optimal — as closed by neither arm. The "improved on 6, worsened on none" claim is therefore scored over the wrong denominator.

### 7. MEDIUM — `:1769` / `:2107` — two new sections both claim the identifier "B1"

The plan already defines `B1. Node selection and plunging` in Track B (`main:702`). This PR adds `### B1 — #1183, an external reproducer` *and* `### Track B1 — node selection`. Downstream text then uses the label ambiguously: A9 opens "B1 falsified four levers" (the #1183 section) while A11's ledger says "A11 and B1 (node selection)", and the §4 table strikes `~~B1~~`. A reader cannot tell which experiment the struck gate row refers to.

### 8. MEDIUM — `:2324` — a measured no-op is presented as a distinct panel arm with a better result

The table lists "+ neighborhood dedup" at 7 better / +44.9 % wall against "stride 16, no backoff" at 6 better / +45.9 %. But the text at `:2340` reports `RinsSkippedDup = 0` over 1422 considerations — the guard never fired once, so the arm executed the identical code path. The 7-vs-6 difference is run-to-run variance (the doc concedes large spread at 20 s at `:2358`), yet the table reads as evidence dedup helped, which is the opposite of the paragraph's own conclusion that the guard was removed.

### 9. MEDIUM — `:2062` — A10's headline 1.43× excludes the arm's worst failure by construction

The pre-registered geomean is taken "over the instances **every arm** drives to optimality" (20 instances). `gt2` — where the wide-SB arm goes from 0.06 s / 2643 nodes to *unproven after 828,901 nodes* — is solved only by the shipped arm, so it drops out of the geomean entirely. That is the solved-only survivorship bias CLAUDE.md warns about: "a solved-only statistic flatters whichever solver times out most". The doc notes the lost instance qualitatively but still reports 1.43× uncorrected, and A10's next section ("nothing is built on the 1.43×") then treats it as a real magnitude.

### 10. MEDIUM — `:2369` — A13 graduates a default-ON change into ground Track C marks "Do not spend here"

Track C (`main:757-762`, untouched here) records: "Integral-objective cutoff tightening — **already built and wired** ... Both source reviews flagged HiGHS's `computeNewUpperLimit` as a possible gap; **it is not one**." A13 extends exactly that detector, measures −11.6 % wall / −7.8 % nodes / +2 solves, and ships it default-ON — without amending Track C. The plan now simultaneously instructs future work to skip this ground and to have taken it.

### 11. LOW-MEDIUM — `:2350` — a pre-registered kill criterion fired and the mechanism was kept

> "the criterion registered was 'retain ≥ 6 of the 7 improved instances and wall ≤ 1.10'. It retained **4**. The backoff is kept."

Under CLAUDE.md §4 a criterion that fires kills the hypothesis or forces re-registration. The justification offered (large 20 s run-to-run spread) is itself an argument that the panel could not test the criterion — which argues for re-running, not retaining. The retained backoff is then the arm declared to "pass both §5 bars" at `:2363`, so that eligibility claim inherits the unresolved criterion.

### 12. LOW — `:1802` — stale line references

`validate_seed_incumbent` is cited as "rejects silently by design (`milp_driver.rs:3729`)". In the PR branch's own tree the function is defined at **:3651** and called at **:824**; line 3729 is inside `try_dive_repair`'s objective accumulation. Same at `:2221`, citing `milp_driver.rs:2678` for `c.2 < ctx.reliability` (actual **:2657**) — that one is inherited from `:1657`, but this PR propagates it into a new section.

### 13. LOW — `:2452` — the §4 gate row keeps its target after removing a contributor

`| ~~B1~~ + B3 | 33/38 | ... The target now rests on restarts alone and is not carried forward on B1's share`. Striking B1 while leaving 33/38 unchanged makes 33/38 a B3-alone target, contradicting the sentence that says B1's share is not carried forward. Either the number moves or the row should state the restart-only target explicitly.

---

Nothing here touches solver behaviour, so there is no certificate risk. The issue is that this is a *correction* document in which several retracted measurements remain in force elsewhere in the same section, and two arithmetic results (`4.6×`, the 21-instance population) do not follow from their own inputs. Findings 1 and 2 are the ones to fix before merge — a retraction banner that restates the error it retracts will mislead exactly the reader who goes looking for the correction.

Issue #1060, opened 2026-08-17
mip-nlp: lp_nlp_bb hard-requires Gurobi, so the fastest convex-MINLP path cannot be the default route


jkitchin:
`mip_nlp_method="lp_nlp_bb"` is the fastest convex-MINLP algorithm in the tree —
1.5-3.4 s to optimality on instances where the default path burns 60 s and lands
14-88% off — but it refuses to run without Gurobi:

```
rsyn0820m02m  lp_nlp_bb  milp=highs  RAISED RuntimeError:
  mip_nlp_method='lp_nlp_bb' requires milp_solver='gurobi' because LP/NLP
  branch-and-bound uses single-tree lazy constraint callbacks. Backends 'auto', ...
```

The refusal itself is correct behaviour (CLAUDE.md §3: refuse loudly rather than
silently degrade). The problem is what it implies for #1059: the natural
auto-route target for convex MINLPs cannot be the default, because it depends on
a commercial solver. In this environment the Gurobi license is additionally
size-limited, so even when it is selected it fails on larger models:

```
rsyn0820m02m  lp_nlp_bb  RAISED GurobiError: Model too large for size-limited license
```

## Why it matters

Plain `oa` is free but materially weaker on the same set (60 s budget):

| instance | `oa` (free) | `lp_nlp_bb` (Gurobi) |
|---|---|---|
| rsyn0805m | optimal 9.8 s | optimal 2.0 s |
| syn40m | optimal 59.2 s | optimal 1.5 s |
| syn20m02m | optimal 23.5 s | optimal 2.3 s |
| rsyn0840m | **unknown** | optimal 3.4 s |
| rsyn0820m02m | **unknown** | (license limit) |

So routing convex MINLPs to `oa` fixes roughly 6 of the 10 externally reported
rows; closing the rest needs single-tree LP/NLP branch-and-bound on a backend
that ships with discopt.

## What "done" means

Single-tree LP/NLP-BB (Quesada-Grossmann) available without Gurobi — the master
MILP solved by the in-house Rust B&B (or HiGHS) with a lazy-constraint /
callback hook at integer-feasible nodes, so OA cuts are added inside one tree
instead of re-solving the master per iteration. Gurobi stays supported as a
backend; it stops being a requirement.

Acceptance: `lp_nlp_bb` with `milp_solver` set to a free backend reaches
`optimal` on `rsyn0840m` and `rsyn0820m02m` inside 60 s, matching the published
[...]

PR #1186, opened 2026-09-06
perf(milp): A12 -- RINS improvement heuristic behind DISCOPT_RINS (default off)

jkitchin:
## What

Adds **RINS** (Relaxation Induced Neighborhood Search) as the driver's first
*improvement* heuristic, behind `DISCOPT_RINS` and **default off**.

## Why — the measurement, before the code

The dual-side ledger in `docs/dev/milp-competitiveness-plan.md` is exhausted:
A1 (cut aging), A2 (in-tree aging), A4 (SB budget), A6→A7 (pseudocost default),
A8 (SB discrimination), A9a–A9c (relaxation), A10 (SB budget at corpus scale),
A11 (SB gate shape) and B1 (node selection) are **all falsified**. What A5
located instead is on the **primal** side: on the 38-instance parity panel,
17 instances were unsolved at an **89.6 % median primal share**, and **12 of
those held a merely poor incumbent rather than none**. discopt has a rounding
heuristic (`try_rounding_csc`) and a no-incumbent dive (#1060, which stops
firing the moment an incumbent exists) but **no improvement heuristic at all** —
its first incumbent is very nearly its last.

A12 ran RINS's precondition as an **entry experiment on the real corpus before
any implementation** (the #727 RLT lesson: synthetic root-gain 0.68, real gain
0.0). At every node with an LP solution, a real incumbent and ≥1 integer
variable, it measured the fraction of integer variables where the node LP value
equals the incumbent value within 1e-6:

```
INSTANCES RUN: 38
EXECUTED: 2023762 incumbent-present nodes measured
  RinsFixLt30            28,746    1.4%
  RinsFix30to50         312,942   15.5%
  RinsFix50to70         714,150   35.3%
  RinsFix70to90         934,272   46.2%
  RinsFix90to99          33,275    1.6%
  RinsFixGe99               377    0.0%
USABLE BAND [0.30, 0.99): 1,994,639 / 2,023,762 = 98.6%
VERDICT: SURVIVES
NON-VACUITY: 33/38 instances contributed at least one node
```

Pre-registered: ≥30 % SURVIVES, 10–30 % WEAK, <10 % FALSIFIED. Split by
outcome, the precondition holds on the population the gap actually lives in:
**unsolved 17 instances / 1,792,775 inc-nodes / 98.4 % in band** (solved:
21 / 230,987 / 99.7 %). 15 of the 17 unsolved carry an incumbent, all at ≥84 %.

## Mechanism

Danna, Rothberg & Le Pape, *Exploring relaxation induced neighborhoods to
improve MIP solutions*, Math. Program. 102(1), 2005. At a batch boundary, fix
the integer variables on which the most promising node's LP relaxation and the
incumbent already agree, and solve the remainder as a small MILP. The agreement
test is the standard exact-equality-within-tolerance test, **not** a rounding
test; the 0.30 refusal threshold is the customary minimum fixing rate.

## Soundness

Purely primal: RINS can only propose an incumbent, never move a dual bound, so
it cannot loosen or falsify a certificate. Three concrete guards rather than an
argument:

1. the sub-MILP is a **restriction** of the **base** problem — same matrix, same
   rhs, integer bounds only narrowed to values the incumbent already takes. A
   pin that would fall outside the base box is **declined, not widened**;
2. the returned point is nevertheless **re-validated** through
   `validate_seed_incumbent`, the same gate a caller-supplied seed passes, which
   re-checks integrality and bounds, verifies every original row and
   reconstructs the slacks itself. A rejected point is dropped and counted in
   `RinsRejected`, **which must stay zero** — trusting the sub-solve instead
   would turn a bug there into a false `Optimal` at the top level;
3. it runs against the **base** problem snapshotted before the root cut loop,
   not the cut-augmented working matrix, so a wrong cut could not launder itself
   into an incumbent.

RINS declines entirely **under a lazy separator** (on that path a point must be
screened before it may become the incumbent, and bypassing the screen would
enforce a model the caller did not ask for) and **inside a sub-solve**.

## Cost control

A flat per-call node cap leaves the **total** unbounded, because attempts scale
with the batch count. `rins_node_budget` gives RINS 5 % of the parent's nodes
plus a flat 500 allowance, capped at 500 nodes per attempt, and attempts are
spread on a batch stride. Sub-solve time is also clamped to a quarter of the
parent's remaining budget, so a heuristic can never convert itself into a
timeout.

A RAII `MilpDepthGuard` keeps a nested sub-solve from calling
`profile::begin_solve`, which **resets** the counters — a nested reset would
wipe the parent's measurement mid-run and read as "this path never executed"
(CLAUDE.md §6). RAII specifically because the driver has many early returns and
a manual decrement would be missed by one of them.

`DISCOPT_RINS` **refuses an unrecognized value loudly** rather than silently
picking an arm: a panel that exports `DISCOPT_RINS=on` and is quietly handed the
OFF arm produces a tidy A/B table in which both arms ran the same code.

## Testing

- `cargo test -p discopt-core` — **694 passed, 0 failed**, all integration
  binaries green.
- Two new regression arms, separate binaries because `DISCOPT_RINS` is
  process-wide, sharing one fixture so they cannot silently drift onto different
  models:
  - `rins_on_improves_incumbent` — a capacitated facility-location instance
    seeded with a feasible but deliberately poor incumbent (objective 230,
    optimum 148). Asserts RINS ran, **improved** the incumbent,
    **rejected nothing**, and still certifies the optimum. **Fails before this
    change**, because RINS never runs.
  - `rins_default_off` — with the flag unset, asserts RINS is never even
    reached, and independently pins the optimum the ON arm asserts against so
    that assertion is not circular.
- `parse_rins_flag` and `rins_node_budget` are pure and unit-tested directly, so
  the refusal and the quota are asserted rather than inferred from a solve.
- `cargo clippy -p discopt-core --all-targets` — no new warnings.
- `pytest -m smoke` and the adversarial suite: results appended below once the
  Python extension build finishes.

## Status

**Default off.** Per CLAUDE.md §5 this stays off until the ON-vs-OFF
differential panel over the corpus clears **both** bars: cert-clean
(`incorrect_count = 0`, no bound above its reference optimum, no certification
regression, incumbents independently feasibility-verified, and `RinsRejected`
exactly 0) **and** net-positive. That panel is running; its numbers will be
posted here, and the flag does not graduate on this PR.

Contributes to the parity plan's Track B2.

### DIFF
--- a/crates/discopt-core/src/bnb/milp_driver.rs
+++ b/crates/discopt-core/src/bnb/milp_driver.rs
@@ -361,6 +361,7 @@ fn dive_batch_eligible(
 }
 
 /// Options for the MILP driver.
+#[derive(Clone)]
 pub struct MilpOptions {
     /// Number of structural (model) variables; columns `[n_struct, n)` are slacks.
     pub n_struct: usize,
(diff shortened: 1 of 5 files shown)


jkitchin:
## A12 graduation-gate panel — results

Four interleaved ON-vs-OFF panels over the 38-instance MIPLIB set at 20 s/instance
(the same regime A5/A10/A11/A12 used). Arms run **per instance, OFF then ON**, so
machine-load drift lands on both rather than on one (CLAUDE.md §9). Every incumbent
is re-verified in numpy against the original model — column bounds, integrality, row
activities, objective recomputed from `c` — and *not* through the Rust
`validate_seed_incumbent` that RINS itself calls, so the check cannot inherit a bug
in the thing it is checking.

### Bar 1 — cert-clean: PASS, all four runs

0 violations. `RinsRejected = 0` on every instance-arm, i.e. RINS never once proposed
a point the driver could not independently verify. No dual bound above its reference
optimum, no certification regression, no incumbent beating the oracle. 72–74
incumbents verified per run.

### Bar 2 — net-positive: mixed, and the first run's PASS was wrong

The net-positive predicate I first wrote gated on solved-count and primal quality and
**omitted wall time**, so it reported PASS while both-solved wall regressed 45.9%.
CLAUDE.md §5 names wall explicitly; the bar now includes it (`wall_ratio <= 1.10`).
Retracting that first PASS.

| mechanism | better / 17 unsolved | worse | both-solved wall | both-solved nodes |
|---|---|---|---|---|
| stride 16, no backoff | 6 | 0 | +45.9% | −14.9% |
| + neighborhood dedup | 7 | 0 | +44.9% | −14.9% |
| backoff, cap 6 doublings | 4 | 0 | +8.9% | +0.8% |
| backoff, cap 2 doublings | 4 | 0 | +6.4% | −6.2% |

### A hypothesis that was falsified, recorded rather than deleted

I attributed the wall cost to RINS re-solving identical neighborhoods (121 runs /
0 improvements on `enlight8`, 68/4 on `neos-3611689-kaihu`, 49/2 on `bppc8-02`) and
built a dedup guard keyed on the fixed columns and their pinned values.

**The panel measured `RinsSkippedDup = 0` over 1422 considerations.** RINS is invoked
on a different node's LP each batch, so the agreement set differs even while the
incumbent is frozen — those were 121 genuinely distinct neighborhoods, not repeats.
Wall was unchanged (43.2 s → 43.0 s). The guard was dead code by measurement and is
not in the branch. The real diagnosis is in the funnel: **1401 runs → 35 improvements,
a 2.5% hit rate**, at 107,933 sub-solve nodes. Low yield, not duplicated work.

### What shipped instead

A failure backoff: the stride doubles on each attempt that does not improve the
incumbent and resets on every one that does, so an instance where RINS is working
keeps firing at the base rate while one where it is not fades out.
`rins_backoff_stride` is pure so the schedule is asserted directly rather than
inferred from a process-wide counter (§6); `RinsBackoffSkips` proves it engaged.

The cap was measured, not chosen: at 6 doublings the stride reaches 1024 batches,
which inside a 20 s solve disables the heuristic outright — that run lost all three
`mik-250-*` gains. At 2 it tops out at 64 batches, a quarter rate.

`enlight8` 121 → 6 runs (2.26 s → 2.17 s); `neos-3611689-kaihu` 68 → 11 runs
(+8.80 s → +0.06 s) while keeping its node reduction (97269 → 71765).

### The pre-registered kill criterion FAILED

Before running the backoff panel I registered: *retain ≥ 6 of the 7 improved
instances and wall ≤ 1.10*. It retained **4**. Reporting that as a failure.

The backoff is kept because it removes a pathological cost without ever worsening an
incumbent, but I make **no claim** that it preserves the primal gains. At a 20 s limit
the run-to-run spread is large enough that 4 vs 7 is not a clean mechanism comparison
— `enlight_hard` found an incumbent in three runs and not the fourth, and the
identity of the improved instances shuffles between runs.

### What is solid across all four runs

- Cert-clean every time; `RinsRejected = 0` in 152 instance-arms.
- **RINS never made an incumbent worse — 0 WORSE, in every run.**
- It improves 4–7 of the 17 instances that time out, and the improvements land *on*
  the reference optimum: `beavma` 593880 → **383285** (opt), `mik-250-20-75-2`
  −29347 → **−50768** (opt), `mik-250-20-75-3` −40559 → **−52242** (opt),
  `nexp-50-20-1-1` 45 → 29/30 (opt 29).

That is the A5 gap this was built for: 17/38 unsolved at an 89.6% median primal
share, 12 holding a merely poor incumbent rather than none.

### Recommendation

**`DISCOPT_RINS` stays default OFF.** Both bars passing is *eligibility* to graduate,
not a decision, and the primal criterion I set for the backoff failed. Flipping the
default is a call for the owner with the table above in hand. The trade on offer:
roughly +6% wall on a both-solved set whose total is ~30 s (median 0.13 s/instance)
in exchange for materially better incumbents on instances that hit the time limit.

### Verification

- `cargo test -p discopt-core` — **695 passed, 0 failed** (all integration binaries green)
- `cargo clippy -p discopt-core --all-targets` — no new warning; the 4 remaining errors
  are pre-existing PI-constant lints in `expr.rs`/`nl_parser.rs`, confirmed by `git stash`
- `pytest -m smoke` — **1182 passed, 1 skipped, 2 xpassed**
- Build §8-verified before every panel: `discopt.__file__` under `wt-rins`, the counter
  set asserted present, and the superseded `RinsSkippedDup` asserted *absent* after the
  rename

jkitchin:
## Code review — PR #1186

True scope is the last two commits against `origin/main`: RINS in `crates/discopt-core/src/bnb/milp_driver.rs` (+456), the new profile counters, and two test binaries. I read the whole diff plus the surrounding driver — node export paths, `validate_seed_incumbent`, `inject_incumbent`, options cloning, deadline plumbing.

**The soundness spine holds up.** The base snapshot really is taken pre-cut-loop; `MilpResult.x` is exactly `n_struct` so it matches `validate_seed_incumbent`'s length gate; objective accounting is structural-only on both sides, so `sobj + obj_const` is consistent with `inc_obj`; `inject_incumbent` independently enforces strict improvement; and the deferred-node record (`solution: Vec::new()`) is correctly excluded by the length check.

---

### 1. `milp_driver.rs:2159` — candidate selection prefers nodes whose `solution` is an untrusted box midpoint

RINS picks its candidate with `min_by(lower_bound)`, which always prefers `lower_bound = -inf` records. Those are exactly the records that carry an explicitly untrusted **box midpoint** as `solution` (IterLimit/Numerical exits, and unverified-Farkas nodes). Any batch containing one such node makes RINS run its agreement test against a fabricated point rather than a relaxation solution — so the neighbourhood it fixes is derived from noise. Filter to records with a finite `lower_bound`, or tag midpoint records and skip them.

### 2. `milp_driver.rs:1205` — RINS declines under a lazy hook but not under a fractional `MilpNodeHook`

The base snapshot excludes hook-contributed rows. Under a lazy hook RINS correctly declines; under a fractional `MilpNodeHook` it does not, so an injected incumbent can violate hook rows where an ordinary incumbent cannot. Reachable from Python with `node_callback` set and `lazy_callback` unset. The same decline should cover both hook kinds.

### 3. `milp_driver.rs:742` — the sub-solve can run unbounded when the caller passes no time limit

The sub-solve time clamp sits inside `if let Some(d) = deadline`. With `time_limit_s: None` the sub-MILP gets no time limit at all, and the inherited `cut_rounds` / `root_cuts` / `root_cut_time_s: None` mean `max_nodes = 500` bounds nothing before the first node is reached. A root cut loop with no clock and no node pressure can then run arbitrarily long inside what the caller believes is one node's work.

### 4. `milp_driver.rs:2186` — `inject_incumbent`'s return is discarded and `RinsImproved` is incremented before it

A marginal improvement that `validate_seed_incumbent` then wipes out via snap/clamp still counts as a hit, and still resets the failure backoff. So the counter overstates RINS's success rate and the backoff — the mechanism that is supposed to stop RINS retrying a barren neighbourhood — is defeated by exactly the failures it exists to detect. Increment on the returned boolean.

### 5. `milp_driver.rs:774` — sub-solve nodes never reach `tm`

`MilpResult.nodes` and `opts.max_nodes` therefore exclude all RINS work. The PR pre-registers an ON/OFF node-count comparison as its evidence, and that comparison is biased toward the flag: the ON arm does strictly more work per reported node. Either accumulate sub-solve nodes into `tm`, or report the comparison on wall time and say explicitly that node counts are not comparable across arms.

### 6. `milp_driver.rs:1204` — `rins_enabled()` runs after the presolve-infeasible early return

So a malformed `DISCOPT_RINS` value does not produce the promised loud refusal on instances that return early — it is silently treated as the default arm on exactly the instances where nobody is watching. Read and validate the flag before the early return.

---

Findings 1 and 3 are the ones I would want closed before merge: the first feeds RINS a fabricated point, the second removes the only bound on a sub-solve. 4 and 5 both bear on whether the graduation measurement means what it claims, which matters given the flag is proposed for a §5 panel.

Issue #1187, opened 2026-09-06
deterministic=True does not make clay0303hfsg reproducible: same 27 nodes, three different incumbents


jkitchin:
Found while running the bound-neutrality gate for #1180. Filed separately because it is a `deterministic=` contract defect in the primal path, not a per-node-cost question — but it is a live hazard for **every** bound-neutral gate in this repo, since those gates trust that flag.

Successor in kind to #912 ("the solver is nondeterministic: node counts depend on machine speed"), which was closed `not_planned` with its own verdict recorded: *"Direction 1 [deterministic work budgets] is unimplemented and is the real fix."* This issue is a concrete, cheap reproduction of what is left, on the *primal* side, with the dual side measured clean.

## The finding

`clay0303hfsg`, `deterministic=True`, `max_nodes=20`, `time_limit=120`, four identical repeats in one process on **unmodified `main`**:

| repeat | nodes | objective | dual bound |
|---|---:|---:|---:|
| 1 | 27 | 55092.52061935072 | -1.2309090606600805e-05 |
| 2 | 27 | 55092.52061935072 | -1.2309090606600805e-05 |
| 3 | 27 | 55092.52061935072 | -1.2309090606600805e-05 |
| 4 | 27 | **46785.551237214655** | -1.2309090604747465e-05 |

A separate four-repeat run in a different process produced a **third** incumbent, 41573.26258778966, again at 27 nodes.

Two things are worth separating:

* **The dual side is clean.** The bound agrees to 12 significant figures across every run (`-1.23090906...e-05`). Whatever is moving is not the certificate arithmetic.
* **The incumbent is not.** Three distinct objectives — 55092.52 / 46785.55 / 41573.26, a 25 % spread — at an *identical* node count. So the nondeterministic component is a primal heuristic choosing differently, not the tree taking a different shape.

`deterministic=True` is documented as making "the search a function of the model rather than of machine speed". On this instance the dual search is; the primal search is not.

## Reproduction

```python
from discopt.modeling.core import from_nl
for i in range(4):
    m = from_nl("python/tests/data/minlplib_nl/clay0303hfsg.nl")
    r = m.solve(time_limit=120.0, gap_tolerance=1e-4, deterministic=True, max_nodes=20)
    print(i, r.node_count, repr(r.objective), repr(r.bound))
```

(Discard the first solve as a warm-up; the divergence appears from repeat 2 onwards and is not order-dependent — it reproduces with the repeats interleaved against another arm in either order.)

[...]

Issue #1026, opened 2026-08-14
perf: re-scope the BARON gap onto node-NLP cost and Python-JAX marshaling (successor to #1008)


jkitchin:
Supersedes #1008. That issue asked why the in-house simplex was 8.7x–29x slower than HiGHS; it is answered (fill in `LuNumeric`) and largely fixed (feral 0.16.0, #1025: aggregate fill 8.32x → 1.28x, factor nonzeros −81.8%). The useful residue is the negative result: **fixing it changed nothing end-to-end**, because the LP layer was never where the wall is.

This issue re-scopes onto where the measurements actually put the gap.

## The decomposition (`docs/dev/baron-gap-plan.md` §1)

`wall_gap ≈ floor × per_node_cost × node_count`. TX0's binding-constraint histogram over a 62-instance panel (`docs/dev/data/tenx-attribution.json`): **floor 26, node_count 18, no_bound 8, overrun 4**.

Panel-wide wall split:

| layer | share |
|---|---:|
| node-NLP | **69%** (600.6s of ~870s) |
| node-LP | **0.06%** |

Single-instance layer split on `nvs05` (cProfile aggregated across binding boundaries, §1.3):

| layer | share of 20s | what it is |
|---|---:|---|
| python | **82.5%** | `pounce.Problem.solve`, 515 calls, 16.0s cumulative |
| jax/xla | 12.3% | kernel execution |
| rust (LP) | 3.4% | `solve_lp_warm_csc_py`: 0.67s |
| pounce native | 0.1% | the optimizer itself |

Per-node gap vs BARON on that instance: **~90x** (20.5 nodes/s vs 1,874).

Inside those 515 NLP solves: `evaluate_constraints` 402,669 calls, `evaluate_objective` 400,329 — **~780 evaluator callbacks per NLP solve**, each a Python→JAX→`float()` round-trip. Interop alone: `np.asarray` 1.71s, `jax array._value` 1.53s, `array.__float__` 2.17s cumulative.

And the per-process floor, fresh subprocess, median of 3, trivial instance (§1.1):

| phase | ms |
|---|---:|
| `import jax` | 299 |
| `import pounce` | 148 |
| `import discopt` | 66 |
| parse `.nl` | 2 |
| **actual solve (21 nodes)** | **80** |
| **total** | **595** |

86% import tax. The solve itself is BARON-competitive; the process is not.

## What has already landed against this, and what it left

[...]

PR #1188, opened 2026-09-06
perf(#1180): re-measure the per-node layer split post-tape, and take the one lever it names

### LINKED ISSUE #1180: perf: re-measure the per-node layer split post-tape (successor to #1026)
Successor to #1026, which is closed as superseded: three of its four parts are finished or void, and its remaining deliverable is the one this issue carries. Read that issue's thread for the audit; the short version is that the JAX removal (#74/#75, `docs/dev/jax-removal-plan.md`) deleted the premise it was scoped on.

## Why the old attribution can no longer be cited

`docs/dev/baron-gap-plan.md` §1 has been annotated accordingly (PR #1179). What changed underneath it:

- The POUNCE-tape evaluator (`python/discopt/_tape_nlp_evaluator.py`, `DISCOPT_NLP_EVAL=tape`, opt out with `=jax`) is **default-ON** since `a2fb90d2`. **66 of 66** in-repo corpus instances solve with `jax` never entering `sys.modules` — asserted in clean subprocesses with a control arm that checks the JAX path *does* import jax, so the check cannot pass vacuously.
- The §5 panel that graduated it: cert-clean on every check (0 unsound bounds/statuses, 0 false optima, 0 certification regressions, 0 infeasible incumbents over 103 cross-engine verifications) and net-positive on wall — **10 faster / 0 slower, median 1.80×, total −43.7 %** on instances doing identical work (44 of 46 node counts unchanged).

So the pre-tape numbers — `python 82.5 % / jax 12.3 % / rust 3.4 %` on nvs05, the 299 ms `import jax` row in the 595 ms floor, `np.asarray` 1.71 s / `array._value` 1.53 s / `__float__` 2.17 s — measure a code path that no longer executes.

## Why this is still worth doing

A 1.80× median does not close a **~90×** per-node gap (nvs05: 20.5 nodes/s vs BARON's 1,874). Per-node cost is still the dominant factor in the BARON gap; what is gone is any current account of *where inside it* the time goes. That account is the deliverable.

## Deliverable

1. **Re-run the §1.3 layer split at current `main`**, aggregating cProfile across binding boundaries (pounce / rust / python) — never a single-layer profiler label, per §8 item 5, which is on the record for producing June's wrong "jax=97.7 %".
2. **Re-run the §1.1 fresh-subprocess floor decomposition** and report the new total. Do not subtract 299 ms from the old table — the remaining phases were measured alongside an import that no longer happens, and the arithmetic would carry that error forward.
3. **Separate the candidates that survived the engine swap:**
   - POUNCE iteration count per node-NLP;
   - Python frame overhead in `_IpoptCallbacks` / `_BoundOverrideEvaluator`;
   - warm-start quality — is every node re-solving from scratch?

   "Per-iterate JAX dispatch latency", the fourth candidate in #1026, no longer exists.
4. **Then build against whatever the measurement names** — not before. Entry experiment first, on real corpus instances, per CLAUDE.md §4 and the #727 RLT lesson (synthetic root-gain 0.68, real gain 0.0).

## Binding constraints

**The node-NLP solves are the bound source, not a heuristic.** With `DISCOPT_NODE_NLP_STRIDE=1000000` (strided node-NLP fully off), nvs05's wall and bound are essentially unchanged. The fix is making each solve cheap, or not needing an NLP at the node at all — **never skipping bound-producing work**.

From `baron-gap-plan.md` §8, still binding, new data required to reopen:

1. Do not truncate bound-producing native solves (TX2/PF5) — capping `casctanks`' OBBT probe flipped its bound −99.09 → +5.70.
2. Do not re-add the Rust LP native deadline (TX2b, reverted, preserved at `b8c591f`).
4. Do not blanket-disable the node-NLP (TX1) — `tls2`'s bound went looser, `tspn12` lost its incumbent, panel wall got *worse* (870 → 893 s).
5. Do not trust single-layer profiler labels.
6. Do not drop or reorder the root-relaxation fallback's bound candidates — the `max` is load-bearing.

**§8 item 3 (the TX3 JAX-free relaxation swap) is deliberately absent** — it is OBE, and PR #1179 marks it so in the doc. Its premise was that the evaluator is an immovable JAX dependency; the evaluator was replaced.

Also relevant, from #1026's own record: **G1's lesson.** Co-occurrence fusion (#645) bought 2.45× on the access-pattern replay but only 1.26× end-to-end on nvs05. A callback-level win does not imply a node-level win; measure end-to-end before believing a micro-benchmark.

## Gates

Per CLAUDE.md §5. Measurement alone needs no solver gate. A change that computes the same numbers more cheaply is **bound-neutral**: `node_count` and certified `objective` exactly unchanged on a certifying panel, any drift treated as a defect. An engine or evaluator *replacement* is **bound-changing** and needs the cert-clean + net-positive pair on an instrument that can see the change.

**Reporting rule, carried over from #1026:** any perf claim in a PR here must carry the layer split before and after, not a wall number alone. That is the specific rule that would have caught the June mis-attribution.

Also carried over: measurement discipline per CLAUDE.md §6–§11 — the probe prints an executed-assertion count and exits non-zero at zero, no swallowed exceptions in an instrument, assert `module.__file__` plus a version-unique marker before trusting any worktree measurement, and timing claims need a load gate, an interleaved control, and a spread.

jkitchin:
## Summary

Closes #1180.

`baron-gap-plan.md` §1.1/§1.3 described a solver that no longer exists — they measured the JAX evaluator, and a default solve now imports zero `jax` modules. This re-runs both measurements at `main` over the 66-instance in-repo corpus, separates the three surviving #1026 candidates, and builds against what the measurement actually names.

**The layer split has inverted.**

| layer | corpus (wall-weighted) | `nvs05` | §1.3's `nvs05` |
|---|---:|---:|---:|
| POUNCE native (IPM + tape) | **47.7 %** | 26.1 % | 0.1 % |
| `discopt._rust` (LP/MILP) | 15.0 % | **50.8 %** | 3.4 % |
| `discopt` Python | 15.4 % | 8.3 % | — |
| evaluator callback glue | 9.2 % | 4.7 % | — |
| other Python + numpy/scipy | 12.8 % | 10.2 % | — |
| **jax / XLA** | **0.00 %** | **0.00 %** | 12.3 % |

The dominant seam is the NLP solve itself (57.4 % of corpus wall; the native IPM alone is 44.9 % of *all* self time). 1.95 M derivative callbacks cost **11.5 s of tape arithmetic wrapped in 47.7 s of Python frames** — so the entire marshaling story is now worth at most 1.10× corpus-wide, against the 82.5 % it used to be.

**Deliverable 2, the floor** (fresh subprocess, `alan`, median of 5): startup 94 ms + `import discopt` 248 ms + `import pounce` 210 ms + parse 2 ms + solve 1211 ms = **1752 ms**. Re-measured, not back-subtracted. The missing `import jax` is proven by a control arm that forces `DISCOPT_NLP_EVAL=jax` and does load 217 jax modules — without it, a probe blind to imports would report the same thing.

**Deliverable 3, the three candidates:**
1. *Iteration counts are ordinary* — median 11–53 on nine of ten instances. No pathology.
2. *Python frame overhead is real but small* — on `alan` the tape arithmetic is 5 % of the callback delivering it, but the whole path is 9.2 % of wall.
3. *Warm start* — "is every node re-solving from scratch?" **No**: the parent point already beats a cold midpoint on 7 of 10 instances (`gkocis` 3.4×, `tanksize` 3.2×). A full `pounce.WarmStart` across nodes is **falsified**: 0.53×–1.95×, median 0.99×. It is also structurally blocked — on `tls2`, 10 of 12 consecutive node NLPs have different `(n, m)` because the cut pool changes the row count.

**What shipped**, the one in-repo lever left (the native IPM is outside this repo; the LP engine's per-pivot gap is external per `performance-plan.md` §16; the OBBT probe loop is already 1–3 % Python):
- `_timing.charge` as a `__slots__` context manager rather than a generator: **1.93 µs → 1.05 µs** per `with`, entered once per derivative callback.
- `TapeNLPEvaluator._x` hands pounce a contiguous `float64` array instead of a per-callback Python list: `nvs05` (n=15) **2.11 µs → 0.20 µs**, `4stufen` (n=157) **12.29 µs → 1.52 µs**. The cost scaled with `n` while the arithmetic under it did not, so the benefit grows with model size rather than sitting on one family.

Full record: `docs/dev/issue-1180-per-node-layer-split-2026-09-05.md`; house-style entry at `performance-plan.md` §27; `baron-gap-plan.md` §1 marked superseded (keeping #1179's note and answering it).

## Correctness

- [x] Cannot produce a false certificate — the change is bound-neutral by construction: it computes identical numbers by a cheaper route. Pinned by **1610 bit-identity comparisons** (66 instances × 5 points × all five tape entry points, old list path vs new array path, nan-aware exact equality, **0 mismatches**).
- [x] No validation, fallback, or safety guard was weakened. `_timing.charge`'s parent/child self-time accounting is line-for-line identical; a fully-inlined variant that would have been 0.3 µs faster was **rejected** because duplicating that accounting in a second place is a defect factory. Every entry on `baron-gap-plan.md` §8's do-not-do list is untouched and was re-confirmed against the new data.

One deliberate behavior change in an error path: `charge("bad-bucket")` now raises `ValueError` at call time rather than at `__enter__`. Strictly earlier, same exception, same place in any `with` statement.

## Tests run

- [x] `pytest -m smoke` → **1146 passed, 21 skipped, 2 xpassed, 0 failed**
- [x] `ruff check` / `ruff format --check` → clean over `python/` and the new scripts
- [x] `test_74_layer_time_attribution.py` → 9 passed (the accounting the rewrite had to preserve)

## Regression test

- [x] `python/tests/test_1180_callback_marshaling.py` — **7 failed before / 10 pass after**, verified by checking out the pre-change sources and re-running. Pins float64, contiguity, zero-copy on an already-contiguous vector, bit-identity across all five tape entry points (with a count assertion so the loop cannot pass vacuously), and `charge`'s accounting.

## Performance change

Bound-**neutral**, so no flag: the gate is exact invariance, not a graduation panel.

- Measurement: A/B with both arms interleaved **in one process**, alternating order, a discarded warm-up, and each arm asserting which variant is live before it runs.
- **Callback-heavy panel** (14 instances, 3 reps): median **1.221×**, from 1.394× (`ex1226`) to 1.010× (`alan`); every row neutral on nodes/objective/bound.
- **Whole corpus** (66 instances): 13 rows terminate on a limit and carry no neutrality or wall information; on the **53 comparable rows, 52 are exactly neutral** on node count, objective *and* bound, median **1.142×** (**1.198×** on the 30 rows with ≥ 1 s of wall). Aggregate wall is only 1.027× because two long instances that are not callback-bound dominate the sum — the change concentrates where the callbacks are and is close to free elsewhere.
- The 53rd row, `clay0303hfsg`, is **pre-existing nondeterminism**: unmodified `main` returns three different objectives at an identical 27 nodes with the dual bound stable to 12 significant figures. Filed as #1187.

### Three measurement traps found the hard way

Recorded because each nearly produced a wrong verdict, and one of them is the 2026 form of the trap §8 item 5 already warns about:

1. **A deterministic budget makes node counts comparable but makes wall clock meaningless for any phase whose stopping rule *is* the wall budget.** `beuster` measured a reproducible **0.516×** that was pure artifact — the two arms issued **3858 OBBT probes against 942** for the same 3 nodes. On the ordinary wall budget the same instance honours its limit and returns **5× the nodes at a 30 % tighter dual bound**.
2. **`node_callback` is a routing signal, not an observer.** `alan`, fresh subprocesses, both orders, same 13 nodes and objective: **54 POUNCE solves and 11 130 tape evaluations without it, 1 and 0 with it.** `profile_instance.py` attaches one by default.
3. **A replay-based "pure-binding floor" compares the wrong population.** An earlier version of the probe-LP instrument reported a *negative* Python overhead, because the captured call cold-starts while the in-loop population is warm-started. Two nested timers in the same run have no such failure mode — and this is the same shape as the measurement #764's "~70 % marshaling" figure came from, which this PR withdraws (it is **1–3 %** on four instances today).

### What this PR does **not** claim

The ~90× per-node gap to BARON was **not** re-measured. It cannot be from this box: `alan`'s solve is 15× slower here than on the machine §1 used, and no BARON licence is reachable. Every attribution figure here is a *composition* — a share, a call count, a per-call cost — and the one wall-clock claim is an A/B of two arms interleaved on the same box.

https://claude.ai/code/session_01LgfG2zbk15x3SRxZYBUNMm

---
_Generated by [Claude Code](https://claude.ai/code/session_01LgfG2zbk15x3SRxZYBUNMm)_


jkitchin:
Two things after the `main` merge into this branch (`18ef9bb`): the red check, and a re-validation of the perf claim on the new base.

## The red check is not this PR's

`Python fast` failed on `3722b25` with:

```
FAILED python/tests/test_modeling_tail_m4_m8.py::test_m7_name_check_is_not_quadratic
AssertionError: name check looks super-linear: 9.0 us/var at n=1k vs 130.6 us/var at n=4k
```

Three pieces of evidence, in order of strength:

1. **The failing path never reaches this diff.** This PR changes exactly two things: `_timing.charge` and `TapeNLPEvaluator._x`. I instrumented both and ran 2000 `m.continuous()` declarations — the shape the test measures: `_timing.charge` fired **0 times**, `_x` fired **0 times**.
2. **The same code passed 26 minutes earlier.** `Python fast` was green on `18ef9bb` at 15:25Z and red on `3722b25` at 15:52Z. The only difference between those two commits is a `.gitignore` edit.
3. **It is not actually super-linear.** Five consecutive runs of the test's own measurement on a quiet box:

   | run | n=1k | n=4k | ratio (bar: < 2.5) |
   |---|---:|---:|---:|
   | 1 | 7.80 µs/var | 7.70 | **0.99** |
   | 2 | 7.80 | 7.71 | 0.99 |
   | 3 | 8.01 | 7.74 | 0.97 |
   | 4 | 7.68 | 7.47 | 0.97 |
   | 5 | 7.25 | 7.08 | 0.98 |

   Flat — which is what the test asserts. Note also that CI's n=1k figure (9.0 µs/var) matches this box, and only the n=4k measurement exploded to 130.6. That is a stall during the second measurement, not a scaling property of the code.

The test is a bare wall-clock scaling assertion (`large < small * 2.5`) with no load gate, taking two single unrepeated samples on a shared runner — it will keep doing this to unrelated PRs. It is not mine to widen this PR with, but a `min` over a few repetitions, or gating on `os.getloadavg()`, would fix it; happy to send that separately if you want it.

I have re-run the failed job once, per the "confirm it isn't this PR's" rule. No further action from me unless it reproduces.

## The perf claim, re-validated on the merged base

Since the merge brought in 2,533 lines of Rust plus changes to `interval_eval.py` / `modeling/core.py` / `mpec.py` / `status.py`, the numbers in the PR body were measured on a base that is no longer this PR's head. I rebuilt and re-ran rather than let them stand:

- **Bit-identity**: 1610 comparisons over 66 instances × 5 points × all five tape entry points — **0 mismatches**, unchanged.
- **`test_1180_callback_marshaling.py` + `test_74_layer_time_attribution.py`**: 19 passed.
- **Callback-heavy A/B panel** (14 instances, 3 interleaved reps, `deterministic=True`): median **1.232×**, against 1.221× before the merge. Every instance exactly neutral on node count, objective and bound, except `4stufen`, which is time-truncated in both arms at the 120 s limit (the caveat already in the PR body).

| instance | old | new | speedup |
|---|---:|---:|---:|
| ex1226 | 5.18 s | 3.66 s | **1.415×** |
| ex1221 | 2.30 | 1.65 | 1.393× |
| gkocis | 1.16 | 0.88 | 1.326× |
| oaer | 1.58 | 1.22 | 1.295× |
| nvs21 | 2.87 | 2.25 | 1.276× |
| tls2 | 5.80 | 4.57 | 1.270× |
| tspn08 | 58.85 | 47.05 | 1.251× |
| st_e29 | 4.40 | 3.63 | 1.214× |
| tspn10 | 43.61 | 36.23 | 1.204× |
| ex1225 | 5.50 | 4.67 | 1.180× |
| tanksize | 50.22 | 46.82 | 1.073× |
| nvs05 | 51.99 | 50.70 | 1.026× |
| alan | 0.03 | 0.03 | 0.998× |
| 4stufen | 120.18 | 120.14 | 1.000× (time-truncated) |

The claim holds on the merged base; the PR body's 1.221× is the conservative figure and I have left it, with this comment as the record.

---

jkitchin:
Follow-up, and a correction to my previous comment.

**Correction.** I wrote "I have re-run the failed job once". That was wrong about my own action: my `rerun-failed-jobs` call was **refused** (`403: This workflow is already running`) because a re-run had already been triggered on `3722b25` at 16:01Z by something other than me. That re-run was then **cancelled** mid-flight when my docs push moved the head to `ae15ced`. So at the time I wrote it, no completed second observation existed — the only evidence was the local characterisation.

**The confirmation has now arrived anyway.** `ae15ced` differs from `3722b25` only by a docs commit, so its CI is the second observation of identical test code:

| | `3722b25` | `ae15ced` |
|---|---|---|
| `Python fast` | **failed** — `test_m7_name_check_is_not_quadratic`, 130.6 µs/var at n=4k | **success** |

Same code, red then green. Together with the local runs (ratio 0.97–0.99 over five repetitions, against a 2.5 bar) and the fact that the failing path never touches this diff (`_timing.charge` and `_x` both fire **0** times across 2000 `m.continuous()` declarations), `test_m7_name_check_is_not_quadratic` is a load-sensitive assertion, not a regression from this PR or from the merged `main`.

**Current state of the PR:** green on `ae15ced` — Lint + typecheck, Rust, Python fast, claim-boundary, AMP coverage and both codecov gates all passing; `Python correctness (fast subset)` still running at the time of writing. No open review threads. Nothing outstanding from my side.

The standing offer from the previous comment holds: that test takes two single unrepeated wall-clock samples on a shared runner with no load gate, and it will keep doing this to unrelated PRs. A `min` over a few repetitions would fix it. Out of scope to widen this PR with, but say the word and I will send it separately.

---

PR #1189, opened 2026-09-06
perf(bnb): A13 recover the objective lattice through a defining equality row

jkitchin:
## What this fixes

`objective_granularity` refuses the moment a continuous column carries objective
cost. The common shape

```
min z   s.t.   z = <integer expression>
```

routes the whole objective through one continuous column, so the lattice is lost
entirely and neither objective-integrality fathoming nor dual-bound rounding is
available. On `neos-3610051-istra` the search would sit at `bound = 48.53`
against an integral optimum of `49` until the clock ran out.

## The lever

`objective_lattice` resolves such a column through an equality row that pins it
to integer columns. If row `i` is `Sum_k a_ik z_k = b_i` and column `j` has
`a_ij != 0`, then at *every* feasible point

```
z_j = (b_i - Sum_{k!=j} a_ik z_k) / a_ij
```

so when every other column of that row is integer-constrained or bit-exactly
fixed, `c_j z_j` is itself a lattice value. Other constraints only remove
feasible points, which cannot invalidate a claim quantified over all of them.
Every unproven case returns `None`.

**The model is not rewritten.** This reads the row to compute a spacing, so
there is no substitution to undo and no postsolve obligation. That is the
distinction from the free-column-singleton substitution HiGHS performs in
presolve (`HPresolve::freeColSubstitution`, which removes the column and must
restore it), and from the aggregation SCIP performs before
`SCIPprobCheckObjIntegral` runs. Ours is the cheaper, narrower move: a detector,
not a transformation.

### The anchor is load-bearing

The substitution introduces a constant, so the return type is
`ObjLattice { spacing, shift }` rather than a bare spacing — the lattice is
`k*g + obj_const + shift`. A wrong anchor lifts the dual bound past the optimum,
which is a false certificate. Five panel instances carry a nonzero shift
(`istra` 24, `iskar` 15, `itata` 80, `markshare*` 3082-7286, `rout` 2375.25).
For the same reason `is_fixed` takes **no tolerance**: a column declared fixed
when it is merely narrow lets the objective drift off the claimed lattice.

## Entry experiment, run before the implementation (CLAUDE.md §4)

Over a 104-instance MIPLIB draw: 20 instances already had a directly-integral
objective, and a further **13** are of exactly this shape — the whole `neos-361*`
family, `misc07` (spacing 5), `markshare*`, `rout` (spacing 0.01). The stated
kill criterion was "fewer than ~5 instances gain a usable lattice"; 13/104 = 12.5%
cleared it.

All 13 derived lattices were then verified against the instance's true optimum,
including the non-unit cases (`misc07` g=5 puts the optimum at k=562 exactly;
`rout` g=0.01 at k=-129769 exactly). Zero violations.

`b-ball`, one of the three instances that motivated the work, is honestly *not*
resolved — its objective column has no defining equality row over integers, so
the detector refuses rather than guessing.

## Graduation panel (CLAUDE.md §5, both bars)

44 MIPLIB instances, 20 s each, arms interleaved per instance, re-run on the
rebased branch so it describes the code being merged. Script:
`scratchpad/miplib/a13_panel.py`; bars pre-registered before the run.

**BAR 1 — cert-clean: 0 violations.** 83 incumbents independently
feasibility-verified against the engine (column bounds, integrality, row
activities, recomputed objective). No bound above its reference optimum, no
certification regression, and `ObjLatticeSubst = 0` on every OFF arm.

**BAR 2 — net-positive: PASS.**

| | OFF | ON |
|---|---|---|
| solved | 23/44 | **25/44** |
| newly solved | — | `neos-3610051-istra`, `neos-3610040-iskar` |
| lost | — | **none** |
| both-solved wall | 38.5 s | **34.0 s (-11.6%)** |
| both-solved nodes | 440,413 | **406,037 (-7.8%)** |
| dual bound, 21 unsolved | — | **6 better, 0 worse, 15 unchanged** |

The lever engaged on 11/44 instances. The two headline cases go from timed-out
feasible to certified optimal:

| instance | status OFF → ON | nodes OFF → ON | bound OFF → ON |
|---|---|---|---|
| `neos-3610051-istra` | feasible → **optimal** | 180,395 → 41,795 | 48.68 → **49** |
| `neos-3610040-iskar` | feasible → **optimal** | 358,223 → 97,377 | 36.80 → **37** |
| `neos-3611689-kaihu` | optimal → optimal | 97,269 → 70,651 | 118.99 → **119** |
| `neos-3610173-itata` | optimal → optimal | 14,717 → 10,753 | 147.99 → **148** |

Both bars pass, so per CLAUDE.md §5 the flag graduates **default-ON** in this
PR. The `DISCOPT_OBJ_LATTICE_SUBST=0` opt-out and the legacy path stay intact
and are covered by their own test binary.

## An instrumentation failure worth recording (CLAUDE.md §6)

The **first** run of this panel measured nothing. `obj_lattice_subst_enabled()`
cached its env read in a `OnceLock`, so the first solve in the process fixed the
arm for every later one and the entire ON arm silently ran OFF code. The
signature was `neos-3611447-jijia` reporting an *identical* 313,441 nodes in both
arms with the counter at 0 throughout. It was caught by reading the interim log,
not by trusting the summary.

The fix is a fresh `getenv` per solve (not measurable against a solve), plus a
**hard refusal** on an unrecognized value — a panel that exports
`DISCOPT_OBJ_LATTICE_SUBST=on` and is quietly handed the OFF arm measures nothing
and reads as a result. `obj_lattice_subst_reread.rs` pins this and was verified
to fail when the `OnceLock` is reintroduced.

## Tests

New behavior, so new regression tests that fail before the change:

- 8 unit tests on the detector: substitution through a defining row, nonzero RHS
  moving the anchor and not the spacing, refusal on a second free continuous
  column, a fixed column as a constant, a *nearly* fixed column refused, an
  inequality row refused, agreement with the base detector when nothing needs
  resolving, and a property test asserting `checked > 400` executed comparisons.
- A flag-parse unit test pinning the refusal of unrecognized values.
- Four arm binaries — default, explicit-on, opt-out, re-read — each with an
  anti-vacuity assertion on the `ObjLatticeSubst` counter, so a build where the
  lever never ran cannot pass. The opt-out arm asserts `nodes > 1` so the on-arm's
  "1 node" keeps its meaning.

The `LatticeRows` refactor that followed the panel was checked bound-neutral per
CLAUDE.md §5: `istra` 64,327 and `misc07` 28,499 nodes, bit-identical to the
panel's ON arm.

## Verification

- `cargo test -p discopt-core` — **712 passed, 0 failed**
- `cargo clippy -p discopt-core -- -D warnings` and the `--no-default-features`
  job — **both clean**
- `pytest -m smoke` — **1182 passed, 1 skipped, 2 xpassed**

## Scope note

This closes no issue — it comes from the MILP competitiveness plan directly, so
under the standing merge rule it needs an explicit call rather than an automatic
merge. The plan-doc record for A13 goes on the `docs/a1-retraction` branch
(#1185), where the A11/A12 records already live, to keep that narrative
contiguous and this PR scoped to code.


jkitchin:
## Code review — PR #1189

I read every hunk, opened `obj_integral.rs`, `milp_driver.rs`, `tree_manager.rs` and `lp/simplex/presolve.rs`, ran the four new test binaries (all pass), and ran a scratch probe against `objective_lattice` to confirm two findings empirically.

**The core algebra is sound.** The substitution `z_j = (b_i − Σ_{k≠j} a_ik z_k)/a_ij`, the anchor bookkeeping, the "no two pending columns share a defining row" argument, and the gcd-over-split-terms behaviour (splitting a coefficient can only make the lattice *finer*, the safe direction) all check out. The findings are on the numeric guards around it — and the first is a false-certificate path.

---

### 1. HIGH — `crates/discopt-core/src/bnb/obj_integral.rs:339` — row eligibility drops columns by absolute coefficient magnitude, ignoring bound width, yielding a false lattice

The `.any(...)` disqualifier skips any entry with `v.abs() <= ZERO_TOL` (1e-12) before it ever asks whether that column is free continuous, and the term/shift loop at `:350` skips it again. But a free continuous column contributes `−(v/a_ij)·z_k` to `z_j` at every feasible point, and with this repo's `INF = 1e20` sentinel `z_k` is not small. Verified with a probe:

```
min z  s.t.  z − 2x − 1e-13·w = 0,  x integer in [0,10],  w continuous in [-1e20, 1e20]
-> ObjLattice { spacing: 2.0, shift: 0.0 }
```

`z` can move by `1e-13·w ≈ 1e7` off that lattice. `round_bound_up` then lifts the dual bound onto a lattice the optimum is not on — **a false certificate**.

This is precisely the class CLAUDE.md documents from the gear4 incident, and which `presolve.rs:contrib` quotes: *test unboundedness on the bound, never on a product.* `|a_ij| < 1` makes the product of an unbounded bound an ordinary finite number. The test needs to be on `|v| · width(lo[k], up[k]) / |a_ij| · |c_j|`, or simply refuse on any structural nonzero in the row.

Related and worth fixing in the same pass: `:284` skips a costed *continuous* column when `|c_j| <= 1e-12` instead of refusing. That is inherited from `objective_granularity:127`, but it survives into the new path.

### 2. MEDIUM — `obj_integral.rs:213` — `is_fixed` treats the `1e20` INF sentinel as a finite bound

`lo.is_finite() && up.is_finite()` is exactly the check CLAUDE.md says never to use in this layer: `1e20.is_finite()` is `true`. Verified — with a costed continuous column at `lo == up == 1e20`, `objective_lattice` returns `ObjLattice { spacing: 1.0, shift: 1e20 }`. `set_objective_lattice` accepts it (the offset is finite), and `round_bound_up` then computes `k = (v − 1e20)/g`, carrying `LATTICE_EPS·|k| ≈ 1e11/g` of slack plus catastrophic cancellation on the way back to objective units — the "lift" is numeric garbage that can land above `v` and be published as a dual bound.

Guard on `lo > -INF && up < INF` (the sentinel), not `is_finite()`. Same at `:288` (`shift += cj * lo[j]`) and `:356` (`shift -= s * v * lo[k]`), which propagate the sentinel. Note also that the `!shift.is_finite()` guard at `:366` is unreachable for the `pending.is_empty()` early return at `:334`, so that path has no shift sanity check at all.

### 3. MEDIUM — `crates/discopt-core/src/bnb/milp_driver.rs:751` — with the flag ON (now the default), a `None` from `objective_lattice` silently discards a valid `base` lattice

The ON branch returns `subst` unconditionally. Today the two agree whenever `pending` is empty — except through the length guard at `obj_integral.rs:311`, which returns `None` first. More importantly it makes the ON arm strictly weaker than OFF the moment anyone adds a refusal to `objective_lattice` — such as the magnitude guard finding 1 calls for. A model whose objective is *directly* integral would then lose its lattice entirely on the default path, costing pruning and certification with no signal. `subst.or(base)` costs nothing and makes the "strictly more objectives" claim in the comment above it structurally true rather than incidentally true.

### 4. LOW — `milp_driver.rs:2853` — `Err(_) => true` swallows `VarError::NotUnicode`

The function's own doc says an unrecognized value is a hard refusal, "for the same reason" a cached read is fatal: a panel that sets a value it thinks means ON and is quietly handed the other arm measures nothing. A non-UTF-8 value takes exactly that silent-default path. Match `Err(VarError::NotPresent) => true` and refuse on `NotUnicode`.

### 5. LOW — `tests/obj_lattice_subst_on.rs:7,9` and `tests/objlat_common/mod.rs:16` — stale headers contradict the change this PR makes

Both say "the flag is read once per process through a `OnceLock`" — the exact caching this PR removes and adds `obj_lattice_subst_reread.rs` to pin. `obj_lattice_subst_on.rs:7` also points at `obj_lattice_subst_default_off.rs`, which does not exist (the file is `obj_lattice_subst_optout.rs`). Under CLAUDE.md §6 these headers *are* the instrumentation rationale; leaving them wrong invites the next reader to reinstate the `OnceLock`.

### 6. LOW — `tests/objlat_common/mod.rs:23` — `OPTIMUM` is unused in the reread binary

Confirmed in the build output: a `dead_code` warning on every `cargo test -p discopt-core`. CI's clippy job does not pass `--all-targets` so it will not fail the gate, but it is permanent noise. `#![allow(dead_code)]` on the shared module, or importing `OPTIMUM` into the reread assertion, clears it.

---

### Verified as *not* problems

- The presolved-bounds rationale holds: `tighten_bounds_csc` is pure FBBT (no dual or objective fixing), and its `res_err`/`tol` widening means an FBBT-derived continuous fixing essentially never satisfies the bit-exact `lo == up`, so `is_fixed` is conservative in practice.
- Independence of resolutions: a pending column is neither `integral` nor `is_fixed`, so a row containing a second pending column is always disqualified — no two resolutions can interact.
- A variable appearing in both a direct objective term and a substituted term yields `gcd(c_k, −s·v)`, which divides the true combined coefficient — finer, and sound.
- `objective_granularity`'s refactor through `spacing_of` is behaviour-identical.
- Coefficient amplification through a small pivot is implicitly capped: `denominator_of` bails once `|floor(t)| > MAX_DENOM`, so no term above ~1000 enters the gcd.
- The `ObjLattice` offset (`obj_const + shift`) is in the units `TreeManager` uses, and `Option<ObjLattice>` is `Copy`, so the double `.map` at `milp_driver.rs:759-760` is fine.

Finding 1 is the blocker: it is default-ON and it can publish a dual bound above the optimum.

jkitchin:
Conflict resolution note (merge of `origin/main` at c74c7d54 into this branch, commit 4691d430):

- `crates/discopt-core/src/profile.rs` conflicted where A13's `ObjLatticeSubst`
  counter and A12's RINS counter block (`RinsConsidered` … `RinsBackoffSkips`,
  landed with #1186) were added at the same point in the `counters!` list.
  Both sides are purely additive; the resolution keeps **both**, A13's first.
- `crates/discopt-core/src/bnb/milp_driver.rs` auto-merged (A13 and A12 touch
  different call sites).

Verified after the merge, locally in the `perf/a13-obj-lattice-substitution`
worktree: `cargo test -p discopt-core` → **EXIT=0**, 704 unit tests passed,
0 failed, plus all four A13 integration binaries
(`obj_lattice_subst_default_on`, `_on`, `_optout`, `_reread`) passing.

jkitchin:
## Review findings addressed — `13d3bb66`

All six findings fixed. Each was reproduced with a failing test before the change; the exact-zero control test passed throughout, so the tightened guard is not over-tightening.

### The two soundness ones

**1 (HIGH) — the false lattice.** The disqualifier now tests `v != 0.0` rather than `|v| > ZERO_TOL`. Reproduced first:

```
min z  s.t.  z - 2x - 1e-13 w = 0,  x int [0,10],  w cont [-1e20, 1e20]
before -> ObjLattice { spacing: 2.0, shift: 0.0 }     # z moves ~1e7 off it
after  -> None
```

**I also tightened the term loop, which the review did not name.** Skipping a small-but-nonzero coefficient there removes a term from the gcd and returns a *coarser* lattice than the truth — unsound in the same direction as the finding itself. A tiny term now flows into `spacing_of`, where `denominator_of` refuses it; refusal is the safe outcome. An exact structural zero contributes nothing at any width and stays skippable, pinned by `an_exact_zero_coefficient_still_resolves_the_row`.

**2 (HIGH) — the INF sentinel.** Added `const INF: f64 = 1e20` beside `ZERO_TOL`; `is_fixed` now compares `lo > -INF && up < INF`. Reproduced first: a costed continuous column at `lo == up == 1e20` returned `ObjLattice { spacing: 1.0, shift: 1e20 }`.

### The rest

- **3** — `subst.or(base)`. This matters more now than when the review was written: finding 1 *adds* a refusal path to `objective_lattice`, so without `.or(base)` the ON arm would newly drop valid directly-integral lattices on the default path.
- **4** — only `VarError::NotPresent` takes the default; `NotUnicode` panics, as the function's own doc promises.
- **5** — both headers corrected, including the dangling `obj_lattice_subst_default_off.rs` reference (the file is `obj_lattice_subst_optout.rs`).
- **6** — targeted `#[allow(dead_code)]` on `OPTIMUM`. The inner-attribute form suggested itself has to lead the file and broke all four test binaries, so it is an outer attribute on the const.

### Verification

Rebased onto `4691d430` (merged `main` + the #1186 RINS work) and re-run there, not on the pre-merge base:

- `cargo test -p discopt-core` — **707 passed in the lib**, every test binary green, 0 failed
- the two new tests fail before / pass after
- `cargo fmt --check` and `cargo clippy -p discopt-core --lib` clean; the `dead_code` warning is gone

### Not addressed, flagged in the review and left alone

`obj_integral.rs:284` skips a costed *continuous* column when `|c_j| <= 1e-12` instead of refusing. It is inherited from `objective_granularity:127` and predates this PR, so changing it would alter the OFF arm's behaviour too — it wants its own change with its own differential, not a quiet ride here.

PR #1190, opened 2026-09-06
perf(milp): A14 -- RENS, a primal heuristic that needs no incumbent

jkitchin:
## Why

A13's parity panel measured *where* the remaining MILP gap lives, and the answer
contradicts the competitiveness plan's own §0 ordering. Over the 38-instance MIPLIB
panel at 20 s the median unsolved instance's gap is **91.1% primal / 8.9% dual**.
Re-run at 3x the budget — the pre-registered kill criterion — it does not move: 91.8%.

The within-instance comparison is sharper than the shares and is the finding of record:

> Of the **14** unsolved instances that already held an incumbent at 20 s, **zero**
> improved it at 60 s, while the **dual bound improved on 10 of 15**.

`b-ball` explored 1,242,219 nodes without moving its incumbent. The primal side is not
slow, it is **frozen**. `enlight_hard` is the natural control: the one instance holding
no incumbent at 20 s — and therefore the one where the primal machinery was still
permitted to run — reached the optimum by 60 s.

Root cause, read out of `milp_driver.rs` rather than inferred: the repair dive is gated
on `!has_incumbent`, `DIVE_STRIDE_DEFAULT` is `0`, and the only off-root heuristic is
nearest-rounding, which never re-solves the continuous columns. **Once any incumbent
lands, every improvement mechanism switches itself off.** A12/RINS does not reach this
case either — it fixes the columns where the incumbent and the relaxation agree, so it
is anchored to the incumbent it starts from, and with incumbents 22–140% off the
optimum that anchor is the defect rather than the fix.

RENS (Berthold, *RENS — the optimal rounding*, Math. Prog. Comput. 6(1):33–54, 2014) is
the mechanism that needs no incumbent: fix every integer column already integral in the
LP **relaxation**, restrict each fractional one to `{floor, ceil}`, solve the sub-MIP.

The §0 falsification and its retraction are recorded in `docs/dev/milp-competitiveness-plan.md`
(section A14, plus in-place annotations on §0 and §3) — that lands via PR #1185.

## Entry experiment (CLAUDE.md §4)

Bars declared **before** the run, in `scratchpad/a14/rens_entry.py`'s docstring; 15
scored unsolved instances from the A13 panel.

| metric | bar | measured |
|---|---|---|
| hit rate | ≥ 5/15 | **7/15** |
| median primal-gap reduction | ≥ 25% | **80.3%** |
| certificate violations | 0 | **0** |
| median LP-integral fixing rate | — | 70.0% |

Every feasible sub-MIP solved to **optimality** inside 20 s. `gsvm2rl3` 0.586355 → the
exact optimum 0.336528; `sp150x300d` 93.8%; `beavma` 88.3%; `neos17` 80.3%.

Honest limit, recorded rather than buried: the four `mik-250-20-75-*` came back *worse*,
because the root LP fixes 70% of their integers at values that force a near-zero objective.

## What changed

Reuses A12's sub-MIP subsystem rather than duplicating it, so the soundness argument
lives in exactly one place.

- `RinsBase` / `rins_node_budget` / `RINS_*` → `SubMip*`; the backoff stride is now
  parameterized by a base stride.
- **`run_sub_mip(kind, …)`** extracted from `try_rins`'s tail: budget clamping, the
  recursion guard, the strict-improvement test, and re-validation through
  `validate_seed_incumbent` against the **base** rows and bounds. A point the sub-driver
  calls feasible and this driver cannot verify is never injected (`RensRejected`).
- **`rens_box`** is pure and separate from the solve, so the contraction property is
  asserted directly instead of inferred from a node count. It narrows only: a pinned
  value outside the base box is *declined*, never widened — a widened bound is exactly
  what would make the restriction argument, and with it the soundness argument, false.
- **`try_rens`** seeds the incumbent into the sub-solve **only** after verifying it lies
  inside the RENS box. It generally does not — its integer values come from a different
  assignment than the relaxation's — and seeding a point outside the box would have the
  sub-solve prune against an infeasible incumbent, which is how a heuristic deletes the
  solution it was sent to find.
- Separate node quota, failure streak and cadence from RINS, so a panel can attribute
  nodes to one mechanism or the other.
- Eight new profile counters plus a `Rens` phase, with the funnel-discipline doc comment.

`DISCOPT_RENS` is **default OFF**; an unrecognized value is a hard **refusal**, not a
silent arm choice. Bound-changing under CLAUDE.md §5, so it stays off until a
differential panel clears **both** bars — the entry experiment shows the mechanism *can*
pay, never that it pays broadly (the `DISCOPT_CUT_INHERIT` lesson).

## A note on the test fixture

RENS was tried first on the existing CFL fixture and **every box it built was
integer-infeasible** — pinning the assignment columns the relaxation had driven to 0
leaves the free columns unable to serve every customer within capacity. Re-solving each
box with presolve, propagation and reduced-cost fixing all disabled still proved
infeasibility in 3–7 nodes, so the infeasibility is the *model's*, not the presolver's.
Loosening the fixing-rate floor until CFL passed would have been precisely the
tolerance-tweak CLAUDE.md §3 forbids.

The new fixture is a minimum-weight vertex cover on an odd cycle plus a forced block,
where both properties it needs are **provable rather than observed**: summing the five
C5 rows gives `2·Σx ≥ 5`, so the LP cannot beat 2.5, and attaining it forces every row
tight — which on an *odd* cycle admits only `x = 1/2`. The fixing rate is therefore
exactly 0.5, and the RENS box leaves all five half-valued columns free, so the sub-MIP
is an unrestricted C5 cover and is feasible **by construction**.

## Verification

```
cargo test -p discopt-core                                        698 lib + 9 integration binaries, 0 failed
cargo clippy -p discopt-core -- -D warnings                       clean
cargo clippy -p discopt-core --no-default-features -- -D warnings clean
```

New arms (each a separate binary — `DISCOPT_RENS` is process-wide):

- `rens_on_finds_without_incumbent` — no seed, `heuristics: false`; funnel
  `considered=1 gated=1 run=1 improved=1 rejected=0 subnodes=3` → certified optimum 18.
- `rens_default_off` — `RensConsidered == 0` (stronger than `RensRun == 0`: it proves the
  schedule was never reached), same certified optimum without RENS, so the ON arm's
  objective assertion is not circular.

Unit tests: `parse_rens_flag` refuses garbage; `rens_box` contraction property over every
column with an executed-comparison count (CLAUDE.md §6), plus the decline cases —
all-integral, below the fixing-rate floor, no integer columns, non-finite value, pinned
value outside the base box, short slices.

`pytest -m smoke` and the adversarial suite are left to CI here: this branch's only
Python-visible change is behind a default-OFF Rust env flag, and installing a second
build over the venv would invalidate the A13 measurement environment (CLAUDE.md §8).

## Caveats

Machine load was 21–76 during both A13 panels, so **no wall-clock claim is made from
them**. Bound and incumbent *values* are load-independent, and every number above is one
of those.

**Stacking:** this branch is cut from `perf/a12-rins` (PR #1186), which is not yet
merged. Review the A12 diff there; this PR's own changes start at `ab95615b`.

Contributes to the MILP competitiveness track. Does not finish it, so no closing keyword.


jkitchin:
## Correction: this PR's base gets no automatic CI

My PR description said `pytest -m smoke` and the adversarial suite were "left to
CI". That was wrong as written, and I am retracting it here rather than leaving it
to be discovered (CLAUDE.md §11).

`.github/workflows/ci.yml` triggers `pull_request` only on
`branches: [main, "release/**", feature/mip-nlp-solver]`. This PR targets
`perf/a12-rins` (it is stacked on #1186), so **no CI ran on it at all** —
`gh pr checks 1190` reported "no checks reported on the 'perf/a14-rens' branch".
Deferring verification "to CI" on a stacked PR defers it to nothing.

This is the same hole the comment above that trigger list already anticipates for
release branches ("Without it those PRs get no CI at all, and the release lands on
main having been validated only as one big merge"); a stacked perf branch has it too.

Fixed for this branch by dispatching the workflow explicitly against the branch ref:

```
gh workflow run ci.yml --ref perf/a14-rens
```

→ https://github.com/jkitchin/discopt/actions/runs/34035707199

The Rust verification I reported was run locally and stands unchanged:

```
cargo test -p discopt-core                                        698 lib + 9 integration binaries, 0 failed
cargo clippy -p discopt-core -- -D warnings                       clean
cargo clippy -p discopt-core --no-default-features -- -D warnings clean
```

When #1186 merges, rebasing this branch onto `main` will also restore automatic CI.

jkitchin:
## Graduation panel result: **`DISCOPT_RENS` stays default-OFF**

The CLAUDE.md §5 differential panel has run. Recorded in
`docs/dev/milp-competitiveness-plan.md` §A14 with the script, log and JSON under
`scratchpad/a14/` (commit c55bb170). **The flag does not graduate.**

### BAR 1 — cert-clean: PASS, and non-vacuous

```
reference comparisons executed : 152
ON incumbents feasibility-verified against the original model : 72
PASS: 0 violations
```

No dual bound above a reference optimum, no incumbent below one, `bound <= incumbent`
throughout, no `optimal` status lost from OFF to ON, and every ON incumbent
re-derived against the original rows/bounds/integrality rather than trusted.

### BAR 2 — net-positive: FAIL on 2c

18 instances not `optimal` in both arms.

```
median primal gap  OFF 38.74%   ON 1.64%
incumbent          wins 12   losses 0
dual bound         wins  1   losses 9   (cap: 3 of 18)

2a median primal gap strictly better : PASS
2b incumbent wins > losses           : PASS
2c dual-bound losses <= 20%          : FAIL
```

The primal effect is the largest this plan has measured — `mik-250-20-75-2`
42.19% → 0.00%, `beavma` 54.94% → 0.29%, `neos17` 140.35% → 40.00%, twelve wins
and **zero** losses — and it is exactly what A5 said the residual gap needed.
It is still not a graduation. 2c was pre-registered in the script docstring
before the run, it failed, and the kill criterion fires. Two bars out of three
does not flip a default.

### Post-hoc, and labelled as such

Written *after* seeing the result, so it argues for a re-run, not for re-reading
this one. The panel's `pick()` compares `rs[-1]` — a single rep per arm — while
the dual bound at a fixed time limit is itself noisy. Comparing each instance's
between-arm difference against the spread between two reps of the *same* arm:

| | count |
|---|---|
| dual-bound difference inside same-arm rep-to-rep spread | 5 |
| difference exceeding it | 5 |

`gsvm2rl3` moves 15.94% between two OFF reps against a 4.28% OFF↔ON difference;
`beavma`'s ON bound is *better* than OFF in rep0 and worse in rep1. Of the five
that do exceed the noise, four are 0.077–0.339%; only `neos17` (10.02%) is both
real and large. So there is a genuine dual-side cost, but it is much smaller
than "9 losses" implies, and a zero-tolerance *count* under a shared wall-clock
budget conflates "RENS harmed the search" with "RENS spent part of the 20 s in
sub-MIPs".

**This does not license graduating the flag.** A bar discovered to be badly
specified after it fails is not thereby satisfied. The right next step is a
re-run with the dual-side bar pre-registered on magnitude and repetition — more
reps, a per-instance spread, a relative threshold — on an unloaded machine.
Until such a panel passes, A14 ships default-OFF with its `=0` opt-out and the
legacy path intact, which is exactly what this PR already does.

### Load disclosure

`load average 14.33` throughout, another session holding a core. Per §9 **no
wall-clock claim is made from this run**; every bar above is stated on bounds,
incumbents and statuses.

jkitchin:
Rebased-by-merge onto `main` and **retargeted from `perf/a12-rins` to `main`** (that base merged with #1186).

Two things this fixes, both worth stating plainly:

1. **This PR now gets CI.** While it targeted `perf/a12-rins` it got *no checks at all* —
   `.github/workflows/ci.yml` fires `pull_request` only on `branches: [main, "release/**",
   feature/mip-nlp-solver]`. `gh pr checks` on such a PR prints "no checks reported", which
   reads as nothing-failing rather than nothing-ran. That vacuous pass is exactly the CLAUDE.md
   §6 failure mode, and it was in my own monitor, not in the repo's instruments.
2. **The `docs/dev/milp-competitiveness-plan.md` conflict is resolved without dropping
   anything.** `main`'s A14 entry-gate section (`§0's ordering principle is FALSIFIED …
   RENS clears its entry gate`) is kept verbatim, and this branch's graduation-panel verdict
   follows it, renamed `### A14 (continued) — the graduation panel: …` so the two read as
   entry-then-verdict rather than as competing A14 sections. `crates/discopt-core/src/{bnb/milp_driver.rs,profile.rs}`
   auto-merged against A12 (#1186) and A13 (#1189).

Local verification on the merge commit (057d5677): `cargo test -p discopt-core` → **EXIT=0**,
707 unit tests passed / 0 failed, plus all integration binaries.

The verdict is unchanged by any of this: **`DISCOPT_RENS` stays default-OFF.** It is cert-clean
and the largest primal win this plan has measured, and it fails its own pre-registered
net-positive bar 2c. The corrected dual-side bar (`rens_panel2.py`) is pre-registered in this
branch and has not been run yet.

Issue #1182, opened 2026-09-05
mpec: exact continuous simplex/CNF lowering of logic constraints (deferred from RFC #1123)


jkitchin:
## Background

RFC #1123 discussed an exact continuous reformulation of logic constraints under
"A second continuous target: exact simplex/CNF projection", and its closing comment
**deferred** the lowering. #1123 remains the design record but is closed, so this issue
exists to hold the requirements somewhere they can be picked up.

The requirements below were written by @bernalde in
[#1148 (comment)](https://github.com/jkitchin/discopt/issues/1148#issuecomment-5554603766)
and are reproduced here with attribution. They were explicitly **not** acceptance criteria
for the reporting work of #1148/#1158 — that comment says so in its last line. Filed
separately so they survive #1148 being closed.

Reference: Jad Wehbeh and Eric C. Kerrigan, *Exact Continuous Reformulations of Logic
Constraints in Nonlinear Optimization and Optimal Control Problems*
([arXiv:2601.03906v1](https://arxiv.org/abs/2601.03906v1)).

## What the paper offers, and what it does not

Theorem 1 replaces each CNF clause `OR_j [p_ij(z) <= 0]` with

```
sum_j lambda_ij * p_ij(z) <= 0,   lambda_i >= 0,   sum_j lambda_ij = 1
```

Read carefully before building anything on it:

- Exactness concerns **projection onto the original variables**. The lifted problem
  remains generally nonconvex.
- §3.1 represents strict negation with an existential exponential lift. Replacing that by
  a closed inequality or a fixed margin **changes the contract**.
- §5 evaluates quadrotor and two-tank control with *local Ipopt* solves, and its "Big M"
  comparison also eliminates binaries continuously — so those results do **not** establish
  superiority over certified mixed-integer solving.

## Requirements to carry forward

1. **Preserve the original predicates** and evaluate their truth/boundary residuals at the
   returned **source point**, independently of the weighted rows. Simplex weights are
   *existential witnesses*: fractional weights must not be reported as failed Boolean
   integrality, nor exposed as recovered named Boolean assignments.
2. **Reuse the local-versus-certified result distinction** (`discopt.status`, from #1148).
   Promote a candidate only after fresh source validation and any required assignment
   recovery/polishing; local convergence supplies no global dual bound.
3. **Reuse the overlapping-disjunct and strict-boundary fixtures** from the
   semantic-contract work. Add a **fractional-witness control** and an **infeasible source
[...]

PR #1195, opened 2026-09-06
fix(determinism): #1187 deterministic=True did not hold on the primal path

### LINKED ISSUE #1187: deterministic=True does not make clay0303hfsg reproducible: same 27 nodes, three different incumbents

jkitchin:
## Summary

Closes #1187.

`clay0303hfsg` at `deterministic=True, max_nodes=20, time_limit=120` returned three different incumbents — 55092.52 / 46785.55 / 41573.26, a 25 % spread — across repetitions of the same binary in one process, at an *identical* 27 nodes and with the dual bound agreeing to 12 significant figures. The tree was not moving; a primal heuristic was choosing differently, under a flag whose whole promise is that this cannot happen.

Two role-2 wall gates survived #1116's sweep, and the second was only visible once the first was fixed.

**1. The NLP-BB root RENS budget.** `_RENS_BUDGET_FRAC * (time_limit - elapsed)` capped at `_RENS_BUDGET_CAP_S`, handed to a nested `solve_model` as its `time_limit` — a fraction of what is left on the clock, never routed through `_role2_*`. As the process warmed up (the same solve went 25.5 s → 17.0 s) the same nominal slice bought more sub-MINLP. Now routed through a new `solver._role2_slice(seconds, whole=…)`, whose no-clock value is the caller's own `time_limit` rather than `None`/`math.inf`: a nested solve's `time_limit` is its own role-1 contract, and removing it would trade a reproducibility bug for a broken promise.

**2. The GDP-config plan wave** (`one_hot_config_subnlp`) — the more instructive half. It *already carried* a deterministic bound, `_WAVE_SOLVE_CAP = 48`, but `_gdp_config_deadline`'s 15 s slice always expired first:

| repeat | plans attempted | cap | slice granted | wall used |
|---|---:|---:|---:|---:|
| 1 | 36 | 48 | 15.0 s | 15.06 s |
| 2 | 37 | 48 | 15.0 s | 15.25 s |
| 3 | 38 | 48 | 15.0 s | 15.00 s |
| 4 | 37 | 48 | 15.0 s | 15.12 s |

A different prefix of the plan order picks a different disjunct. **A cap that is never reached is not a bound** — recorded on the flag's docstring, since it generalizes past this gate. Under the flag the wave's extent is now its own solve cap; the caller's deadline is left to the *dive* that follows it, which is what still bounds the stage.

Also closes the class hole that hid this. `test_912_wall_budget_inventory` knew two shapes — `clock() + budget` and `clock() - origin <cmp> budget`. A budget carved from what is *left* matches neither, because the subtraction feeds arithmetic rather than a comparison. `_scan_carved_slice` + `KNOWN_SLICES` add that third shape, discriminating a *fraction or capped piece* of the remaining time (role 2, recorded) from *all* of it (role 1).

**Part 2 of the issue — the harness.** `deterministic=True` cannot equalise work on a run that terminates on the wall clock, because the terminating condition *is* the clock. `cert_neutrality.wall_limited_rows` names rows wall-limited in **both** arms, `check_neutrality(..., exclude=…)` skips exactly that set, and both gate scripts print them as UNMEASURED — an excluded row yields no verdict rather than a silent pass. Status alone does not identify them: a run cut off by `time_limit` while holding an incumbent reports `feasible`, so the helper also takes the per-instance budgets and treats an unsettled row that spent ≥98 % of its budget as wall-limited. A certified baseline is never excluded, so a genuinely lost certificate stays a violation.

### Negative results recorded rather than shipped (CLAUDE.md §4/§11)

- **`_deadline_wall_cap`'s 3 s sub-NLP clamp** — the leading suspect. Suppressing it under the flag left the alternation *exactly* unchanged over five repetitions, at no measurable wall cost.
- **Local branching's 2 s sub-MIP slice**, routed through `_role2_horizon` — same two values, same wall correlation, no effect.
- **Freeing the config dive as well as the wave** — removed the only bound on the stage: 241 s against a 120 s `time_limit`, still nondeterministic.

The trace pointed at the wrong place twice because it resolved every read in `primal_heuristics` to the `_now` seam's own frame instead of the calling site, collapsing fifteen call sites into one label.

## Correctness

Both gates are **primal-only**: RENS and the GDP-config constructor propose incumbents that `subnlp` re-verifies for integer- and constraint-feasibility, and `inject_incumbent` enforces strict improvement. Neither touches a bound, a cut, or a relaxation, so no certificate can be affected. The dual bound is bit-identical across all five post-fix repetitions.

Both changes are behind `deterministic` (**default OFF**); with the flag off `_role2_slice` returns its argument and `wave_deadline` is the caller's deadline, so the shipped path is unchanged.

- [x] Cannot produce a false certificate — primal-heuristic extent only; no solver-math change
- [x] No validation, fallback, or safety guard was weakened. The wall deadline is *demoted*, not removed: freeing it entirely was measured to overrun `time_limit` and was rejected.

- [x] `pytest -m smoke python/tests` → **1253 passed, 13 skipped, 2 xpassed** (330 s)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` (+ `test_1187_deterministic_primal.py`) → **20 passed** (467 s)
- [x] `cargo test -p discopt-core` → **N/A — no Rust touched**
- [x] `ruff check` / `ruff format --check` clean on changed files; `mypy` clean
- Also: `pytest discopt_benchmarks/tests/ -m smoke` → **51 passed**; inventory / determinism / harness units → **44 passed**

## Regression test

`python/tests/test_1187_deterministic_primal.py` — `test_clay0303hfsg_primal_is_reproducible_under_the_flag` (slow marker) is the issue's own reproduction. It **fails before** (three incumbents 25 % apart at 27 nodes) and **passes after**, asserting **bit-equality** on node count, incumbent and dual bound rather than a tolerance.

Four ratchets guard the specific gates, each verified to fire by re-introducing the pre-fix shape:

- `test_the_rens_slice_is_routed` and `test_role2_slice_is_wired_into_the_solve_path`
- `test_the_gdp_config_wave_is_count_bounded_under_the_flag` (also asserts the dive keeps the caller's deadline)
- `test_no_unrecorded_carved_slice` / `test_the_rens_slice_is_no_longer_in_the_carved_inventory`

`discopt_benchmarks/tests/test_1187_wall_limited_rows.py` covers part 2, including the `tls2`-shaped `feasible`-at-the-wall case and the guard that a lost certification is still a violation.

Default-OFF flag, not graduating — no default flip is proposed here.

- [x] Default-OFF flag: `DISCOPT_DETERMINISTIC` / `Model.solve(deterministic=True)` (flag-OFF path byte-identical)
- [ ] Differential-bound test — N/A: no bound, cut or relaxation is touched; the change alters a *primal heuristic's extent* only
- [x] Not graduating — default-off only
- **Measurement:**
  - **Issue reproduction, 5 repeats:** bit-identical — `nodes=27`, `obj=26669.109572842823`, `bound=-1.2309090595471407e-05`, ~76 s against the 120 s limit; 48 plans attempted every time. Better incumbent than any wall-truncated repetition (was 41573–55092), the #1116 result again.
  - **Default-path bound-neutrality:** 20 in-repo MINLPLib instances, HEAD vs this branch, `status`/`node_count`/`objective`/`bound` compared as exact reprs → **19 bit-identical, 1 excluded**. The exclusion is `tls2`, which does not reproduce against itself: an interleaved A/B gave **245 / 217 / 179 nodes with three distinct dual bounds across three *baseline* runs**, every run ending `feasible` at the 30 s wall. That is the row class part 2 of this PR teaches the harness to refuse.
  - **Generality of the wave fix:** where the slice never bound, nothing changed — `flay03m` (72 plans in 2.62 s) and `flay02m` (66 in 1.69 s) return identical objectives, node counts and plan counts before and after; `syn05hfsg` never enters the constructor.

https://claude.ai/code/session_014zXckH9iftZ4dD6Tfi4gGz

---
_Generated by [Claude Code](https://claude.ai/code/session_014zXckH9iftZ4dD6Tfi4gGz)_


jkitchin:
**Python fast** went red on `cd81a4e` for a failure that is **not this PR's**, fixed here rather than handed back since it was blocking. Flagging it because it touches #1182's test and is worth knowing about generally.

```
ERROR python/tests/test_1182_simplex_lowering.py::test_a_nested_disjunction_is_refused
AssertionError: no source-predicate comparison was executed ... (CLAUDE.md §6)
assert 0 > 0
```

**Why it is not this PR's.** `test_1182_simplex_lowering` accumulates executed comparisons in a module global and asserts the total in `teardown_module`. CI runs `-n <workers> --dist loadgroup`, and that module's tests carry no `xdist_group` — so a worker handed only the *refusal* tests, which never call `_residuals`, reaches the teardown with a zero counter and errors while all 26 tests pass. Purely worker-count dependent:

| workers | result |
|---|---|
| `-n 2` | 26 passed |
| `-n 4` | 26 passed, **1 error** |
| `-n 8` | 26 passed, **2 errors** |

It reproduces identically on unmodified `main` at `-n 8` (checked in a clean worktree at `09fa315`). It surfaced here only because the previous commit turned three failures into passes and so reshuffled the distribution — this PR changed *when* it shows, not *whether* it exists.

**The fix** (`3c6818a`) keeps both halves of the original §6 intent, neither now depending on which tests a worker receives:

- *the machinery measured something* → asserted **per call** inside `_residuals`, so a silently-zero report fails immediately in whatever worker ran it. Strictly stronger than the old module-total check.
- *somebody still calls it* → asserted against the module's own source, identical in every worker. This is the case a per-call assertion cannot see: deleting every call site would otherwise leave the suite green and measuring nothing.

Both guards were **proven to fire** rather than assumed to. The per-call one raises when `disjunction_residuals` is stubbed to return `comparisons=0`. The source one needed correcting first — the obvious `count("_residuals(")` also matches `disjunction_residuals(` (the import, and the wrapper's own body), so it stayed above any threshold even with every call site deleted, which is exactly the vacuous instrument the guard exists to prevent. A bare-name regex counts 13 genuine call sites, falling to 1 when they are removed.

Verified: the module passes at `-n 1 / 2 / 4 / 8 / 12`; `pytest -m smoke python/tests -n 4 --dist loadgroup` → 1253 passed, 13 skipped, 2 xpassed; the four determinism/budget files → 30 passed; pinned ruff 0.14.6 clean over `python/`.

Happy to split this into its own PR against #1182 if you'd rather keep this one to #1187 — say the word and I'll move it.

---

jkitchin:
**Blocking status: the `cert-neutrality regression guard` is red on this PR, and I do not yet have a sound attribution.** Recording what is measured so far rather than merging on a guess.

## The check is not reproducible against itself

Two runs of **the same PR code**, differing only in which main they sat on:

| flag | run 21 (`e919fa2`, base `9509c1c`) | run 24 (`097692a`, base `28e8e5c`) |
|---|---|---|
| `root_fixpoint` | FAIL | FAIL |
| `psd_cost_gate` | ok | **FAIL** |
| `lift_zero_spanning` | **FAIL** | ok |
| `lift_loose_products` | FAIL | FAIL |
| `lu_density_route` | ok | **FAIL** |
| `node_numerical_dual_bound` | ok | ok |
| `root_build_deadline` | FAIL | FAIL |

Four arms fail in both — but **not the same four**. A deterministic defect in this diff would fail the same arms every time. A verdict set that reshuffles between runs is the signature of a noisy instrument.

That is consistent with what the gate's own flag-OFF control reports before any flag is switched on: `nvs05` losing its `optimal` status, `tls2` 153 → 255 nodes (+67 %), `st_e36` +7 %, `tspn05` +19 %. `tls2` in particular I measured independently while working on this issue — **245 / 217 / 179 nodes with three different dual bounds across three baseline runs of identical code**, every one ending `feasible` at the wall. `soundness=FAIL` here is a *lost certification* verdict, and on instances that flip certification run-to-run at a 60 s budget, that verdict is a coin toss.

## Why I doubt the diff is causal, stated as reasoning rather than proof

With `deterministic` off — which is how the gate runs — this PR's solver changes are inert by construction: `_role2_slice` returns its argument, and `wave_deadline` is the caller's deadline. Independently measured: 20 in-repo instances, HEAD vs this branch, 19 bit-identical (the exception being `tls2`, above). The harness change only ever *removes* rows from comparison via `exclude=`, which can subtract violations, never create them.

So there is no mechanism by which this diff produces a soundness violation. But mechanism-reasoning is not evidence, and the one main sample I have (run 23 on `0c3e293`: **success**) is a single draw from an instrument I have just shown to be noisy. It is not enough to conclude either way.

## What I am doing about it

A second dispatch of the same workflow on `main` is running now. Two main samples against two PR samples is the smallest comparison that can distinguish "this PR breaks four arms" from "this gate fails ~4 arbitrary arms per run on today's main".

If main fails on the second draw, the check is not this PR's and I will say so with both samples. If main passes twice while this PR fails twice, the failure is this PR's despite the absent mechanism, and I will root-cause it rather than merge.

Not merging this PR until that resolves.

---

jkitchin:
## Retracting my previous comment on this gate (CLAUDE.md §11)

In [my earlier comment](https://github.com/jkitchin/discopt/pull/1195#issuecomment-5563942495) I argued that the `cert-neutrality regression guard` failure was a noisy instrument rather than this PR's, on the grounds that runs 21 and 24 failed on *different* arms from identical PR code. **That reasoning does not survive a second control.** I said at the time it was reasoning, not proof; it is now disproved and I am withdrawing it.

### Two samples per side

| run | ref | head | event | conclusion |
|---|---|---|---|:-:|
| 23 | `main` | `0c3e293` | dispatch | **success** |
| 25 | `main` | `378abf2` | dispatch | **success** |
| 21 | this PR | `e919fa2` | pull_request | failure (4 arms) |
| 24 | this PR | `097692a` | pull_request | failure (4 arms) |

Main is green twice; this PR is red twice. Under my "noise" reading, that split has probability ~1/16. The burden is on the PR.

### A second, independent disconfirmation

I also ran the entry experiment the noise claim needed and had skipped: the cert panel measured **twice on the same machine, same code, flag OFF in both arms**, compared against itself with the gate's own `check_neutrality` under the `bound_changing` regime. If identical code failed against itself, the arm verdicts would carry no information.

```
=== A0 vs A1 (same code, flag OFF, both arms) ===
  wall-limited-in-both (excluded): ['nvs05', 'nvs17', 'tanksize']
  violations total=2  HARD(objective/status/missing)=0
   perf clay0303hfsg  node_regression  node_count 235 -> 321 (+37%)
   perf tls2          node_regression  node_count 153 -> 333 (+71%)
comparisons=1
```

**Zero soundness-class violations.** The panel does reproduce against itself at the level the gate hard-fails on; only the perf-class `node_count` notes move. The probe asserts `discopt.__file__` and the presence of the `_role2_slice` marker before measuring, and exits non-zero if it makes zero comparisons.

### What I do and do not know

I have re-read the diff for a mechanism and cannot find one:

* `python/discopt/solver.py` — `_role2_slice` returns `float(seconds)` unless `deterministic`; the RENS site computed the same float before.
* `python/discopt/_relax/primal_heuristics.py` — `wave_deadline` *is* `deadline` unless `deterministic`.
* `discopt_benchmarks/utils/cert_neutrality.py` — purely additive (`git diff` shows no removed lines); `exclude=` can only *subtract* comparisons.
* `discopt_benchmarks/scripts/graduation_gate.py` — the worker gained `wall_limited_rows` + `exclude=skipped`; per-instance `budgets` and the oracle were already there on `main`.

The gate runs with the flag off, so the solver half should be inert. That is an argument, and the argument is what just failed me once. **So I am not merging on it.**

### Why the gate could not be root-caused, and what I changed

The guard exits 1 having printed only `cert=FAIL` per arm. The rows that produced the verdict lived solely in the uploaded JSON artifact, which is not reachable from this session. A CI gate whose failure detail cannot be read from its own log is a gate nobody can act on — which is exactly why two runs' worth of failures could only be compared by *arm name*, the comparison that misled me.

`5eb38f9` makes it name them: the soundness-class violations as `FAIL` and the `node_count` notes alongside as context. Proven to fire rather than assumed to — the test asserts the rows appear for a failing arm **and** that the report stays silent for a clean one, and removing the print makes it fail. The branch is also refreshed onto `378abf2`, the exact `main` that just passed the gate, so the next run is a like-for-like comparison.

Next run will say which instances fail. I'll root-cause from that rather than merge past it.

---

jkitchin:
## Gate resolution: run 26 green, and the real mechanism (with one correction)

**Run 26 passed** — all 7 arms `soundness=ok cert=neutral` — on `5eb38f9`, refreshed onto `378abf2`, the same `main` that passed run 25. So the earlier failures were transient, which is what my first comment guessed and my second comment argued against. Neither comment had it right for the right reason, so here is the mechanism, measured.

### Correction: it is not runner speed

My first reading of run 26 was "faster runner, fewer rows tip." **That is wrong**, and run 25's log disproves it. Run 25 (`main`, **passed**) drifted *more* than run 24 (this PR, **failed**):

```
run 25 (main, PASS)   drift 5:  nvs05 status=feasible, tanksize status=time_limit, + 3 node notes
run 24 (PR,   FAIL)   drift 4:  nvs05 status=feasible,                             + 3 node notes
```

A slower runner that passed. Speed alone predicts the opposite of what happened.

### The actual tripwire is asymmetric certification

The arms compare against the live flag-OFF control, and `check_neutrality`'s status check is reference-relative. So a row that **fails to certify in the control** is *disarmed* for every arm: `base.status != "optimal"` suppresses the status check, and `nb is None` suppresses the objective check. It cannot fail anything.

The tripwire is the opposite: a row that **certifies in the control and tips in an arm** measured minutes later. Then `status` fires (`baseline optimal`) and, when the objective goes to `None`, the lost-certificate branch fires too — both soundness-class, both hard.

| run | control certified `tanksize`? | outcome |
|---|:-:|---|
| 25 (main) | no (`time_limit`) — disarmed | PASS |
| 24 (PR) | yes — armed | FAIL, 4/7 arms |

That is why the failing arms differ run to run, and why the arm count is uncorrelated with which branch is under test. Locally I confirmed the flag side is innocent: `root_fixpoint` — red in *both* CI runs — is **neutral over 4 comparisons** against two independently measured OFF controls, `HARD=0` with and without `exclude=`.

### A real bug found on the way, fixed in `3ed1519`

Running the panel at half budget (what a ~2× slower runner does to the edge rows) surfaced something separate and genuine:

```
nvs05  OFF status=feasible obj=1107.8904814191037 wall=30.2
       ARM status=feasible obj=1107.8904814191037 wall=30.3
→ HARD nvs05 objective "certified obj 1107.89… disagrees with true optimum 5.4709… — FALSE CERTIFICATE"
```

Both arms returned **the same number to the last bit** and the guard hard-failed the arm. `check_neutrality` bracketed *every* row against the oracle, including rows that never certified — but a `feasible` row holds an incumbent, and an incumbent above the optimum is an open gap, not a wrong answer. `graduation_gate` already states the correct rule for its own drift check ("Only a certified row carries a certificate … so it is skipped"); the library did not implement it. Now it does. Full strength is kept and asserted, not assumed: a *certified* objective disagreeing with the oracle still fails on the same wrong number, a lost certificate still fails on both axes, and the `bound_neutral` byte-reproducibility path is untouched. This is a pre-existing defect on `main` — re-scoring the same measurement gives `HARD=0` on the no-`exclude` path too.

**What that fix does not do:** it does **not** fix the flakiness above. The asymmetric flip still hard-fails through the `status` check, which I deliberately left alone. I am not claiming credit for solving a problem this does not solve.

### What I am not doing here

Neutralising the asymmetric flip means ruling that a certification lost *at the wall* is a perf fact, not a soundness one. The gate already says exactly that about its own drift report, and the soundness argument is available (`optimal` is in `_SETTLED_STATUSES`, so a row excluded because the *arm* hit the wall is provably never certified and can hide no false certificate). But making a guard more permissive is the change this repo warns about hardest, and it is a policy call on the gate rather than part of this determinism fix. It belongs in its own reviewable change with this evidence attached, not folded in here — I will open it separately.

### Verification on `3ed1519`

- regression test **fails before** the fix (on two identical inputs), passes after; a companion test asserts the certified case still fails
- `test_1187_wall_limited_rows.py` → **12 passed**
- `pytest discopt_benchmarks/tests/` → **484 tests, 389 passed, 95 skipped, 0 failed**
- `pytest -m smoke python/tests -n 4 --dist loadgroup` → **1253 passed, 13 skipped, 2 xpassed**
- `ruff check` + `ruff format --check` clean under the pinned 0.14.6 on both touched files

---

Issue #1198, opened 2026-09-06
relax: _flatten_sum recurses per term and hits RecursionError on t1000


jkitchin:
`_flatten_sum` in `nonlinear_bound_tightening.py` recurses once per `+` node, so
a left-deep sum of a few thousand terms exhausts the Python stack and
`tighten_nonlinear_bounds` dies. `t1000` is such a model.

## Reproducer

```python
import os
from discopt._relax.model_utils import flat_variable_bounds
from discopt._relax.nonlinear_bound_tightening import tighten_nonlinear_bounds
from discopt.modeling.core import from_nl

CORPUS = os.path.expanduser("~/Dropbox/projects/discopt-minlp-benchmark/minlplib/nl")
m = from_nl(os.path.join(CORPUS, "t1000.nl"))
lb, ub = flat_variable_bounds(m)
tighten_nonlinear_bounds(m, lb, ub)
```

```
  File "python/discopt/_relax/nonlinear_bound_tightening.py", line 3360, in tighten_nonlinear_bounds
    cand_lb, cand_ub = rule.tighten(model, prev_lb, prev_ub, metadata, **rule_kwargs)
  File "python/discopt/_relax/nonlinear_bound_tightening.py", line 2906, in tighten
    terms = _cached_flat_terms(model, con.body)
  File "python/discopt/_relax/nonlinear_bound_tightening.py", line 179, in _cached_flat_terms
    _flatten_sum(expr, 1.0, terms)
  File "python/discopt/_relax/nonlinear_bound_tightening.py", line 725, in _flatten_sum
    _flatten_sum(expr.left, scale, out)
  File "python/discopt/_relax/nonlinear_bound_tightening.py", line 721, in _flatten_sum
    _flatten_sum(expr.left, scale, out)
RecursionError: maximum recursion depth exceeded
```

Note the recursion is entirely down `expr.left` -- a left-deep `((a+b)+c)+d…`
chain, which is what a long linear row parses to.

## Why it matters

Same shape as the `ex7_3_6` report: the exception escapes
`tighten_nonlinear_bounds`, so the instance drops out of any panel that touches
this path. It is a scaling limit, not a soundness bug, but it is a *silent size
cliff* -- the rule works on a 500-term row and dies on a 5000-term one, and
nothing in the model says which side of the line an instance is on.

## What "done" looks like

`_flatten_sum` becomes iterative -- an explicit work stack of `(expr, scale)`
[...]

Issue #1160, opened 2026-09-05
correctness: a constraint on dm.sum(X, axis=k) is certified optimal at the axis-collapsed answer (false bound)


jkitchin:
## Summary

A constraint whose body is an **axis-reduced** sum, `dm.sum(X, axis=k)`, is solved as if the reduction were over *every* axis. The solver returns `status="optimal"` with a dual bound that **excludes the true optimum**, on a model whose own NLP evaluator handles the rows correctly.

Found while building a Python-side differential panel for jkitchin/discopt#1158; it is **pre-existing and unrelated to that PR** — measured identically on `1e75919` (`main`) and on the PR head, so nothing in #1158 causes or fixes it.

## Reproducer

```python
import numpy as np
import discopt.modeling.core as dm
from discopt.validation.feasibility import verify_point

m = dm.Model("sum_axis")
A = m.continuous("A", shape=(2, 3), lb=0, ub=1)
m.subject_to(dm.sum(A, axis=1) <= 2)     # per-ROW cap
m.minimize(-dm.sum(A))                   # maximize the total

r = m.solve(time_limit=60, gap_tolerance=1e-6)
print(r.status, r.objective, r.bound)
# -> optimal -2.0000000 -2.0000000

# ... but this point is feasible and twice as good:
better = np.array([1., 1., 0., 1., 1., 0.])   # row sums 2 and 2
print(verify_point(m, better, with_objective=True))
# -> VerifyResult(ok=True, objective=-4.0, reason='')
```

Each row sums to at most 2, so the total is at most 4 and the true optimum is **-4**. The solver certifies **-2**, and reports it as a *dual bound* — for a minimize, a valid lower bound must be `<= -4`, so `-2` is above the optimum and fathoms it away. `incorrect_count` would count this row.

## What is and is not at fault

The NLP evaluator is **correct** — it is not a row-mapping problem:

```
evaluator n_constraints: 2            (expected 2 rows: one per row of A)
row map:                [(0, 2, 'rowcap')]
g(x) at the true optimum [1,1,0,1,1,0]: [0. 0.]     (both rows satisfied)
```

The failure is downstream, in the relaxation/branch-and-bound path. The returned value is exactly the answer of the **axis-collapsed** model:

```python
m2 = dm.Model("total")
B = m2.continuous("B", shape=(2, 3), lb=0, ub=1)
m2.subject_to(dm.sum(B) <= 2)          # TOTAL cap, not per-row
[...]

PR #1200, opened 2026-09-07
fix(relax): #1197 decide crossed intervals in DefinedVariableForwardRule

### LINKED ISSUE #1197: relax: DefinedVariableForwardRule raises "Interval lo > hi" on ex7_3_6 from a 3e-15 rounding crossing
`DefinedVariableForwardRule` builds an `Interval` from a forward-propagated
range whose endpoints have crossed by one ulp of rounding, and the `Interval`
constructor's own invariant check raises out of `tighten_nonlinear_bounds`.

## Reproducer

```python
import os
from discopt._relax.model_utils import flat_variable_bounds
from discopt._relax.nonlinear_bound_tightening import tighten_nonlinear_bounds

CORPUS = os.path.expanduser("~/Dropbox/projects/discopt-minlp-benchmark/minlplib/nl")
m = from_nl(os.path.join(CORPUS, "ex7_3_6.nl"))
lb, ub = flat_variable_bounds(m)
tighten_nonlinear_bounds(m, lb, ub)
```

```
  File "python/discopt/_relax/nonlinear_bound_tightening.py", line 3360, in tighten_nonlinear_bounds
    cand_lb, cand_ub = rule.tighten(model, prev_lb, prev_ub, metadata, **rule_kwargs)
  File "python/discopt/_relax/nonlinear_bound_tightening.py", line 2953, in tighten
    box[var] = Interval(upd_lo, upd_hi)
  File "python/discopt/_relax/convexity/interval.py", line 152, in __post_init__
    raise ValueError(f"Interval lo > hi: lo={lo}, hi={hi}")
ValueError: Interval lo > hi: lo=1.0, hi=0.999999999999997
```

The crossing is 3e-15 -- pure floating-point noise on a variable whose true
range is the single point 1.0, not a genuine infeasibility proof.

## Why it matters

The exception escapes `tighten_nonlinear_bounds` entirely, so any caller that
tightens this model dies rather than getting a (possibly untightened) box back.
It is not a soundness bug -- nothing wrong is *returned* -- but it takes the
instance out of every panel that touches the nonlinear tightening path.

## What "done" looks like

`DefinedVariableForwardRule` decides explicitly what a crossed interval means
before constructing the `Interval`, rather than letting the dataclass invariant
decide for it. The two defensible readings are:

1. the crossing is within a rounding tolerance (as here, 3e-15): collapse to the
   midpoint, or widen to `[min(lo,hi), max(lo,hi)]`, and continue; or
2. the crossing exceeds any rounding tolerance: that is a genuine empty
   interval, which the module already has a first-class representation for --
   report infeasibility through `stats.infeasible` / `_prove_infeasible` (see
   `_EMPTY_INTERVAL_FEAS_TOL = 1e-6`), not by raising.

Per CLAUDE.md §3 the fix must not be a bare `except ValueError` around the
construction -- that would convert "this rule is broken on this shape" back into
"this rule is fine".

## Provenance

Found while panelling #1193 over the MINLPLib corpus. Reproduces identically
with `DISCOPT_ELLIPSOID_BOUNDS` off, i.e. it is present on `main` and is not
related to that flag. Same rule as the sibling report for `t1000`.

jkitchin:
## Summary

`DefinedVariableForwardRule` intersected the forward-substituted enclosure of a defined
variable with its box and then built `Interval(upd_lo, upd_hi)` from the result. When the
two crossed by rounding noise the dataclass invariant raised a bare `ValueError` out of
`tighten_nonlinear_bounds` entirely, so any caller that tightened the model died rather
than getting a (possibly untightened) box back — `ex7_3_6` crosses by 3e-15 on a variable
whose true range is the single point 1.0. The old guard
`upd_lo <= upd_hi + _EMPTY_INTERVAL_FEAS_TOL` admitted exactly the sub-tolerance crossings
that then blew up in the constructor, and silently skipped the row for anything larger.

The rule now decides what a crossed interval means *before* constructing it:

- **crossing beyond `_EMPTY_INTERVAL_FEAS_TOL`** → `_prove_infeasible`, the first-class
  representation every sibling rule already uses, which the driver turns into
  `stats.infeasible`. This was previously a silent skip — inconsistent with the driver,
  which already classifies the same crossed box as infeasible when a rule returns one.
- **crossing within tolerance** → collapse to the degenerate box at the enclosure's own
  endpoint (the same outward repair `_snap_tolerant_crossovers` applies box-wide), and
  continue.

No `except ValueError` around the construction, per CLAUDE.md §3. `tightened` is still
computed from the pre-snap values, so non-crossing rows are byte-identical.

Closes #1197

## Correctness

The sub-tolerance repair snaps *outward* — to the enclosure's own endpoint, i.e. below the
declared lower bound — never to the tighter declared bound, so eps-scale rounding cannot
cut a feasible point. The new infeasibility path is an intersection of a valid interval
enclosure with the incumbent box, tested with the module's existing
`_EMPTY_INTERVAL_FEAS_TOL`, identical to how every sibling rule in the file proves
infeasibility; it is also the same verdict the driver already reaches on its own when a
rule returns a crossed box. The corpus panel below confirms no instance newly reports
infeasible.

- [x] `pytest -m smoke python/tests` → 1209 passed, 21 skipped, 9445 deselected, 2 xpassed (315 s)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → 19 passed (160 s)
- [x] `ruff check` / `ruff format --check` → clean (ruff 0.14.6)

Also run: `test_defined_variable_forward_fbbt.py` → 5 passed; the FBBT-adjacent files
(`test_fbbt_aux_chain_bounds`, `test_fbbt_bindings`, `test_fbbt_bounds_suspect_parity`,
`test_issue_1066_fbbt_row_sweep`, `test_nonlinear_bound_tightening_tol`) → 26 passed.

**Differential panel.** `tighten_nonlinear_bounds` over all 66 `.nl` instances in
`python/tests/data/minlplib_nl/`, `main` vs this change, comparing the full returned
`(lb, ub)` arrays plus `infeasible`, `infeasibility_reason`, `n_tightened` and
`applied_rules`: **66 compared, 0 differing**. Bound-neutral, and no instance newly
reports infeasible. Each arm asserted `nbt.__file__` and the presence/absence of a marker
string unique to the patched version before measuring (CLAUDE.md §8), and the panel exits
non-zero on zero instances compared (§6).

## Regression test

Two tests in `python/tests/test_defined_variable_forward_fbbt.py`, both confirmed to fail
before the change and pass after (`git stash` on the source file alone; before → 2 failed,
3 passed with `ValueError: Interval lo > hi: lo=1.0, hi=0.9999999999999972`; after → 5
passed):

- `test_sub_tolerance_crossing_snaps_instead_of_raising` — the `ex7_3_6` shape returns a
  degenerate box instead of raising. It asserts the crossing was actually exercised
  (`0 < 1.0 - hi < _EMPTY_INTERVAL_FEAS_TOL`) so it cannot pass vacuously.
- `test_beyond_tolerance_crossing_is_reported_as_infeasible` — `s == 2.7/3` with `s >= 1`
  (genuinely infeasible) surfaces through `stats.infeasible` naming the rule, not as an
  escaping `ValueError` and not as a silently skipped row.

`ex7_3_6.nl` is not in the in-repo corpus and minlplib.org is blocked by this
environment's network policy (`CONNECT tunnel failed, 403`), so the reproducer is a
synthetic instance of the same shape — `s == x/3` with `x` fixed just below 3.0 and `s`
free above a declared `lb = 1.0`. It reproduces the issue's trace values exactly
(`lo=1.0, hi=0.999999999999997`). Confirming on the real `ex7_3_6` needs a machine with
the MINLPLib snapshot.

N/A — the panel above shows the returned boxes are byte-identical on every corpus
instance; the only behavior change is on rows that previously raised or were silently
skipped.

https://claude.ai/code/session_017ZQBztHoGuWEP4VS53o3Eg

---
_Generated by [Claude Code](https://claude.ai/code/session_017ZQBztHoGuWEP4VS53o3Eg)_

### DIFF
--- a/python/discopt/_relax/nonlinear_bound_tightening.py
+++ b/python/discopt/_relax/nonlinear_bound_tightening.py
@@ -2861,9 +2861,12 @@ class DefinedVariableForwardRule(NonlinearBoundTighteningRule):
     subtree stay bounded, so the spatial dual bound becomes *certifiable* instead of
     being dropped as un-rigorous on an unbounded node (issue: nvs/gear unbounded
     auxiliaries). Sound: an interval enclosure of the defining expression is a valid
-    enclosure of the variable; bounds are only ever intersected, never loosened. The
-    fixpoint loop in :func:`tighten_nonlinear_bounds` resolves chained definitions
-    (``x7`` defined via ``x6`` resolves once ``x6`` is bounded).
+    enclosure of the variable; bounds are only ever intersected, never loosened --
+    the one exception being a sub-tolerance crossing, which is snapped *outward* to
+    the enclosure's own endpoint (see ``tighten``) so eps-scale rounding cannot
+    manufacture an empty box. The fixpoint loop in :func:`tighten_nonlinear_bounds`
+    resolves chained definitions (``x7`` defined via ``x6`` resolves once ``x6`` is
+    bounded).
     """
 
     name = "defined_variable_forward"
@@ -2940,7 +2943,29 @@ class DefinedVariableForwardRule(NonlinearBoundTighteningRule):
                 or upd_lo > out_lb[iso_idx] + 1e-12
                 or upd_hi < out_ub[iso_idx] - 1e-12
             )
-            if tightened and upd_lo <= upd_hi + _EMPTY_INTERVAL_FEAS_TOL:
+            # Decide here what a crossed (``upd_lo > upd_hi``) interval means,
+            # rather than constructing the ``Interval`` and letting its dataclass
+            # invariant raise out of the whole tightening pass (#1197).
+            if upd_lo > upd_hi + _EMPTY_INTERVAL_FEAS_TOL:
+                # Beyond any rounding slack the defining expression's enclosure
+                # and the variable's box share no point. That is a genuine empty
+                # interval, reported the way every sibling rule reports one --
+                # the caller turns it into ``stats.infeasible``.
+                _prove_infeasible(
+                    self.name,
+                    con,
+                    f"defining interval [{new_lo}, {new_hi}] for flat variable "
+                    f"{iso_idx} is disjoint from its box "
+                    f"[{out_lb[iso_idx]}, {out_ub[iso_idx]}]",
+                )
+            if upd_lo > upd_hi:
+                # Sub-tolerance crossing: floating-point noise on an effectively
+                # degenerate range (ex7_3_6 crosses by 3e-15 on a variable whose
+                # true range is the single point 1.0), not an infeasibility proof.
+                # Collapse to the degenerate box at the enclosure's endpoint, the
+                # same repair ``_snap_tolerant_crossovers`` applies box-wide.
+                upd_lo = upd_hi
+            if tightened:
                 out_lb[iso_idx] = upd_lo
                 out_ub[iso_idx] = upd_hi
                 var = idx_to_var.get(iso_idx)
--- a/python/tests/test_defined_variable_forward_fbbt.py
+++ b/python/tests/test_defined_variable_forward_fbbt.py
@@ -24,7 +24,13 @@ from pathlib import Path
 import discopt.modeling as dm
 import numpy as np
 import pytest
-from discopt._relax.nonlinear_bound_tightening import tighten_nonlinear_bounds
+from discopt._relax.nonlinear_bound_tightening import (
+    _EMPTY_INTERVAL_FEAS_TOL,
+    DefinedVariableForwardRule,
+    NonlinearBoundTighteningInfeasible,
+    _cached_flat_metadata,
+    tighten_nonlinear_bounds,
+)
 from discopt.modeling.core import from_nl
 from discopt.solver import _extract_variable_info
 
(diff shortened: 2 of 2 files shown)


jkitchin:
CI is green on `ade44b0` (all 14 checks, `mergeable_state: clean`). Recording why the earlier run on `77018e6` was red, since it was not this PR's and it will recur on other PRs.

**The failure.** `Python correctness (fast subset)` on `77018e6`:

```
FAILED python/tests/test_issue_1160_axis_sum_certificate.py::test_axis_sum_with_an_equality_sense
       - Failed: Timeout (>120.0s) from pytest-timeout
ERROR  ...::test_axis_zero_on_a_one_dimensional_operand_is_a_full_reduction
       - AssertionError: 1 certificate check(s) entered but never reached their
         objective/bound comparison; assert 5 == 6
```

The ERROR is the `teardown_module` guard cascading off the timeout, not a second defect.

**Root cause: a zero-margin race, pre-existing.** `_solve()` in that module passes `time_limit=120.0`; the lane runs `--timeout=120`. The solve never certifies this instance, so it always burns the full budget and the test wall lands at 120.0 + ε — on the wrong side of the harness deadline whenever scheduling noise goes the wrong way. On the last green `main` run (`9509c1c`, run 2754) the same test reports **120.01 s** in `--durations` and passed; here it reported 120.00 s and was killed. Same code path, opposite coin flip. It is the hazard already documented on `python-correctness-slow`: *"a per-test timeout below the budget would fail the harness instead of the assertion."*

**Not this PR's, three ways.** (1) `DefinedVariableForwardRule` cannot fire on this model at all — `_isolated_defined_var` only considers variables unbounded on a side, and every variable here is `lb=0, ub=1`. (2) Interleaved local A/B of that test, marker-gated per arm (`nbt.__file__` plus a string unique to the patched version, asserted present on one arm and absent on the other): branch 120.04 / 120.04 s, `main` 120.03 / 120.02 s — both saturate the budget. (3) Node-limited A/B on the same instance (`max_nodes=2000`, `time_limit=3600`), two replicates each arm: `nodes=2013` both, `obj=-3.999999945479066` both, `bound=-6.000000000000008` both — bit-identical.

**Proposed fix, not taken here** (out of this PR's scope, and it belongs to whoever owns #1160): the solve returns `status="feasible"` with `bound=-6.0` at every budget tried, up to 600 s / 15 647 nodes — it never closes the gap, and `_assert_min_certificate` asserts objective and bound, never certification. So the 120 s buys nothing and could drop to ~20-30 s, which removes the race and gives the lane back ~90 s. The alternative — raising the lane's `--timeout` above the largest in-lane budget — leaves a 120 s burn in a lane advertised as "~5 min serially". Happy to open that as its own issue/PR if you want it.

Re-validated end to end after the merge of `main` (which brought #1201's iterative `_flatten_sum` into the same module): differential panel over the 66-instance in-repo corpus against the new base — 66 compared, 0 differing; `pytest -m smoke` 1209 passed / 21 skipped / 2 xpassed; `pytest -m slow test_adversarial_recent_fixes.py` 19 passed; FBBT-adjacent files plus #1198's new module 35 passed; `ruff check python/` and `ruff format --check` clean.

---

PR #1217, opened 2026-09-09
feat(#1216): dm.argmin / dm.argmin_kkt — an inner NLP as a block of an outer model, and the active set in pounce_sensitivity

### LINKED ISSUE #1216: Embed an inner NLP as a block in an outer model: POUNCE forward solve with sIPOPT sensitivities (1st and 2nd order)
## Summary

I would like to embed an inner NLP as a block inside an outer model: pass the inner
optimization problem as it is — objective and constraints, no reformulation — have its
forward pass solved by POUNCE, and have POUNCE's existing sensitivity machinery supply the
derivatives the outer solver needs.

Most of the machinery is already present. `discopt.solvers.sipopt.pounce_sensitivity` takes
a `dm.Model` unchanged, solves it with POUNCE, and returns the sIPOPT sensitivity system

```
[W  Jᵀ] [dx*/dp]  =  -[∂²L/∂x∂p]
[J   0] [dλ*/dp]      [∂g/∂p   ]
```

as `SensitivityResult.dx_dp` (n × n_params) and `.dlambda_dp` (m × n_params).

Two gaps prevent using it as an inner block:

1. It returns arrays to Python. There is no construct that makes the solve a node in an
   outer `Model`, so an outer POUNCE solve cannot consume the derivatives.
2. It stops at first order. There is no `d²x*/dp²`. `ReducedHessian` exposes only
   `is_positive_definite`, and `QpSensitivity.reduced_hessian` is QP-only.

## Why the existing workaround is insufficient

Without such a block, embedding a follower means writing its KKT conditions by hand into
`dm.implicit_full_space`. For a non-convex follower this is a relaxation rather than a
reformulation: the KKT set contains every stationary point, and the leader is free to select
whichever one favours its own objective.

The follower below is a projection onto the unit circle. The objective is strictly convex,
but the nonlinear equality makes the feasible set non-convex, so the KKT system has two
roots — the near side (the minimizer, `y* = (p,1)/√(p²+1)`, hence `y0 > 0` for `p > 0`) and
the far side (a maximizer). The leader asks for `y0 = -0.9`, which no projection can produce.

```python
import numpy as np
import discopt.modeling as dm
from discopt.modeling import Model

def kkt_residual(u, v):
    p = u[0]
    y0, y1, lam = v[0], v[1], v[2]
    return [2 * (y0 - p) + 2 * lam * y0,
            2 * (y1 - 1.0) + 2 * lam * y1,
            y0 * y0 + y1 * y1 - 1.0]

m = Model("leader")
p = m.continuous("p", lb=0.05, ub=3.0)
v = dm.implicit_full_space(m, kkt_residual, [p], n_unknowns=3,
                           bounds=(np.array([-2.0, -2.0, -10.0]),
                                   np.array([2.0, 2.0, 10.0])))
m.minimize((v[0] + 0.9) ** 2)
r = m.solve()

pv = float(np.asarray(r.value(p)).ravel()[0])
yv = np.asarray(r.value(v)).ravel()[:2]
true = np.array([pv, 1.0]) / np.sqrt(pv ** 2 + 1.0)
print(r.status, pv, yv, true)
```

```
status         : optimal
p*             : 2.064741
y* returned    : [-0.9     -0.43589]
true projection: [ 0.9      0.43589]
leader objective at returned y* : 1.415e-15
leader objective at true y*     : 3.240e+00
```

The returned `y*` is not the projection of any point; it is the far-side KKT root, reported
as optimal. `dm.implicit` shows the same behaviour and is additionally dependent on the
starting point: from a near-side `x0` it recovers the correct outer optimum
(`p* = 2.064741` against an analytic `2.064742`), and from a far-side `x0` it terminates at
`p* = 0.05`, the lower bound.

Solving the same follower as an optimization problem does not have this failure mode,
because only minimizers are representable. Using `make_nlp_layer`, which wraps the same
POUNCE machinery for JAX:

```
exact  y* = [0.4472136  0.89442719]   dy*/dp = [ 0.71554175 -0.35777088]
layer  y* = [0.4472136  0.89442719]   err = 8.02e-13
layer dy*/dp = [ 0.71554175 -0.35777088]   err = 3.78e-11

p=0.2: y* = [0.196116 0.980581]   minimizer
p=1.0: y* = [0.707107 0.707107]   minimizer
p=2.0: y* = [0.894427 0.447214]   minimizer
```

The correct behaviour is therefore already implemented; it is not reachable from the
algebraic modelling layer.

## Request

**1. An argmin block that takes the inner model unchanged.**

```python
inner = Model("projection")     # objective and constraints, untouched
...
v = dm.argmin(inner, bind={p_inner: outer_expression})
m.minimize(f(v))
```

Forward pass: a POUNCE solve of `inner`. Derivatives passed to the outer solver: the KKT
sensitivity system `pounce_sensitivity` already assembles.

**2. Second-order sensitivities.** `d²x*/dp²` from the same factorized KKT system, so an
outer NLP solver can use exact Hessians instead of a quasi-Newton approximation. First order
alone is already useful, so an `order=1|2` switch would be a reasonable compromise if second
order proves expensive.

**3. Exact right-hand side.** `pounce_sensitivity` currently forms `∂²L/∂x∂p` and `∂g/∂p`
by central finite differences on parameter values. Exact derivatives from the tape would
improve accuracy and cost, and matter more once second order is built on top.

## Relation to a Pyomo grey box

The contract is that of Pyomo's `ExternalGreyBoxModel`, with the forward pass being an
optimizer rather than a simulator.

| | Pyomo `ExternalGreyBoxModel` | requested `dm.argmin` |
|---|---|---|
| what is passed | callback plus hand-written Jacobians | the model, unchanged |
| forward pass | user callback, typically a simulator | POUNCE solve |
| derivatives | user-supplied; Hessian optional | POUNCE sIPOPT machinery, 1st and 2nd order |
| outer solver | cyipopt through PyNumero only | POUNCE |
| `.nl` export | not possible (Python callback) | desirable |

The Pyomo route requires cyipopt and cannot be written to `.nl`, which prevents comparing
solvers on an identical formulation. discopt has neither restriction; the gap is only the
missing connection between an inner solve and the outer model.

## Existing components this would compose

- `solvers/sipopt.pounce_sensitivity` — model as-is, POUNCE solve, `dx_dp` and `dlambda_dp`.
- `_relax/pounce_layer.make_nlp_layer` — the same idea already bridged to JAX through
  `custom_vjp`. Reverse-mode only: `jax.hessian` raises *"can't apply forward-mode autodiff
  (jvp) to a custom_vjp function"*.
- `dm.implicit_full_space` — demonstrates how a block can be lowered so that the Rust tape,
  and therefore POUNCE, handles it.

## Environment

discopt 0.8.0, pounce-solver 0.11.0, Python 3.10, macOS arm64.

jkitchin:
## Summary

Closes #1216 — embedding an inner NLP as a block of an outer model — and, on the way, fixes two bugs the issue's machinery was standing on.

The block ships in two arms, the same trade `dm.implicit` / `dm.implicit_full_space` already make, so it is a choice the caller makes rather than a lowering that changes under them:

**`dm.argmin(inner, bind={q: p})` — solve the follower.** The inner problem is passed as a `Model`, unchanged: its forward pass is a POUNCE solve, and its derivatives are the sIPOPT KKT sensitivity wired into the outer solve through `jax.lax.custom_root` over the inner KKT residual rather than returned as arrays. Because that rule is expressed in ordinary differentiable operations, **first and second order both work** — the outer solver's Lagrangian Hessian is a forward-over-reverse pass through the node — and the right-hand side is exact, differentiated off the compiled DAG instead of finite-differenced. Takes any follower discopt can solve; reports its answer as local, per the `CustomCall` contract.

**`dm.argmin_kkt(model, inner, bind={q: p})` — lower the follower.** Its variables become real variables of the outer model and its KKT conditions become real constraints, so the Rust tape, FBBT, a **global certificate** and **`.nl` export** all come back. Measured on a convex follower (`min (y-q)² s.t. y ≤ 0.5`) under a leader wanting `y* ≈ 0.3`:

| arm | `p*` | status | `gap_certified` | `.nl` |
|---|---|---|---|---|
| `dm.argmin` | 0.297030 | feasible | False | refused |
| `dm.argmin_kkt` (`kkt`/`gdp`) | 0.297030 | optimal | True | no `.nl` form (GDP/SOS1 constructs) |
| `dm.argmin_kkt` (`strong_duality`) | 0.297030 | optimal | True | **664 bytes, valid `g3` header** |

`method="strong_duality"` emits pure algebra — stationarity, primal/dual feasibility, and the single bilinear equality `Σ μᵢgᵢ == 0` — so the identical formulation can be handed to BARON or SCIP, which is what the issue's `.nl` row asks for.

The cost of lowering is not negotiable: KKT conditions characterize a follower's optimum only when the follower is **convex in its own variables**. For a nonconvex follower the KKT set contains every stationary point, so lowering it would reproduce exactly the failure the issue reports — hand-writing that system into `dm.implicit_full_space` on the reporter's projection follower (`min ||y-(p,1)||² s.t. ||y||²=1`, leader asking for `y0 = -0.9`) returns `y* = (-0.9, -0.436)`, the far-side root, and certifies it, because that point really is optimal *for the model as stated*. `argmin_kkt` therefore runs `BilevelProblem`'s convexity certifier and **refuses** what it cannot prove, with no flag to override; `dm.argmin` takes that follower instead and lands on the honest optimum (`p* = 0.05`, objective `0.9024`, matching a 20 001-point brute-force scan). Both behaviours are pinned by tests.

Two bugs found while building it:

* **`pounce_sensitivity` ignored the active set.** It assembled `[W Jᵀ; J 0]` over *all* constraint rows — as though every constraint were an active equality — and did not represent variable bounds at all. On `min (y-q)² s.t. y ≤ 1` at `q = 0.5`, where the bound is slack and the true `dy*/dq` is 1, it returned `-2e-10`. On this repo's own sIPOPT tutorial portfolio at a slack return target it returned `[8.45, 0.55, -3.85, 4.44, -3.39, -6.20]` where a black-box re-solve gives `0`. `_relax.pounce_layer.make_nlp_layer` shared the assembly and so had the same defect. Identification is now primal-**dual**, not a distance threshold: POUNCE is an interior-point method, and on that same portfolio at a *binding* target two weights sit 2.0e-6 and 1.5e-6 off their lower bound while carrying bound multipliers of ~1e-3.
* **`x ** k` broke third-order forward-mode AD at a negative base.** The DAG compiler emitted the exponent as an array, which JAX lowers to `pow`, whose exponent-argument derivative carries `log(base)`. Values, gradients and Lagrangian Hessians are unaffected — measured, not assumed — but one nesting deeper the `0 * NaN` poisons everything, and that is exactly where a `custom_root` sensitivity rule lives. Constant integral exponents now lower to `integer_pow`.

Also on the branch: `order=2` on `pounce_sensitivity` (returns `d2x_dp2`), an exact right-hand side by default with `method="fd"` kept as an independent cross-check, and the sIPOPT tutorial notebook re-executed (its portfolio example used a slack return target, so every sensitivity it printed was noise; its equality-QP cell claimed `dλ*/dc = +1` while printing `-1`).

A sixth commit (e6dd1d2) fixes an unrelated CI coin flip it surfaced: `test_issue_1160_axis_sum_certificate.py`'s solve budget was `time_limit=120` under the fast-correctness lane's `--timeout=120`, so the watchdog rather than the assertions decided that test. See the comment below for the measurements.

All four of the issue's asks are covered — the block, second-order sensitivities, an exact right-hand side, and `.nl` export for every follower where an `.nl` form exists at all. For a nonconvex follower none does: the only algebraic form is the KKT system, which is a relaxation and not the argmin, so the refusal is the answer rather than a missing feature.

## Correctness

No change to any bound, cut, relaxation, or branch-and-bound path. `dm.argmin` is a `CustomCall`, which the existing contract already confines to the local NLP path with no dual bound — it cannot make a global claim. `dm.argmin_kkt` emits ordinary constraints and reuses `discopt.bilevel`'s existing, already-tested reduction and its convexity gate; it adds no new reformulation of its own, and its own gate refuses (never assumes) convexity. The `pounce_sensitivity` change moves an *analysis* API from wrong numbers to right ones; it does not feed the solver.

Every soundness gate refuses rather than approximates. `dm.argmin`: integer/binary inner variables and a maximize inner objective raise; a failed inner solve, a KKT residual that does not check out at the returned multipliers, or a reduced Hessian that is not PSD evaluate to NaN, which the outer NLP solver reports as a failed evaluation (a traced solve cannot raise — the same choice `dm.implicit` documents); a degenerate active set warns that `dx*/dp` is one-sided. `dm.argmin_kkt`: a follower the certifier cannot prove convex, an opaque node inside the follower, a multi-dimensional follower variable, and a vector-valued follower row are each refused by name — and the expression rewriter that transplants the follower raises on any node type it does not handle explicitly, since passing one through unchanged would leave the inner model's own variable in an outer constraint.

- [x] Cannot produce a false certificate (no solver-math change; `argmin` is local-NLP-only by the `CustomCall` contract, `argmin_kkt` reuses the existing gated bilevel reduction)

- [x] **CI is green on the current head (e6dd1d2)**: every job of run 2796 succeeded — fast correctness, PR-fast, claim-boundary, Rust (build, test, `--no-default-features`, clippy ×2, parallel≡serial), AMP coverage, lint+typecheck.
- [x] `pytest -m smoke` → 1209 passed, 20 skipped, 9590 deselected, 2 xpassed
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → 19 passed
- [x] `cargo test -p discopt-core` → N/A locally (no Rust touched); CI ran it anyway, green
- [x] `ruff check` / `ruff format --check` clean; `mypy` clean on all four changed modules
- Targeted, `-m ""` (argmin + sipopt + pounce_layer + all four bilevel suites + both implicit suites + dag_compiler) → **210 passed**
- I do not have a local full-non-slow-suite count: three attempts were each cut off by an environment restart mid-run (green to ~3700 tests, ~70% and ~65% respectively). CI's lanes above cover that ground and are green, so I stopped re-running it rather than leave a promise I could not keep.

## Regression test

`python/tests/test_argmin_block.py` (27 tests) and `python/tests/test_sipopt_sensitivity.py` (13 tests) are new; `TestIntegralExponentLowering` is added to `test_dag_compiler.py`. Fail-before/pass-after confirmed by stashing the compiler change and re-running:

```
FAILED test_dag_compiler.py::TestIntegralExponentLowering::test_integral_exponent_lowers_to_integer_pow
FAILED test_argmin_block.py::test_layer_matches_exact_solution_and_both_derivatives[0.2]  (Obtained: nan)
FAILED ... [1.0]  FAILED ... [2.0]  FAILED ... [2.5]
```

The sensitivity tests fail the same way against the old assembly — measured before the fix: `dx/dq = -2e-10` where the closed form is `1.0`, and `[8.45, …]` where a re-solve gives `0`.

Derivative assertions compare against closed forms or an independent finite difference of a *re-solved* problem, never against the other code path alone, and the parametrized grid carries an executed-comparison count (`test_derivative_cases_all_ran`) so it cannot silently stop running. The two arms are cross-checked against each other on a convex follower, where they must agree.

https://claude.ai/code/session_013NKdeG57FFByJUtqqFEsVe


jkitchin:
CI was red on `Python correctness (fast subset)`. Root cause and fix pushed in e6dd1d2.

**What failed.** One test, `test_issue_1160_axis_sum_certificate.py::test_axis_sum_with_an_equality_sense`, hit `Failed: Timeout (>120.0s) from pytest-timeout`. The reported `ERROR` on the neighbouring test is that file's `teardown_module` vacuity guard firing because the timed-out test entered `_assert_min_certificate` and never reached its comparison — one failure, not two.

**Why it is not this PR's diff.** The failing model is `min -sum(A*A) s.t. sum(A, axis=1) == 2` — no `**`, no argmin block, nothing this PR touches. The immediately preceding commit on this branch (1f48fc2), which carries *every* library change in the PR, passed the same job at run 2794; the failure appeared at 2795, whose only new content is a new module, its tests, and a CHANGELOG entry. The two jobs took 833 s and 831 s.

**The actual cause.** That test's solve budget is `time_limit=120.0` and this lane runs with `--timeout=120`, so the per-test watchdog and the test's own wall time are the same number. The solve is a concave minimization that never closes its gap — it spends whatever budget it is given:

| `time_limit` | wall | nodes | objective | bound |
|---|---|---|---|---|
| 5 s | 5.12 s | 383 | −3.999999945 | −6.0 |
| 120 s | 120.04 s | 5297 | −3.999999945 | −6.0 |
| 300 s | 300.03 s | 9101 | −3.999999945 | −6.0 |

So the test always runs ≈120 s and the watchdog, not the assertions, decides it. That is a coin flip on any branch.

**Fix.** Give the solve a 30 s budget — a 4× margin under the watchdog, and a ~30× margin over the first budget that reaches the answer at all (1 s works; 0.5 s returns `time_limit` with no objective, which the file's own assertions catch). The incumbent (−4) and the dual bound (−6) are bit-identical at 1, 2, 5, 10, 20, 40, 120 and 300 s, so nothing is given up: a shorter budget can only make these assertions harder to satisfy, never easier. No assertion, tolerance, or guard is touched. `ci.yml` already states this principle for the slow lane — *"the per-test timeout must not be what decides a test"*, which is why that lane sets `--timeout=1800` against its 600/1500 s deadlines.

Verified locally under the lane's exact setting: `pytest python/tests/test_issue_1160_axis_sum_certificate.py -m "" --timeout=120` → **17 passed in 30.96 s** (was 131 s, decided by the watchdog).

**One thing left for you.** This is the instance; the class is that the fast-correctness lane's `--timeout=120` is not above the largest solve budget its tests declare — several other `correctness` tests pass `time_limit=120`, and any of them that stops converging becomes the same coin flip. Raising that lane's `--timeout` to 300 (the repo's default per CLAUDE.md) would close the class, with `timeout-minutes: 30` still the real hang backstop, exactly as the slow lane is set up. I have not made that change here because it is CI policy for every user of the lane rather than something this PR needs — say the word and I will.

---

jkitchin:
CI is green on `e6dd1d2`. Run 2796, every job:

| job | result |
|---|---|
| Lint + typecheck (ruff, ruff-format, mypy) | success |
| Python correctness (fast subset) | success — 12:26, was 13:54 |
| Python fast (PR-fast, parallel) | success |
| Python claim-boundary (serial) | success |
| Rust (build, test, `--no-default-features`, clippy ×2, parallel≡serial) | success |
| Python AMP coverage | success |

The fast-correctness lane — the one that was red — now passes, and 88 s faster, which is the 120 s → 30 s budget on the axis-sum test minus what it actually needs. `mergeable_state` is `clean` against `main`.

Nothing outstanding on my side: no open review threads, no conflict. The one item still with you is the class fix in the previous comment (raising that lane's `--timeout` above the largest solve budget its tests declare, so the next test that stops converging doesn't become the same coin flip). I'll keep watching this PR until it merges or closes.

---

PR #1220, opened 2026-09-11
perf(#1215): close the construction gap — 1.09× Pyomo like-for-like, 0.922× on memory

### LINKED ISSUE #1215: perf(modeling): model construction is ~2.1x slower and ~1.9x heavier than Pyomo — tracking issue
Tracking issue for **model construction cost** in the Python modeling layer — building a `Model`, before any solve. This is a layer nobody in this repo had benchmarked against an external baseline, and the result is not what the project has assumed.

All numbers below measured on one quiet machine (4 cores, load ≤ 1.3, 1 runnable), `main` @ `c052e85`, 5 reps each, same model in every arm: **40 distinct equation forms × 5,000 instances = 200,000 constraint instances** (bilinear / `exp` / square / `log`).

## The headline

| | build (median of 5) | µs / instance | retained RSS | B / instance |
|---|---:|---:|---:|---:|
| **oximo** 0.6.0 (Rust, arena) | **0.132 s** | 0.66 | ~41 MB \* | ~205 |
| **Pyomo** 6.10.1 | **1.474 s** (mean 1.516 ± 0.147) | 7.4 | **107.5 MB** | 538 |
| **discopt** @ `c052e85` | **3.278 s** (mean 3.211 ± 0.119) | 16.1 | **203.3 MB** | 1017 |

\* oximo's memory is peak-RSS delta from its own probe, not the retained-RSS metric used for the other two — treat as approximate.

**discopt is ~2.1× slower and ~1.9× heavier than Pyomo at construction.** Memory is deterministic (203.3 MB on every rep, sd 0.02).

This matters beyond raw speed: construction cost is paid on every solve, every benchmark run, and every B&B rebuild, and it is entirely upstream of the solver.

## What is actually slow — mechanism, not speculation

Attribution over 20,000 instances of one form (min of 3 reps):

| stage | wall | µs / instance |
|---|---|---:|
| model setup (sets + vars) | 0.005 s | — |
| expression tree only (`x*y`) | 0.254 s | 12.70 |
| + `Constraint` wrapper (`<= 4.0`) | 0.329 s | 16.43 |
| full `m.constraint(...)` | 0.375 s | 18.75 |

**89 % of construction is inside expression-node creation**; indexed-family bookkeeping is 11 %. So the cost is per-node, and there are four identified contributors:

**1. A node is allocated for every variable reference.** Pyomo pre-creates leaf objects at declaration, so `m.x[i]` is a dict lookup returning an existing object. discopt constructs a fresh `IndexExpression` each time.

```
pyomo   x[0] is x[0] : True   -> VarData
discopt x[0] is x[0] : False  -> IndexExpression
```

Measured cost of the indexing operation alone: **2.73 µs (discopt) vs 0.37 µs (Pyomo) — 7.4×.** This is the single largest item.

**2. 2.5× the expression nodes per constraint.** For `x[i]*y[i] <= c`:

```
discopt: IndexExpression 2.00, BinaryOp 2.00, Constant 1.00, Constraint 1.00  = 6.00
pyomo:   ProductExpression 1.00, InequalityExpression 1.00, (+2 tuples)       = 4.00
```

Five *real* nodes against two. Two extras are the `IndexExpression`s above; the other two come from `<=` normalization — discopt folds the bound into the body (`body - rhs`, `rhs == 0.0`), adding a `BinaryOp` **and** a `Constant`, where Pyomo keeps the bound as a plain float on the relational node:

```
discopt <= body : (x[0] - 4)  BinaryOp     rhs normalized to 0.0
pyomo   <= args : ['VarData', 'float']     bound stays a float
```

The normalized-body invariant is deliberate and load-bearing (26 modules read `.body` and never `.rhs`), so this is a real trade, not an oversight — but it costs two nodes per constraint at construction.

**3. Every scalar literal becomes a numpy 0-d array.** Retained allocation by site, 40,000-instance model, **944 B per constraint instance**:

| B/inst | site | what |
|---:|---|---|
| **182** | `core.py:841` | `Constant.value = np.asarray(value, dtype=np.float64)` |
| 160 | `core.py:747` | `IndexExpression(self, idx)` |
| 126 | `core.py:1286` | `Constant(x)` in `_wrap` |
| 110 | `core.py:652` | `BinaryOp("-")` from `__le__` normalization |
| 72 | `core.py:684` | `Constraint(...)` |
| 49 | `core.py:3652` | `c.name = f"{name}[{key_label(member)}]"` |

`Constant(4.0).value` is an `ndarray` at 104 B where a Python float is 24 B — **19 % of retained memory**.

**4. Per-node shape inference and unslotted nodes.** discopt runs static shape inference on every node; Pyomo does none. Pyomo's expression nodes carry `__slots__` (48 B, no `__dict__`); discopt's do not (~96 B).

Net effect on the same constraint: **12.42 µs (discopt) vs 2.08 µs (Pyomo) — 6.0×** on that form.

## Why oximo is 12× faster than Pyomo — it is *not* compilation

Worth recording because it determines which fixes are worth attempting. Holding the language constant and varying only the representation, same 200,000 constraints:

| | µs / instance | |
|---|---:|---|
| Python, one object per node | 2.18 | |
| Python, flat arena (parallel lists, integer handles) | **0.25** | 8.7× |
| Python, flat arena **+ names + collision check** | **0.72** | 3.1×, like-for-like |
| oximo (Rust, flat arena, same bookkeeping) | 0.66 | |

**A flat representation in Python lands at 0.72 µs/instance against Rust's 0.66 — a language factor of 1.10×.** Essentially all of oximo's advantage is representation; compilation buys ~10 %.

Corollary: **most of discopt's construction gap is addressable without leaving Python.** (Caveat: these two arms are microbenchmarks — no sets, bounds, sparsity or queryable structure — so a production Python arena would be slower than 0.72. The ordering is robust; the exact figure is a floor.)

Second corollary, visible in the same table: once nodes are cheap, *bookkeeping* dominates — name formatting plus a collision check alone costs 0.25 → 0.72 µs. That is consistent with the 49 B/instance of constraint-name strings in the allocation table.

## Candidate improvements, roughly by value/risk

1. **Cache the `IndexExpression` per position on `IndexedVar`.** Targets the 7.4× on the hottest operation. The leaf is immutable, so caching should be sound. Plausibly the largest single win available and well-scoped. **Not attempted.**
2. **Stop wrapping scalar constants in numpy arrays.** 182 B/instance, 19 % of retained memory. Wider blast radius — consumers read `.value` expecting an ndarray — so it needs its own scoping. **Not attempted.**
3. **Make constraint names lazy.** 49 B/instance of pure debugging metadata materialized for all 200,000 rows, plus the formatting cost.
4. **Reduce redundant shape work.** `_known_shape` is called ~4× per `x[i]`, some of it recomputing the same value in `Expression.__getitem__` and again in `IndexExpression.__init__`.
5. **A flat/arena-shaped construction path** — the 8.7× lever from the table above. Much larger change; listed for completeness, not proposed here.

Explicitly **not** proposed: `__slots__` on the `Expression` hierarchy. It was implemented, fully gated, and reverted — measured at 3.8 % memory and a build-time 95 % CI of [−0.4 %, +6.8 %] at n = 20 per arm (crosses zero). See `docs/dev/performance-plan.md` §30.

## Already landed

- **#1208** — pure-integer index shapes answered arithmetically instead of via a `np.broadcast_to` probe per `x[i]`: **1.92×** on a 20k bilinear family, **1.32×** on the 200k model. Bound-neutral, bit-identical solves.
- `performance-plan.md` **§30** records the attribution, the `__slots__` falsification, and the allocation table.

## Correctness framing

Everything here is **bound-neutral** work (CLAUDE.md §5): construction changes must leave `node_count`, certified `objective` and status **exactly unchanged**. The gate used for #1208 — paired LP/MILP/NLP/MINLP solves, bit-identical on status/objective/node_count — is the pattern to reuse.

## Reproduction / method

Interleaved same-process A/B where applicable, marker assertion on the loaded module (§8), load gate + spreads reported (§9), probes counting executed comparisons (§6). Probe scripts are ad hoc and not in-tree — **a standing construction benchmark with Pyomo (and ideally oximo) as recorded baselines is itself a deliverable of this issue**, since the absence of an external baseline is why this went unmeasured for so long.

## Open question worth settling early

`Model.solve()` lowers the Python expression DAG to the Rust arena via `model_to_repr`: on this model, **5.78 s to build the DAG + 2.22 s to lower it, 203.4 MB + 44.1 MB**. The arena — the representation the solver actually consumes — is 44 MB, comparable to oximo's for the same model. Whether construction should keep producing an intermediate that is then converted, or emit arena-shaped data directly, is the framing question behind items 1–5 and should be decided before investing heavily in any of them.

jkitchin:
## Summary

Closes #1215 (model construction ~2.1× slower and ~1.9× heavier than Pyomo).

**What that issue asked for, and where this leaves it** — measured on #1215's own model (40 forms × 5 000 = 200 000 constraint instances) via the standing benchmark, arms interleaved, 5 reps, load 0.89:

| | #1215 measured (`c052e85`) | now |
|---|---:|---:|
| discopt build | 3.278 s · 16.1 µs/inst | **0.944 s · 4.72 µs/inst** (sd 0.0328 s) |
| discopt retained | 203.3 MB · 1017 B/inst | **86.9 MB · 435 B/inst** |
| Pyomo | 1.474 s · 107.5 MB | 1.769 s · 94.3 MB (sd 0.0218 s) |
| **vs Pyomo, time, like-for-like** | **2.1×** | **1.09×** |
| **vs Pyomo, retained** | **1.9×** | **0.922×** |

**Memory is closed and then some** — "~1.9× heavier" → **0.922×**, discopt is now leaner than Pyomo. **Time went 2.1× → 1.09× like-for-like** (0.534× as written; quote the like-for-like number, because the as-written one credits discopt for tuning GC inside `Model.constraint` where Pyomo leaves that to the caller). The residual 0.080 s is 2.3 pooled sd — small, but real, and named in [What is NOT done](#what-is-not-done).

Every candidate #1215 named is now closed or has a measurement saying why it should not be taken; see `docs/dev/performance-plan.md` §63.

> **Numbers corrected after review.** An earlier body quoted 10.1–12.1× vs Pyomo and oximo parity, both measured at 100 000 rows — past the MINLPLib p99. Re-measured at corpus row-count percentiles, interleaved, the Pyomo margin is ~2× smaller and **the oximo claim reverses**. See [Performance](#performance-measured-at-corpus-row-counts) and §62.

The bulk of the gain here is not from making the per-element path fast — it is from
establishing that **the action is the idiom, not the tool**, and then making
discopt's array-valued idiom reach all the way through to a written file.

A cross-tool panel (discopt / Pyomo / oximo) showed that oximo is **not
vectorised**: it is per-element with a very cheap element (linear fusion in its
operator overloads, 16-byte `Copy` handles, a flat `Vec<ExprNode>`). discopt
narrows the gap by the opposite route — no per-element cost at all.

What landed, in rough order:

- **Modeling layer** — canonical `x[i]` handles, interned scalar literals, inlined
  shape inference, GC scoped during bulk construction, per-row call count cut,
  and finally #1215's **candidates 1, 3 and 4** (below).
- **`.nl` writer moved to Rust** (`crates/discopt-core/src/nl_writer.rs`), plus a
  Rust array fan-out (`discopt_core::expand`) so array-valued bodies expand
  without a Python round trip. The writer now also serves **builder-resident
  rows**, which it previously refused outright.
- **An array-valued objective can be exported at all.** On the merge base,
  `m.minimize(dm.sum(x))` — the documented vectorised idiom — raises `ValueError:
  Cannot write array variable x without indexing`. For that idiom end to end,
  this is less "faster" than "newly possible".
- **Tape lowering** reads the Rust arena instead of walking the Python DAG.
- **NN emitters vectorised**: `full_space`, `reduced_space`, `relu_bigm`.
- **The standing construction benchmark**, which #1215 names as a deliverable in
  its own right: `bench_model_construction.py` (Pyomo baseline) plus
  `issue1215_cross_tool_panel.py` (oximo arm), the latter now taking `--sizes` so
  a panel can sample the corpus row-count distribution as §62 requires.
- **Eight defects fixed**, one a certified false optimum, one a soundness guard
  that silently switched itself off, one a writer panic (see Correctness).

## The last three #1215 candidates

The entry experiment (§63) re-ran #1215's own attribution and allocation probes
before anything was built, and **the bottleneck had moved**: expression-node
creation is **37%** of the build, not the 89% #1215 measured, and the issue's top
allocation site (`Constant.value = np.asarray(...)`, 182 B/instance) is gone from
the table entirely — the linear-fusion work removed the per-row `Constant` from
linear bodies. Profiling put 68% of one family's build, and 25% of the whole
headline model, in the *indexing chain* instead.

- **Candidate 1 was only half-done.** The per-element cache had been put on the
  flat `Variable` (keyed by position), so `x[i]` was canonical — but
  `IndexedVar[key]` still reached it through `Set.ordinal(key)` and then through
  `Variable.__getitem__`'s own type test and dict read, on every access. A set's
  members are fixed at construction, so `key -> node` is a fixed mapping;
  `IndexedVar._key_cache` memoises it.
- **Candidate 4.** A plain `int` into a 1-D variable was re-deriving an answer
  fixed by the shape — `_known_shape` twice (a tuple copy each),
  `_pure_integer_index`, `_index_result_shape`, `_integer_index_out_of_range`.
  `Variable.__getitem__` now does the bounds test itself and hands the result
  shape to `IndexExpression` through `shape_hint=`. One family: 9.175 → 5.657
  µs/instance.
- **Candidate 3.** Formatting one `"family[label]"` string per row at build time,
  for metadata the solve path never reads. Timed alone: **+0.497 µs/instance**
  (median of 15 reps, warm-up discarded, `gc.collect()` per build; +0.510 on the
  minima). `Model.constraint` now records the family string and the member —
  both already alive, so the store allocates nothing — and `Constraint.name`
  formats on first read. The site is gone from the allocation table.

Worth recording: the headline barely moved for candidate 4 while a single family
measured 1.62×, because the headline model shares **one** index set across all 40
forms, so 39 of 40 index operations already hit the flat cache. That
disagreement is structural, not measurement error.

## Performance, measured at corpus row counts

MINLPLib row counts are p25=9, **p50=111**, p75=945, p90=4009, p99=68072. The
in-repo corpus is smaller still — p50=**6**, max 718. Measured with arms
interleaved, 728 points plus a 336-point confirmation run:

| claim | at corpus sizes | |
|---|---|---|
| **vs discopt before this branch**, vectorised | **median 4.75×** (2.02× at p50 → 7.80× at p99) | the win this PR actually delivers |
| vs discopt before, per-element | median 1.21× | compatibility idiom, modestly faster |
| vs Pyomo, vectorised | **median 5.89×** | true; the original "10.1–12.1×" was a 100 000-row figure |
| vs Pyomo, per-element | median 1.12×, **Pyomo wins 3 of 16 cells** | idiom-dependent — real `.nl` instances arrive one row at a time |
| **vs oximo, vectorised** | **oximo faster in all 16 cells**, median 1.55×, worst 3.16× at p50 | **the original parity claim is retracted** |
| retained memory | 0.0000–0.025× oximo | true, and understated |

**The oximo reversal is the honest headline.** Parity (0.90–1.15×) holds only as a
large-n asymptote: the gap closes monotonically with size (≈3.0× at p50 → ≈1.4× at
p99), and even at p99 = 68 072 rows the ratio is 1.23–1.59×. discopt's vectorised
path is **not** at oximo parity at the sizes that occur.

Two honest caveats on the memory row: it is real (the vectorised arm is flat in
`n` — 0–1.1 MB at both n=111 and n=68 072) but **it is not this branch's win**.
Construction memory was already O(1) on the merge base. What this branch adds is
that such a model can now be written.

The original panel's *method* was the problem, not just its numbers — sizes past
the corpus p99, arm-major rather than interleaved, no spread reported, and a
memory half that only runs on Linux. §62 records that and the binding rule that
follows from it; `--sizes` on the panel is how that rule is now honoured in-tree.
The default is left where it is, with the §62 citation on it, because both arms
hardcode the sizes and the combiner joins on `rows` — moving the Python side
alone would make every recorded oximo TSV uncombinable.

## Correctness

All 66 of `python/tests/data/minlplib_nl/`: **66 byte-identical** between the Rust
and Python `.nl` writers, 0 declined, **66/66 semantic round-trip**. The 16 with a
reference optimum in `known_optima.toml` solve with **`incorrect_count = 0`**,
uncertified = 0, all 16 `optimal` and matching.

Review independently extended this to the **full MINLPLib snapshot: 1610
instances attempted, 1570/1570 byte-identical, 0 mismatches**, 1601 semantic
round-trips, plus **24/24** on the builder-resident path the corpus sweep
structurally cannot reach, **GAMS 70/70**, and **LP 21/21 / MPS 21/21 / GAMS
46/46** on real linear-quadratic instances. It also ran a 70-instance interleaved
certificate panel: **0 false certificates, 27/27 node counts equal, 28/28
objectives bit-equal, 0 `gap_certified` regressions**.

Three correctness defects:

- **C-43** (`7632fb6`), a pre-existing **certified false optimum**:
  `_extract_disjunct_bounds` keyed hull-GDP bounds by variable *name* but reached
  through an `IndexExpression` to its `.base`, so a bound on `x[0]` capped all of
  `x`. Returned `optimal` obj 2.0 against a true optimum of 20.0. An existing
  unit test was asserting the defective behaviour and has been corrected.
- **C-45** (`63cde51`), introduced by this branch: the arena-tape lowering named a
  `pounce.NlExpr.log2()` method that does not exist. On the solve path the
  `AttributeError` was caught upstream and downgraded to `RuntimeWarning:
  incumbent-verification snapshot failed …; the false-primal guard is disabled
  for this solve` — so every `log2` model solved with that guard silently off.
- **C-46** (`3be6432`), introduced by this branch and **found by review**: the
  same class of bug in a *different* table. `expand.rs::func_code` admits
  `MathFunc::Log2`; `nl_writer.rs::nl_func_opcode` had no arm for it; the emit
  site's `.expect("mapped function")` **panicked**, so `to_nl()` raised
  `PanicException` on any `log2` model. Worse, `PanicException` derives from
  `BaseException`, not `Exception`, so `_rust_nl_text`'s fallback could not catch
  it — a writer designed to degrade to a slow path crashed instead.

C-46 is fixed at three levels: `Log2` rewritten as `log(x)/ln 2` exactly as
`export/nl.py` writes it; `write_expr` made **fallible**, so any future unmapped
code is a refusal the Python writer picks up rather than a panic; and
`_rust_nl_text` catching `BaseException`, logging at WARNING and falling back.

Three `.nl` changes landed on `main` while this was open — `nlvo` (#1224), the
`x` section (#1224/#1225), the reader carrying it (#1226) — and **all three
diverged in the Rust writer** until ported. Each is now byte-identical, and the
differential test asserts the Rust writer actually ran rather than silently
comparing the Python writer against itself.

One regression was introduced and caught while verifying the new work:
`IndexedVar[["a", 1]]` — a *list* key, which `_normalize_member` coerces to a
tuple, and which indexed fine before the key cache — began raising `TypeError:
unhashable type: 'list'` from the cache lookup. An unhashable key now takes the
uncached path; `Set.ordinal` still raises for a list that is not a member.

No validation, fallback, or safety guard was weakened. The one guard removed — the
Rust writer's refusal to handle builder rows — was replaced by an implementation,
not by silence, and byte-compared against the Python writer.

- [x] Cannot produce a false certificate. C-43 was pre-existing and is fixed here;
      C-45 and C-46 were introduced by this branch and are fixed here

- [x] `pytest -m smoke` → **1397 passed, 14 skipped** (524 s)
- [x] `cargo test -p discopt-core` → **742 passed, 0 failed**; `clippy -D warnings` clean
- [x] `ruff check` / `ruff format --check` → clean, 1011 files
- [x] `mypy python/discopt/` (pinned 2.1.0, the version CI's pre-commit hook uses)
      → **Success: no issues found in 341 source files**
- [x] `.nl` byte-parity, 66 in-repo instances → **66/66 identical, 0 declined,
      66/66 round-trip**, with the `x` section live
- [x] **§5 bound-neutrality A/B for the construction changes** → all 16 in-repo
      instances with a reference optimum, solved on this tree and on a `fdca93a`
      worktree, `(status, objective, bound, node_count)` compared as exact
      strings: **64 comparisons, 0 drift**. Each arm asserts the three markers
      unique to the change are present (new) / absent (old) before solving, so a
      run against the wrong tree cannot read as a pass (§8)

## Regression test

Every fix carries a test that fails before and passes after.

| test file | cases | covers |
|---|---:|---|
| `test_1215_hull_element_bound_leak.py` | 10 | C-43, the false optimum |
| `test_1215_arena_tape_func_table.py` | 5 | C-45, the missing `log2` method |
| `test_1220_nl_writer_func_coverage.py` | 96 | **C-46**, every math function × {scalar, array} |
| `test_1224_rust_writer_nlvo_parity.py` | 4 | the `nlvo` prefix bound in both writers |
| `test_1215_index_fast_path_equivalence.py` | 18 | **candidates 1 + 4** — differential grid against the bypassed `Expression.__getitem__` |
| `test_1215_lazy_constraint_name.py` | 9 | **candidate 3** — every name vs the eager formatter, verbatim |
| `test_1215_deep_sum_export.py` | 39 | `RecursionError` on `sum()` over 1000+ vars |
| `test_1215_gams_row_numbering.py` | 5 | GAMS numbered family rows from 1 |
| `test_1215_nn_vectorised_emitter.py` | 39 | `full_space` byte-equivalence |
| `test_1215_reduced_space_vectorised.py` | 28 | `reduced_space` byte-equivalence |
| `test_1215_relu_bigm_vectorised.py` | 15 | `relu_bigm` byte-equivalence |
| `test_1215_gdp_idiom_preservation.py` | 36 | GDP emitter idiom |
| `test_1215_documented_vectorised_idiom.py` | 22 | the documented idiom stays valid |
| `test_1215_cross_tool_panel_combiner.py` | 15 | the panel's own combiner |
| `test_1215_construction_bench_arms.py` | 5 | the benchmark arms |
| `test_nl_writer_rust.py` | 91 | Rust/Python parity over all 66 corpus instances |

Both new grids count their executed comparisons and fail at zero (CLAUDE.md §6).
Checked for non-vacuity by breaking the fast paths deliberately: a wrong axis in
the `IndexError`, a wrong shape hint, and a removed list-key guard fail 5 of them.

## Performance change

Not bound-changing — no flag. The `.nl` writer change is **bound-neutral by
construction**: byte-compared against the previous writer on all 66 corpus
instances (1570/1570 in review's full sweep), so `node_count` and objective
cannot drift — independently confirmed by review's 70-instance interleaved
certificate panel (27/27 node counts equal). The construction changes are
bound-neutral by the paired-solve A/B above (64 comparisons, 0 drift).

## What is NOT done

**On #1215 itself** — all five candidates are now closed or have a measurement
saying why they should not be taken (§63):

1. **The per-element path is 1.09× Pyomo like-for-like, not below it.** Down from
   2.1×; memory is at 0.922×. The residual 0.080 s is 2.3 pooled sd — small, but
   real. The vectorised idiom is far faster than either.
2. **Candidate 2 (stop wrapping scalar constants in numpy arrays) is deliberately
   left.** #1215 measured it at 182 B/instance, 19% of retained memory. It is no
   longer a top allocation site — the linear-fusion work removed the per-row
   `Constant` from linear bodies — and memory is now *below* Pyomo's.
   `Constant.value` is still an `ndarray`, and the issue's own caveat stands
   (consumers read `.value` expecting one), so there is no measurement justifying
   the blast radius.
3. **Candidate 5 (a flat/arena construction path) stays out of scope**, as #1215
   itself frames it. The vectorised idiom answers the same question from the
   other side at 0.00465 µs/instance.
4. **The top allocation site is now the `<=` normalization** —
   `Constraint(BinaryOp("-", self, rhs), sense="<=", rhs=0.0)`, 216 B/instance.
   #1215 already diagnosed this as "a real trade, not an oversight": 26 modules
   read `.body` and never `.rhs`. Changing it is a normalized-body contract
   change, not a construction optimisation.

**Elsewhere (unrelated to #1215's scope):**

5. **`tree_ensemble` is not vectorised** (§60) — blocked on a *structural*
   constraint: split rows alternate `≤`/`≥` and a `Constraint` carries one sense,
   so separating them **reorders rows**, and row order is how a `.sol` maps duals
   back to constraints. 99.4% of build cost is in those rows.
6. **The LP, MPS and GAMS writers are still Python.** The `.nl` port justified
   itself with a measurement; these have not been measured. (Review found the
   deep-`sum()` `RecursionError` fix reaches them too.)
7. **`MatMulExpression` relaxes more weakly than the equivalent expanded sum.**
   `W.T @ prev_z` produced a **byte-identical** `.nl` but `feasible`/121 nodes
   where the expanded form gave `optimal`/1 node. The emitters here use
   `dm.sum(W.T * x, axis=1)` to avoid it, so nothing in this PR is affected.
8. **Shape-`(1,)` handling has not been audited.** Five instances of one bug
   family turned up during this work; two are fixed here.
9. **`MathFunc` indices past `Softplus` are not lowered by the arena tape**
   (`Erf`, `Sign`, the inverse hyperbolics) — correct fallback, but leaves speed
   on the table. The `.nl` writer now covers all 18 `.nl`-expressible functions.
10. **The per-element path stays ~9× off oximo at large n.** Deliberate; §53
    records the falsification that linear fusion cannot rescue it.
11. **The cross-tool panel's default sizes are still 1 000 / 10 000 / 100 000.**
    `--sizes` exists and the §62 citation sits on the default, but the recorded
    oximo TSVs were measured at the old sizes; re-running the panel at corpus
    percentiles needs the Rust arm re-run to match.

**Settled since the original body:** the broad-corpus sweep (closed by review at
1570/1570) and the ASL row-order question — row order is fine. The `.nl` header
defects are pre-existing and were fixed on main in #1224/#1226, whose changes
this branch has merged and carried into the Rust writer.

### Retractions recorded in this PR

Per CLAUDE.md §11, in `docs/dev/performance-plan.md`:

- **§62 — the headline multipliers.** 10.1–12.1× vs Pyomo is really ~5.89× at
  corpus sizes, and oximo parity **reverses** to oximo being faster in all 16
  cells. The panel that produced them was arm-major, unspread, and sized past the
  corpus p99.
- §51 — the GDP emitter does **not** loop over rows.
- §52 — a quoted 22.83 µs/row was never replicated; replicated median is 18.72
  (sd 1.12).
- §57 — §56's "160/160 byte-identical" compared the vectorised emitter **against
  itself**: `git stash push` on an already-committed file stashed nothing.
- §63 — #1215's own attribution no longer holds on the current tree:
  expression-node creation is 37% of the build, not 89%, and its top allocation
  site is gone. The candidates were re-scoped against the re-measurement, not
  against the issue text.

https://claude.ai/code/session_01YG8b2752hhac9a37LsgcNY


jkitchin:
# Review of #1220

Reviewed at head `cdb2de0f` against merge-base `c052e85d`, in two worktrees built
side by side. Every arm is gated on both `discopt.__file__` and a PR-unique
marker (`hasattr(model_to_repr(...), "write_nl")`), asserted `True` on the PR arm
and `False` on the base arm in every run below.

**Verdict: one blocker, otherwise this holds up well — and it holds up on
evidence the PR itself could not gather.** The central claim ("the action is the
idiom, not the tool") is correct and the writer port is remarkably clean: I could
not find a single byte-difference it introduces on real instances. Two of the
three performance premises need their numbers restated, and one is wrong in
direction.

---

## 1. Blocker: `log2()` makes `to_nl()` panic

```python
m = discopt.Model("p"); x = m.continuous("x", lb=0.5, ub=10.0)
m.subject_to(dm.log2(x) <= 3.0, name="c"); m.minimize(x)
to_nl(m)
```

| arm | result |
|---|---|
| base `c052e85d` | writes 420 bytes, `o3 o43 … n0.6931471805599453` |
| PR `cdb2de0f` | `pyo3_runtime.PanicException: mapped function` at `nl_writer.rs:301:61` |

Root cause is table drift between two independently-maintained opcode maps:
`expand.rs::func_code` maps `Log2 → 2`, and `nl_writer.rs::nl_func_opcode` has no
arm for `2`, so the call site's `.expect("mapped function")` fires.

Two things make this worse than a missing case:

- **The fallback cannot catch it.** `_rust_nl_text` ends in `except Exception:
  return None`, but `PanicException` derives from `BaseException`
  (`isinstance(e, Exception)` is `False`, verified). So the Python-writer
  fallback — which handles this correctly — is bypassed, and a Rust panic
  reaches the user.
- **The cross-reference points at the wrong file.** `expand.rs::func_code`'s doc
  comment says ``Must match `expr_bindings.rs::math_func_code` `` and never
  mentions `nl_writer.rs`, which is the table that actually drifted.

An exhaustive probe (15 functions × {scalar, array}, 30 per arm) found **exactly
2 panics on the PR (log2 scalar + array) and 0 on the base**; nothing else
regressed. The fix is the rewrite the Python writer already does:
`o3 o43 <x> n0.6931471805599453`. Reproducer:
`test_pr1220_log2_panic.py` — 3 failed on PR, 2 passed + 1 skipped on base.
No test currently covers `log2 → .nl`.

I'd also suggest catching `BaseException` in `_rust_nl_text` (re-raising after
recording), so the next opcode gap degrades to a slow path instead of a crash.

## 2. The three performance premises, measured on real problem sizes

The PR's panel uses 1 000 / 10 000 / 100 000 rows. The MINLPLib snapshot this
solver targets has row counts p25=9, **p50=111, p75=945, p90=4009, p99=68072** —
so the PR's headline size is past the 99th percentile of the real corpus. I re-ran
the PR's *own builders*, unmodified, at the corpus percentiles, with the arms
**interleaved** (§9; the PR's `_measure` is arm-major) and a standard deviation
reported. 728 executed points, then a 336-point confirmation run at lower load.

The base cannot write a vectorised model at all (§4 below), so both trees get an
identical `x[0]` objective substitution to give the vectorised idiom a base arm;
it is O(1) on both and leaves every constraint row untouched.

```
 size family       baseS/prS(vec)  base-elem/pr-elem  pyomo/pr-vec  pyomo/pr-elem  pr-vec/oximo
  p50 linear                2.31x              1.02x         3.47x          1.01x         2.71x
  p50 sep_nl                2.02x              1.01x         3.94x          1.22x         2.76x
  p50 coupled_nl            2.68x              1.11x         3.63x          1.22x         3.16x
  p50 minlp                 2.22x              1.10x         3.75x          1.29x         2.71x
  p75 linear                4.61x              1.04x         4.90x          0.75x         1.50x
  p75 sep_nl                4.50x              1.16x         6.04x          1.12x         1.50x
  p75 coupled_nl            4.86x              1.25x         5.38x          1.03x         1.82x
  p75 minlp                 3.81x              1.17x         5.28x          1.12x         1.66x
  p90 linear                5.96x              1.36x         5.74x          0.69x         1.27x
  p90 sep_nl                5.25x              1.22x         6.41x          1.08x         1.43x
  p90 coupled_nl            5.68x              1.68x         6.23x          1.04x         1.68x
  p90 minlp                 4.64x              1.20x         6.18x          1.17x         1.51x
  p99 linear                7.80x              1.40x         8.29x          0.94x         1.23x
  p99 sep_nl                7.11x              1.66x         8.61x          1.29x         1.32x
  p99 coupled_nl            7.42x              1.65x         8.35x          1.28x         1.59x
  p99 minlp                 6.06x              1.75x         8.14x          1.51x         1.48x
```

**(a) "Much faster than discopt was before" — TRUE, and worth stating explicitly,
because the PR never quantifies it.** Vectorised: **median 4.75×** (2.02× at p50
rising to 7.80× at p99). Per-element: **median 1.21×** (1.01–1.75×) — the
compatibility idiom got modestly faster, consistent with the PR's own framing.

**(b) "Faster than Pyomo" — TRUE for the vectorised idiom, but the margin is
5.89× at real sizes, not 10.1–12.1×.** The 10–12× figure is a 100 000-row number
and does not survive the trip down to corpus sizes. Per-element it is **median
1.12×, and Pyomo actually wins 3 of 16 cells** (0.69× / 0.75× / 0.94×, all in the
`linear` family). Since real MINLPLib instances arrive one row at a time, "faster
than Pyomo" is idiom-dependent and should say so.

**(c) "Competitive with oximo" — NOT supported at real sizes; this is the one
that changes direction.** The PR reports 0.90–1.15× (parity) at 100 000 rows. At
corpus sizes **oximo is faster in all 16 cells**: median **1.55×**, worst
**3.16×** at p50. The gap closes monotonically with size (3.0× at p50 → ~1.4× at
p99), so the parity claim is a large-n asymptote, not a property users will see.
Even at p99 = 68 072 rows — close to the PR's own 100 000 — I measure 1.23–1.59×,
not 0.90–1.15×.

Confirmed on a quieter machine (load 4.2 vs 5.7–10.3): p50 oximo 2.73–3.05×,
p99 1.24–1.69×, Pyomo/per-element 0.87–1.34×. Every headline is stable.

**(d) Memory, "0.02–0.03× oximo" — TRUE, and understated. But it is not this
PR's win.** The PR's table is unreproducible off Linux (both halves read
`/proc/self/status`); I ported it to `psutil`, kept the protocol exactly, and
added a total-bytes column, because B/row cannot distinguish "O(1)" from "small
per row" (one 16 KiB page over 68 072 rows prints as 0).

| n=68 072, total retained | PR vec | **base vec** | PR elem | pyomo | oximo |
|---|---:|---:|---:|---:|---:|
| linear | 0 | **0** | 90.2 MB | 83.4 MB | 76.9 MB |
| sep_nl | 1.1 MB | **0** | 70.5 MB | 63.9 MB | 44.4 MB |
| coupled_nl | 16 KB | **0** | 82.5 MB | 80.6 MB | 56.9 MB |
| minlp | 1.1 MB | **0.5 MB** | 75.6 MB | 74.3 MB | 58.2 MB |

The vectorised arm is flat in `n` (0–1.1 MB at n=111 and at n=68 072), so the
ratio is 0.0000–0.025× of oximo — better than claimed. **The base column is
identical**, though: construction memory was already O(1) before this PR. The
memory table sits under "what landed", which reads as a PR contribution. What the
PR actually adds is that such a model can now be *written*.

## 3. Correctness: the solve path is bound-neutral, tested rather than assumed

The PR calls the writer change "bound-neutral by construction". The branch also
edits `_arena_tape.py`, `_tape_nlp_evaluator.py`, `_nl_expr_compiler.py`,
`modeling/core.py` and `_relax/gdp_reformulate.py` — all on the solve path — and
turns arena-tape lowering **on by default**, so I tested it: 70 stratified real
MINLPLib instances, both arms **interleaved per instance**, 20 s limit.

- **0 false certificates** against `minlplib.solu` (on either arm).
- **28/28 objectives bit-equal** where both arms terminated definitively.
- **27/28 node counts exactly equal.** The one difference (`tls2`, PR 219 vs base
  153) is **not a PR effect**: five repeats *on the base alone* give
  219/219/153/219/153. The 219-node runs consume the full 20 s limit; the
  153-node runs finish in 13.5 s — node count there is a function of wall clock,
  so that instance cannot carry a regime-1 equality test. Excluding it: 27/27.
- **0 `gap_certified` regressions**, 0 status disagreements.

Also — the user's explicit ask — the **smoke benchmark suite** (`run_benchmarks.py
--suite smoke`) run on both arms: 9/10 solved, 9/10 proved, `incorrect 0`, SGM
0.19, total 2.3 s, and a programmatic diff of the two JSONs gives **10/10
identical on (status, objective, node_count, bound, correct)**. `ex8_5_4` is
unproved on both arms — not a regression.

## 4. Byte-parity: much stronger than claimed, with two real gaps

The PR's "main outstanding validation" was the broad-corpus sweep. I ran it:
**1610 instances attempted, 1570/1570 byte-identical Rust vs Python, 0
mismatches**, 1601 semantic round-trips, 8 too big, 1 correctly refused by both
(`fuzzy.nl`). That is the PR's "What is NOT done" item 1, closed.

I also closed a parity hole the PR's own sweep **structurally cannot reach**: its
sweep goes through `from_nl`, so it never exercises the **builder-resident** rows
— the newest and riskiest code in the port, and the path whose previous refusal
the PR removed. Hashing the PR's own panel builders across 2 idioms × 4 families ×
sizes 7/111/945: **24/24 byte-identical** between the Rust and Python writers.

The other writers, which the PR flags as unmeasured (its item 3), show no
regression either: **GAMS 70/70 byte-identical** on real nonlinear instances, and
on 50 real linear/quadratic instances **LP 21/21, MPS 21/21, GAMS 46/46
identical, 0 differences anywhere**.

Two gaps in "byte-compared against the previous writer":

- **`tan` is not byte-identical.** Inside the PR tree the Python writer emits
  `o3 o41 v0 o46 v0` (sin/cos) and Rust emits `o38 v0` (native tan). Semantically
  equivalent and arguably better, but it contradicts the claim, and no corpus
  instance uses `tan` so the sweep could not see it.
- **6 of 13 functions never reach the Rust writer at all.** `abs`, `tanh`,
  `acos`, `log1p`, `sigmoid`, `softplus` are counted "identical" only because the
  Rust writer **declines** and the Python writer does the work. Corpus-wide that
  is **32/1602 real instances (2.0%)** — Abs 27, Tanh 4, Min/Max 1, including the
  entire `gastrans*` family. Those users get zero speedup and no signal that they
  are on the old path.

## 5. Unadvertised fixes worth claiming

- **The base cannot export an array-valued objective at all.**
  `m.minimize(dm.sum(x))` — the documented vectorised idiom, and what the PR's own
  builders write — raises `ValueError: Cannot write array variable x without
  indexing` on `c052e85d`. Isolated 2×2: the base fails both array-objective
  cases and passes both element-objective ones; the PR passes all four. So for
  the vectorised idiom end-to-end, "much faster than before" is really "newly
  possible".
- **A `RecursionError` is fixed beyond `.nl`.** On 3 of 50 real instances
  (`graphpart_clique-30`, `squfl015-080`, `unitcommit_200_100_1_mod_8`) the base
  raises `RecursionError: maximum recursion depth exceeded` from the **LP, MPS and
  GAMS** writers; the PR writes all three. The PR only claims the deep-`sum()` fix
  for `.nl`.

## 6. "What is NOT done" item 4 is settled — and it is a live silent-wrong-answer defect

The PR lists the ASL question as unresolvable ("no ASL-linked solver is installed
in this container"). There are three here. Both findings below are **identical on
the base** — pre-existing, not caused by this PR — but they are real and I could
not find them recorded in `performance-plan.md`.

- **`nlvo` is under-declared.** discopt writes the raw count of
  objective-nonlinear variables where AMPL writes the ASL prefix bound. On `fuel`,
  changing **only** `nlvo` from discopt's 3 to AMPL's 6 moves Ipopt from
  7269.4545 to a bit-identical 8457.6875914168377 — and Ipopt reports "Optimal
  Solution Found." **either way**. That is a silent wrong answer for any
  third-party ASL consumer. Prevalence **32/374 instances (8.6%)**; `nlvc` always
  agrees; column *ordering* is already correct. Verified formula
  `nlvo = len(nl_cons) + len(nl_objs) - len(nl_both)` on fuel (6), batch_nc (22),
  arki0017 (2326), powerflow0009r (57), ann_fermentation_tanh (4). One-line fix at
  `python/discopt/export/nl.py:376`.
- **The `x` (initial primal guess) block is dropped** on 202/202 instances that
  had one. Removing only that block from the original `ex1221.nl` reproduces
  Ipopt's SIGSEGV exactly.

Blast radius is limited — `runner.py::_find_nl_file` feeds external solvers the
*original* corpus `.nl`, so in-repo benchmarks are unaffected; this bites users
exporting discopt models to third-party ASL solvers. Worth its own issue.

## 7. Methodology

The panel violates the repo's own §9: arm-major rather than interleaved, median
of 3 with no spread reported, sizes past the corpus p99, and a memory table that
cannot run off Linux. None of this makes the PR's conclusions wrong — my
interleaved re-run reproduces the *direction* of every claim — but the specific
multipliers in the PR body do not survive a move to real problem sizes, and the
oximo one reverses.

## What I ran

| check | result |
|---|---|
| Rust/Python `.nl` byte-parity, full MINLPLib (1610) | 1570/1570 identical, 0 mismatches |
| Builder-resident path parity (24 combos) | 24/24 identical |
| GAMS parity, 70 real nonlinear | 70/70 identical |
| LP / MPS / GAMS, 50 real linear-quadratic | 0 differences; PR fixes 3 base `RecursionError`s |
| Certificate panel, 70 real instances, interleaved | 0 false certs, 27/27 nodes, 28/28 objectives, 0 cert regressions |
| Smoke benchmark suite, both arms | 10/10 identical; `incorrect 0` |
| `pytest -m smoke` | 1441 passed, 1 skipped |
| `cargo test -p discopt-core` | 735 passed |
| `ruff check` / `ruff format --check` | clean, 1001 files |
| `test_export_cli_roundtrip.py` + `test_nl_writer_rust.py` | 156 passed, 1 xpassed |
| 6 PR-added test files absent from the PR's table | 95 passed |
| C-43 false-optimum regression test | base 5 failed/5 passed; PR 6 passed |
| Timing panel, 4 sizes × 4 families × 8 arms | 728 points + 336 confirmation |
| Memory panel, both arms + oximo | 48 + 48 + 16 points |
| Function coverage probe, 15 fns × 2 idioms | PR 2 panics, base 0 |

Blocking: §1. Everything else is a request to restate numbers or file follow-ups.

---

jkitchin:
Thank you — this found a blocker I shipped and reversed a headline I had wrong. All of it is addressed; details below.

## §1 blocker — fixed (`3be6432`), and at three levels rather than one

Reproduced first at head `2845bd5`, not taken on faith:

```
RAISED (<class 'pyo3_runtime.PanicException'>, <class 'BaseException'>, <class 'object'>) mapped function
```

Your diagnosis is exactly right, including the `BaseException` part. Fixed as:

1. **The gap.** `Log2` is rewritten at the emit site as `log(x) / ln 2` — the exact form `export/nl.py` uses. Verified byte-identical: `o3 o43 <arg> n0.6931471805599453`, 412 bytes on both writers.
2. **The class.** `write_expr` is now fallible: an unmapped function code is an `ExpandError` **refusal** the Python writer picks up, never a panic. A Rust test, `nl_func_opcode_covers_every_expanded_func`, matches exhaustively on `MathFunc` with **no wildcard arm** — a new variant fails to compile until it is classified — and asserts every code `func_code` admits is mapped or explicitly rewritten.
3. **The blast radius.** Took your suggestion on `_rust_nl_text`: it now catches `BaseException`, logs at **WARNING** ("this is a writer bug — please report it") and falls back, re-raising `KeyboardInterrupt`/`SystemExit`. A panic is now loud but survivable.

I also fixed the cross-reference you flagged: `expand.rs::func_code`'s doc comment now names `nl_writer.rs::nl_func_opcode` as a table it must stay in step with, and points at the test that enforces it.

**Your `tan` finding is fixed too, in your direction.** I agree native `o38` is arguably better output — but `sin(x)/cos(x)` is a *deliberate* rewrite in `export/nl.py`, in the same family as `log2`, `log1p`, `sigmoid` and `softplus`, and this port's whole safety argument is byte-identity with that writer. Changing what discopt writes for `tan` is a user-visible output change that deserves its own justification, not a port side effect. Rust now rewrites too, and there is a test pinning the choice so it is a decision rather than drift.

Regression test is `python/tests/test_1220_nl_writer_func_coverage.py` — your probe's shape (every math function × {scalar, array}), plus a byte-parity assertion where Rust writes. **8 cases fail before and pass after**, verified by reverting the Rust change and *rebuilding*, not by assuming. One correction to my first draft: it originally asserted `to_nl()` never raises, which wrongly flagged `erf`/`sign`/the inverse hyperbolics — those have no `.nl` opcode and the Python writer refuses them by name, which is correct and pre-existing. The test now forbids only *panics*, and requires a refusal to name the function it is refusing.

Recorded as **C-46** in `docs/dev/correctness-issues.md`. Worth noting it is the same class as **C-45**, which CI caught two hours earlier in a *different* table (`_arena_tape.py` naming a `pounce.NlExpr.log2()` method that does not exist, silently disabling the false-primal guard). Two tables, one root cause: an opcode name assumed instead of looked up.

## §2 performance — you are right, and §62 retracts it

The oximo reversal is the one that matters and I have retracted it in writing (`8101d10`, `performance-plan.md` §62). I did not just take the numbers: I checked your central premise against the in-repo corpus, and it is **worse than your figure** — p50 = **6** rows, p90 = 151, max 718. Not one in-repo instance reaches 1 000 rows, which was my panel's *smallest* size. The methodological criticism lands cleanly.

| claim | body said | §62 now says |
|---|---|---|
| vs Pyomo, vectorised | 10.1–12.1× | **median 5.89×** |
| vs Pyomo, per-element | "within noise" | median 1.12×, Pyomo wins 3/16 |
| vs oximo | 0.90–1.15× parity | **oximo faster in all 16 cells**, median 1.55× |
| memory | 0.02–0.03× | 0.0000–0.025×, understated |

The PR body is rewritten to lead with **"median 4.75× vs discopt before this branch"** — the win this PR actually delivers, which as you note it never quantified — and to state the oximo reversal plainly rather than bury it. I have also taken your point that **the memory win is not this branch's**: construction was already O(1) on the base, and what this branch adds is that such a model can be *written*. Same for the array-objective `ValueError` on the base — "much faster" really is "newly possible" for that idiom, and the body now says so.

§62 ends with a binding rule: a modelling-layer figure without its row count is not a measurement, and any future panel must sample the corpus row-count distribution, interleave, report a spread, and name its idiom.

## §4, §5, §6 — logged, with one split out

- **Broad-corpus sweep (was my "NOT done" item 1): closed by your 1570/1570.** Thank you — that was the validation I flagged as most wanted and could not run here.
- **6 of 13 functions decline to the Python writer** — now an explicit item in "What is NOT done" with your 32/1602 (2.0%) prevalence and the `gastrans*` note. Correct behaviour, but silent, and worth surfacing.
- **§6 ASL findings → #1223.** `nlvo` under-declared and the dropped `x` block are both pre-existing (identical on base) and are genuine silent-wrong-answer defects, so they get their own issue rather than widening this PR — your call, and I agree with it. I recorded them as *not independently reproduced here*, since this container has no ASL-linked solver, which is exactly why I had listed the question as unresolvable.
- **Unadvertised fixes** (§5) now claimed in the body: the base cannot export an array-valued objective at all, and the `RecursionError` fix reaches LP/MPS/GAMS, not just `.nl`.

## Gates at `8101d10`

`cargo test -p discopt-core` **737 passed, 0 failed** (+2 new) · `pytest -m smoke` **1395 passed, 14 skipped** · adversarial **19 passed** · `ruff` clean over 1003 files · `mypy` **Success, 340 files** · `.nl` byte-parity re-run after the writer change: **66/66 identical, 66/66 round-trip**.

---

jkitchin:
Correction to my previous comment: the ASL findings were **already** filed as #1222, seven minutes before I opened #1223. I should have checked for an existing issue before filing — that is the "avoid issue proliferation" rule in CLAUDE.md and I missed it.

#1223 is closed as a duplicate. #1222 is the better issue on every point: the exact fix site and expression (`nl.py:314`, `len(nl_cons) + len(nl_objs) - len(nl_both)`, where I had guessed `nl.py:376`), the replacement verified **32/32** against AMPL rather than 5, AMPL's header line 4 for four named instances, the mechanism (ASL sizes its nonlinear-column prefix as `max(nlvc, nlvo)`), and it settles the row-order half of the question — **row order is fine, the header count is not** — which mine left open.

This PR's "What is NOT done" now points at #1222.

---

jkitchin:
Merged `main` (now `aa33a26`, carrying #1224) into this branch — and the merge turned up something worth flagging, because a clean auto-merge here would have been silently wrong.

## #1224's `nlvo` fix had to be carried into the Rust writer

#1224 corrected header line 4 in `export/nl.py`. It could not know this branch adds a **second** `.nl` writer, in Rust, that `to_nl()` prefers. Merging alone would have left the fast path emitting the **old, wrong `nlvo`** — reintroducing exactly the silent wrong answer #1222/#1224 are about, for precisely the models that take that path.

Measured, by reverting only the port and rebuilding:

| | byte-identical | disagreeing |
|---|---:|---:|
| without the port | 61/66 | **5** |
| with the port | **66/66** | 0 |

The five are `casctanks`, `contvar`, `heatexch_gen2`, `heatexch_gen3`, `st_e11` — some by a wide margin (`heatexch_gen3` writes `nlvo` 510 where the raw count is 85). So 7.6% of the in-repo corpus, consistent with your 8.6% across full MINLPLib.

`nl_writer.rs` now mirrors `_NLWriter._nlvo_prefix_bound` exactly, including the conditional — I took care to port the *conditional* form and not the unconditional `nlvc + nlvo - nlvb` from the issue, since your commit message records that it breaks all 29 linear-objective instances.

## The one merge conflict

`nl.py`: main added `initial_point`, this branch added the Rust fast path. Resolved so **an explicit initial point takes the Python path** — the Rust writer emits no `x` section, so preferring it would silently drop the guess the caller asked for. That is asserted in the new tests rather than left to a comment.

## New test

`test_1224_rust_writer_nlvo_parity.py`, 4 cases; **2 fail before the port and pass after**, verified by reverting the Rust change and *rebuilding*. Written against a constructed model rather than those five instance names, so it checks the rule and not the sample — and one case asserts the prefix-bound branch actually fires, since a model where the two formulas agree would pass vacuously.

## Gates at `b523aa6`

`.nl` byte-parity **66/66 identical, 0 disagreeing, 66/66 round-trip** · `cargo test -p discopt-core` **737 passed, 0 failed** · `pytest -m smoke` **1395 passed, 14 skipped** · adversarial **19 passed** · `ruff` clean over 1009 files · `mypy` **Success, 341 files**.

Worth noting for whoever lands this: while both writers exist, a header change to one is a silent divergence in the other unless ported. The 66-instance byte-parity check catches it — it caught this — but it is not in the test suite, which is why the new file pins the specific rule.

---

jkitchin:
`47ff201` removes "NOT done" item 5 — **the Rust `.nl` writer now declines nothing the Python writer can write.**

The framing in the review (and in my own PR body) treated "6 of 13 functions never reach the Rust writer" as a property of the port. It wasn't. It was three lookup tables that stopped at code 11. Every one of the six is either a plain opcode or a short composition of opcodes the writer already emits:

```
acos     -> o53              tanh -> o37              abs -> o15
log1p    -> o43 o0 n1 <x>
sigmoid  -> o3 n1 o0 n1 o44 o16 <x>
softplus -> o43 o0 n1 o44 <x>
```

All three sequences are `_function_call_sequence`'s, byte for byte — including the bare `n1` constants, which that writer emits **literally** and not through float formatting (`py_float(1.0)` would give `n1.0`, and the two writers are diffed byte for byte).

I widened the shared `MathFunc` encoding in **both** tables that define it (`expr_bindings.rs::math_func_code` and `expand.rs::func_code`) rather than one, since that exact drift is what caused C-46. `_arena_tape.py` is deliberately not extended — it has no mapping for the new codes, falls back as before, and touching the solve path needs its own bound-neutrality verification.

## Measured

23 functions × {scalar, array} = 46 cases:

| | count |
|---|---:|
| handled by Rust **and byte-identical** | **36** |
| Rust declines | **0** |
| byte mismatch | **0** |
| no `.nl` form, both writers refuse by name | 10 |

The 10 are `erf`, `sign`, `asinh`, `acosh`, `atanh` — genuinely not expressible in `.nl`, refused loudly by both. Corpus unchanged: **66/66 byte-identical, 66/66 round-trip**.

The test no longer skips on decline; a decline is now a failure, with a dedicated case asserting the "only Python can write it" set is **empty** and the writable count is exactly 18. The Rust-side exhaustive `MathFunc` match keeps its no-wildcard arm, so a new variant still fails to compile until classified.

## On deleting one of the writers

This was prompted by asking whether we still need both. We do, for now — but the blocker list is shorter than it was:

| blocker | status |
|---|---|
| ~~6 functions only Python writes~~ | **gone, this commit** |
| `x` initial-point section | Rust doesn't emit it (#1224) |
| `dm.custom` | not representable in the arena |
| the Python writer is the **oracle** | the port's whole safety argument is byte-identity against it |

The last one is the one that doesn't go away by writing more Rust. For what it's worth, the Rust writer is worth keeping on its own merits — measured on a quiet box (load 0.03), interleaved, median of 9 over 108 runs: **6.3× / 7.1× / 9.6×** on the vectorised idiom at n = 111 / 945 / 4009, and a steady **1.8×** per-element.

Separately: the 66-instance byte-parity sweep that caught both C-46 and the `nlvo` divergence **is still not in the test suite**. While both writers exist, a change to one is a silent divergence in the other, and that sweep is the only thing that catches it. Happy to add it here or file it — say which you prefer.

---

jkitchin:
## Re-review at head `106bd8cb` — regression check only

The head moved since my earlier review (`cdb2de0f` → `106bd8cb`), so I re-ran the
panel against **current `origin/main` `d8a41078`**, which is the exact merge-base of
this branch, and scoped the question narrowly: **does merging this break anything
that works today?**

**Answer: no.** Zero regressions across every comparison below, and five things it
fixes that `main` currently fails.

Both arms were gated per CLAUDE.md §8 on `discopt.__file__`, `discopt._rust.__file__`,
and the branch-unique marker `hasattr(_rust.model_to_repr(m, m._builder), "write_nl")`
(True on this branch, False on the base) — asserted in every child process, not once
at start.

### Export parity

| check | result |
|---|---|
| `.nl` output, full MINLPLib snapshot | 1610 instances — **1559 byte-identical, 0 different**, 51 skipped identically on both arms, 0 asymmetric |
| LP / MPS / GAMS / NL cross-arm, 119 instances | 468 comparisons — nl 117/0, gams 117/0, lp 114/3, mps 115/2. **Regressions (main OK → PR fails): 0.** All non-matches are main failing and the PR succeeding |
| MathFunc matrix, 23 functions × {scalar, array} × {Rust, Python} | 92 writes/arm, **0 panics on either arm**, rust==python 36/36 on both, cross-arm 46/46 identical |

This also retires the item I raised last round: the `log2` Rust-writer blocker is gone,
`tan` now matches the Python writer, and the six functions that previously declined in
Rust now write byte-identically. `erf`, `sign`, `asinh`, `acosh`, `atanh` decline on
**both** arms — pre-existing and correct.

### Solve path (CLAUDE.md regime 1)

70 real MINLPLib instances, one process per (instance, arm), interleaved, 20 s limit,
120 s hard cap. 69/70 paired (`johnall` hard-timed-out symmetrically on both arms).

```
FALSE CERTIFICATES vs minlplib.solu               : 0
node_count comparisons (optimal/infeasible only)  : 27
objective comparisons                             : 27  → 27 bit-equal (rel 1e-9)
gap_certified regressions (base True → PR False)  : 0
status disagreements (any)                        : 0
```

Node counts were 25/27 exactly equal, with two apparent differences —
`clay0203hfsg` 243 vs 245 and `tls2` 165 vs 153. Both finished `optimal` in 16–19 s
against a **20 s** limit, i.e. right on the clock, where wall-clock jitter lands
directly in the node count. Re-running both with the time pressure removed (300 s
limit), interleaved, 3 reps per arm:

```
clay0203hfsg   PR 245 245 245    base 245 245 245    objectives identical
tls2           PR 331 331        base 331 331        objectives identical
```

Identical search path. The 20 s differences were the clock, not this branch, so
regime 1 is satisfied: `node_count` and certified `objective` exactly unchanged.

### Gates, both arms

| gate | this branch | main |
|---|---|---|
| `cargo test -p discopt-core` | 737 passed, 0 failed, 14 binaries | — |
| `ruff check` / `format --check` | clean, 1007 files | — |
| `pytest -m smoke` | 1441 passed, 1 skipped | — |
| adversarial suite | 19 passed | — |
| directed export/NN/GDP/arena run (20 files) | 629 passed, 13 skipped, 1 xpassed | — |
| `mypy python/discopt/` | 6 errors in 4 files | **the identical 6 in the same 4 files** |

mypy adds nothing; only the checked-file count differs (341 vs 339).

Smoke BENCHMARK suite on both arms: 9/10 solved, 9/10 proved, `incorrect_count = 0`,
SGM 0.19 — and a programmatic diff is **10/10 identical** on
(status, objective, node_count, bound, correct, gap_certified).

### Compatibility surface

- No public `def`/`class` removed, no files deleted, no `__all__` entries removed.
  One addition: `bulk_construction_gc`, opt-in, and it restores the old thresholds in
  a `finally`. GC is raised, never disabled — no `gc.disable`/`gc.freeze` anywhere in
  the diff.
- Both opt-outs still work: default, `DISCOPT_RUST_NL=0`, `DISCOPT_ARENA_TAPE=0`, and
  both together all produce identical 3712-byte output, with no WARNING output and
  without tripping `warnings.simplefilter("error")`.
- `docs/notebooks/modeling_guide.ipynb` is valid JSON (32 cells, 13 code) and every
  `dm.*` name it newly uses exists on this branch.

### What it fixes that `main` currently fails

| instance | writer | main | this branch |
|---|---|---|---|
| `graphpart_clique-30` | lp, mps | `RecursionError` | writes |
| `squfl015-080` | lp, mps | `RecursionError` | writes |
| `pooling_sppa5stp` | lp | `RecursionError` | a correct named `ValueError` refusal |
| `unitcommit_200_100_1_mod_8` | gams | `RecursionError` | 5 947 659 bytes |
| `unitcommit_200_100_1_mod_8` | lp, mps | `RecursionError` (lp at 1.7 s, mps after 268 s of work) | 3 617 703 B in 278.7 s / 6 675 268 B in 338.9 s |

### Two things worth knowing going in — neither a regression

1. **LP/MPS on a 3.9 MB instance now takes ~5 minutes.** `main` cannot write it at
   all, so nothing got slower; but the failure mode changes from an immediate
   `RecursionError` to a long wait. Worth a note in the writer docs, or a size
   warning, if anyone exports at that scale.
2. **The two ASL export defects are untouched and tracked separately in #1222** —
   `nlvo` under-declared in the `.nl` header (32/374 instances, one silent wrong
   answer), and the `x` initial-guess block never emitted (202/202 lost, one
   reproducible SIGSEGV). Both pre-existing on `main`; they are not caused by this
   branch and are out of scope for it. #1222 also settles this PR's own open item 4:
   the row order is fine, the header count is not.

### Machine

Apple M4 Pro, 14 cores. The corpus sweeps ran under heavy background load; the two
timing comparisons above (node-count determinism, LP/MPS export) were run at load
2.7 with interleaved arms and repeats, per CLAUDE.md §9.

Nothing here blocks a merge.

jkitchin:
### Correction to the comment above (CLAUDE.md §11)

Both refs moved while that panel was running, so the comment I just posted is
already out of date in two specific ways. Retracting them in writing before
anything acts on them:

1. **"The two ASL export defects are untouched and tracked separately in #1222"
   is wrong.** #1222 was fixed and merged to `main` in #1224 (`03c3dfe9`) at
   11:53 UTC today, and #1225 followed in #1226 (`692f621b`). This branch has
   already merged #1224 in `b523aa62`. So `main`'s own `.nl` output has changed
   (the `nlvo` header count and the `x` initial-point section), which means the
   byte-parity comparison in my table was against a `main` that no longer exists.

2. **The head I reviewed is stale.** My panel ran `106bd8cb` vs `origin/main`
   `d8a41078`. Current head is `47ff2018` and current `main` is `682417e7` — two
   commits on each side. `47ff2018` in particular is substantive: it extends the
   Rust `.nl` writer to `acos`/`tanh`/`abs`/`log1p`/`sigmoid`/`softplus`, which
   is exactly the set my function matrix reported as "declines on both arms,
   correct, pre-existing". That row of my table no longer describes the branch.

What is **not** affected: the gate results, the API-surface check, the opt-out
check, and the solve-path regime-1 result all concern code neither of the four
new commits touches. The node-count determinism finding (the 20 s differences
were the clock, not this branch) stands.

I am re-running the export parity and corpus sweeps against `47ff2018` vs
`682417e7` and will post the result. Until then, treat the export tables in the
previous comment as measured against superseded revisions, not as current.

jkitchin:
## Re-run against current head — **one regression found, please hold the merge**

Per the correction above, I re-measured against head `47ff2018` and `main`
`682417e7`. Result reverses my earlier "nothing blocks a merge": as of `main`
`682417e7`, **merging this branch does break something that works today.**

### 1. The branch no longer merges cleanly

`git merge origin/main` gives **4 conflicts**, all in
`crates/discopt-python/src/expr_bindings.rs` — both sides add one field to
`PyModelRepr` (this branch's `n_builder_constraints`, main's `initial_point`
from #1225), plus main's rename of `from_model_repr_with_complementarity` to
`from_parsed_nl`. Mechanical, not semantic. I resolved it locally (keep both
fields at every construction site; a `.nl`-parsed repr has no builder rows, so
the boundary is 0 there) and it compiles clean. Everything below is measured on
that resolved merge, which is what a reviewer would actually get.

### 2. The Rust `.nl` writer drops the `x` initial-point block

`main` learned to emit the `x` segment four hours ago (#1224, then #1225/#1226
carried the parsed starting point through the reader). The **Python** writer was
taught; the **Rust** writer — which is this branch's default path — was not. So
the merge silently un-fixes the defect `main` just fixed.

Corpus-wide, MINLPLib, 1610 instances (1558 writable on both arms):

| arm | vs `main` `682417e7` |
|---|---|
| merged branch, default (Rust writer) | **786 different**, 772 byte-identical |
| merged branch, `DISCOPT_RUST_NL=0` | **0 different, 1558 byte-identical** |

That pins it to the Rust writer with nothing else in the way. The defect is also
*exactly* the `x` block and nothing more — on 80 differing instances, deleting
the `x` segment from the Python output makes it byte-equal to the Rust output,
**80/80, zero with a second difference**.

Concretely, on `alan`, same tree, same model:

```
$ diff <(DISCOPT_RUST_NL=1 to_nl) <(main's to_nl)
77a78,82
> x4
> 0 0.302884615384618
> 1 0.0865384615384593
> 2 0.504807692307693
> 3 0.10576923076923

$ diff <(DISCOPT_RUST_NL=0 to_nl) <(main's to_nl)
(identical)
```

This is Defect 2 of #1222 reappearing on the default path — the one whose
reproduction was an `ex1221` SIGSEGV in an ASL reader, and which is the initial
guess every NLP solver starts from. It is silent: the file parses, the shape is
right, only the starting point is gone.

**The fix is contained**: `nl_writer.rs` needs to emit the segment the way
`export/nl.py` now does, and `PyModelRepr::write_nl` needs the `initial_point`
handed to it the way `n_builder_constraints` already is. The 1558/1558 parity of
the `DISCOPT_RUST_NL=0` arm is the target to hit, and it is an exact one.

This is the same two-implementations-drift shape as C-46 and as the `nlvo` fix
itself — worth a parity test that writes both writers and byte-compares, so the
next `.nl` feature cannot land in one writer only.

### 3. Everything else re-measured clean

| check | result |
|---|---|
| LP / MPS / GAMS, 119 instances | gams 117 same / 0 differ, lp 114/3, mps 115/2 — **all 3+2 are main failing and this branch succeeding**; 0 regressions |
| Function matrix, 23 funcs × {scalar, array} × {Rust, Python} | 92 writes/arm, **0 panics**, rust==python 36/36, cross-arm **46/46 identical** |
| `47ff2018`'s new coverage (`acos`, `tanh`, `abs`, `log1p`, `sigmoid`, `softplus`) | all six now write through the Rust writer, **byte-identical to the Python writer**, and identical to what `main` produces |
| Fixes still standing | `graphpart_clique-30` and `squfl015-080` lp+mps (`RecursionError` on main), `unitcommit_200_100_1_mod_8` lp/mps/gams |

The solve-path result from the previous comment is unaffected — none of the four
new commits touch it, and the node-count finding (the two 20 s differences were
the clock, not this branch) stands.

### Verdict

Everything this PR set out to do still holds up. The single blocker is the `x`
block in the Rust writer, introduced by `main` moving underneath the branch
rather than by anything the branch did wrong. Once the Rust writer emits it and
the corpus sweep reads 1558/1558 against `main`, I have nothing else against
merging.

jkitchin:
Merged `main` again (`682417e`, carrying #1226) in `8cde1a7`. **Third `.nl` change on main today, third silent divergence in the Rust writer** — this one the worst of the three.

## #1226 would have silently dropped the starting point

#1226 makes `from_nl` carry a file's `x` segment through to `Model._initial_point`, so the Python writer now emits an `x` section on a read/write round-trip. The Rust writer emits no `x` section. Merging alone left the fast path dropping it.

Measured before the fix: **52 of the 66 corpus instances (79%)** lost their `x` section — the two writers disagreed on four files in five, against exactly what #1226 had just made work.

Fixed by refusing in `_rust_nl_text`, i.e. **with the writer, not in `to_nl`**. A caller reaching for the Rust writer directly would otherwise get text that dropped the point silently; this is the same shape as its existing `dm.custom` refusal. `to_nl` keeps a separate check for an explicit `initial_point=` argument, which the Rust writer never sees.

Four conflicts in `expr_bindings.rs`, all additive struct-field unions (#1226's `initial_point` beside this branch's `n_builder_constraints`) — both kept in all four places.

## It also exposed a vacuous test of mine

`_assert_identical` in this branch's own differential compared `to_nl(model)` against the Python writer. But `to_nl` **falls back to the Python writer when the Rust one declines** — so for the 52 models carrying a start point it was comparing Python with Python and passing. That is `performance-plan` §57's mistake a second time, in code I wrote to avoid it.

It now calls `_rust_nl_text` directly and **asserts a decline rather than tolerating one**, and the corpus differential clears the start point so it still exercises the writers on every other section of all 66 files. Two new tests: one pins that a model with a start point routes to the Python writer *and* that `to_nl` still emits the section with the right entry count; one guards the guard, asserting a plain model still takes the Rust path — otherwise the refusal could quietly send everything to the slow writer.

Note for the record: a Python-built model has no `_initial_point` unless you set one, so the fast path is intact for the case this PR targets. The refusal costs the read/write round-trip, not model construction.

## Verification at `8cde1a7`

Raw writers **66/66 byte-identical** with the start point cleared, 0 differing, 0 declined · `to_nl` **66/66 identical** on the user-visible path · main's own `test_1225_nl_initial_point_roundtrip.py` **129 passed** · export suites **296 passed** · `cargo` **742 passed, 0 failed** · smoke **1397 passed, 14 skipped** · adversarial **19 passed** · `ruff` clean over 1010 files · `mypy` clean over 341.

## Unrelated: `test_863_sparse_qp_extraction` is flaky, and it is not this branch's

`Python fast` failed once on `b523aa6` with:

```
test_cost_does_not_scale_quadratically_with_unrelated_variables
extraction cost grew 39.3x for 4x the variables
assert 39.25766636345212 < 14.0
```

It exercises `_relax/problem_classifier.py`, which **this branch does not touch**, and `b523aa6` differed from the previous (passing) head only in `nl_writer.rs` / `nl.py` / `core.py` / tests.

I measured it on an idle box (load 0.22), 12 runs: ratio **median 4.44, min 3.63, max 22.51, sd 6.56**, with `t_small` at ~8 ms. So it **exceeds its own 14.0 threshold about one run in twelve with nothing else running** — it is a ratio of two short wall-clock measurements with no repeat, no spread and no load gate, which is the hazard `CLAUDE.md` §9 is written about.

I have not touched it: it is main's test and outside this PR. It can be made robust without weakening it — take the median of a few repeats and leave the 14.0 threshold alone, which *increases* discriminating power (a quadratic regression is ~29×). Happy to do that here or file it — your call.

---

jkitchin:
Our clocks crossed — you found this and I fixed it within about a minute of each other. Your re-run measured `47ff2018`; the fix landed in `8cde1a7` at 13:32, and your comment posted at 13:32:50. Same defect, same diagnosis, independently.

**The output regression is gone as of `8cde1a7`.** Measured on the 66 in-repo instances:

```
default (Rust preferred) vs pure-Python:  66 identical, 0 different
instances carrying an x block:            52 of 66
```

So your acceptance criterion — the corpus sweep matching the `DISCOPT_RUST_NL=0` arm exactly — is met. Worth re-running your 1558-instance sweep against `8cde1a7` to confirm at your scale.

## But I reached it a different way than you asked for, and yours is better

You asked for the Rust writer to **emit** the segment. I made it **refuse** any model carrying a starting point, so `to_nl` falls back to the Python writer. Both give byte-correct output; the difference is the fast path:

```
instances still taking the Rust fast path: 14 of 66
```

So my fix costs the fast path on **52 of 66** round-trip instances. That's the right trade for correctness-now, and it's wrong as an endpoint. I'm implementing the emit you described — `nl_writer.rs` writing the segment, with `initial_point` handed to `write_nl` the way `n_builder_constraints` already is — and will post when the sweep reads full parity *with the Rust writer actually running*.

One note on where I put the refusal, which I'd keep either way: it lives in `_rust_nl_text`, not in `to_nl`. My first attempt guarded `to_nl`, which left the raw writer still returning text that silently dropped the point — a trap for any other caller. The capability check belongs with the writer, same as its `dm.custom` refusal.

## Your parity-test suggestion — agreed, and it caught me

> worth a parity test that writes both writers and byte-compares, so the next `.nl` feature cannot land in one writer only

Strongly agreed; this is the third such divergence today (C-46, `nlvo`, now `x`). Worse, this one exposed that my *existing* differential had gone **vacuous**: `_assert_identical` compared `to_nl(model)` against the Python writer, and since `to_nl` falls back when Rust declines, those 52 models were comparing Python with Python and passing. That is `performance-plan` §57's mistake a second time, in code I wrote to avoid it. Fixed in `8cde1a7`: it calls `_rust_nl_text` directly and **asserts a decline instead of tolerating one**.

## Also in `8cde1a7`

- The 4 `expr_bindings.rs` conflicts, resolved the same way you did — both fields at every construction site, boundary 0 for a `.nl`-parsed repr.
- Gates: cargo 742/0, smoke 1397 passed / 14 skipped, adversarial 19 passed, ruff clean over 1010 files, mypy clean over 341, main's own `test_1225_nl_initial_point_roundtrip.py` 129 passed.

## One item from your earlier comment I have not actioned

`test_863_sparse_qp_extraction` failed once on `b523aa6`. It exercises `problem_classifier.py`, which this branch does not touch. Measured on an idle box (load 0.22), 12 runs: ratio **median 4.44, min 3.63, max 22.51, sd 6.56** against its **14.0** threshold, `t_small` ≈ 8 ms — so it exceeds its own threshold about **1 run in 12 with nothing else running**. A ratio of two short wall-clock measurements, no repeats, no spread, no load gate. I have left it alone as out of scope; it can be made robust without weakening it (median of repeats, threshold untouched — a quadratic regression is ~29×, so discriminating power goes *up*). Say the word and I'll do it here or file it.

---

jkitchin:
Done — `fdca93a`. **The Rust writer now emits the `x` section**, which is the fix you asked for rather than the refusal I shipped in `8cde1a7`.

Your acceptance criterion, on the 66 in-repo instances:

```
raw writers, x section live:  66 identical,  0 differing,  0 declined
to_nl vs DISCOPT_RUST_NL=0:   66 identical
instances taking the Rust path:  14 → 66
```

Worth re-running your 1558-instance sweep; the target should now be met **with the Rust writer actually running**, not by falling back.

## How it is plumbed

Exactly as you suggested — the way `n_builder_constraints` already travels. `PyModelRepr::write_nl` takes `(column, value)` entries in the repr's own column numbering, remaps them to `.nl` columns and sorts on the remapped index, as the Python writer does; the section is emitted between `b` and `k`, matching `export/nl.py::write`'s order.

Resolution of `(name, element)` keys to columns stays in **Python**, the layer that owns that convention. A key naming a variable the model does not have refuses the fast path, so the Python writer raises its own clearer error rather than this one silently dropping an entry.

## One real bug the intermediate version had, caught by your test

`_rust_nl_text` now carries the same **three-way contract** as `_NLWriter.__init__`:

| argument | behaviour |
|---|---|
| `None` | use the point attached to the model |
| `{...}` | the explicit point wins |
| `{}` | **suppressed**, even for a model that carries one |

The `{}` case was genuinely broken in my first pass: it let the Rust path emit the model's attached point *after the caller had asked to drop it*. `test_empty_dict_suppresses_the_attached_point` in your `test_1225_nl_initial_point_roundtrip.py` caught it. That is precisely the silent class — an `x` block that still parses, just not the one the caller wanted — and it is why the contract now lives in one place for both writers rather than being re-derived per call site.

## Tests updated, not worked around

- The corpus differential **no longer clears `_initial_point`**. Clearing it was my stopgap and it hid the one section the two writers had most recently disagreed on.
- New: the Rust writer emits the section on all 52 instances that carry one, with the right entry count.
- New: absent stays absent — an empty point must not become `x0`, which would tell an ASL reader "starts at 0.0" where the truth is "no starting value given".

## Gates at `fdca93a`

export + `test_1225` suites **516 passed** · `cargo` **742 passed, 0 failed** · `clippy -D warnings` clean · smoke **1397 passed, 14 skipped** · adversarial **19 passed** · `ruff` clean over 1010 files · `mypy` clean over 341.

Three `.nl` features landed on main today (`nlvo`, the `x` section, the reader carrying it) and all three diverged in the Rust writer. Your parity-test suggestion is the thing that would have caught each one at the source; it is now in `test_nl_writer_rust.py` as a byte-compare of the two writers over all 66 instances with nothing cleared, and `_assert_identical` asserts the Rust writer actually ran rather than silently comparing Python against itself.

---

jkitchin:
### Correcting the summary: "the issue's stated goal is met" overstates this

The body's opening line claimed #1215's goal is met. Re-measuring against the issue's **own** model — 40 forms × 5,000 = 200,000 constraint instances, the size #1215 and `performance-plan.md` §30 report against — shows that is true for memory and not for time. Load 0.35, interleaved A/B/A/B, 5 reps, via `bench_model_construction.py --mode headline`:

| | #1215 measured (`c052e85`) | now |
|---|---:|---:|
| discopt build | 3.278 s · 16.1 µs/inst | **1.152 s · 5.76 µs/inst** |
| discopt retained | 203.3 MB · 1017 B/inst | **93.6 MB · 468 B/inst** |
| Pyomo | 1.474 s · 107.5 MB · 538 B/inst | 1.554 s · 94.3 MB · 471 B/inst |

**Memory — closed.** "~1.9× heavier than Pyomo" → **0.992×**, parity.

**Time — improved 2.85× in absolute terms, but not closed.** The as-written ratio is 0.741× (discopt faster), but the honest one is the benchmark's own **like-for-like 1.37×**: discopt tunes GC inside `Model.constraint`, which Pyomo leaves to the caller via `PauseGC`, and the instrument warns in its own output that quoting the as-written figure "would overstate discopt by ~1.6×". So "~2.1× slower" is now "~1.37× slower", not parity.

### Against #1215's five candidates

| candidate | status |
|---|---|
| 1. Cache the `IndexExpression` per position on `IndexedVar` | **done** — `54a0688` |
| 2. Stop wrapping scalar constants in numpy arrays | **done** via interning — `4bef6b2`; memory reached parity |
| 3. Make constraint names lazy | **not done** — `core.py:3935` still formats eagerly; the issue measured 49 B/instance plus the formatting cost |
| 4. Reduce redundant shape work | **done** — `4bef6b2`, `93a8f86` |
| 5. A flat/arena-shaped construction path | the vectorised idiom lands at 0.0008 s / 0.14 MB for the same 200k rows, but it is a **different idiom**, not the per-element path the issue's headline measures |

The issue's one explicit deliverable — *"a standing construction benchmark with Pyomo (and ideally oximo) as recorded baselines is itself a deliverable of this issue"* — is delivered (`bench_model_construction.py`, plus `issue1215_cross_tool_panel.py` for the oximo arm). Its "open question worth settling early" (intermediate DAG vs arena-shaped emission) is settled with measurements in `performance-plan.md` §41–§48.

### So: `Contributes to #1215`, and now for a stated reason

I picked "Contributes to" early out of caution. It turns out to be right on the evidence, and the remainder is nameable rather than vague:

- like-for-like per-element construction is **1.37× Pyomo**, not parity;
- candidate 3 is untouched.

The PR body's summary is corrected accordingly. Nothing else in it changes — the measured figures elsewhere were already stated against the right baselines.

---

PR #1232, opened 2026-09-13
fix(lp): one standard-form marshaler for both LP entry points

jkitchin:
# One LP standard form, shared by both marshalers

## Why

Two marshaling paths reach the same Rust driver (`solve_milp_csc_py`), and they
laid the columns out **differently**:

* the `Model.solve()` extractors in `_relax/problem_classifier.py` gave a
  logical (slack) column only to *inequality* rows, so any model with an
  equality row produced a basis short of `m` columns;
* `solvers/milp_simplex.py::_marshal_std_form` split every equality into two
  opposing `<=` rows so both halves got a logical — the shape the engine wants,
  at the cost of an always-tight opposing pair (the degenerate configuration
  `_dual_start_slack_basis`'s docstring already blames for cold-primal stalls).

Three parts of the engine index the basis by row and refuse to work with a short
logical block:

| component | counter | consequence |
|---|---|---|
| `lp/gomory.rs` | `SepGomoryShortBasis` | no tableau row to read a GMI cut from |
| `lp/simplex/dual.rs` `PreparedDual::prepare` | `DualPrepRejectShape` | every child LP solved cold, never dual-warm-started |
| `bnb/milp_driver.rs` `BaseRows::build` | `SubstDropNoSlack` | cut slack substitution mis-maps after the first equality; cut dropped |

Measured on `p1183_n8` before this change: `DualPrepAccept` **0 of 4650**
prepares, `RootCutsGenerated` **0**, 821k phase-1 pivots.

A measurement on one path therefore said nothing about the other. This is the
"one correct path" prerequisite — no cut family, node-selection or inference
work can be evaluated until both entry points marshal the same problem.

## What changed

`python/discopt/_relax/std_form.py` (new) is now the single definition of the
layout. Columns are `[structural (n_struct) | one logical per row, in row
order]`; row `r`'s logical is column `n_struct + r` and is that column's only
nonzero.

| sense | row as marshaled | logical coef | logical box |
|---|---|---|---|
| `le` | `body + s = rhs` | `+1` | `[0, inf)` |
| `ge` | `body - s = rhs` | `-1` | `[0, inf)` |
| `eq` | `body + s = rhs` | `+1` | `[0, 0]` |

An equality's logical is **fixed at zero**: it adds a column but no freedom, so
the feasible set, the optimum and every valid bound are unchanged.

All five sites now derive the layout from that module: the four extractor rungs
(`_extract_constraints_algebraic`, `_lp_rows_from_repr`, `_extract_lp_data_tape`,
`_extract_lp_data_autodiff`) and `_marshal_std_form`.

`DISCOPT_LP_ROW_LOGICALS` selects the layout and is **default OFF** pending this
panel (§5). With it off, both paths keep their pre-existing behavior exactly.

### The QP/IPM path is deliberately excluded

POUNCE's interior-point QP path takes `A x = b` directly; a column fixed at zero
only adds a structurally singular direction to its KKT system. The QP extractors
pin `for_qp=True`, which forces the inequality-only block regardless of the flag.

### A soundness hazard the consolidation created, and closed

Under the row-logical layout *every* row carries a logical, so
`solver._decompose_eq_slack_form` — which classified rows by sparsity ("has a
slack column ⇒ inequality") — would have turned every equality into a one-sided
inequality and silently stopped checking the other direction. That is a weakened
MILP feasibility gate.

Fixed by classifying on the logical's **box** (`ub == 0.0` ⇒ equality) via
`std_form.logical_fixed_mask`, threaded to all 11 call sites as `col_ub`. A
regression test pins both directions, including what happens when `col_ub` is
omitted (the legacy layout, where sparsity is sufficient).

### Instrumentation: the driver's work was invisible on deferred instances

When `_solve_milp_simplex` exhausts its budget with no incumbent it returns
`None` and `solve_model` falls back to `_solve_milp_bb`. The `SolveResult` the
caller sees is then the *fallback's*, so the driver's node and LP-iteration
counts were dropped — on exactly the hard instances where a work metric matters
most. `_merge_engine_stats` folds them across under `milp_driver/*` keys
(namespaced so they can never be read as the fallback's own), mirroring the
existing `_merge_engine_bound`.

Two related fixes in the same area:

* the #698 re-entry runs the driver a second time and folded `_nodes2` into
  `nodes` while dropping `_iters2` — a two-run search reported run 2's nodes
  against run 1's iterations alone. Now accumulated in the same branch.
* `lp/iters` and `lp/driver_nodes` are surfaced on `SolveResult.solver_stats`
  for the optimal/feasible, unbounded and infeasible returns.

**Retraction (CLAUDE.md §11).** Earlier in this work I reported bell5 as using
**28,895 nodes**. That was the *fallback's* count. With the driver's counters
plumbed through, bell5 actually spends **100,021 nodes** (its node cap) and
**106,819 LP iterations** before deferring. The earlier figure is withdrawn.

## Verification

Build under test frozen for the whole panel: base `66d341df`, `solver.py`
md5 `aabf892597014114e47152f0b1ecab6d`. Every run asserts it loaded `discopt`
and `discopt._rust` from the worktree and that the marker module
`discopt._relax.std_form` is present (absent on the baseline arm).

One edit landed *after* the panel: `std_form.py`'s `logical_fixed_mask`
gained a local type annotation to satisfy `mypy --strict`'s `no-any-return`.
It changes no expression and no value; the 22 consolidation tests were re-run
against it. Every measurement below predates that annotation.

### 1. Flag OFF is byte-identical to the pre-consolidation baseline

`dump_form.py` fingerprints `(A, b, c, x_l, x_u)` of the marshaled standard form
from both entry points, run against a detached worktree at `66d341df` and
against this branch with `DISCOPT_LP_ROW_LOGICALS=0`:

```
FORMS_EMITTED 80   (40 instances x 2 entry points)
diff base_form.txt new_form.txt  ->  identical
```

### 2. The QP/IPM path is a structural no-op under the flag

`qp_unchanged.py` extracts the same instance with the flag off and on **in one
process** and compares fingerprints; instances drawn by filtering
`qplib_manifest.csv` on `usable_oracle`, smallest first (never hardcoded names).
It also asserts the tuning flag actually followed the env var in both
directions, so the comparison cannot be vacuous.

```
QP_INSTANCES_COMPARED 12   all identical
```

### 3. Flag ON vs OFF panel (17 MIPLIB instances, 60 s each, 4 arms)

Arms run `off1, on1, on2, off2` — each setting gets an early and a late slot, and
the repeats are the spread. One instance at a time behind a lock directory.
Oracle: the HiGHS reference run, which proved 16 of the 17 optimal.

**Cert-clean — 0 violations over 269 executed checks:** no incumbent below the
reference optimum; every *certified* objective at the optimum; no dual bound
above the optimum; no bound crossing its own incumbent; no certification
regression OFF→ON; and every deterministic instance reproduced exactly within
each arm (`off1 == off2`, `on1 == on2`, node and iteration counts alike).

**Net-positive**, over the 13 deterministic instances (the three wall-clock-
limited ones are excluded — a run stopped by the clock measures the machine):

| | OFF/ON geomean | ON better | worse | tied |
|---|---|---|---|---|
| nodes | **2.04x** | 6 | 1 | 6 |
| LP iterations | **4.30x** | 7 | 3 | 3 |

Largest movements: `egout` 2439 → 11 nodes; `p1183_n8` 1,067,158 → 12,244
iterations; `misc03` 257,598 → 19,394; `enigma` 4379 → 1447 nodes.

The wall-clock-limited three, on bound and incumbent (minimization):

| inst | optimum | bound OFF | bound ON | incumbent OFF | incumbent ON |
|---|---|---|---|---|---|
| bell5 | 8966406.49 | 8960997.35 | 8960997.35 | none | none |
| gesa2 | 25779856.37 | 25652140.29 | **25771832.44** | none | **26344182.32** |
| p1183_n10 | 337.00 | 265.00 | **329.00** | **876.00** | none |

**One quality regression, recorded rather than smoothed over:** on `p1183_n10`
the ON layout finds no incumbent in 60 s where OFF finds 876 (the optimum is
337). Its dual bound is much stronger under ON (329 vs 265), and it reaches the
driver's 100k node cap because nodes there are ~7x cheaper — but the primal side
is worse on that instance.

Because of that, and because §5's net-positive bar asks for the corpus rather
than a 17-instance slice, **this PR lands the flag default-OFF**. The
graduation panel is the next step, not this one.

### 4. Tests

* `pytest python/tests/test_lp_std_form_consolidation.py` — 22 passed
* `pytest -m smoke` — **1443 passed, 1 skipped** (181 s)
* `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — **19 passed** (147 s)
* `ruff check` / `ruff format --check` on all changed files — clean
* No Rust touched, so `cargo test -p discopt-core` is not applicable.

The re-entry accumulation test was confirmed to fail before the fix
(`1300.0 != 1390`) and pass after.

---

https://claude.ai/code/session_012SiuDHKe9AbD8tqsCKAhvz


jkitchin:
### CI round 1: one failure, fixed in `c72fc54e`

`Python correctness (fast subset)` failed with

```
TypeError: _counting() got an unexpected keyword argument 'for_qp'
```

`test_75_jax_free_solve_path.py::test_vector_valued_constraint_body_solves_without_jax`
replaces `_extract_lp_data_tape` with a counting stub of pinned arity, and the
new `for_qp` keyword reached it. **The production path was never broken — only
the stub**; it now forwards `**kwargs` verbatim. Verified by running the test's
driver directly: `REACHED:1 ANSWERED:1 STATUS:optimal JAXMODS:0`, i.e. the tape
rung is reached with `for_qp=True`, answers, and imports zero jax modules.

Recording the verification gap this exposed, per §6/§11: `qp_unchanged.py`'s 12
QPLIB instances all resolve via the *algebraic* rung, so my own QP probe never
exercised the tape rung's `for_qp` threading. CI's jax-free probe is what
covered it. No other test replaces one of the four extractor rungs (checked);
every other reference calls them positionally, which is unaffected because
`for_qp` has a default.

PR #1240, opened 2026-09-15
feat(serialize): native round-trippable model format (Model.save / discopt.load)

jkitchin:
## Summary

Adds `discopt.serialize` — a native format that saves a model, and optionally the fitted state that belongs to it, and loads it back as the same model. Motivated by the ordinary workflow of fitting parameters today and wanting the model back tomorrow for a different analysis.

Until now the only way to persist a model was `.nl` / `.gms` / MPS / LP export. Those are right when the consumer is another solver and wrong when the consumer is you, later: `.nl` renames every variable positionally (`x0..xn-1`), none of them carry the solution, and they refuse the non-algebraic relations outright.

**Format.** One JSON document, optionally gzipped, with gzip detected on read by magic bytes rather than by file name. Chosen over npz/pickle deliberately: no numpy format version, no Python version, readable from any language, greppable and diffable in git. Float text compresses well (4.8× on the notebook's fit), so gzip covers size without a binary container; an `{"$array": ...}` escape hatch is reserved should a large weight tensor ever need one.

Two structural decisions worth review:

- **The expression DAG is a flat node table, not a nested tree.** Array positions are the node ids. Shared subexpressions — what a collocation transcription or a surrogate evaluated at many points produces — are written once rather than re-expanded per reference (51 references → 1 node in the table, asserted in a test).
- **`_constraints` is written as one ordered, kind-tagged row list.** It is a *heterogeneous ordered* list with relations interleaved between algebraic rows. My first draft used a section per relation kind and silently reordered such models on reload; `test_interleaved_rows_keep_their_declaration_order` pins it.

Carried: variables (names, type, shape, element-wise bounds, and the `fix()` bound stack), parameters, the objective, algebraic constraints in declaration order, the fast-construction (Rust builder) linear blocks and builder objective, indicator / SOS / disjunctive relations, complementarity relations with this model's lowering marks, the decomposition annotations and the rest of `Model`'s declared state, the starting point, and optionally the `SolveResult`.

Refused loudly, never silently dropped: `CustomCall` (an arbitrary Python callable — nothing faithful to write, and writing its *name* would reload into a different model), propositional `_LogicalConstraint`, an unknown node `op`, operator or function name on read, and a future schema major. `_LogicalConstraint` is the only structural gap left.

Not tied to an issue — this came out of a direct request, not the backlog.

### Commits

- [`5c28de3`](https://github.com/jkitchin/discopt/commit/5c28de3) the format, `Model.save()` / `discopt.load()`, and the round-trip suite
- [`9f8fe6e`](https://github.com/jkitchin/discopt/commit/9f8fe6e) `docs/notebooks/model_serialization.ipynb`
- [`fb0c514`](https://github.com/jkitchin/discopt/commit/fb0c514) complementarity relations and their lowering marks
- [`4ba1989`](https://github.com/jkitchin/discopt/commit/4ba1989) merge current `main` (#1241)
- [`11f9021`](https://github.com/jkitchin/discopt/commit/11f9021) review fixes: `fixed()` scope, dropped model state, operator validation

### Three details that shaped the implementation

**Builder-resident rows.** The fast-construction API emits rows straight into the Rust builder; they never appear in `_constraints`. A writer reading only `_constraints` emits a strict subset — the X-1 hazard documented in `export/_common.py` — and would reload a model that solves to a wrong answer. They are encoded as their own CSR section and restored through the same fast API so they stay builder-resident.

**Complementarity lowering marks carry their method.** `LoweringRecord.is_exact` is derived from the method alone, and `scholtes` is a *relaxation*. A mark that kept the relation but lost the method would read as exact, and the solver would certify a relaxation as the declared model. `rows` is deliberately not encoded — it holds `Constraint` objects of the emitting model, and `LoweringRecord` documents `None` as what a rebuilding pass records; a reload is a rebuilding pass.

**The `fix()` bound stack travels with the box.** Since #1241 `lb == ub` can be a *temporary* pin, with the declared domain at the bottom of `_bound_stack`. Writing only the live box persisted a temporary pin as the permanent declared domain: a model saved inside `Model.fixed(x=1.0)` reloaded solving a different problem (18.0 → 13.0) with no warning, and the pin could not be undone. The stack is now written alongside the box, so the reloaded model is pinned exactly as deeply as the saved one and `unfix()` restores the declared domain.

Also: floats are asserted to round-trip through their repr at write time, non-finite values are tagged strings under `allow_nan=False`, and the `1e20` INF sentinel is preserved literally rather than normalised to infinity.

## Correctness

No solver math changes — this is a serialization layer. Two places touch certificate-relevant state, and both are explicitly tested:

- **Complementarity lowering marks.** A `scholtes` (relaxing) lowering reloads with `is_exact` still `False` and still reported by `relaxed_relations`, and a declared-but-never-lowered relation reloads still unlowered so the solver boundary still refuses it (`NotImplementedError: … no lowering`) rather than quietly solving a model with the condition missing.
- **Declared variable domains.** Covered by the `fix()` bound stack above — a reloaded model solves the problem that was saved, and its declared box is recoverable.

Fidelity is judged by byte-diff rather than by same-optimum: `docs/dev/performance-plan.md` §7226 records three export defects that a same-optimum round-trip test reported as passing and a byte-diff caught.

- [x] `pytest -m smoke` → **1434 passed, 21 skipped** (0 failures)
- [x] `ruff check` / `ruff format --check` clean; `mypy python/discopt/` clean (343 files)

New suite `python/tests/test_serialize_roundtrip.py` → **107 passed** (37 smoke + 70 slow). Existing complementarity/MPEC tests → **135 passed, 17 skipped**. `jupyter-book build docs/` (pinned `jupyter-book<2`, as CI does) → **0 warnings**.

Corpus evidence, over all 66 in-repo MINLPLib instances:

| Check | Result |
|---|---|
| `.nl` → native → `.nl`, byte-identical | 66/66 |
| Document byte-stable under reload (`dumps(loads(dumps(m))) == dumps(m)`) | 66/66 |
| Bound-neutral solve: `node_count` **and** certified `objective` exactly equal | nvs01, st_e13, st_miqp3, nvs04, both fast-API models, MPCC and vector-MPCC models |

Three guards so these cannot pass while measuring nothing: `test_corpus_probe_actually_fired` asserts the corpus glob is non-empty (a parametrized test over an empty glob passes silently), `test_dropping_a_builder_block_changes_the_nl` is a negative control proving the `.nl` comparison can actually detect a loss, and `test_every_model_attribute_is_accounted_for` fails when `Model` gains a field the serializer does not classify.

## Regression test

Verified rather than asserted, twice, by swapping the pre-change `serialize.py` into the tree with the loaded module asserted in both directions (the new marker absent for the baseline, present after):

- complementarity support (`fb0c514` vs `9f8fe6e`): **6 failed before, 6 passed after**
- review fixes (`11f9021` vs `fb0c514`): **8 failed before, 9 passed after** — the 9th covers the saved-outside-a-`fixed()`-scope case, which was never broken

A first attempt at the baseline used a git worktree and reported a bogus pass, because the editable install resolved `discopt` to the main tree — the failure mode CLAUDE.md §8 documents. Hence the marker assertions.

The format commit itself introduces a new module, so its tests fail before only in the trivial sense that the API did not exist.

https://claude.ai/code/session_015jrg515FoUppj3Bgsj8mKB


jkitchin:
## Review: not merging yet — merge conflicts plus two correctness issues

CI is green and the design is solid: flat node table, one ordered row list, builder blocks, and complementarity marks that keep their method. The `.nl` byte-identity corpus check with a negative control is the right test. But this can't merge as-is, and two issues break the PR's own promise to "refuse loudly, never silently drop".

### 0. Merge conflicts (blocking, mechanical)

#1241 merged first (`a837ffb8`). GitHub reports `mergeable=false`. A test merge against current `main` conflicts in three files:

- `docs/references.bib`: keep **both** sides (`Fogler2016`, `BoxLucas1959` from main; `Wilkinson2016` from this PR).
- `python/discopt/modeling/core.py` (the `sum` docstring `**kwargs` line and the `bulk_construction_gc` table) and `python/discopt/status.py`: #1241 fixed **the same four Sphinx warnings** differently. Take **main's** version of all three hunks and drop this PR's copies; the scope note's docstring changes become unnecessary. Rebuild the docs afterwards to confirm the count is still 0.

### 1. Saving inside a `fixed()` scope bakes the fix in as the declared domain (blocking)

Since #1241, `lb == ub` can be a *temporary* fix recorded on `Variable._bound_stack`. `_enc_variable` writes the live `v.lb`/`v.ub` and ignores the stack, so a model saved inside `Model.fixed(...)` / `Variable.fixed(...)` reloads with the pinned value as its **permanent declared bounds**. That is exactly the silent leak #1241 exists to prevent, now persisted to disk.

Measured on current `main` with this PR's `serialize.py` (import path asserted to be this tree and to include #1241; 6/6 probe checks executed):

```text
max 3x + 2y  s.t. x + y <= 6,  0 <= x,y <= 10

with m.fixed(x=1.0): doc = dumps(m)
reloaded declared box for x : (1.0, 1.0)   fix_depth 0
original solve              : optimal 18.0
reloaded solve              : optimal 13.0      <- different problem, no warning
back.x.unfix()              : ValueError: has no fix() to undo
back.fixed(x=5.0)           : ValueError: outside its declared bounds [1.0, 1.0]
```

Once reloaded, the fix can't be undone through the API.

**Suggested fix**, either of:
- **Refuse**: raise `SerializationError` in `dumps` when any variable has `fix_depth > 0`, naming the variables. This is the smallest change and matches the refusal policy.
- **Carry it**: write the declared box (`_bound_stack[0]`) as `lb`/`ub`, plus the stack or current pins as a separate field, and restore through `fix()`.

Either way, add a regression test that dumps inside `m.fixed(...)` and asserts the refusal, or asserts that the reloaded declared box and solve match the original.

### 2. Model state `loads` silently drops (should be refused or carried)

`loads` builds a fresh `Model` and restores only the listed sections. These `Model.__init__` attributes are never written and are silently reset on reload:

| attribute | set by (public API) | effect of dropping |
|---|---|---|
| `_decomp_stages`, `_decomp_blocks`, `_coupling_keys` | `set_stage` / `first_stage` / `second_stage` / `set_block` / `mark_coupling` | `discopt.decomposition` (`structure.py:120,226`, `advisor/analyzer.py:170`) falls back to **auto-detecting** structure, so the user's declared Benders/Lagrangian decomposition is quietly replaced |
| `_sets` | `Model.set(...)` | named index sets vanish |
| `_aux_counter` | `if_else` | the counter restarts, so adding an `if_else` after reload can reuse an auxiliary name |
| `_zero_spanning_factor_auxes` | factorable reform (read at `solver.py:11392`) | internal-model state; probably only matters if a reformulated model is saved |
| `_simplex_lowerings` | GDP simplex lowering (#1182 size accounting) | the lowering record is lost |

The decomposition annotations are the important row: user-declared intent, set through public API, dropped with no error. Note that `_coupling_keys` can hold `id(constraint)` values, which need mapping to a row index before they can be written. As with #1: carry them, or refuse when they're non-empty. Worth adding a guard test that fails when `Model.__init__` gains an attribute the serializer doesn't account for, so the next field added to `Model` can't slip through the same way.

### 3. Unknown operator / function names are accepted at load (non-blocking)

`_decode_nodes` validates the node `op` but not `binop.o`, `unop.o` or `call.f`, and the `BinaryOp`/`UnaryOp`/`FunctionCall` constructors don't validate either. A corrupt or newer document loads fine and fails later at solve time. The first symptom there is also bad:

```text
INCUMBENT-VERIFICATION SNAPSHOT FAILED: ValueError: Unknown binary operator: 'frobnicate'.
RuntimeWarning: ... the false-primal guard is disabled for this solve
ValueError: Unknown binary operator: 'frobnicate'
```

This contradicts the module docstring ("On read, an unknown `op` is a hard error"). Validating the operator and function names against their known sets in `_decode_nodes` would make the read-time refusal complete; add a test next to `test_unknown_op_is_refused_rather_than_skipped`.

### Minor

- `_enc_index` refuses array/list (fancy) indices and `None`/`Ellipsis`. That's a loud refusal, so fine, but worth a sentence in the Scope section.
- `SCHEMA_MAJOR` comparison: `"discopt.model/1.1"` would be refused as major `1.1`. Fine if minors are never planned; otherwise split on `.`.

Happy to re-review after the rebase and fixes for #1 and #2.

jkitchin:
Thanks — all five points addressed. Merge in `4ba1989`, fixes in `11f9021`.

## 0. Merge conflicts

Merged `main` (`a837ffb`). `references.bib` keeps both sides — the conflict boundary fell mid-entry and dropped `BoxLucas1959`'s closing brace, so I repaired that by hand and checked the file parses: 238 entries, braces balanced, all three keys present. Took **main's** version of all three docstring hunks and dropped this PR's copies; `status.py` went back to main's wholesale since the docstring was its only change here. Docs rebuild: **0 warnings**.

## 1. `fixed()` scope (blocking)

Reproduced exactly as you measured — 18.0 → 13.0, `fix_depth` 0, `unfix()` refused.

Took the **carry** option rather than the refusal, since refusing would block a legitimate save and the stack is cheap to write. `_enc_variable` now writes `_bound_stack` alongside the live box; `_dec_variable` restores it before registering the variable:

```text
inside scope: x box = (1.0, 1.0) fix_depth 1
reloaded:     x box = (1.0, 1.0) fix_depth 1
reloaded solve: 13.0          <- faithful: the model WAS saved pinned
back.x.unfix()  -> (0.0, 10.0), fix_depth 0
solve after unfix: 18.0       <- declared domain intact
back.x.fix(5.0) -> (5.0, 5.0) <- accepted against the true domain
```

The reloaded solve is still 13.0 by design: a model saved inside the scope *is* a pinned model. What's fixed is that the pin is now recoverable and the declared domain survives. `test_saving_inside_a_fixed_scope_keeps_the_declared_domain` asserts the whole sequence, and a second test covers save-outside-the-scope solving identically.

## 2. Dropped model state (blocking)

All seven attributes now travel in a `state` section. `_coupling_keys` needed the treatment you flagged: it holds a mix of name strings and `id(constraint)`, so each id is written as the constraint's **row index** and turned back into the rebuilt row's `id` on load — an id matching no row is refused rather than dropped.

`_MODEL_STATE` is now the explicit registry you suggested, classifying every `Model.__init__` attribute as carried-in-its-own-section, derived-on-load, or carried-in-`state`. `test_every_model_attribute_is_accounted_for` compares it against a live `Model` **in both directions**, so a new field fails the test until it's classified, and a removed one doesn't leave a stale entry.

## 3. Operator names at load (non-blocking)

Fixed — and worth flagging, because my first attempt was wrong in a way that would have been worse than the bug. I hand-wrote the accepted function set and it both invented names (`where`, `if_else`, `hypot`) and **omitted real ones** (`centropy`, `entropy`, `signpower`, `sum`). An omission there is a false refusal of a valid document.

So the set is derived from `core._ELEMENTWISE_FUNCS` plus an explicit `_EXTRA_FUNCS`, and `test_every_produced_operator_name_is_accepted` scans the package for every literal `BinaryOp` / `UnaryOp` / `FunctionCall` name and fails if any is unaccepted. (The scan then matched my own comment describing it — reworded the comment rather than narrowing the scan.)

## Minor

- Scope section now states that fancy / `None` / `Ellipsis` indices are refused.
- `SCHEMA_MAJOR` splits on `.`: `discopt.model/1.1` now reads as major 1, `2.0` is still refused. Test covers both.

## Verification

Fail-before/pass-after verified by swapping the pre-fix `serialize.py` (`fb0c514`) into the tree, with the loaded module asserted in both directions (new markers absent for the baseline, present after): **8 failed before, 9 pass after** — the 9th covers the saved-outside-a-scope case, which was never broken.

| Check | Result |
|---|---|
| `pytest -m smoke` | 1434 passed, 21 skipped, 0 failures |
| `test_serialize_roundtrip.py` | 107 passed (37 smoke + 70 slow) |
| Corpus: 66 MINLPLib instances, `.nl` byte-identical + document-stable | 66/66, unchanged |
| `ruff check` / `ruff format --check` / `mypy python/discopt/` | clean |
| `jupyter-book build docs/` | 0 warnings |

The notebook was re-executed so its printed document shows the new `state` section rather than a stale one.

CI on `11f9021` is still running as I write this — Lint + typecheck is green, nothing red so far. I'm watching it and will push if anything goes red.

One open question for you: the four Sphinx docstring fixes are now redundant with #1241 and gone from this branch, so the scope note in the PR description no longer applies — I've left the description as-is rather than rewriting history on it, but say the word if you'd like it edited.

---

Issue #1230, opened 2026-09-12
Consolidate the LP/QP Rust–Python boundary: one safe bound, one equilibration, no standard-form round trip


jkitchin:
## Context

The LP/QP/MILP/MIQP paths compute solver numerics on **both** sides of the FFI. The contract that says they should not, plus the audit below and a ratchet test that pins it, landed on `claude/bold-lovelace-9f1wzw` (commit `5c90d2b`, **not yet merged**):

- `docs/dev/lp-qp-boundary.md` — the binding contract (§0), the audit (§2), the migration order (§3)
- `python/tests/test_lp_qp_boundary.py` — ratchets each violation class; reads source, so it runs without a built `discopt._rust`
- `scratchpad/ns_diff.py` — the V1 measurement probe

This issue is the consolidation work itself. The doc/test commit changed no solver behavior.

> **Re-scoped 2026-09-18** (see the status comments). Step 1's entry experiment has now been run and the hypothesis is **confirmed**; steps 1–3 are done. The "Done when" section below was rewritten because it was stated entirely in terms of the ratchet test on `claude/bold-lovelace-9f1wzw`, which was never merged — as originally written the issue could not be closed or even measured. The V1/V2/V3/V4 audit text is kept unchanged as the historical record; where the tree has moved since, the status comments say so.

**The line (`docs/dev/lp-qp-boundary.md` §0):** Python turns a `Model` into a problem; Rust turns a problem into a certified result; nothing that enters the certificate is computed in Python. Testably: *if a function's output can change when floating-point arithmetic changes — rounding, summation order, a margin constant — it belongs in Rust.*

### V1 — the safe dual bound has five production implementations

The Neumaier–Shcherbina bound `g(y) = bᵀy + Σ_k min_{z_k∈[l,u]} (c−Aᵀy)_k z_k` **is** the certificate.

| # | location | form | precision | margin | open-bound handling |
|---|---|---|---|---|---|
| 1 | `_relax/obbt.py:60` `_ns_safe_lp_lower_bound` | `A_ub x ≤ b`, `n_eq` tail | float64 | magnitude-scaled | `rc_snap_tol` snap |
| 2 | `solvers/milp_simplex.py:365` `_safe_lp_lower_bound_std` | `Az = b` | float64 | `1e-9·(1+\|bᵀy\|+Σ\|contrib\|)` | FBBT recovery, then abstain |
[...]

Issue #1212, opened 2026-09-07
kernel: standalone `discopt-solve` binary (.nl → certificate, zero Python) + apples-to-apples SCIP comparison


jkitchin:
## Summary

`docs/dev/scip-parity-kernel-plan.md` names the end metric as **wall-clock vs SCIP/BARON per family**, and gates every phase on it. But that metric is currently measured *through the Python harness*, so discopt's number includes orchestration cost SCIP never pays. The comparison is not apples-to-apples, and no phase gate can be fully trusted until it is.

This issue proposes the missing artifact: **a standalone Rust binary that reads a `.nl` file and solves it to a certificate with no Python in the process**, plus a head-to-head harness that runs it against SCIP from the same input.

The modeling layer is deliberately not involved. `.nl` is sufficient — it is already the corpus format (66 in-repo instances, ~4,800 in the MINLPLib snapshot), SCIP reads it natively (per CLAUDE.md's 3-way procedure), and `parse_nl_file` is already pure Rust returning `ModelRepr`.

## Why this is the right artifact now

**It is a forcing function, not an optimization.** You cannot have a standalone binary while the node loop lives in the interpreter. The kernel plan's §1 F3 is explicit that the ingredients already exist in Rust and "sit idle because the loop that would drive them is Python." A binary that must run without Python converts that architectural aspiration into a build-or-fail artifact.

**E0 already passed and E1's harness exists.** E0 measured the kernel pattern at **25,928 warm re-solves/s on nvs09 (38 µs p50)** and **1,419/s on rsyn0805m** after the P1.0 basis-finalization fix, against the ~500/s gate. `e1_row_family_census.py` and `e1_template_refresh.py` are built, with several envelope families already proven exact (0 ulp). The plan is live; this issue supplies its measuring instrument.

**Precedent exists in the ecosystem.** `pounce-cli` 0.11.0 already "solves built-in TNLPs and AMPL `.nl` files" in pure Rust, so the shape of the deliverable is proven.

## Relationship to prior issues — read before re-litigating

[...]

Issue #907, opened 2026-07-31
correctness: a sub-ulp bound crossing fathoms a live node on the default path


jkitchin:
## Symptom

Two sites test box emptiness with a strict `lo > hi` (zero tolerance) while every other site in the tree gates on `is_empty_beyond(FEAS_TOL)` — and the two strict sites are precisely the two that **act** on the verdict. The B&B loop treats `in_tree_presolve`'s `infeasible` flag as a *rigorous fathom*, so a floating-point crossing far below any meaningful tolerance can discard a region that contains feasible points.

This is a false-bound mechanism on the **default path**. No flag required.

## Evidence (measured)

- `heatexch_gen3` aborts root presolve at a crossing of **8.53e-14** with no flag set.
- The in-tree kernel declares that instance's own certified root box empty by the same amount.
- Corpus probe over the 119 in-repo instances: **22,116 executed comparisons**, spurious aborts **3 → 0** after the fix, genuine crossings **0 → 0** (i.e. nothing real was being caught by the strict test).
- Affected instances: `heatexch_gen3` (no flag), `casctanks` (1.1e-16) and `util` (6.8e-13) under `DISCOPT_FBBT_SEED`.
- `util` was **invisible** to the containment-only probe that first surfaced this: it scored a net +76-bound tightening while silently doing this, so the original measurement reported 1 affected instance where there are 3.

## Reproduction

```bash
python -u discopt_benchmarks/scripts/card3e_infeasible_root_cause.py
```

Compare the strict and tolerance-gated emptiness verdicts over the corpus; the script prints executed comparison counts and the per-instance crossing magnitudes. On the unfixed tree it reports 3 spurious aborts.

## Where it is

- `crates/discopt-core/src/presolve/orchestrator.rs` — `any_empty`
- `crates/discopt-core/src/bnb/in_tree_presolve.rs`
- Contrast with the correct gating in `crates/discopt-core/src/presolve/fbbt.rs`, `fbbt_fp.rs`, `probing.rs`, which all use `is_empty_beyond(FEAS_TOL)`.

## What a fix looks like

Gate both acting sites on `is_empty_beyond(FEAS_TOL)` rather than strict `lo > hi`, and repair a sub-tolerance crossing in place instead of returning an inverted box. Two further holes surfaced while testing the first fix: the returned box could stay inverted, and the opt-in probing branch had the same defect.

[...]

PR #1258, opened 2026-09-15
feat(lp-milp): #1229 route pure LP/MILP to HiGHS with verified certificates

### LINKED ISSUE #1183: MILP performance: large branch-and-bound tree on a big-M GDP (about 28 to 46x the nodes of SCIP and HiGHS)

### LINKED ISSUE #1229: correctness: free columns come back at the 1e20 INF sentinel, so an infeasible point is reported optimal (issue-2388.lp)
## Summary

On `ref/HiGHS/check/instances/issue-2388.lp` the MILP driver returns
`status = optimal, obj = 0.0, bound = 0.0` after 1 node — and the incumbent it
hands back is **infeasible**. Eight of the 42 structural entries come back
holding the `INF` sentinel `1e20`.

The reported objective happens to equal the true optimum (HiGHS: 0.0), so
nothing downstream notices: the objective is right *because those eight columns
all have zero cost*, not because the point is feasible.

This is the `INF`-sentinel class the repo `CLAUDE.md` already warns about
("`INF` in the Rust LP layer is the sentinel `1e20`, not `f64::INFINITY`"), but
in the **solution readback** rather than in FBBT.

## Reproduction

```python
import numpy as np, mloader
from discopt import _rust
from discopt.solvers.milp_simplex import _marshal_std_form

p = "ref/HiGHS/check/instances/issue-2388.lp"
d = mloader.read_arrays(p)
A_ub, b_ub, A_eq, b_eq, bounds = mloader.to_ub_eq(d)
std = _marshal_std_form(d["c"], A_ub, b_ub, A_eq, b_eq, bounds, d["integrality"])
n = len(d["c"])
st, x, obj, bd, nodes, iters = _rust.solve_milp_csc_py(
    std.c, std.m, n + std.m, std.col_ptr, std.row_idx, std.vals,
    std.b, std.lb, std.ub, std.int_cols, n, 0.0, 1_000_000, 1e-4, 1e-9,
    root_cuts=16, cut_rounds=1, cut_select=False,
    sb_max_cands=6, sb_node_budget=48, time_limit_s=15.0)
# -> optimal, obj 0.0, nodes 1, iters 12
```

## What comes back

```
columns at the 1e20 sentinel: 8 of 42   (free columns in the model: 14)
  x5    x[ 6] = 1e+20   clo=-inf  chi=inf   std.lb=-1e+20 std.ub=1e+20
  x27   x[ 8] = 1e+20   ...
  x17   x[ 9] = 1e+20
  x37   x[11] = 1e+20
  x6    x[15] = 1e+20
  x28   x[17] = 1e+20
  x18   x[21] = 1e+20
  x38   x[23] = 1e+20
```

All eight are **free** columns (`lb = -1e20`, `ub = +1e20` in standard form),
returned nonbasic at the sentinel *upper* bound. The other six free columns
(`x31 x32 x41 x42 x43 x44`) come back at sensible near-zero values, so this is
not "the whole model is unbounded" — it is specific columns parked on the
sentinel.

## Why it slips the internal feasibility check

The sentinels cancel **pairwise and exactly** in the rows that pair them:

```
c1:  x5  + x11 - x27 = 0     ->  1e20 + 0 - 1e20  = 0   exactly
c41: -a x5 + a x37 + b x43 = 0  ->  same magnitude, cancels
```

so the residual on those rows is genuinely 0 and the point looks feasible.

The row that catches it does not pair them:

```
c43:  x17 - x37 >= 10.2      ->  1e20 - 1e20 = 0  <  10.2
```

**Violated by exactly 10.2.** HiGHS on the same arrays returns a point with
`max |A_eq x - b_eq| = 6.6e-17` and `max (A_ub x - b_ub) = 0`.

## Note on the reported violation magnitude

A verifier built on `scipy.sparse` matvec reports `viol = 173.82` on row `c41`
for this point. That number is an **artifact of the checker, not of the
solver**: scipy's CSR matvec contracts to FMA, so evaluating
`-a*1e20 + a*1e20` accumulates the *exact* second product against the
*rounded* first one and returns the first product's rounding residual instead
of 0. The dense dot of the identical row and vector gives exactly 0:

```
A_eq row 9 coefs:  x5 -0x1.dfb29f4d88341p-3,  x37 +0x1.dfb29f4d88341p-3  (exact negatives)
sparse  A_eq @ x  ->  -173.8205348
dense   row @ x   ->   0
```

The `c43` violation of 10.2 is FMA-independent and is the real one. Any
feasibility verifier in this repo that may see sentinel-valued entries should
avoid the sparse-matvec path for the same reason.

## Scope

- Found by the 46-instance MILP differential panel run for #2122 (PR #1228).
- Reproduces **identically on `origin/main`** and on the #2122 branch — it is
  pre-existing and unrelated to that fix.
- Only `issue-2388` in that 46-instance panel exhibits it, but the mechanism is
  not instance-specific: any model with free columns that the simplex leaves
  nonbasic at the sentinel bound can produce the same readback.

## What "done" looks like

A free column must never be returned at `±1e20`. Either the simplex must not
leave a free column nonbasic at a sentinel bound, or the readback must detect
it and refuse (loudly) rather than reporting `optimal` — per `CLAUDE.md` §3, a
refusal is preferable to a silently wrong certificate. A regression test should
assert that the returned incumbent for this instance satisfies every original
row within `1e-6`, checked **densely** (see the FMA note above).

jkitchin:
## Summary

Routes pure LP/MILP models to HiGHS by default, with every HiGHS certificate re-verified by discopt. The plan and falsification log are in `docs/dev/lp-milp-highs-routing-plan.md`. MINLP code paths are unchanged, apart from two defects in the shared Rust LP engine fixed below.

Closes #1229. Resolves #1183 as re-scoped in plan §2.3: pure MILP goes to HiGHS, and Rust driver search work is closed unless new evidence reopens it. The evidence is below.

#1212 (standalone `.nl` → certificate binary) and #1230 (LP/QP boundary consolidation) are not part of this PR and stay open.

### What changed

- **Verified HiGHS route** (`python/discopt/solvers/lp_milp_highs.py`, dispatch in `solver.py`), plan §3:
  - LP `optimal` is backed by a Neumaier–Shcherbina safe bound, over an FBBT box, with exact dual correction.
  - LP `infeasible` needs a verified Farkas ray, falling back to a phase-1 proof.
  - LP `unbounded` needs a verified primal ray plus a feasible point.
  - MILP needs a readback guard, an exact feasibility and integrality check, and a root NS check.
  - Each label carries provenance. HiGHS never falls back to Rust: anything unverifiable is `error` with a reason.
- **Default flipped (P4):**
  - `DISCOPT_LP_MILP_BACKEND` defaults to `highs`; `=rust` keeps the legacy route.
  - Not rerouted: MILPs with callbacks, `lagrangian_bound`, `nlp_bb=True`, `nlp_solver="simplex"`, or a model that is not exactly linear.
  - A default MINLP solve never imports highspy (test below).
- **Root presolve bypass (H5 part):**
  - discopt's root presolve is skipped for models the HiGHS route takes.
  - A/B, 3 interleaved rounds: issue-2446 went from 5.603 s to 0.581 s, and 2122 from 2.510 s to 0.675 s. Status, objective, bound and nodes were identical.
- **HiGHS global scheduler fix:**
  - The route no longer pins `threads=1`. HiGHS sizes one process-wide scheduler and refuses later runs with a different `threads`.
  - Before, every route solve after another highspy user in the process was an `error`.
  - The M and L panels are bound-neutral: 48 instances, 0 drift.
- **#1229 root cause** (Rust, `milp_driver.rs` `dual_slack_basis`): the free-column sentinel is fixed at the driver entry, with regression tests.
- **Rust route: 2122 false `infeasible`, fixed** (`lp/simplex/presolve.rs` `fbbt_row`):
  - 2122's rounded row data leave a 3.2e-9 bound crossing, and it was declared an empty box.
  - Emptiness now needs a crossing > `FEAS_TOL`. A smaller crossing is widened to its endpoints (the #907 idiom), never replaced by a midpoint.
  - On `=rust`, 2122 now ends at `time_limit` with a sound bound, −187609.32 (maximize; optimum −187612.94).
- **Shared dual engine: sentinel defect, fixed** (`lp/simplex/dual.rs` `select_leaving`):
  - A ±1e20 sentinel side was tested as a bound, so a drifted free basic column was pivoted onto 1e20.
  - On qplib2170 this was the only trigger of the unstable-pivot bail/recovery and a 17 794-pivot stall. After the fix: `optimal 0` in 613 pivots.
  - Tests built on that fixture were re-targeted without weakening. A new real-corpus fixture, `bchoco06_unstable_pivot_lp.json`, keeps the unstable-pivot recovery covered.
  - The `COST_PERTURB_EPS` doc now retracts its qplib2170 evidence.
- **Swallowed exceptions** in `mccormick_lp.py`: 14 silent `except Exception:` fallbacks now log at debug with `exc_info`. Control flow is unchanged.
- **Packaging (P3):**
  - `highspy>=1.10` is a core dependency.
  - `release.yml` smoke-solves an LP and a MILP from each wheel.
  - `ci.yml` gains a highspy floor/latest lane that also runs the cross-backend soundness files on `=rust`.
- **P7 (re-scoped):** no binding deletions (§12). The import probes now name the CSC bindings their wrappers actually call.
- **Docs (P8):**
  - `solve-routing.md`, closure/status notes, and the routing plan's §12.
  - **The CLAUDE.md Architecture note changed; owner review please.**
- Benchmark harness: `discopt_benchmarks/scripts/lp_milp_panel.py`, `lp_milp_loader.py`, and the panel manifest.

### Retraction, CI fixes, and adversarial testing (commit 964aea71)

**Retraction:** I earlier called this branch complete. That was before CI ran on the merged branch, and CI then failed 14 tests in 9 files. Plan §12 records this.

**CI fixes:**
- Tests whose subject is discopt's own LP/MILP machinery are pinned to `DISCOPT_LP_MILP_BACKEND=rust`, because the new default sends them to HiGHS. No assertion was weakened. Files: `test_933`, `test_952` ×2, `test_lagrangian_bb_hook`, `test_lp_backend_seam` ×4, `test_milp_engine_routing`.
- **#912 inventory:** the exact dual correction's carved wall-clock slices are gone. It is now capped by a deterministic work count, `EXACT_MAX_WORK`, plus the caller's own `time_limit`.
- **GP gate:** a GP that is also a pure LP now goes to the HiGHS route. The log-space NLP answered `dm.sum(y)` only to IPM accuracy (3 + 7.5e-9), which failed `test_sets_adversarial`.
- **`test_convex_fast_path` ×2:** passes locally after these changes.

**Adversarial testing:** a separate session ran 5,085 random LP/MILP comparisons and 45 hand-built edge cases.
- The oracle was exact `Fraction` vertex and integer enumeration for n ≤ 6, and scipy `milp` otherwise.
- Every wrong-answer claim was confirmed with an exactly checked witness, on both backends.

Defects found and fixed on this branch, each with a regression test that fails before and passes after:

| # | Defect | Fix |
|---|---|---|
| F1 | Declared-box tightening certified a feasible LP `infeasible` on both backends. A ±9.999e19 box contribution absorbed a row's small terms. | Widen by a floating-point error bound (`nonlinear_bound_tightening.py`) |
| F2 | HiGHS MIP returned a false `kInfeasible` with the finite default box, and the route certified it. 21 saved instances. | Pass that box to HiGHS as ±inf. This is a relaxation, so HiGHS's `kInfeasible` and tree bound stay valid. |
| F3 | A proved `infeasible` became certified `optimal` on both backends. The integer column is boxed in [−1.6, −1.07]; the #844 fallback still ran, and the `lp_spatial` verifier accepted a rounded-then-clipped non-integer point. | The fallback skips a certified infeasible or unbounded result. The verifier refuses non-integral integer columns. |
| F4 | An integer column with no integer in its box and no rows made `=rust` raise `MILP-BB returned an infeasible point`. | Tightening now proves this infeasible, within the 1e-5 integrality tolerance. It fires on 0 of the 66 in-repo `.nl` files. |
| F5 | `Model.solve` raised `RuntimeError` on HiGHS passModel `kError` (row side at the 1e20 sentinel; 124 saved instances) and `kWarning` (dropped tiny coefficients; 47). | `kError` is now an `error` result. On `kWarning` the model is re-passed at `small_matrix_value=1e-12`. A MILP still perturbed at that value is `error`; an LP continues, because every LP certificate is re-verified against the unperturbed form. |

- **F5, what I tried first:** setting 1e-12 on every model. That was falsified: `test_lp_milp_highs_route` caught netlib klein1 turning from a proved infeasible into `kUnknown`. So the option changes more than the drop threshold, and it is now applied only after a drop.
- **Checked and not defects:**
  - The oracle-artifact mismatches, measured in the re-adjudication below.
  - `lb > ub`, which is refused loudly at model build.
- **Weak but honest, left as is:**
  - `error` where HiGHS rejects a 1e20-scale right-hand side.
  - MILP `kSolveError` raised inside HiGHS (`transformNewIntegerFeasibleSolution`).
- **Adversary reproducers on the final tree:**
  - `repro_milp_false_infeasible` 4/4 correct
  - `repro_empty_int_box_false_optimal` 2/2
  - `repro_passmodel_crash` 4 runs, 0 crashes
  - `repro_fixed_eq_false_infeasible` 9/9
  - The Rust empty-integer repro: 1 of 2 wrong before the F4 fix; after it, `test_integer_box_without_an_integer_and_no_rows_is_infeasible` passes on both backends.
- **Re-adjudication of all 768 saved findings on the final tree** (`adjudicate.py`, 767 compared, default route):
  - 391 optimal-ok, 8 infeasible-ok.
  - **0 FALSE-INFEASIBLE, 0 FALSE-OPTIMAL (x infeasible), 0 BOUND-CROSSES-OPT.**
  - 1 CRASH: the loud `lb > ub` ValueError at model build.
  - 354 `weak:error` and 6 `weak:feasible`. These are no-certificate outcomes, not answers; mostly HiGHS `kUnbounded` / `kUnboundedOrInfeasible` against the default box, and sentinel-magnitude readback refusals.
  - 7 OBJ-MISMATCH against the exact oracle, each re-checked independently (`check_mismatch.py`: integer enumeration plus scipy `linprog`, discopt run on both backends). All 7 are oracle artifacts:
    - 3 LP near-parallel-row cases: the independent optimum equals discopt's on both backends; the vertex oracle is wrong.
    - 3 MILP cases: the exact oracle rejects discopt's point because a row is tight to float rounding. Exact relative row violation of discopt's point is 4.5e-17 to 1.1e-16, and enumeration equals discopt on both backends.
    - 1 edge case: an integer at declared ub 2.9999999, accepted within the 1e-5 integrality tolerance, which is the existing convention.

### claim-boundary CI failure: hda bound regression, fixed

CI on 964aea71 failed only `test_current_root_lp_matches_committed_baseline`: hda's root LP bound went from −64675.25 to −5710326.49 at an identical relaxation fingerprint, with the same value on Linux CI and macOS.

- **Cause:** the #1229 `dual_slack_basis` free-column fix. I bisected with a separate build per arm and asserted each `.so` md5. Reverting only that rule restores −64675.25. The `select_leaving` and `fbbt_row` fixes don't move it.
- **Not unsound:** I re-solved every LP the relaxer solves with HiGHS. The branch's vertex is primal feasible at −64675.25 (row violation 1.6e-10). Only the Neumaier–Shcherbina bound computed from its duals is loose. Before the fix, the simplex hit its iteration limit, the #671 failure-triggered row filter fired, and HiGHS confirms that filtered LP at −64675.2492.
- **Solve impact before the fix:** hda, 200 nodes, TL 120 s, 2 interleaved rounds. Main −64509.8; branch −1.40e10. Both `time_limit` at 3 nodes.
- **Fix:** `relax_row_filter_loose_bound`, default ON; `DISCOPT_RELAX_ROW_FILTER_LOOSE_BOUND=0` opts out. When an `optimal` node's certified bound is more than 1e-3 relative below its vertex objective, it re-solves a row-filtered copy and keeps the larger certified bound. Sound because the filtered LP is a superset of the original; inert when no row is float64-intractable.
- **Results:**
  - The root partition shows hda `unchanged` again. The committed baseline reproduces, so it is not regenerated.
  - Locally, flag ON gives 4 unreproduced (alan, contvar, nvs08, tanksize), and main shows the same 4 on this host. Flag OFF gives 5, the extra one being hda. CI's list was alan, contvar, hda, tanksize, so it should drop to 3, within the cap.
  - The hda 200-node solve bound is −64509.8, equal to main.
  - `test_row_filter_loose_bound.py` fails with the flag off (−5710326) and passes with it on.
- **Cert-clean panel:** 66 `.nl` files, arms OFF/ON/ON2, TL 20 s, max_nodes 300, oracle `minlplib.solu`. Ran 66; 156 oracle-checked arms; **0 oracle violations, 0 certification regressions**. The 3 changed rows (clay0303hfsg, nvs05, tls2) are all noisy: ON2 differs from ON.
- **Slow row-filter tests** (`test_relax_row_filter.py`, in no CI lane):
  - `test_row_filter_fires_where_the_corpus_says_it_does[bchoco07, bchoco08]` fails on main too; it depends on how much the host gets through in its 15 s budget.
  - `test_hda_optout_restores_loose_candidate_a_floor` pinned the legacy opt-out bound below −1e7. The #1229 Rust fix makes that path return −141697, which is valid (≤ opt) and tighter, so the test now asserts soundness.
- **Verification of this commit:** `pytest -m smoke` 1531 passed, 1 skipped; adversarial suite 19 passed; `ruff check` and `ruff format --check` clean. No Rust was touched.

### LP / QP / NLP category smoke benchmarks

`run_category_benchmarks.py --level smoke --time-limit 60` on this branch; `discopt` loaded from the branch tree.

| Category | Arms | Result |
|---|---|---|
| LP (5 problems × ipm, pounce, ipopt, highs) | 20 | 20/20 expected status after the harness fixes below; 0 incorrect |
| QP (5 × ipm, pounce, ipopt) | 15 | 15/15 optimal at the known optimum; 0 incorrect |
| Convex NLP (2 × 3) | 6 | 6/6; 0 incorrect |
| Nonconvex NLP (4 × 3) | 12 | 12/12; 0 incorrect |

The first LP run surfaced two harness defects, both present on main and fixed here:

- **`highs` arm always "timed out".** `category_runner.py` imported `discopt.solvers.lp_highs`, removed in #365. The `ImportError` was caught and recorded as wall = inf, shown as a 60 s timeout. It now calls `discopt.solvers.lp_milp_highs.solve_lp`.
- **`lp_unbounded` expected `unbounded` on a bounded model.** It used the default `continuous` box, which is the finite ±9.999e19; `min -x` there is `optimal` at the corner by design (pinned by #850/#937; main returns the same). It now declares `ub=np.inf` and certifies `unbounded` on the HiGHS route.

Known `error` case found while probing (sound refusal, not changed here): a MILP whose optimum sits on the default ±9.999e19 box (e.g. `min -x + z`, `x >= 0` default box, `z` integer) returns `error` on the HiGHS route; `DISCOPT_LP_MILP_BACKEND=rust` returns the corner at −9.999e19. The route opens the sentinel-magnitude box to ±inf for HiGHS MIP (adversarial fix F2), HiGHS then says `kUnboundedOrInfeasible`, and the declared-box root LP is optimal, which cannot decide the MILP.

### Route differential probe (final tree 9bd6aa64)

1000 random small LP/MILPs, seed 2026. They mix `np.inf`, one-sided, free, default-box and finite bounds; min/max; `<=`/`>=`/`==` rows; and integer columns. Arms: H is the default route, R is `DISCOPT_LP_MILP_BACKEND=rust`, and S is scipy `milp`, run in a child with a 30 s wall timeout. Only witness-backed violations count.

- The route marker fired on H for 880/1000 models and on R for 0/1000. There were 4259 comparisons.
- Raw result: 24 violations and 26 disagreements. The probe's first versions reported 167 "violations". Those came from two probe defects: points at the ±9.999e19 corner were counted as infeasible, and the scipy reference hung (model model 897; scipy's bundled HiGHS overran its own `time_limit`). Both are fixed. An earlier attribution of that hang to the discopt route was wrong and is retracted.
- **All 23 flagged models were re-checked** (`triage.py`) in fresh processes, with the loaded tree asserted per arm: H and R on this branch, and main (marker absent). The reference was re-run on discopt's real declared box: integers without bounds get the documented fallback [0, 1e6]; the probe had assumed they were unbounded.
  - **14 were probe artifacts** from that assumption. With the right box, the reference agrees with H (and R, and main) on every one: model 91, model 103, model 264, model 395, model 410, model 439, model 489, model 524, model 591, model 713, model 749, model 790, model 804, model 809.
  - The other 9 are model 841, model 81, and the 7 `error` cases below.
  - **No wrong certificate on the default H route.** H's other outcomes were a verified-better optimum (model 841: 7.6667 at a checked point; scipy's 3.0 is wrong) or `error` (model 42, model 72, model 97, model 201, model 305, model 349, model 548).
  - **Rust-route wrong certificates, all present on main, all unchanged by this PR, all fixed on the default route:**
    - model 841: `optimal −0.0` at a ±9.999e19 corner where the −x1 and −x2 terms cancel. The true optimum is 7.6667.
    - model 81: `unbounded` on an infeasible model.
    - model 42, model 201, model 305, model 349: `infeasible` where a feasible optimum exists on the default box.

    They only reach users on the `=rust` opt-out. They are outside this PR's scope and recorded here.
- **Usability cost:** 283/1000 H results are `error`, versus 39 on R. They are almost all MILPs whose continuous variables use the default ±9.999e19 box (see the known `error` case above). They are refusals, not answers.

Row-filter probe (`relax_row_filter_loose_bound`): 66 instances, 264 nodes, 10 filter events, 49 compared, **0 violations**.

### #1183 evidence

The issue's own Pyomo reproducer (`gdp.bigm`, 8 blocks) through `SolverFactory("discopt")`. Each run was a fresh subprocess, with 3 interleaved rounds per arm. Node count is the one Pyomo reports, the issue's metric.

| Arm | Nodes (all 3 rounds) | Status | Objective | `gap_certified` | Wall mean ± sd |
|---|---|---|---|---|---|
| Before: `DISCOPT_LP_MILP_BACKEND=rust` (old default) | 4973 | optimal | 224 | True | 2.673 ± 0.060 s |
| After: default (env unset) | 7 | optimal | 224 | True | 0.463 ± 0.010 s |

- The before arm reproduces the issue's 4973 exactly.
- The after arm's route is `highs-milp: verified HiGHS route (HiGHS kOptimal)`.
- The issue reports HiGHS 1.14 at 109 nodes and SCIP at 179, both at 0.6 s. The default now matches or beats both.
- Load was 6.3 → 6.8, above the plan's timing gate, so the walls are indicative. Nodes, status, objective and certification do not depend on load.
- **Not addressed here:** the issue's closing remark that the same search weakness would affect OA masters and spatial B&B. Stage 2 (HiGHS as the OA/GDP master) was not built. That is optional under plan §2.3.

### Before / after: the old default against the new one

Final run on 9bd6aa64, both panels, arms interleaved per round, 3 rounds, TL 20 s, gap 1e-4, highspy 1.15.1, each solve in a fresh subprocess. R is the old default (`DISCOPT_LP_MILP_BACKEND=rust`); H is the new one. 144 executed per panel, **`incorrect_count` 0 on both**.

| Panel | Arm | Total wall | Median | Max sd | Nodes | Certified |
|---|---|---|---|---|---|---|
| M (24 MILPs) | **Before (R)** | 104.05 s | 0.202 s | 0.150 s | 610 256 | 19/24 |
| M (24 MILPs) | **After (H)** | **5.35 s** | 0.140 s | 0.011 s | **217** | **24/24** |
| L (24 LPs) | **Before (R)** | 116.76 s | 0.202 s | 0.214 s | — | 21/24 |
| L (24 LPs) | **After (H)** | **3.16 s** | 0.037 s | 0.006 s | — | **24/24** |

- **Total wall: 19.4× faster on the MILP panel, 36.9× on the LP panel.** Per-instance geometric mean of the R/H ratio is 6.25 (M) and 11.91 (L).
- **Certification is the bigger result.** Before: 5 MILPs and 3 LPs finished without a certificate (2 `time_limit` and 3 `feasible` on M; 3 `time_limit` on L). After: every instance on both panels is `optimal` and certified.
- Node counts collapse: 610 256 → 217 on M. sp150x300d went from `feasible` at 44 163 nodes to `optimal` at 1; p0548 from 11 309 to 1; lseu from 8 241 to 15.
- Largest per-instance ratios — M: sp150x300d 93×, bell5 42×, issue-2446 35×. L: standmps 89×, shell 78×, scrs8 63×. The only instance where the old route wins is issue-2173 (0.6×).
- **Load gate: this run does not meet it, and I am not restating the gate to fit.** §6.3 wants load < 2 throughout. It started at 1.88 but 21 of 25 samples ran above 4 (max 9.13). A re-run cannot fix this: with nothing of mine running, 12 samples over 2 minutes give an idle load of min 2.65, median 3.06, max 4.53 — never below the gate. The load average here is dominated by uninterruptible system daemons, not CPU contention: `top` concurrently reports 89.7% CPU idle on 14 cores with 1 runnable thread. So **treat the walls as indicative, not gate-valid.** What does not depend on load: `incorrect_count`, statuses, certification and node counts, all of which carry the result above. Wall spread is also small (H max sd 11 ms on M, 6 ms on L), and the effect is 6-12× per instance, far outside that noise.
- **Retraction:** the table in the next section reports M total wall 85.0 s (R) / 5.3 s (H) and geo-mean ratios 5.68 / 11.89, from an earlier run whose load was also out of gate. The numbers above supersede it. Neither run is gate-valid; the direction and magnitude agree.

### Benchmarks (earlier run, superseded by the section above)

Setup: 3 rounds per instance, interleaved arms R (Rust route), H (HiGHS route) and H0 (raw highspy); TL 20 s, gap 1e-4, highspy 1.15.1. The oracle is HiGHS at gap 1e-4.

| Panel | Executed | Incorrect | H | R |
|---|---|---|---|---|
| M (23 MILPs + i1183) | 216 | 3, all R | 24/24 optimal, certified; total wall 5.3 s | 19/24 certified; total wall 85.0 s; 2122 false `infeasible` ×3 (fixed on this PR) |
| L (24 LPs) | 216 | 0 | 24/24 optimal, certified; total wall 3.2 s | 21/24 certified; total wall 117.0 s |
| Infeasible | 36 compared | 0 wrong | the 6 former `error` cases are now `infeasible` | |

- Geo-mean wall ratio R/H: M 5.68, L 11.89.
- Objectives: across 88 comparisons against H0, the worst relative difference is 1.7e-5.
- Node counts, R / H: i1183 4965 / 19, lseu 8241 / 15, p0548 11309 / 1, dcmulti 21485 / 1.
- **Caveats (plan §12 audit):**
  - Panel load rose past the §6.3 gate (M 2.7 → 8.0), so the walls are not gate-valid.
  - H1 (M-MIPLIB) and H2 (L-Netlib) were run on the 24-instance proxy panels, not as registered.
  - H3 has no H arm.
- **H5 fails as registered:** H − H0 > 20 ms on 20/24 MILPs and 17/24 LPs.
  - Attributed by wall-clock stage timers. All certificate checks are ≤ 44 ms per instance, and < 10 ms on the MILPs.
  - The excess is the pre-dispatch `Model.solve` pipeline that every route pays: declared-box NBT, convexity classification, std-form extraction.
  - On MILPs it also includes the route's three HiGHS runs.
  - This is recorded, not changed here. **Retraction:** an earlier version of this description said E1 was not measured. The R/H0 comparison was measured; the H arm is missing.

### MINLP guards

- **Routing guard (H4)** (`lp_milp_minlp_guard.py`): 66 `.nl` files × {rust, highs, rust again}.
  - compared=66, drift=0, noisy=3.
  - The route is `None` on all 66: no MINLP reaches HiGHS.
- **Rust LP fixes, cert-clean A/B:** 66 `.nl` files, arms old rules / new / new again, `backend=rust`, TL 20 s, max_nodes 300, oracle `minlplib.solu`.
  - 156 oracle-checked arms, **0 oracle violations, 0 certification regressions**. Counter profiles differ on 14 instances, so the toggle fired.
  - Changed rows: beuster and tspn10 are noisy (the repeat differs).
  - casctanks: 17 → 9 nodes, with the same bound and incumbent.
  - nvs05: 73 → 63 nodes, same bound, incumbent 1107.89 → 8.116. Both are time-limited under load.
- **hda "floor-as-fallback":** falsified before building. Both stale-floor nodes decline in `_certify` on magnitude (2.1e11 > 1e7), so the change could not tighten hda.

### Verification of the final tree (commit 964aea71)

- `pytest -m smoke`: 1531 passed, 1 skipped.
- Adversarial suite: 19 passed.
- The 9 previously failing CI files plus the route and new tests pass.
- Fail-before / pass-after, with the old modules stashed: 7 route and #844 tests. The no-rows integer test also failed on the rust arm before the F4 fix.
- NBT-related test files: 236 passed.
- `ruff` and `ruff format` are clean; pre-commit passed. No Rust was touched in this commit.
- `test_875_root_setup_budget::test_watercontamination0202_honours_its_time_limit[60]` (needs the local MINLPLib snapshot; skipped on CI) took 79.7 s solo against 75 s allowed, at load ~5. The instance is MBQP, so the LP/MILP route never takes it. Interleaved A/B against the previous commit 9d38881a (module path and marker asserted per arm), 2 rounds at load 3.9–4.2: HEAD 76.7 ± 0.8 s, this commit 76.0 ± 0.4 s. Both end in `time_limit` with 0 nodes. No regression from this commit. The overrun against the 75 s allowance is load-related and predates it.

### Verification (after merging `main`, before 964aea71)

- `cargo test -p discopt-core`: 745 passed, 0 failed (730 unit + 15 in other binaries). `cargo fmt --check` is clean. No clippy warnings in touched files.
- New Rust regression tests:
  - `crossing_within_feasibility_tolerance_is_not_an_empty_box`: failed before, passes after.
  - `free_basic_column_is_never_pivoted_onto_the_sentinel`: fails under the old rule (`Numerical`), passes after.
  - `unstable_pivot_recovery_is_not_gated_on_a_deadline` (bchoco06 fixture).
- `pytest -m smoke`: 1478 passed, 12 skipped, on the new default.
- Adversarial suite (`pytest -m slow python/tests/test_adversarial_recent_fixes.py`): 19 passed.
- Route plus #1229 tests (`test_lp_milp_highs_route.py`, `test_issue_1229_free_column_sentinel.py`): 34 passed.
- Earlier on this branch, each of these failed before and passes after:
  - `test_highs_route_skips_discopt_root_presolve[lp,milp]`
  - `test_highs_route_runs_after_another_highs_user_in_the_process[lp,milp]`
  - `test_default_route_imports_highspy_only_for_pure_lp_milp[milp]`
- `mypy python/discopt/`: no issues in 346 files. `ruff check` and `ruff format --check`: clean.

### Optional stages not built

1b (reformulated routes) and Stage 2 (HiGHS as the OA/AMP/GDP master; E2 first). Both are optional under plan §2.3.


python/discopt/solvers/lp_milp_highs.py
```diff
@@ -0,0 +1,1274 @@
...
+    lo = (np.abs(sf.xl) >= READBACK_LIMIT) & (sf.xl > -INF)
+    hi = (np.abs(sf.xu) >= READBACK_LIMIT) & (sf.xu < INF)
+    return np.asarray(lo | hi, dtype=bool)
+
+
+def _relax_huge_box(sf: StdForm, huge: np.ndarray) -> StdForm:
+    """``sf`` with the huge finite bounds opened to infinity: a relaxation of ``sf``."""
+    return dataclasses.replace(
+        sf,
+        xl=np.where(huge & (sf.xl < 0), -INF, sf.xl),
```
jkitchin, line 718: **Blocking: this drops *declared* bounds, and turns solvable MILPs into `error`.**

`_huge_box` flags a column if **either** side is sentinel-magnitude, and `_relax_huge_box` then opens the other, declared side too, because it tests only the sign (`sf.xl < 0` / `sf.xu > 0`). `solve_milp_std` (L1064) applies this unconditionally, so an ordinary one-sided bound is discarded before HiGHS ever sees the model.

Measured on this branch (3.12 venv, markers asserted per CLAUDE.md §8 — `ns_safe_bound_csc_py` present in the `.so`, `_relax_huge_box` present in the module):

```python
m.continuous("x", lb=-5.0); m.binary("z")
m.subject_to(x + 3*z <= 10); m.minimize(x)
```

```
default route                 -> status=error, objective=None
DISCOPT_LP_MILP_BACKEND=rust  -> optimal -5.0        # i.e. main's answer
```

`x >= -5` is gone, HiGHS returns `kUnboundedOrInfeasible`, and the declared-box root LP is `optimal` so it cannot decide the MILP. Symmetric for a finite positive `ub` with a sentinel `lb` (`m.continuous("x", ub=5.0)` loses `x <= 5`, `max x` → `error`). Exhaustive on the mask:

| declared box | after `_relax_huge_box` | |
|---|---|---|
| `lb=0`, default ub | `[0, inf]` | ok |
| `lb=2`, default ub | `[2, inf]` | ok |
| `ub=-3`, default lb | `[-inf, -3]` | ok |
| `lb=-5`, default ub | `[-inf, inf]` | **loses `x >= -5`** |
| `ub=5`, default lb | `[-inf, inf]` | **loses `x <= 5`** |

`lb=0` survives only because `0 < 0` is false — the common case is saved by accident, not by design. This is a behavior regression against the old default on an ordinary modelling pattern, and it is plausibly a large share of the 283/1000 `error` rate the description reports.

Suggested fix — use the per-side masks `_huge_box` already computes internally:

```python
def _huge_box(sf: StdForm) -> tuple[np.ndarray, np.ndarray]:
    lo = (np.abs(sf.xl) >= READBACK_LIMIT) & (sf.xl > -INF)
    hi = (np.abs(sf.xu) >= READBACK_LIMIT) & (sf.xu < INF)
    return lo, hi

def _relax_huge_box(sf: StdForm, huge_lo: np.ndarray, huge_hi: np.ndarray) -> StdForm:
    return dataclasses.replace(
        sf, xl=np.where(huge_lo, -INF, sf.xl), xu=np.where(huge_hi, INF, sf.xu)
    )
```

I checked this still opens **both** sides of the fully-free column, so the F2 case keeps working:

```
lb=-5 / default ub   -> [-5, inf]      (declared lb kept)
default lb / ub=5    -> [-inf, 5]      (declared ub kept)
free col (F2 case)   -> [-inf, inf]    (unchanged)
```

Worth noting why the suite missed it: `test_milp_on_the_default_box_is_not_falsely_infeasible` (the F2 regression test) uses a **fully free** column `y`, which is exactly the one case where opening both sides is harmless. A one-sided declared bound on a MILP has no coverage.

The LP path gets this right, incidentally — it tries the declared box first and relaxes only to hunt a Farkas ray, verifying the ray against the declared box regardless. Mirroring that staging on the MILP side would also work, and would avoid handing HiGHS a weaker model than necessary.

---

python/discopt/solvers/lp_milp_highs.py
```diff
@@ -0,0 +1,1274 @@
...
+        return False
+    y = y / np.max(np.abs(y))
+    for s in (1.0, -1.0):
+        r = sf.A.T @ (s * y)
+        yb = float(sf.b @ (s * y))
+        pos, neg = r > 0.0, r < 0.0
+        if np.any(pos & (sf.xu >= INF)) or np.any(neg & (sf.xl <= -INF)):
+            continue
+        terms = np.where(pos, r * sf.xu, 0.0) + np.where(neg, r * sf.xl, 0.0)
+        sup = float(np.sum(terms))
```
jkitchin, line 344: **Soundness gap in the Farkas kernel — witness attached; no end-to-end reproducer.**

`r = sf.A.T @ (s * y)` is formed in plain float64, and the guard `yb - sup > RAY_REL * scale` derives `scale` from those same rounded terms. Roundoff in a column dot enters `sup` at full magnitude while it raises the margin by only `RAY_REL` (1e-9) times itself, so the guard can be passed by a ray that proves nothing.

Constructed witness — a StdForm that is **exactly feasible** and is nevertheless certified infeasible:

```python
P = 2.0**53
A  = [[P, 1, 0], [1, 0, 0], [-P, 0, 1], [-1, 0, 0]]
b  = [P + 2.0, 1.0, -P + 2.0, -1.0]
xl = xu = [1.0, 2.0, 2.0]
y  = [1.0, 1.0, 1.0, 1.0]
```

`x* = (1, 2, 2)` satisfies all four rows exactly — I verified each row in `Fraction` arithmetic, not float — and lies in the box. Summed in stored order the `+1` of column 0 is absorbed by `2^53`, so:

```
computed (A^T y) = [-1.  1.  1.]
exact    (A^T y) = [ 0   1   1 ]
farkas_verified(y, sf) -> True        # on a feasible problem
```

The spurious `-1` drops `sup` from 4 to 3 against a margin of 1.1e-8.

This kernel is the sole gate on the LP `infeasible` label, and on the `farkas-root-lp` promotion in `solve_milp_std` that turns an *unusable MILP incumbent* into a certified `infeasible` — the path where a wrong accept converts an undecided outcome into a wrong certificate.

What makes this look like an oversight rather than a deliberate tolerance choice: this module already corrects for exactly this absorption twice. `fbbt_box` widens every derived side by an explicit `(nnz_i + 8)·eps·(Σ|terms| + |b_i|)/|a_ij|` error bound, and `_on_open_side` (L476) zeroes a roundoff reduced cost against a sentinel-scale side, citing "a roundoff reduced cost of 1.4e-17 times that side cost 1.9e3 of bound on nlp_cvx_001_010". Same mechanism, two different treatments.

**Caveat, stated plainly:** I did **not** find a model where HiGHS hands the route such a `y`, so I have no end-to-end reproducer — only the kernel-level witness above. Per §4 that makes this a demonstrated defect in the verifier, not a demonstrated wrong answer from the route. Given §1's "the product is the certificate", I'd still want the error bound added here rather than relying on HiGHS never producing an adversarial ray.

---

python/discopt/solvers/lp_milp_highs.py
```diff
@@ -0,0 +1,1274 @@
...
+    stats: dict[str, float] = field(default_factory=dict)
+    labels: dict[str, str] = field(default_factory=dict)
+
+
+# ─────────────────────────────────────────────────────────────
+# Solver-independent verification kernels
+# ─────────────────────────────────────────────────────────────
+
+
+def _logical_columns(sf: StdForm) -> np.ndarray:
```
jkitchin, line 169: **Non-blocking (pre-existing): the logical-column correction is computed but never written back.**

`_logical_columns` matches any continuous, zero-cost column in exactly one row — which includes **structural model variables**, not just the slacks `_decompose_eq_slack_form` adds. `readback_problem` waives the sentinel guard for them, and `_exact_rows_problem` then proves the row feasible by computing the value the logical *would need* and checking that value against its bounds. That value is never written into `x`, so the returned point is not the point that was verified.

```python
m.continuous("x"); m.continuous("w")
m.subject_to(x + w == 1e4, name="link"); m.maximize(x)
```

```
status=optimal  gap_certified=True
x = 9.999e+19   w = -9.998999999999998e+19
row 'link' residual at the RETURNED point: 6384
```

`need = 1e4 - 9.999e19` is exact and inside `w`'s box, so the check passes; float64 cannot represent it (ulp(1e20) = 16384), so the vector handed back violates a user-named equality by 6384.

The optimal **value** is sound — `c_w = 0` and a feasible completion exists at the same objective — so this is not a false certificate. But `x` is consumed downstream (incumbents, MIP seeding, `result.x`), and the docstring's claim that "the point is feasible iff that value is" is true of the corrected point, not the returned one.

**Not a regression:** `DISCOPT_LP_MILP_BACKEND=rust` returns the same 6384 residual, so this is a pre-existing gap that the new verifier declines to close rather than one this PR introduces. Raising it because the PR's premise is that every HiGHS certificate is re-verified, and this is the one place the verification and the artifact diverge. Fix looks small: have `_exact_rows_problem` return the corrected `need` and write it into `x` on acceptance.

---

python/discopt/solver.py
```diff
@@ -9753,6 +9763,19 @@ class _DcbSkip(Exception):
     # above overran ``time_limit``): presolve only tightens bounds, so declining
     # it leaves a looser-but-valid box and lets the wall track ``time_limit``
     # (#654). ``propagate_bounds_to_model`` is a no-op when skipped.
+    #
+    # #1229 (plan §12, H5): a pure LP / MILP bound for the HiGHS route skips root
+    # presolve and the presolve-gated passes below; HiGHS presolves internally.
+    if presolve and _highs_takes_pure_lp_milp(
```
jkitchin, line 9769: **Nit (measured): this gate re-classifies the model on every solve.**

`_highs_takes_pure_lp_milp` runs an unmemoized `classify_problem` (full `model_to_repr` + per-constraint linearity scan), and `_milp_is_exactly_linear` runs `model_to_repr` again for its degree-analysis backstop. Neither result is cached for the dispatch site at L10228, which re-derives the same facts.

Instrumented on a 60-variable / 31-row MILP (wrapping `classify_problem` and `discopt._rust.model_to_repr`):

| arm | `classify_problem` | `model_to_repr` |
|---|---|---|
| default (highs) | 3 | 9 |
| `DISCOPT_LP_MILP_BACKEND=rust` | 2 | 6 |

One extra classification and three extra model conversions per solve. Small in absolute terms, but it lands in exactly the pre-dispatch pipeline that §12's H5 attribution blames for the `H − H0` excess (>20 ms on 20/24 MILPs, >10 ms of it outside the certificate checks) — so it's part of the thing already identified as the route's fixed cost. The H4 guard can't see it either, since that panel compares status/objective/bound/nodes only.

Memoizing the classification on the model (or threading the gate's result through to the dispatch site) would remove it. Not blocking.

---

jkitchin (commented):
## Review — 4 findings, 1 I'd fix before merge

Reviewed the code at `9bd6aa6` (the newer `24399fa` is docs-only). Built the branch in a clean 3.12 venv and asserted the loaded tree per §8 before measuring anything: `discopt.__file__` in the branch tree, `_relax_huge_box` present in the route module, and `ns_safe_bound_csc_py` — new in this PR — present in the compiled `.so`. That check caught a first build that had silently resolved against `main`'s `pyproject.toml` (no `highspy`), which would have invalidated everything below.

### What I ran

| | result |
|---|---|
| `pytest -m smoke` | **1478 passed, 12 skipped** |
| `pytest -m slow python/tests/test_adversarial_recent_fixes.py` | **19 passed** |
| New/changed test files (route, #1229 sentinel, #844, NBT absorption, row filter) | **54 passed** |
| CI on the PR | Rust, Lint+typecheck, and both HiGHS-route lanes green at review time |

Each probe below prints an executed-check count and exits non-zero on zero (§6).

### What holds up

The two Rust fixes are right and properly tested. `select_leaving` testing a ±1e20 sentinel as though it were a bound is a real defect and the fix is the correct one; `fbbt_row`'s widening of a sub-`FEAS_TOL` crossing to `[min, max]` is a relaxation and therefore sound; `dual_slack_basis`'s free-column rule is correct. On the Python side `exact_ns_bound`'s rational correction and downward rounding, the F1 NBT float-error widening, the F3 `#844` guard, and the new empty-integer-box proof all check out. `copy.copy(milp)` in the row-filter re-solve is safe — I confirmed `_filter_unresolvable_rows` rebinds `_A_ub`/`_b_ub` rather than mutating them, and `solve()` only rebinds attributes on `self`, so the node's own LP can't be corrupted by the shadow solve.

The LP path's staged handling of the huge box — try the declared box, relax only to hunt a Farkas ray, verify the ray against the declared box regardless — is the right shape. Finding 1 is what happens where the MILP path doesn't do that.

### Findings

Details are in the four inline comments.

1. **`_relax_huge_box` drops declared bounds on MILPs** — `lp_milp_highs.py:718`. **The one I'd fix before merge.** A column with one sentinel-magnitude side has its *other*, declared side opened too. `m.continuous("x", lb=-5.0)` in a MILP returns `status=error` on the default route where the old default returns `optimal -5.0`. Regression against `main` on an ordinary modelling pattern; fix is a few lines, and I verified it preserves the F2 free-column case.
2. **`farkas_verified` certifies a feasible StdForm infeasible** — `lp_milp_highs.py:344`. Witness attached (feasibility proved in `Fraction` arithmetic). No end-to-end reproducer — I did not find a model where HiGHS supplies such a ray — so this is a defect in the verifier, not a demonstrated wrong answer. Flagging it because this module already corrects for the same absorption in `fbbt_box` and `_on_open_side`.
3. **Logical-column correction verified but not written back** — `lp_milp_highs.py:169`. Returns a certified-`optimal` point violating a named equality by 6384. Pre-existing (`=rust` gives the identical residual), so not a merge blocker, but it's the one place the verification and the returned artifact diverge.
4. **Routing gate re-classifies on every solve** — `solver.py:9769`. +1 `classify_problem`, +3 `model_to_repr` per solve, measured. Nit.

### One process question, and one description nit

**`relax_row_filter_loose_bound` ships default-ON.** The mechanism is sound (dropping rows enlarges the feasible set, so the filtered bound is valid for the original, and `max` of two valid bounds is valid). The evidence clears §5 bar (1) *cert-clean* — 66 `.nl`, 156 oracle-checked arms, 0 violations, 0 certification regressions. Bar (2) *net-positive* is where I'd want your read: the case made is that it restores hda's baseline, which this PR itself moved via the `dual_slack_basis` fix, and §12 records the 3 changed rows as noisy (ON2 differs from ON). Defensible as a repair rather than an optimization — the `DISCOPT_CUT_INHERIT` lesson is about flags that buy nothing, and this one demonstrably buys back a regression — but it's a judgement call and the flag was added late in the branch, so worth making explicit rather than implicit.

**Description nit:** "MINLP code paths are unchanged, apart from two defects in the shared Rust LP engine" understates the diff — `nonlinear_bound_tightening.py`, `lp_spatial_bb.py`, `mccormick_lp.py` and the new row-filter flag are all MINLP paths. Everything is documented further down, so it's a summary-line inaccuracy rather than an omission, but that line is what a reader scopes their review by.

Nothing here touches the Rust work, the packaging, or the certificate machinery's core design, all of which look solid. Finding 1 is the merge-blocking one.

---

PR #1259, opened 2026-09-16
fix(lp-milp): #1229 keep declared bounds beside a default-box side on the HiGHS route

jkitchin:
Fixes the blocking finding from the review of #1258, posted after that PR merged. The
regression is on `main` now, so this is a follow-up PR rather than a change to #1258.

## The defect

`_relax_huge_box` decided *which side* of a column to open from the **sign of the bound**
rather than from a per-side mask, and `_huge_box` flagged a column when **either** side sat
at sentinel-adjacent magnitude (the default ±9.999e19 box). A column with one
sentinel-magnitude side therefore had its other, **declared** side opened too:

```python
xl=np.where(huge & (sf.xl < 0), -INF, sf.xl)
xu=np.where(huge & (sf.xu > 0), INF, sf.xu)
```

So `m.continuous("x", lb=-5.0)` in a MILP lost `x >= -5` before HiGHS ever saw the model.
HiGHS returned `kUnboundedOrInfeasible`; the root LP check runs on the *declared* box and
came back `optimal`, so the route could not decide and answered `error`.

`lb=0` escaped only because `0 < 0` is false — the common case was saved by accident, not
by design.

## Reproducer

`m.continuous("x", lb=-5.0)`, `m.binary("z")`, `x + 3z <= 10`, `min x`:

| declared box | before | after | old default (`=rust`) |
|---|---|---|---|
| `lb=-5`, default ub, `min x` | **error** | optimal **-5.0** | optimal -5.0 |
| `ub=5`, default lb, `max x` | **error** | optimal **5.0** | optimal 5.0 |
| `lb=0` / `lb=2` / `ub=-3` | optimal | optimal | optimal |

Both arms assert provenance per CLAUDE.md §8 before measuring: `discopt.__file__` in this
tree, `_relax_huge_box` present, and the H arm on a `highs-*` route with the R arm off it.
`executed=10`, H-vs-R divergences **2 → 0**.

The failing status line names the mechanism exactly:
`HiGHS MILP kUnboundedOrInfeasible; root LP relaxation optimal`.

## The fix

`_huge_box` returns the two side masks it already computes, and `_relax_huge_box` applies
each to its own side, so an ordinary declared bound beside a default side is kept:

```python
xl=np.where(huge_lo, -INF, sf.xl)
xu=np.where(huge_hi, INF, sf.xu)
```

Every side that moves still moves outward, so the result is a relaxation of `sf` exactly as
before, and the fully free column (the F2 case that motivated the relaxation) still opens
both sides.

## A second defect the fix exposed

The LP path in `solve_lp_std` still referenced the old single mask in its re-solve trigger.
Ruff caught it as `F821 Undefined name`; it would have been a `NameError` at runtime.
Nothing in the suite covers that branch — `lp/huge_box_relaxed_resolve` is asserted by no
test — so tests alone would not have caught it. The new kernel unit test pins the helper
contract directly, which is the part that is cheap to cover deterministically.

## Tests

Three regression tests, all **failing before** this change and passing after (verified by
stashing only the source file, so the tests ran against merged `main`: `3 failed`):

- `test_milp_declared_lower_bound_beside_a_default_side_survives`
- `test_milp_declared_upper_bound_beside_a_default_side_survives`
- `test_huge_box_masks_each_side_separately_and_keeps_declared_bounds` (kernel-level: pins
  both masks, the kept declared bounds, and the relaxation property)

The existing F2 test could not catch this: its column is **fully free**, which is exactly
the one case where opening both sides is harmless.

## Verification

- `python/tests/test_lp_milp_highs_route.py` — **41 passed** (3 of them new)
- Before/after falsification — **3 failed** on merged `main`, 41 pass here
- `ruff check`, `ruff format --check`, `mypy` on both changed files — clean
- `pytest -m smoke` and `pytest -m slow python/tests/test_adversarial_recent_fixes.py` —
  running when this PR was opened; results posted in a comment below.

No Rust was touched, so `cargo test` is not implicated.

## The other review findings

- **Finding 2** (`farkas_verified` certifies a feasible StdForm infeasible) — I reproduced
  the reviewer's witness independently: all four row residuals are exactly `0` in `Fraction`
  arithmetic and the point lies in the box, yet `farkas_verified` returns `True`, because
  the float `AᵀY` is `[-1, 1, 1]` against an exact `[0, 1, 1]`. **Confirmed**, and a real
  soundness gap in a certificate kernel. It is a different defect class needing its own
  error bound and tests, and it has no end-to-end reproducer, so it goes in its own PR
  rather than delaying this regression fix.
- **Finding 3** (logical-column correction computed but not written back) — pre-existing;
  `=rust` returns the identical 6384 residual. Not addressed here.
- **Finding 4** (routing gate re-classifies each solve) — nit, not addressed here.

Contributes to #1229.


jkitchin:
Suite results promised in the description, run on this branch at `e962a24f`:

| suite | result |
|---|---|
| `pytest -m smoke` | **1480 passed, 1 skipped**, 10761 deselected (168.4 s) |
| `pytest -m slow python/tests/test_adversarial_recent_fixes.py` | **19 passed** (144.4 s) |
| `python/tests/test_lp_milp_highs_route.py` | **41 passed** (3 new) |
| before/after falsification | **3 failed** on merged `main`, verified by stashing only the source file |
| `ruff check` / `ruff format --check` / `mypy` | clean on both changed files |

No Rust was touched, so `cargo test -p discopt-core` is not implicated.

The two suites ran sequentially rather than concurrently, so neither contended with the
other's time-limited solver tests.

Issue #671, opened 2026-07-16
Tight dual bound for hda-class flowsheets needs higher-precision (iterative-refinement / exact-rational) LP arithmetic


jkitchin:
## Summary

Follow-up to #517 (closed) and #662 (candidate A, merged). #662 gave hda its **first finite dual bound** via a Neumaier–Shcherbina safe bound extracted from the in-house simplex's own dual on a numerically-failed node LP (flag `DISCOPT_NODE_NUMERICAL_DUAL_BOUND`, default OFF). That bound is **sound but loose** (`−1.80e10` vs opt `−5964.53`). This issue tracks the **tight** bound, whose blocker is now rigorously diagnosed.

Full measured trail: `docs/dev/candidate-b-phase2-scaling-entry-2026-07-16.md` (branch `claude/issue-664-phase2-robustness`). This issue supersedes the exploratory tracker #664.

## Verified diagnosis: the root relaxation is FEASIBLE but float64-intractable

hda's root McCormick relaxation LP (2974×1138) is **feasible** — the elastic phase-1 minimum constraint violation in raw units is **1.8e-10** — but **every available float64 LP engine false-infeasibles on it**:

| engine | verdict on hda's root LP |
|---|---|
| in-house feral simplex | `numerical` (FT `Growth` on near-singular bases) |
| HiGHS simplex | infeasible (false) |
| HiGHS interior-point | numerical |
| SCIP / SoPlex (SCIP 10, iterative refinement) | infeasible (false) |

The cause is the relaxation's linear algebra, and it is **scale-invariant**:

- **Rank-deficient** by ~15–43 (SVD numerical rank 1095 of 1138 columns), with ~400 tiny singular values and a true condition number **~1e14**.
- Optimal diagonal scaling (log min-max LP) bottoms out at coefficient range **1.77e9** — a rigorous floor. Confirmed *not* residue: dropping entries 1e-14…1e-8 leaves it unchanged.

### Falsified along the way (do not re-try)

- **Scaling** (better algorithm / more passes / lower noise floor / slack-aware "L1"): L1 cut the coefficient spread 2.8e26→4.1e9, but conditioning is scale-invariant so hda still failed. Optimal scaling is provably 1.77e9.
- **Presolve / rank-reduction:** dropping to 1063 full-rank rows still leaves HiGHS `numerical`.
- **feral's existing double-precision iterative refinement** (`with_numeric_focus`): byte-identical, no effect (residual computed in double can't recover ~1e14 loss).
[...]

PR #1265, opened 2026-09-16
fix(correctness): #1254 #1255 #1256 feasible-distance gate, warm-start completion, log10-of-sum relaxation

### LINKED ISSUE #1254: correctness: absolute feasibility tolerance certifies an infeasible point as optimal on small-magnitude constraints
## Summary

In a constraint whose terms are all tiny, an absolute feasibility tolerance accepts an infeasible point. `Model.solve` then returns `status="optimal"` with `gap_certified=True` at that point, even though the true optimum is far away.

## Reproducer (discopt 0.8.0)

```python
import discopt.modeling as dm

# minimize z  s.t.  10**y1 + 10**y2 <= 10**z,  y1 + y2 >= -14,  y in [-9, -6], z in [-20, 0]
# True optimum: y1 = y2 = -7, z* = log10(2e-7) = -6.69897
m = dm.Model("unscaled_epigraph")
y1 = m.continuous("y1", lb=-9.0, ub=-6.0)
y2 = m.continuous("y2", lb=-9.0, ub=-6.0)
z = m.continuous("z", lb=-20.0, ub=0.0)
m.subject_to(y1 + y2 >= -14.0)
m.subject_to(10.0 ** y1 + 10.0 ** y2 <= 10.0 ** z)
m.minimize(z)
r = m.solve(time_limit=30)
x = {k: float(v) for k, v in r.x.items()}
print(r.status, r.objective, r.bound, r.gap_certified)
print(x, "violation:", 10 ** x["y1"] + 10 ** x["y2"] - 10 ** x["z"])
```

Output:

```text
optimal -20.0 -20.000000000000043 True
{'y1': -7.000000000000022, 'y2': -7.000000000000022, 'z': -20.0} violation: 1.9999999999997977e-07
```

The same constraint divided through by `10**z`, i.e. `10**(y1 - z) + 10**(y2 - z) <= 1`, gives the right answer:

```text
optimal -6.6989699861377465 -6.699218750000013 True
```

## Expected

The point `z = -20` violates the constraint by 2e-7. The right-hand side at that point is 1e-20, so the relative violation is about 10^13. A certified-optimal result should not be returned at a point like this.

Either the certificate should check feasibility relative to the constraint's scale (for example, the magnitude of its terms), or the result should say the certificate holds only up to an absolute tolerance of about 1e-6.

## Context

- **Where it came up:** the discopt-speciation plugin's equilibrium-constrained optimization, where concentrations are naturally 1e-8 to 1e-3 mol/kg. The natural epigraph `sum(m_i) <= 10**z` certified the log-concentration floor, not the true minimum.
- **Workaround there:** every constraint is written in scale-invariant form.
- **Risk:** any model with small-magnitude quantities can hit this, and the failure is silent because the result looks certified.

Possibly related: #952, #954 and #1199 (closed) touched incumbent feasibility checks with absolute tolerances.

### LINKED ISSUE #1255: initial_solution raises 'expected length N, got M' after mpec.reformulate_gdp adds indicator variables
## Summary

`Model.solve(initial_solution=...)` raises `ValueError: objective: x: expected length N, got M` when the model uses `discopt.mpec.reformulate_gdp`.

The cause:
1. The warm start is flattened against the variables the user declared.
2. The complementarity reformulation then adds indicator variables at solve time.
3. The warm-start objective check evaluates the shorter vector against the longer problem.

## Reproducer (discopt 0.8.0)

```python
import discopt.modeling as dm
from discopt.mpec import complementarity, reformulate_gdp

m = dm.Model("warmstart_gdp")
y = m.continuous("y", lb=-8.0, ub=-1.0)
n = m.continuous("n", lb=0.0, ub=1.0)
z = m.continuous("z", lb=-10.0, ub=0.0)
m.subject_to(10.0 ** y + n == 1e-3)
reformulate_gdp(m, [complementarity(n * 1e3, -4.0 - y, name="sat")])
m.subject_to(10.0 ** (y - z) <= 1.0)
m.minimize(z)

m.solve(initial_solution={y: -4.0, n: 9e-4, z: -4.0}, time_limit=20)
```

```text
  File ".../discopt/solver.py", line 9876, in solve_model
    ws_obj = float(evaluator.evaluate_objective(initial_point))
  File ".../discopt/_tape_nlp_evaluator.py", line 555, in evaluate_objective
    return float(self._problem.objective(self._x(x)))
ValueError: objective: x: expected length 5, got 3
```

Without `initial_solution` the model solves (`optimal -4.0`).

A simpler model, a complementarity between two plain variables with a quadratic objective, accepts the warm start. The failure needs the nonconvex spatial branch-and-bound route.

## Expected

The warm start should be extended to the reformulated variables. The indicator values follow from the given point: here `n > 0` implies the "slack = 0" disjunct. Failing that, the unset variables could be filled with defaults, as `validate_initial_solution` already does for variables not provided.

At minimum, the warm start should be dropped with a warning rather than raising.

## Context

- **Where it came up:** the discopt-speciation plugin, which warm-starts a GDP-reformulated equilibrium model from a verified local optimum.
- **Workaround there:** it catches the `ValueError` and re-solves cold.
- **On `main`:** at 03c5218d72, the same `ws_obj = float(evaluator.evaluate_objective(initial_point))` check is still present in `solver.py`.
- Possibly related: #689, where `initial_solution` was silently dropped when a reformulation lifted the model.

### LINKED ISSUE #1256: relaxation: log10 of a sum gets no bound; IncrementalMcCormickLP fails with a reshape error on shaped variables
## Summary

When the objective is `log10` of a sum of exponentials, the incremental McCormick LP declines to build. The solve then returns `status="feasible"` with `bound=None` after one node.

With shaped (vector) variables, the logged reason is a reshape error, `cannot reshape array of size 1 into shape (3,)`, which points to a flattening or indexing bug. With scalar variables the reason is `relaxation has no valid bound / no rows`.

The same problem written as an epigraph gets a tight bound.

## Reproducer (discopt 0.8.0)

```python
import logging
import discopt.modeling as dm

logging.basicConfig(level=logging.DEBUG, format="%(name)s: %(message)s")
for noisy in ("jax", "absl"):
    logging.getLogger(noisy).setLevel(logging.WARNING)

# minimize log10(sum_i 10**y_i)  s.t.  sum_i y_i >= -12,  y in [-8, -1]; optimum y_i = -4, obj log10(3e-4) = -3.5229
m = dm.Model("log10_sum_shaped")
y = m.continuous("y", lb=-8.0, ub=-1.0, shape=(3,))
m.subject_to(y[0] + y[1] + y[2] >= -12.0)
m.minimize(dm.log10(dm.sum(10.0 ** y)))
r = m.solve(time_limit=20)
print(r.status, r.objective, r.bound)
```

```text
discopt._relax.incremental_mccormick: IncrementalMcCormickLP declined to build: ValueError: cannot reshape array of size 1 into shape (3,)
feasible -3.522878736188604 None
```

The error is the same with `dm.sum([10.0 ** y[i] for i in range(3)])`. With two extra variables in the model the message becomes `cannot reshape array of size 2 into shape (3,)`.

**Scalar variables** (`y0, y1, y2` declared separately, objective `dm.log10(10**y0 + 10**y1 + 10**y2)`):

```text
IncrementalMcCormickLP declined to build: ValueError: relaxation has no valid bound / no rows
feasible -3.522878736188604 None
```

**Epigraph form** of the same problem (minimize `z` s.t. `sum_i 10**(y_i - z) <= 1`):

```text
feasible -3.522878727093186 bound -3.523437500000008 gap 0.000159
```

## Expected

`log10` is monotone, and `sum(10**y)` has a convex relaxation, so a finite bound should exist. At minimum:

- the shaped-variable case should not hit a reshape error;
- `log10` of a positive sum should be relaxed, for example through its monotonicity or an automatic epigraph;
- if the relaxation cannot be built, that should be visible above DEBUG level. At present the solve simply returns `bound=None`.

## Context

- **Where it came up:** minimising log10 of total dissolved concentration in the discopt-speciation plugin.
- **Workaround there:** it uses the epigraph form.

jkitchin:
## Summary

Three reports from the discopt-speciation plugin, all on models whose quantities are naturally 1e-8 to 1e-3. They are one PR because they are one session's work on one plugin's failures; each is independently reviewable below.

Closes #1254. Closes #1255. Closes #1256.

**#1254 — an absolute feasibility tolerance certified an infeasible point as optimal.** Every gate asked only whether the residual was small (1e-6 in the incumbent verifier, 1e-4 in the heuristic gates). On `10**y1 + 10**y2 <= 10**z` over `y in [-9,-6]`, `z in [-20,0]` the point `y1 = y2 = -7, z = -20` violates the row by 2.0e-07 against a right-hand side of 1e-20; every gate accepted it, it became the incumbent, it met the root relaxation bound, and `solve` returned `status="optimal"`, `gap_certified=True` at `z = -20` where the true optimum is `-6.69897`.

A residual is not on its own a statement about the point. What makes a small residual tolerable is that the point sits within tolerance of one that *satisfies* the row, and the first-order distance to the row's surface is `violation / ||grad g||_inf`. Every gate now also requires that distance to stay inside the declared `rel = 1e-4` (`validation.feasibility.feasible_distance_cap`), applied as a `min` against the tolerance each gate already computed — so it can only ever reject, never accept. At the bad point every partial derivative is ~2.3e-07, so restoring feasibility needs `Δy ~ 0.87`, a third of that variable's whole box; the point is now rejected by nine orders.

Keyed on the **gradient**, not on the row's term magnitude, and that distinction is load-bearing: the converged Scholtes-regularized MPEC point (`x*y <= t` at `x = 1, y = 1.67e-08`) is violated by 40 % of the row's own magnitude — relatively *worse* than #1254's point — but `dg/dx = 1`, so it is 6.7e-09 away from feasible and `test_mpec_source_residuals.py` requires it to verify. Term magnitude cannot separate the two and ranks the acceptable one worse. A cancellation-noise floor keeps the cap clear of the large-term regime `test_polish_feasibility_gate_1199` pins.

**#1255 — `initial_solution` raised `ValueError: objective: x: expected length 5, got 3`** once the GDP pass lowered a disjunction into selector binaries at solve time. The point is now completed to the working model's columns (`warm_start.complete_initial_point`) right after the lowering, and dropped with a warning when it cannot be — a hint must never be able to fail a solve. The completion is not just padding: the appended discrete columns are repaired by a bounded best-neighbour search on total constraint violation, which recovers the indicator the given point implies. It has to accept one uphill move, since `sum(y) == 1` makes switching disjuncts a swap and plain descent stops on the wrong disjunct (measured: total violation 0.9, the activated wrong branch).

**#1256 — `log10` of a sum of exponentials got no dual bound at all.** Three causes, each a class rather than an instance:

- `b**e` with a constant base and a variable exponent canonicalized to an **opaque** node, and an opaque node has no envelope. For `b > 0` it is `exp(ln(b)*e)` identically, and `exp` is in the univariate envelope table with a certified-convex verdict on every box.
- `sum(<array expression>)` was opaque unless its operand was a bare `Variable`; it is now the sum of the scalar elements `_relax.scalarize` already derives for the constraint path.
- `IncrementalMcCormickLP` built its root box with one entry per `Variable` **object** while the lifted layout has one column per flat **element**, so any model with a shaped variable died on `cannot reshape array of size 1 into shape (3,)` and declined — and a correctly-sized flat box handed in by the caller was rejected for having the "wrong" length. Vectorized models therefore never got the incremental fast path, its cuts or its feasibility pump; they do now.

A solve that ends with no dual bound also now says so at WARNING rather than leaving `bound=None` as the only trace.

All three reproducers from the issues now return the true optimum; #1254's is certified.

## Correctness

- [x] Cannot produce a false certificate. The #1254 cap is applied as a `min` against each gate's existing tolerance, so the accepted set only ever shrinks — it cannot admit a point any gate rejected before. It rejects points, never fathoms nodes, so it cannot produce a false `infeasible` either (the terminal `infeasible` verdict still requires the same rigorous relaxation-empty proofs). The #1256 rewrite `b**e = exp(ln(b)*e)` for `b > 0` is an identity, not an approximation, so every envelope built on it relaxes the same function; `b <= 0` stays unsupported rather than being given an envelope that is wrong off the integers. #1255 is purely primal — the completed point is re-verified by the existing feasibility gate before it may seed anything.
- [x] No validation, fallback, or safety guard was weakened. One existing test changed (`test_row_filter_loose_bound`) and it was made *stronger*, not weaker — see below.

## Tests run

- [x] `pytest -m smoke` → **1483 passed, 12 skipped** (0 failed)
- [x] `pytest -m claim_boundary -n0` → **308 passed, 26 skipped** (0 failed)
- [x] PR-fast suite (the `make test` selection) → **10249 passed, 3 failed, 130 skipped, 4 xfailed, 1 xpassed**. The 3 failures are pre-existing and environmental — verified failing identically on the parent commit: two need the Ipopt C library (`cyipopt` import error) and one is `test_solve_degrades_to_jax_when_pounce_is_missing`.
- [x] `cargo test -p discopt-core` → N/A, no Rust touched.
- [x] `ruff check` / `ruff format --check` clean; `mypy python/discopt/` → Success, 346 source files.

## Regression test

- [x] Three new files, each failing on the parent commit and passing here:
  - `test_1254_feasible_distance_gate.py` (4 tests) — the cap separates #1254's point from the MPEC point, `verify_point` rejects the far point and still accepts the true optimum, and the end-to-end solve neither reports a point that violates its own row nor certifies a bound above the true optimum.
  - `test_1255_warm_start_across_reformulation.py` (3 tests) — the warm-started solve no longer raises and matches the cold optimum; the completed point is *feasible* in the lowered model (worst row violation ≤ 1e-9), not merely the right length; a non-prefix point is declined rather than padded.
  - `test_1256_log10_of_sum_relaxation.py` (6 tests) — no opaque node survives canonicalization for either spelling, the incremental root box is flat and a flat caller box is accepted, both spellings get a finite bound at or below the true optimum, and a boundless solve says so above DEBUG. All 6 fail on the parent (verified by running the new file against the parent source).

Not a flagged relaxation change, but #1256 does change what the relaxation layer emits, so it gets the panel treatment.

**66-instance in-repo differential panel** (`python/tests/data/minlplib_nl`, 10 s/instance, arms run sequentially on a quiet machine, each arm asserting its own `discopt.__file__` and a marker unique to the arm): **0 status changes, 0 certificate lost, 0 certificate gained, 0 objective drift, 0 new exceptions.** Three bound differences (`contvar`, `nvs05`, `tls2`) are time-limit variance, not the change: 10 interleaved runs per arm reproduce *both* outcomes within a single arm (nvs05: 15 nodes in every run of both arms, bound 2.03 or 1.36 in the baseline alone; tls2: 55/75/79/103 nodes and bounds 2.770–2.904 in both arms). `syn05hfsg` reports `bound > objective` identically in both arms and is a MAXIMIZE model, where that is the correct direction.

**Measured trade, recorded rather than hidden.** Giving `b**e` an envelope changes the relaxation of rows that carry one. hda has three `x == -log(c*4**(a+b*x0) - c)` rows whose exp spans 13 decades over the *declared* box, and the envelope coefficients degrade the Neumaier–Shcherbina bound read off a **raw-box** root LP: -5.71e6 → -9.99e11 with the #671 filter off. Sound throughout, and the filter still does its job on it (-9.99e11 → -1.38e9, a 725x recovery where it used to manage 88x). It is a raw-box artefact — every real solve runs FBBT/OBBT first, and hda end to end is unchanged at a 10 s limit (-124296.879066 vs -124296.879048) and **better** at 60 s (-64510.17 on both runs, against -122962 / -141697 across two baseline runs, same status and node count). Tightening the certified bound under ill-conditioned envelope rows is the interval max-combine already recorded as a follow-up.

`test_row_filter_loose_bound` pinned the raw-box constant `-1e5`; it now asserts the **recovery** — the filtered bound must close at least 98 % of the un-recovered bound's distance to zero, a 50x tightening — plus soundness on both arms, which is what #671 actually owns. Also updated: `docs/dev/data/claim-baseline.jsonl`, whose shape gate this change moves on exactly two rows (`contvar`, `hda`); the other 12 rows a full regeneration rewrites are cross-build float noise, established by regenerating from the parent commit on the same host, and are left as committed.

**Two corrections to earlier versions of this section, both from review and both verified before accepting** (CLAUDE.md §11):

1. The first relational form asserted `filtered > 100 * unfiltered`. Both bounds are negative, so that multiplication moves *away* from zero: it passed while the filter made the bound 99x **looser**, and it could not fail at all, since `mccormick_lp` keeps the larger of the two bounds. 10 000 random draws honouring that contract: 0 failures. Fixed to a division.
2. The bar was then stated as 99 % (100x), and this section claimed "the new form passes on the parent commit too". **That claim was false of the corrected form** — it was verified against the tautological one. 88x closes 98.87 %, not 99.9 %, so a 99 % bar rejects the pre-#1256 lowering. The threshold is now 98 % (50x), which clears both measured lowerings (88x with 1.8x headroom, 725x with 14.5x), and the claim is re-verified by *running* the corrected test against the parent commit's source: **3 passed** there and **3 passed** here.

Falsifiability of the final form is measured, not assumed: synthesising a k-fold recovery gives 1x rejected, 40x rejected, 50x exactly rejected (strict inequality), 60x accepted, 88x accepted — 5/5 cases executed.

One counter changed with the shaped-variable fix: a node that takes the incremental fast path returns before the separation chain runs, so the pool-inheritance skip is now counted there too (`pool/skipped_separations`, pure instrumentation) — before #1256 the structure never built for a shaped model, so that branch was unreachable.

https://claude.ai/code/session_01YYKn41XMUuevSPuVqADbBG


python/tests/test_row_filter_loose_bound.py
```diff
@@ -28,12 +46,28 @@ def _root_bound():
...
+    with_filter = _root_bound()
+    assert with_filter.status == "optimal"
+    assert with_filter.lower_bound is not None
+    assert with_filter.lower_bound <= _HDA_OPT, "the certified bound must stay sound"
+
+    monkeypatch.setenv("DISCOPT_RELAX_ROW_FILTER_LOOSE_BOUND", "0")
+    without = _root_bound()
+    assert without.lower_bound is not None
+    assert without.lower_bound <= _HDA_OPT, "the un-recovered bound must be sound too"
+    assert with_filter.lower_bound > 100.0 * without.lower_bound, (
```
jkitchin: **This assertion cannot fail. Given #671's own contract it is a tautology, so the guard `0432181d` created for this mechanism now gates nothing.**

Both bounds here are negative, and multiplying a negative by 100 moves it *away* from zero. With the numbers from your PR body, `100 * -9.99e11 = -9.99e13`, which the filtered `-1.38e9` clears by four orders of magnitude. The condition reduces to `|with| < 100*|without|` — it permits the filtered bound to be up to **100x looser** than the unfiltered one, the opposite of the "large tightening" the docstring says it asserts.

It is not merely loose, though; it is unfalsifiable. `_relax/mccormick_lp.py:2041-2042` keeps the larger of the two certified bounds:

```python
if _filtered_bound is not None and _filtered_bound > bound:
    bound = _filtered_bound
```

So `with >= without` holds by construction, and since `without < 0`, `100*without < without <= with` — always. No implementation of the filter can trip this arm.

Measured:

```
without = -9.99e+11   with_filter = -1.38e+09
with > 100*without                      -> True
filter makes it 10x LOOSER (-9.99e+12)  -> True   (still passes)
filter makes it 50x LOOSER (-4.99e+13)  -> True   (still passes)
filter makes it 99x LOOSER (-9.89e+13)  -> True   (still passes)

9954 random draws honoring "with >= without" plus the soundness cap:
  assertion failures observed: 0
executed=7
```

The fix is to divide rather than multiply:

```python
assert with_filter.lower_bound > without.lower_bound / 100.0, (
```

`without/100 = -9.99e9`, and your measured `-1.38e9` clears it with ~7x room (a 725x recovery against a required 100x). It correctly **rejects** all three degradation cases above.

To be clear about what I am not asking for: the relational reframing is the right call, and I am not suggesting you restore the `-1e5` constant. Pinning a raw-box absolute across an operator-lowering change is exactly the hostage situation your docstring describes, and your reasoning for dropping it is sound. The only problem is that the replacement, as written, is a no-op.

One scope note, take it or leave it. `>= -1e5` was also the only *bound-quality* floor in this file. The corrected relational form restores a tightening check but remains scale-free — it passes at any absolute looseness. If you want the quality half back, the natural anchor is the post-FBBT/OBBT number you already measured end to end (`-124296.879` at a 10 s limit), which is stable across this lowering change, rather than the raw-box root that is not.

Verified at `2c1c233`: the arithmetic above is executed, and the keep-the-larger-bound contract read in `mccormick_lp.py` at this SHA.

jkitchin: Confirmed and fixed in `2755abe`. I re-derived it rather than take it on faith, and every part holds: `mccormick_lp.py` keeps the larger of the two certified bounds, so `with >= without` by construction; with `without < 0` that makes `100*without < without <= with` unconditional. 10 000 random draws honouring that contract gave 0 failures, and the 10x / 50x / 99x-looser cases all passed the old form. The guard gated nothing.

Changed to the division you suggested, stated in the docstring as "the recovery must close at least 99 % of the un-recovered bound's distance to zero" so the direction is not re-derivable-by-accident next time, plus an explicit `without.lower_bound < 0` guard — that form only means what it says in the negative regime, and should fail loudly rather than vacuously if that ever changes.

I also proved the replacement is falsifiable instead of assuming it: running the real test body against a `_root_bound` whose filtered arm is degraded 10x / 50x / 99x trips the assertion in all three cases, with both arms confirmed solved. That harness monkeypatches a test function, so it was a one-off and is not shipped.

**On the scope note — I've left it out, and I'd rather say why than silently skip it.** The end-to-end number is stable *across this lowering* but it is a 10 s wall-clock-limited solve, and that is the one thing I would not want to pin in CI: my own panel put hda at -124296.879066 and -124296.879048 on two runs of the same quiet machine, but at 60 s the same build reaches -64510.17, so the constant is really "wherever the search happened to be when the clock ran out". On a shared runner under load that is a different node and a different number. It would reintroduce exactly the hostage situation, with the clock as the hostage-taker instead of the operator lowering.

If you want the quality half back, I think the durable anchor is `docs/dev/data/claim-baseline.jsonl` — it already carries hda's root LP bound as a reproducible reference (this PR updates that row), so a tightness floor there would be gated on a recorded number rather than on a timer, and would move deliberately with a regeneration. Happy to add that in this PR if you'd like it; say the word and I'll push it.

---

jkitchin: **Verified independently at `2755abee`** — I re-ran the arithmetic against the shipped form rather than accept the fix on its description, which is the same discipline that caught the original.

It now does what it says: it rejects the 10x / 50x / 99x-looser cases the multiplication accepted, rejects a merely-weak recovery all the way up to *exactly* 100x, accepts 200x, and the new `< 0.0` guard trips on a positive un-recovered bound. `executed=10`. Thank you for proving falsifiability by degrading the filtered arm rather than assuming it — that is the part I would have asked for.

**One inconsistency you did not catch, worth a follow-up rather than holding this PR.** The docstring gives the two measured lowerings as closing "99.9 % (88x) and 99.86 % (725x)" against a threshold stated as "at least 99 %". 88x closes **98.86 %**, not 99.9 % (`1 - 1/88.3 = 0.9887`), and 98.86 % is below the 99 % the assertion enforces — so the shipped form would **reject** the 88x lowering, which is the parent commit's. That does not affect this branch: it measures 725x with 7.25x of headroom. But it does mean the assertion is not quite the operator-lowering-independent pin the docstring claims, and the PR body's earlier "the new form passes on the parent commit too" is no longer true of the corrected form, only of the tautological one it replaced. Either restate the threshold (98 %, i.e. 50x, clears both lowerings) or drop 88x from the list of satisfying examples.

**On the tightness floor — your reasoning beats my suggestion and I withdraw that half.** Pinning the 10 s end-to-end constant would reintroduce the hostage situation with the wall clock as the captor instead of the operator lowering, and on a loaded shared runner that is a different node and a different number. You are right. The `claim-baseline.jsonl` anchor you propose is the right shape, but I would rather not widen this PR for it: it is a standalone improvement to the gate and deserves review on its own terms, against the question of what a tightness floor should be gated on, not as a rider here.

Merging on that basis. The hda root-bound loosening itself I am satisfied with: `claim-boundary` ran and passed, its root-LP gate asserts soundness and reproducibility rather than tightness, and regenerating the baseline in the PR that causes the change is what that gate's own docstring requires.

jkitchin: You're right on the inconsistency, and it was worse than a docstring slip — the shipped 99 % bar was wrong, not just mis-described. Reproduced before changing anything: 88x closes `1 - 1/88.29 = 98.87 %`, and `-64675 > -5710326/100 = -57103` is **false**, so `2755abe` would have rejected the pre-#1256 lowering. That is the hostage situation again with the threshold as the captor.

Fixed in `ce52cc8` by restating the bar at **98 % (50x)** rather than dropping 88x from the examples. Dropping it would leave a gate that only the current lowering satisfies, which is the failure mode the reframing exists to prevent; 50x clears both measured lowerings with headroom (88x: 1.8x, 725x: 14.5x). The factor is now a named constant carrying that reasoning, and the docstring states the 98-not-99 choice and why.

Measured rather than asserted, since that is exactly what I got wrong last round:

- the corrected test run against **the parent commit's source** (88x lowering, its own worktree at that SHA): **3 passed**; on this branch (725x): **3 passed**. So the "passes on the parent too" claim is true again — and this time of the form that actually ships.
- boundary behaviour, synthesising a k-fold recovery from the un-recovered arm: `1x rejected, 40x rejected, 50x exactly rejected (strict inequality), 60x accepted, 88x accepted`, 5/5 executed.

One process note, since it briefly looked like a product defect: my first boundary harness scaled the *with* arm by `1/k`, which makes the bound **tighter**, not weaker, so three cases failed for being mis-parameterised. The gate was fine; the harness was not. Deriving the degradation from the un-recovered arm is the correct construction.

I have also **retracted the false claim in the PR body** rather than quietly editing it: that section now records both corrections — the multiply-tautology and the 99 %-bar — and notes explicitly that "passes on the parent commit too" had been verified against the tautological form, not the corrected one.

On the tightness floor: agreed, and thank you for the clear call. It is a standalone question about what such a floor should be gated on, and it deserves its own review rather than riding here. I'll leave it out of this PR entirely.

CI is re-running on `ce52cc8`; the previous head was green across all 16 checks.

---

Issue #1249, opened 2026-09-15
Tracking: discopt features needed by the discopt-calphad plugin


jkitchin:
Tracking issue for discopt features needed by **discopt-calphad**, a new CALPHAD thermodynamic-equilibrium plugin (`discopt.calphad` namespace subpackage, sibling of discopt-mkm and discopt-thermo).

## Context

discopt-calphad computes phase equilibria with a decomposition, plus a certificate:

1. **Restricted equilibrium:** POUNCE solves a restricted multiphase Gibbs-energy minimization over the active composition sets. The duals of the elemental balances are the chemical potentials μ.
2. **Phase pricing:** for every phase, discopt's global solver minimizes the tangent-plane distance `G_phi(y) - mu^T A_phi(y)` over its site-fraction simplices.
3. **Certificate:** if every phase's valid global lower bound is ≥ -eps, the equilibrium is certified to be within eps (times total formula units) of the global minimum, without globally solving the coupled problem. A negative upper bound adds a composition set, which detects miscibility gaps.

The plugin keeps a symbolic, structure-preserving thermodynamic model (x ln x, Redlich–Kister, CEF endmembers), which is lowered to discopt expressions. Phase diagrams and parameter fitting are built on top.

## Requests — all resolved

Boxes re-checked 2026-09-17 against the linked issues' actual state (they had drifted).

**P0: needed for the first certified prototype**
- [x] #1242 `entropy` (x ln x): public `dm.xlogx`, exact interval at `lo = 0`, Rust FBBT, symbolic diff
- [x] #1243 `abs_gap_tolerance` solve option
- [x] #1244 Public `SolveResult.bound_valid` / `bound_source`, with documented semantics at limits

**P1: speed for tracing and fitting**
- [x] #1245 Reuse compiled relaxations across `Parameter` value changes — *closed on measurement*: the premise did not survive (the named compiler is not on the solve path; the rebuild is 0.4–3.8% of a solve, the soundly reusable share ~1%). The real cost was the tape evaluator's staleness check, fixed at 14.7×. Acceptance criteria 2 and 3 ship as permanent tests.
- [x] #1246 `dm.solve_batch` for many small global solves
- [x] #1247 Primal-dual warm start through `Model.solve`, plus a KKT residual on `SolveResult`

**P2: research enablers (one design)**
[...]

Issue #1248, opened 2026-09-15
Design: relaxation extension API (operators, simplex structure, multivariate hook, node cuts, SymPy bridge)


jkitchin:
Part of #1249 (plugin needs from discopt-calphad). Priority **P2**. This is one design issue because the pieces below share a single extension architecture. Split it into implementation issues once the design settles.

## Motivation

Domain plugins know structure that generic factorable relaxations throw away:

- **discopt-calphad:**
  - x ln x mixing entropy;
  - Redlich–Kister excess polynomials;
  - products of site fractions from different sublattices, each on a simplex (y ≥ 0, Σ y = 1);
  - supporting hyperplanes / tangent-plane cuts of Gibbs energy.
- **discopt-mkm / discopt-thermo:** Arrhenius, Langmuir and activity-model forms.

The research question for CALPHAD is how much CALPHAD-specific envelopes improve deterministic global optimization over generic McCormick. That can only be measured if plugins can plug relaxations into the solver.

## Current state (main @ a837ffb8)

- **Operator names are hard-coded** in `_relax/relaxation_compiler.py:983-1002`, `_relax/convexity/interval_eval.py:471`, `_nl_expr_compiler.py:1018`, and Rust `crates/discopt-core/src/expr.rs:50`.
- **The existing catalogue is unused:** `_relax/symbolic/registry.py` has `_REGISTRY` with `available/get/certify/certify_all`, but nothing outside tests reads it. The solver imports domain-pack closures directly.
- **Only one external route:** `dm.custom` with MCBox-traceable bodies gives a reduced-space certified solve, but no Rust FBBT and no `.nl` export (`core.py:1689-1735`). `dm.udf` only composes existing primitives.
- **Cuts only at integer-feasible nodes:** `lazy_constraints` runs there only (`callbacks.py:78,104`), and `NodeCallback` cannot add cuts.
- **No simplex awareness:** multilinear hull separation exists (`_relax/multilinear_separation.py`, `edge_concave.py`) but knows nothing about simplex domains, and there is no way to declare a simplex. `recognize_and_inject` (`solver.py:8062`) is a fixed pattern.
- **Envelope derivation is offline only:** `_relax/symbolic/envelope_deriver.derive_envelope` + `codegen` are design-time. The SymPy-free `_relax/symbolic/runtime.py` could run at solve time.
[...]

PR #1275, opened 2026-09-16
fix(relax): #1277 four native intrinsics had no envelope; #1248 A/D/E/F ship, C and G declined

### LINKED ISSUE #1276: Registered atoms over several arguments (#1248 component C)
Split from #1248 now that component A has shipped (PR #1275) and settled the mechanism.

## What A established

`dm.register_function(name, lower)` names a **univariate** composite. The model carries the lowering — an ordinary primitive expression tagged with the atom's name — so evaluation, `.nl` export, the Rust core and presolve need no new opcode; `canonical_expr` turns the tag into a named `call` node and the relaxer envelopes the atom whole. Everything the relaxer uses is derived from the lowering: `f`/`f'` through the solver's own tape, and the per-box curvature verdict as an **interval enclosure of `f''`**, which is a proof on that box rather than a user assertion.

Measured payoff on the CALPHAD Redlich-Kister binary: 7559 → 35 nodes (216×), 2655 → 47, 1741 → 15, 1729 → 25, 2019 → 19, same optimum in every row.

## What this issue adds

The same mechanism for an atom of **n arguments** — the shape the plugin actually needs for a two-sublattice CEF model, where a site-fraction product spans sublattices.

The pieces that generalise directly:

* the tag already carries `(name, arg)`; it becomes `(name, args)`;
* `canonical_expr._canon_call` already emits a `callN` node for ≥ 2 arguments;
* the curvature derivation becomes a **Hessian** enclosure rather than a scalar `f''` — `convexity/interval_ad.interval_hessian` already computes exactly that, and the existing Gershgorin sign certificate is what turns it into a convex/concave verdict;
* the emitter is the composite-convex lift (`uniform_relax._try_convex_lift`), which already takes a certified-curvature multivariate node, rather than `_emit_1d`.

So the work is: extend the registry to n-ary lowerings, derive the Hessian enclosure from the lowering, and route a certified atom into the existing lift instead of the univariate envelope.

## Entry experiment first (CLAUDE.md §4)

A's payoff does not transfer automatically: the univariate win came from enveloping a composite whose *terms* were being relaxed separately, and the multivariate lift may already be capturing much of that for a CEF product. Before implementing, measure on a real two-sublattice pricing instance:

* root bound and node count for the primitive spelling, versus
* the same with the composite certified convex/concave over the node box.

Kill criterion: if the lift already fires on the primitive spelling and the bounds match, the n-ary registration buys nothing on that class and this should be closed rather than built.

## Acceptance

* A two-argument registered atom solves to the same optimum as its primitive spelling, with a measured node-count reduction on a stated instance.
* The derived Hessian verdict abstains (rather than guessing) on a box where the enclosure straddles, pinned by a test.
* A panel over the parameter family shows no bound above a dense-sampled truth.

jkitchin:
## Summary

The #1248 relaxation-extension components reachable from #1249, plus the defect one of them uncovered. Closes #1276, #1277, #1278; contributes to #1248 and #1249.

Three of the five components ship, two are **declined on measurement**, and the largest result in the PR is not any of them — it is a relaxation gap found while investigating why #1277's premise appeared to hold.

**On scope** (CLAUDE.md says one task per PR): this spans four issues because they turned out to be one measurement. #1277's investigation found the missing-envelope defect; fixing it invalidated the baseline #1248 A was measured against, and it is the corrected baseline that kills #1276 and #1277 B. Splitting now would mean landing a retraction before the measurement that justifies it. Say the word and I'll split it anyway.

| commit | |
|---|---|
| `83cce20` | **#1248 A** — `dm.register_function`, a named composite the relaxer envelopes whole |
| `c59d05a` | **#1277** — four native intrinsics reached the relaxer with **no envelope at all** |
| `6dea494` | retraction of A's published 216× claim, which was that missing envelope |
| `0d4ba36` | perf-plan §66 — #1277's `m.simplex()` and `branch_priority` both declined |
| `466bb06` | perf-plan §67 — #1276's n-ary atoms declined |
| `81fd92a` | **#1278 E** public SymPy bridge, **F** bound provenance |
| `1392f5a` | **#1278 D** — cut callback at every node, with a mandatory validation gate |

### The load-bearing fix (`c59d05a`)

> `entropy` (`dm.xlogx`), `softplus`, `sigmoid` and `tan` are first-class discopt intrinsics — `canonical_expr` emits a `call` node for each — but none had an entry in `uniform_relax._UNIVARIATE_FN`, so `_build_univariate_call` fell through to `Envelope(rows=[], tight=False)`: the aux interval floor, a box with no secant and no tangent.

The floor decouples `w` from `t`, so the LP may take every such term's independent minimum at once. It is *sound*, which is why this hid — and on a bare univariate objective it is even **exact** (the aux *is* the objective, so its enclosure is the answer; measured 0.00% root gap). The cost appears only when the atom sits beside a term `_try_convex_lift` cannot certify.

Measured root gap, floor → envelope (`scripts/entry_1277_univariate_floor.py`, 18 arms, both directions interleaved in one process):

| case | floor | envelope |
|---|---:|---:|
| CALPHAD compound-energy (the #1249 model) | 28.84% | **5.45%** |
| coupled sigmoid | 14.71% | **0.00%** |
| coupled tan | 84.75% | **22.37%** |
| coupled softplus | 5.75% | **2.41%** |
| coupled entropy | 73.57% | 69.31% |
| the four bare univariate objectives | 0.00% | 0.00% |

No new mechanism: each entry is the exact secant/tangent envelope `_emit_1d` already emits for `exp`/`log`/`tanh`. `tan` needed more — `tan''` has the sign of `tan` only *within* a branch, and `domain_ok` only sees `lo`, so the two-sided branch-containment guard lives in `_curv_tan`, which abstains on any box crossing or touching a pole. That also retires the separate `tan` line in `_PIECEWISE_UNIVARIATE_FN`, whose `_curv_by_sign(True)` claimed a definite curvature on pole-straddling pieces where the endpoint chord is not a chord of the graph.

### Two retractions (CLAUDE.md §11)

**1. This PR's own headline.** The body previously led with "7559 → 35 nodes, 216×" for `register_function`. That measured a primitive arm in which every `dm.xlogx` term hit the interval floor. Against the corrected baseline (20 interleaved solves, same optimum in every one):

| L0 | L1 | RT | primitive | registered | ratio | was published |
|---:|---:|---:|---:|---:|---:|---:|
| 3.0 | 0.0 | 1.0 | 51 | **35** | 1.46× | 216× |
| 5.0 | 0.0 | 1.0 | 51 | **47** | 1.09× | 56.5× |
| 3.0 | 1.5 | 1.0 | 23 | **15** | 1.53× | 116× |
| 8.0 | −4.0 | 1.0 | 29 | **25** | 1.16× | 69.2× |
| 20000 | 5000 | 8314 | 29 | **19** | 1.53× | 106× |

A still earns its keep — a strict reduction in every row at the same optimum, from a mechanism general to any user composite — but the honest figure is **1.1×–1.5×**. `operators.py` and the test docstring carry the corrected table; that test's `*5` bar is now a strict-reduction bar.

**2. #1277's premise.** I implemented the simplex marginals and measured 28.8% → 5.65%, with no lift-only control. With the control (truth −10315.85): no-lift **28.84%**, **lift-only 4.72%**, lift+marginals **5.65%**. The marginals are neutral-to-harmful; the gain was the hand-lifting. `m.simplex()` is not implemented.

### Declined on measurement

- **#1277 G** (`branch_priority`): the oracle over declaration orders — what the default selector ties on at the root — is **1.11×** on a 9-var/27-trilinear CEF and **1.00×** on two pooling models. A per-node two-level override (the engine's own `set_branch_deprioritized`, which is what G generalises) is **2×–5× worse** at an equal node budget: gap 8.10% default vs 16.04% / 29.71% / 32.51% / 41.29%.
- **#1276 C** (n-ary atoms): the premise is right — `_sum` flattens, so registration is the only way to create the grouping, and the lift is genuinely lost and recovered. It is worth nothing: on the CALPHAD class the lift never certifies at **any** box width (0/12 at each of six widths, while the convex control fires 12/12 everywhere — bilinear cross terms have Hessian `[[0,c],[c,0]]`, eigenvalues ±|c| on every box); where a composite does certify its terms already carry tight envelopes (0.00% gap both ways); and on log-sum-exp, the best case for it, grouping is worth 0.76 of a 107-point gap.
- **#1278 F's strict mode**: outward rounding and the NS safe bound are already default-ON and already what `_certify` prefers. The provenance tag it was missing now exists, and shows **505/507** corpus node bounds rest on an NS certificate, 2 declined, **0 on trust** — so a `verified_bound=True` refusing the trusted arms would have nothing to refuse.

Both §66 and §67 also record the **measurement traps** that produced wrong intermediate results, since those are the reusable part: a time-limited node count measures throughput rather than work (tell: `status == "feasible"` in every arm), and probing one half of a two-step decision measures nothing.

## Correctness

Two parts touch the certificate, in opposite directions.

**The envelopes (`c59d05a`) tighten**, so a wrong one is a false bound. Each curvature verdict is a closed-form sign argument on `f''` stated at its definition, and the primary evidence is **row-level feasible-point sampling**: for every operator, over boxes that are single-curvature, straddling, very wide, near-degenerate and (for `tan`) pole-crossing and pole-touching, the lifted true point `(t, f(t))` satisfies **every** emitted row — >500 row checks per operator, counted. Plus a differential test (new bound ≥ old AND ≤ dense-sampled truth, control arm popping the entries back out in-process), and the verdicts checked against an independently written closed form of `f''`, itself checked against a central difference on well-conditioned boxes.

**The cut callback (`1392f5a`) accepts user assertions**, which is the opposite problem — A's envelopes are *derived*, so there is nothing to be unsound. `_cut_pool` is applied at every node through `_AugmentedEvaluator`, so every accepted cut is global and an invalid one removes the optimum from the whole tree. Two guards, neither optional: `CutResult(scope="local")` is **refused at construction** (there is no subtree-scoped pool; silently promoting it *is* the false certificate), and every cut is checked against points this solve has already verified feasible, with a violator **raising** rather than being dropped. The gate is a filter, not a proof, and says so; `solver_stats` reports how many points it actually tested so a vacuous gate cannot read as a pass.

- [x] `pytest -m smoke python/tests/` → **1724 passed, 7 skipped, 0 failed** (10:01)
- [x] `pytest -m slow python/tests/test_adversarial_recent_fixes.py` → **19 passed** (2:49)
- [x] `ruff check` / `ruff format --check` clean; `mypy python/discopt/` clean (350 files)

## Regression test

- `test_1277_univariate_envelopes.py` (28) — soundness first: no emitted row cuts the lifted true point; the aux column still encloses `f`; curvature verdicts vs a closed-form `f''`; `tan` abstains on every pole-touching box and keeps an unbounded floor there; the differential bound test; and an assertion of **why the corpus panel is bound-neutral** (its only call names are `log`/`sqrt`/`exp`).
- `test_1278_sympy_bridge.py` (85) — #1248's acceptance: a round trip over the elementary set, `entropy` included, matching **values and gradients** at random points (23 intrinsics × 40 value points and × 25 gradient points, gradients against a central difference of an independent numpy reference), plus the refusals.
- `test_1278_cut_callback.py` (12) — the callback fires at spatial nodes with strict subboxes of the root; an accepted cut reaches the pool with the coefficients returned; a cut excluding a known feasible point aborts the solve, including when returned late; `local` refused; an inert callback leaves status, objective and **node count** byte-identical.
- `test_1278_bound_provenance.py` (7, one `slow`) — every certified node reports which argument certified it, and the corpus certifies from proof rather than trust.
- `test_1248_register_function.py` — bar corrected from `*5` to a strict reduction.

`c59d05a` is bound-changing (CLAUDE.md §5). It ships **default-ON**, deliberately, and here is the reasoning rather than a flag:

- [x] Differential-bound test (new bound ≥ old AND ≤ true box optimum) + feasible-point sampling — both above, per operator.
- [x] Corpus differential panel: **64/66 byte-identical** in status/objective/bound/node_count. The 2 that differ (`nvs05`, `tls2`) are budget nondeterminism, not this change — a same-arm control over 16 runs has **3 of 4 arms disagreeing with themselves** (`tls2`: 4 distinct outcomes in 4 identical pre-change runs).
- [x] Not graduating a flag: the panel is bound-neutral **by construction**, and a test asserts why — `.nl` has no opcode for `entropy`/`softplus`/`sigmoid`, and no corpus instance uses `tan`. A default-off flag would therefore be gated on a panel that cannot move, i.e. a dead flag (CLAUDE.md §3). The soundness evidence is the row-level feasible-point test, not the corpus.

The wider MINLPLib snapshot is not reachable from this container, so the panel is the 66 in-repo instances. Worth a nightly look on the full corpus for `tan`, the one of the four with an `.nl` opcode.

- Measurement: `scripts/entry_1277_univariate_floor.py` (18 arms, both directions, executed-assertion count, non-zero exit if an arm measures nothing), `scripts/entry_1277_branch_priority.py`, `scripts/entry_1276_nary_atoms.py` (296 probes). Numbers in the tables above; falsifications in `docs/dev/performance-plan.md` §66–§67.

https://claude.ai/code/session_01RKHEogcdVoD2gHPrGTuidB


jkitchin:
## Review

Reviewed the soundness-critical parts: the four new `_UNIVARIATE_FN` envelopes, `_curv_tan`, the registered-operator path, and the cut-callback gate.

**`_curv_tan` pole guard: verified.** A probe tried 399,996 floats on the wrong side of a pole, within 3 ulp of `(k±½)π` for |k| ≤ 200000 (true poles computed with mpmath). The guard admitted none, so `math.pi` rounding cannot sneak a pole into an admitted box.

**Soundness defect found and fixed (`4725635b`).** Canonicalization looked a registered atom up by *name*, but the model keeps the body from the registration that tagged it. After `register_function(name, ..., replace=True)`, or clear + re-register, a model built earlier was enveloped with the *new* function. Measured: after replacing `exp(x)+x²` with `exp(x)+x²+100`, the emitted rows cut the true point `(t, exp(t)+t²)` by **99.99**, which can produce a false bound. The fix: the tag now carries the `RegisteredFunction` object, and `atom_of` ignores a tag whose registration has since been replaced. That model is then relaxed term by term from the body it actually holds.

- The new regression test `test_a_replaced_registration_never_envelopes_a_model_built_before_it` **fails** on the unfixed code and **passes** with the fix. Its control arm checks that a model built after the replace still gets the atom envelope.
- The PR's five test files: **150 passed**, run from a worktree after asserting `discopt.__file__` pointed there.
- ruff check and ruff format are clean. Full mypy reports the same 6 errors with and without this change, all in unrelated files and caused by the local numpy version; CI lint passed.

A non-blocking note on the cut gate: a cut returned before any witness exists is accepted unchecked (`cut_validation/unvalidated`). Once it is in the pool, the points it excludes can never become witnesses. The docstring's "filter, not a proof" covers this, and `solver_stats` reports it.

jkitchin:
Confirmed, and thank you — the finding is real and it was my bug. I reproduced it independently before accepting the diagnosis rather than taking the report at face value.

**Reproduction on the pre-fix code (`1392f5a`).** Registered `swap = exp(t) + t²`, built a model, then `replace=True`'d the same name with `exp(t) + t² + 100`. The model keeps the old body; the relaxer enveloped it with the new definition. Checking the emitted rows against the true points of the body the model *actually holds*:

```
tag arity: 2 -> DEFECT (name only)
after replace, atom_of(body) = ('swap', x)
checked=401  worst violation of the TRUE point = 100
VERDICT: ROWS CUT THE TRUE POINT (false bound possible)
```

Violated by exactly the shift — a false-bound generator. The identity check (`_REGISTRY.get(name) is not tag[2]`) is the right fix: it degrades to term-by-term relaxation of the body the model carries, which is sound. End-to-end after the fix, all three scenarios give the correct optimum and a valid bound:

| scenario | atom live | objective | bound ≤ truth |
|---|---|---|---|
| no replace | True | 1.898721 | ✅ |
| replaced | False | 1.898721 | ✅ |
| cleared | False | 1.898721 | ✅ |

My test file had `test_a_dropped_registration_degrades_to_term_by_term` for the *clear* case and `test_duplicate_registration_needs_replace` for the registry, but nothing for a model built **before** a replace. That gap was mine.

### The cut-gate note — acted on, in `980e76d`

You were right that the first cuts are accepted unchecked. Every newly recorded witness is now re-judged against every cut already accepted, so such a cut can still be convicted by a witness found elsewhere. `cut_validation/retro_checks` reports the re-judgements.

It narrows the hole rather than closing it, exactly as you framed it: a cut excluding the optimum *and nothing else* may still go unconvicted, which is why `cut_validation/unvalidated` is still reported.

**Two things I measured rather than assumed**, both worth recording:

1. **How narrow the hole is.** Across two models and 23 validations, **zero** cuts were judged against an empty witness set — the incumbent is recorded before the first cut is validated. The unchecked-first-cut case needs a model whose first batch yields neither an incumbent nor a feasible relaxation solution.

2. **My first regression test for it was worthless.** It drove the solve and returned the bad cut only at invocation 1 — and passed *with and without* the re-check, because a witness already existed by then. A test that cannot fail is not a test (CLAUDE.md §6). Replaced by a direct test of `_recheck_accepted_cuts` (fails before with `ImportError`, passes after), plus one pinning the measured fact that the gate is not vacuous on an ordinary solve, so a change that stopped recording the incumbent as a witness would widen the hole visibly.

`pytest python/tests/test_1278_cut_callback.py python/tests/test_1248_register_function.py` → **33 passed**; ruff / ruff format / mypy clean.

---

Issue #196, opened 2026-06-17
ex1252 end-to-end gap closure: root cost, node throughput, and SOS1 selectors (follow-up to #184)


jkitchin:
Follow-up to #184 (PR #190). #190 landed the three relaxation/search mechanisms that move `ex1252`'s bound in the right direction, but the **global** dual bound still does not lift off 0 within a practical budget. This issue tracks the remaining work, which is now **solver-performance**, not relaxation logic.

## What #190 established (the foundation is in place)

Three sound, gated/tested mechanisms (all on PR #190):
- **Per-node lifted-LP FBBT** (`DISCOPT_LIFTED_FBBT=1`): propagates the McCormick relaxation's own rows to recover bilinear-implied factor bounds, lifting a gating-fixed node bound **0 → ~18987** (sound).
- **Objective-gating priority branching** (`DISCOPT_OBJ_BRANCH_PRIORITY=1`): branches the binaries that gate the objective products first (detector finds exactly `ex1252`'s `{x36,x37,x38}`).
- **Lifted-relaxation LP equilibration** (auto, on ill-conditioned boxes): geometric-mean scaling so HiGHS stops stalling on the >1e15-spread boundary LPs (`x36=1,x37=1` box: **time-out → 4.5s**).

## Why the global bound still doesn't close

The global bound is the **minimum over the open frontier**, so it stays at 0 until *every* shallow node has its objective products forced positive. The final end-to-end run (both gates on, equilibration active, 120s) gave:

```
t= 64.5s nodes=  3 GLOBAL_bound=0.000
t= 97.1s nodes= 15 GLOBAL_bound=0.000
t=121.2s nodes= 31 global=-inf   (a node returned no bound → gap decertified)
FINAL status=time_limit bound≈0 nodes=31
```

Throughput improved ~4× vs. before equilibration (3 nodes/50s → 31 nodes/120s), but the bound is still pinned at 0. The diagnosed blockers:

### 1. Root processing is ~64s before the tree branches at all
The first node callback fires at **t=64.5s**. The root spends that budget on NLP multistart + OBBT + the McCormick **root probe with the full 8+6-round cut separation**. On a hard root this eats most of a short budget. → Bound/cap the root bound-tightening work, or make the McCormick root probe cheaper (it only needs to learn "is there a usable bound", not solve to optimality).

### 2. Node throughput is still ~0.26 nodes/s
[...]

PR #1280, opened 2026-09-17
fix(correctness): #1236 four cut-separator term-deletions, a probe that solved a different problem, and a lift gate falsified by its own metric

### LINKED ISSUE #1236: perf(minlp): diagnose the integer-heavy node-count tail (nvs02/nvs14/nvs13/tls2/fac2, 24-297x SCIP)
## Summary

On the `global50` panel discopt proves more instances than SCIP overall (47/50
vs 44/50, node ratio median 1.00x), but a handful of **integer-heavy** instances
cost 20–300x SCIP's nodes. This issue records the reproduction and the
diagnostic plan for that tail.

The working hypothesis going in was "discopt's missing MILP cuts leak into the
MINLP path." **The reproduction below does not support that as a single
explanation** — the five instances take three different code paths, and the two
worst ones already have a nearly-closed root bound.

## Reproduction (deterministic, ~55 s total)

Node counts reproduce the panel exactly, so this needs no benchmark harness.

```bash
cd <worktree on main>
PYTHONPATH=$PWD/python python -u <<'EOF'
import os, time, logging
logging.basicConfig(level=logging.INFO, format="%(name)s %(levelname)s %(message)s")
import discopt, discopt._rust
root = os.path.abspath("python") + "/"
assert discopt.__file__.startswith(root), discopt.__file__          # CLAUDE.md §8
assert discopt._rust.__file__.startswith(root), discopt._rust.__file__
B = os.path.expanduser("~/Dropbox/projects/discopt-minlp-benchmark/minlplib/nl")
ran = 0
for inst in ["nvs02", "nvs14", "nvs13", "fac2", "tls2"]:
    logging.getLogger("probe").info("=== BEGIN %s ===", inst)
    m = from_nl(f"{B}/{inst}.nl"); t = time.perf_counter()
    r = m.solve(time_limit=60); ran += 1
    print(f"{inst:6s} nodes={r.node_count} root_bound={r.root_bound} "
          f"root_gap={r.root_gap} bound={r.bound} wall={time.perf_counter()-t:.2f}")
    print("   stats:", dict(sorted((getattr(r, "solver_stats", None) or {}).items())))
print("RUNS_EXECUTED", ran); assert ran == 5
EOF
```

**The `assert` on `discopt.__file__` is not optional.** The first attempt at this
run silently imported `discopt` from an unrelated worktree (`wt2122`) because the
editable install points elsewhere; without the assertion it would have measured
the wrong tree.

## Measured, on `main` @ `0587f527`

| inst | discopt nodes | SCIP nodes | ratio | root_gap | path taken |
|---|---:|---:|---:|---:|---|
| nvs02 | 297 | 1 | 297x | **0.40 %** | MILP driver (`lp/driver_nodes=297`) |
| nvs14 | 273 | 1 | 273x | **0.59 %** | MILP driver (`lp/driver_nodes=273`) |
| nvs13 | 637 | 27 | 24x | `None` | native spatial kernel (#764) |
| tls2 | 399 | 15 | 27x | 47.1 % | convex-MINLP auto-route → fallback |
| fac2 | 39 | 1 | 39x | 23.1 % | convex-MINLP auto-route → fallback |

All five return `optimal` with the correct objective (oracle from
`minlplib.solu`), so this is a **work** problem, not a correctness problem.

Wall times are **not** reported: load moved 3.80 → 7.99 during the run
(CLAUDE.md §9). Node counts are deterministic and match the earlier panel
exactly, so they are the metric to use here.

## Why "missing cuts" is not obviously the answer

A missing-cuts story predicts a **weak root bound**. On the two worst instances
the root bound is already within 0.4 % and 0.6 % of the optimum, and discopt
still spends 297 and 273 nodes closing it. Cuts that tighten a bound which is
already 99.6 % closed are not the obvious lever.

tls2 (47 % root gap) and fac2 (23 %) are the two where a bound-strength story is
plausible — and they are on a *different* path from the nvs pair.

### Lead 1 — nvs02 / nvs14: 297 nodes to close 0.4 %

These run on the **MILP driver** (`lp/driver_nodes` equals `node_count`
exactly). 5553 and 5298 LP iterations respectively.

Candidate causes: the termination gap test never fires despite the bound being
closed; integer bound propagation not rounding on the nonlinear integer columns;
branching repeatedly choosing a variable that does not tighten the bound.

*Diagnose*: dump per-node `(depth, branch var, bound)` and find where the bound
stops improving. If the bound reaches the optimum early and the tree keeps
expanding, it is a **termination/propagation** bug, not cuts.
*Kill criterion*: if the bound improves monotonically across most of the 297
nodes, the termination hypothesis is dead and bound strength is back in play.

### Lead 2 — nvs13: the native spatial kernel, no stats at all

nvs13 takes the `native spatial kernel (#764)` path (log line: `exited optimal:
obj=-585.2 bound=-585.2 nodes=637`), reports `root_bound=None`,
`root_gap=None`, and an **empty `solver_stats`**.

*Diagnose*: this path is currently unmeasurable from Python. Surfacing
`root_bound`/`root_gap` and node counters from the spatial kernel is a
prerequisite for diagnosing it at all.
*Kill criterion*: none yet — this is instrumentation work before it is a
performance question.

### Lead 3 — tls2 / fac2: the convex-MINLP auto-route gives up, repeatedly

Both spend budget in the convex-MINLP auto-route, fail to certify, and hand back
to the fallback — tls2 does this **five times** in one solve (0.40 s, 0.13 s,
0.44 s, 0.73 s, 0.74 s), each time "falling back to the default path".

*Diagnose*: measure how much of the 38.8 s (tls2) and 13.8 s (fac2) is spent in
routes that certify nothing, and whether the route's decision point (#1143,
"stopped at 6.0s of 60.0s") is mis-tuned for instances this small.
*Kill criterion*: if the abandoned routes account for <10 % of wall, this is
noise and the nodes are the real story.

## Correctness signal seen while reproducing (may deserve its own issue)

Both tls2 and fac2 emit:

```
discopt.solver WARNING #1059 route objective 5.3 appears to beat the fallback's
CERTIFIED 5.3; keeping the certified result. This should be impossible and
indicates a bound or feasibility inconsistency worth investigating.
```

The guard does the right thing (keeps the certified result), and the final
answers are correct. But the warning says the condition should be impossible,
and it fires on 2 of 5 instances here. Per CLAUDE.md §1 that outranks any of the
performance work above.

## A metadata trap for whoever picks this up

`problem_sizes.csv` reports `integer=0, discrete=0` for nvs02, nvs14 and nvs13.
That is misleading — it appears to count only *linear* integer variables. The
`.nl` headers show the opposite:

| inst | vars | cons | `.nl` discrete line (`nbv niv nlvbi nlvci nlvoi`) | reading |
|---|---:|---:|---|---|
| nvs02 | 8 | 3 | `0 0 3 2 0` | 5 integers, all in nonlinear terms |
| nvs14 | 8 | 3 | `0 0 3 2 0` | 5 integers, all in nonlinear terms |
| nvs13 | 5 | 5 | `0 0 5 0 0` | all 5 vars integer and nonlinear |
| tls2 | 37 | 24 | `31 0 0 2 0` | 31 linear binaries + 2 nonlinear |
| fac2 | 66 | 33 | `12 0 0 0 0` | 12 linear binaries |

So the nvs family is **integers inside nonlinear terms**, not a linear MILP core.
Any framing of these as "small MILPs with a nonlinear wrapper" is wrong, and a
fix aimed at linear cut generation would not touch them.

## Related

- #1215 — model construction cost.
- `solver.py:11199-11204` already names "the nvs* gap" and describes routing
  nonlinear-term integers to the LP-form McCormick relaxation with spatial
  integer branching; worth confirming that path is actually taken here.
- CLAUDE.md §2: the fix must target the class, not these five names. They are
  gate probes.

jkitchin:
## Summary

#1236 asked for a diagnosis of an integer-heavy node-count tail. Reproducing its `fac2` row surfaced something bigger underneath: **the in-house MILP driver returned a certified false `optimal`**, 7839 above a point feasible for the very problem it was handed. Root cause was two cut separators deleting terms from the *left* side of a `≥` cut — a strengthening, not a relaxation — which is how a cut can exclude feasible points.

This PR fixes both separators, ships the instrumentation that made the tail diagnosable, and closes the leads the issue opened. A code review then found two more term-deletions of the same shape, and falsified this PR's own performance claim — both worked through below.

`Closes #1236`

**Commits**

| commit | what |
|---|---|
| `c40555a` | Lead 2 instrumentation — kernel `root_bound`/`root_gap`/node counters + MILP dive funnel |
| `9537862` | **§1** — GMI separator deleted terms from a cut's LHS |
| `dca5d3c` | **§1** — same defect in `cover.rs`; `mir.rs` and `syn05hfsg` cleared as false positives |
| `c1d516a` | Item E — determinism contract pinned |
| `9e9537f` | Item B — cheap-first lift gate (graduated default-ON; **later reverted**, below) |
| `39a6e73`, `3e99f77` | CI fixes: mypy annotation; the #1153 unbounded role-2 carve |
| `9dd0982` | **Review round** — findings 1–3, 6–10: two more term-deletions, probe forwarded 2 of 47 options |
| `5a7220e` | Typecheck error the local gate caught before CI |
| `f53f5dd` | **Item B falsified** — node count was the wrong metric; flag defaulted OFF |

## Correctness

Both original defects are the same shape. A GMI cut is `Σ ψ_j x̃_j ≥ 1` with `ψ_j ≥ 0`, `x̃_j ≥ 0`; a cover cut needs every non-cover column to contribute nonnegatively. Removing such a term without charging its extreme over the box **strengthens** the cut.

- **`gomory.rs`** snapped `ā_j` to the nearest integer (zeroing a *continuous* column's `ψ`) and skipped `|ψ| ≤ tol`, charging nothing to the rhs. Harmless while `x̃` is O(1) — every binary. On the `fac2` master the dropped terms sat on continuous slacks with `u = 1e20` and `x̃ ≈ 7e7`/`2.3e8`, so ~1e-10 coefficients were worth **0.47 and 0.025 against a rhs of 1**. Root cut #1 is violated at the witness by −0.0174.
- **`cover.rs`** skipped `|w| ≤ tol` *before* the binary/sign/bounds validation. On `x0 + x1 − 1e-10·y + s = 1.5` with `y ∈ [0, 1e20]`, the point `(1, 1, 5e9)` satisfies the row and the emitted cut `x0 + x1 ≤ 1` excludes it.

Both now charge the term's extreme over its box to the rhs (a relaxation) and **refuse the cut when that range is unbounded**. The test is on the *bound* against the `1e20` sentinel, never on a product — `1e-10 × 1e20` is an ordinary finite number, which is precisely how this layer has produced a false certified `optimal` before.

### Two more of the same class, found by review

Both are guards that stopped seeing something — the same failure mode as the original defect, reproduced *inside* its fix:

- **An all-tiny cut became emittable.** The `max_c == 0.0` arm used to refuse a cut whose coefficients were all ~1e-12, because every tiny term was *skipped*. Keeping them (soundness requires it) while excluding them from `min_c` left `min_c = INFINITY`, and `max_c / INFINITY == 0` passes the dynamism gate while `max_c = 1e-12 ≠ 0` passes the emptiness gate. Not a weak-cut problem — a **false-prune** problem: `Σ 1e-12 x̃ ≥ 1` reads as `0 ≥ 1` at a 1e-9 LP feasibility tolerance, so the node is fathomed as infeasible although it contains feasible points, feeding a certified bound. The driver is shielded by `substitute_slacks_to_structural`; the convex kernel's `substitute_slacks` drops only exact zeros and would have passed it to the node LP. Fixed by refusing when `!min_c.is_finite()`.
- **`snap_eligible` read on the branch it does not describe.** An integer column pinned at a *fractional* bound falls back to the continuous formula, where an integral `ā_j` gives `ψ = |α|/f₀ ≥ 1` — in the thousands, not tiny. Flagging it `tiny` excluded it from `min_c`, so the dynamism gate stopped seeing the smallest coefficient. Measured: true spread `2e6/2 = 1e6`, gate saw `1.0`, and `max_dynamism = 1e3` waved it through.

- [x] Cannot produce a false certificate — every fix only ever *weakens or refuses* a cut

Cross-solver evidence on the captured master — before, the in-house engine was alone in being wrong:

| backend | before | after |
|---|---|---|
| in-house simplex | **optimal @ 331845337.44** (false) | optimal @ 331837498.18 |
| POUNCE | optimal @ 331837498.50 | unchanged |
| HiGHS | optimal @ 331837498.18 | unchanged |

`fac2` end-to-end now matches the MINLPLib reference to **7e-11**, certifying in **1.4 s instead of 20.2 s** (39 → 0 nodes).

## The cheap-first probe solved a different problem

`_ipx_unlifted_probe`'s result is **returned as the final answer** when it certifies, and it shipped forwarding only `gap_tolerance` and `max_nodes` of `solve_model`'s 47 parameters. With `lazy_constraints` set it solves a **relaxation of the user's problem** and returns that as a certified optimum.

Fixed two ways, because a list of forwarded names is only as good as its last edit:

1. Options it cannot honour faithfully (`lazy_constraints`, both callbacks, `cut_callback`, `initial_point`, `warm_start`, `decomposition_structure`) **decline the probe outright** — the pre-#1236 behaviour, never a wrong answer, mirroring `_sub_blocked`.
2. Everything else is forwarded explicitly, and `test_the_probe_forwards_every_option_that_changes_the_answer` enumerates `solve_model`'s **live signature**, failing on any parameter that is not forwarded, blocked, or exempt *with a stated reason*. Verified non-vacuous by deleting one forwarded option and watching it name it.

Also fixed: `max_nodes` was spent twice (now charged once); the re-entrancy guard was a module global, so with two concurrent solves the formulation choice could depend on an unrelated one (now a `ContextVar`).

## Tests run

- [x] `pytest -m smoke` → **1653 passed, 15 skipped, 0 failed**
- [x] `cargo test -p discopt-core` → **763 passed across 15 binaries, 0 failures**
- [x] `ruff check` / `ruff format --check` clean; **pinned mypy 2.1.0** clean (350 files); `cargo fmt` clean; no new clippy warnings

## Regression tests

All fail before / pass after:

- `gomory.rs::gmi_tiny_coefficient_on_wide_continuous_column_is_not_dropped` — `cut excludes feasible integer point [0, 5e9, 0]`
- `cover.rs::tiny_negative_weight_on_a_wide_column_is_not_skipped` — `cover cut excludes the feasible point [1, 1, 5e9]`
- `gomory.rs::a_cut_whose_every_coefficient_is_tiny_is_refused` — asserts the *spec* (whatever is emitted must be enforceable at LP tolerance), not a cut count
- `gomory.rs::an_integer_column_pinned_at_a_fractional_bound_is_not_tiny` — pre-fix emits `[0, 2, 2e6]`, a 1e6 spread through a 1e3 gate
- `test_1236_gmi_drops_no_term.py` on a checked-in fixture `fac2_master0.npz` carrying a witness, so the defect is assertable without an external oracle

Tolerances against that fixture must be **absolute**: `|z*|max` is 3.3e8, so a 1e-9-*relative* gate is a slack of 0.86 — larger than the 0.017 violation. That is documented in the fixture README because it is how I missed this once — and the tests then broke that very rule (review finding 9): `1e-6 * |attained|` is **331.8**, a 24× margin on a 7839 defect. Now measured absolute slacks (1.0 driver, 10.0 cross-backend, from the worst legitimate excess of +0.32), **detection margin 24× → ~800×**.

**The separator fixes** are correctness, not performance, and ship without a flag. Differential panel, 66 in-repo instances, 20 s, both arms marker-verified:

| | GMI fix | + cover fix |
|---|---|---|
| certificate violations (sense-aware) | 0 / 0 | 0 / 0 |
| certification lost / gained | 0 / 0 | 0 / 0 |
| objective drift > 1e-6 | 0 | 0 |
| total nodes | 4055 → 4072 (+0.4 %) | 4072 → 4208 (+3.3 %) |

The +3.3 % is the price of not emitting invalid cover cuts.

A separate review concern — that keeping unbounded tiny terms makes cuts dense — is **falsified by measurement**. New counters (`SepGomoryCutsEmitted`, `SepGomoryCutNnz`), both arms on the `fac2` master (485 columns), arm identity asserted *behaviourally* (pre-fix returns the known false optimal, fixed returns the true one, so a wrong `.so` fails before measuring): mean nnz/cut **35.3 → 34.5**, i.e. 7.3 % → 7.1 % of available columns. Marginally *sparser*. Most tiny-ψ columns on a big-M master are bounded binaries, still charged to the rhs and dropped, and the driver refuses a cut whose pin is unbounded.

**Item B — `DISCOPT_IPX_CHEAP_FIRST`: shipped default-ON, now DEFAULT-OFF. Falsified by the review.**

> **Retracted:** this PR previously claimed *"total 4208 → 3636 (−13.6 %), 4 fewer, 0 more, no instance regresses"* and graduated the flag default-ON on it. Both the number and the premise behind it were wrong. Kept here rather than quietly edited out, because the failure is instructive.

**Retraction 1 — the −13.6 % omitted the probe's own nodes.** `SolveResult.node_count` covers only the *winning* arm, so a probe that loses is invisible. Measured with the new `solver_stats["cheap_first/probe_nodes"]`: the losing probe spends **2627 nodes on `ex1264`** and **938 on `ex1265`**, against a credited saving of 572. The real figure is **4208 → 7219, +71.6 %**.

**Retraction 2 — the premise.** Item B exists because the lift "is harmful more often than not over the adopted population". *Every number behind that sentence is a node count.* Re-measured in wall clock on the same 12 instances — the complete population the gate can change — interleaved ON/OFF, 2 reps, 60 s, pooled sd ≤ 0.26 s:

| | gate ON | gate OFF (lift adopted) |
|---|---|---|
| total wall | **81.6 s** | **7.4 s** |
| total nodes | 11474 | 12328 |
| faster on | 1 instance | **11 instances** |
| certifications lost / gained | 0 / 0 | 0 / 0 |

An **11.1× slowdown for a 6.9 % node saving**. The lift wins by 30.3× (`ex1264`), 33.0× (`prob03`), 24.2× (`prob02`), 16.5× (`ex1263`); the only rows it loses are `nvs02` (3.4×) and `nvs14` (3.7×) — the very rows in #1236's table that motivated the item.

`nvs02` is the lesson in one row:

```
lifted     297 nodes   0.31 s
un-lifted    3 nodes   1.06 s
```

A node-count panel reads that as a 99 % improvement. It is a 3.4× regression. The lifted model is a pure MILP on the in-house Rust simplex; the un-lifted model pays an NLP relaxation per node — many cheap nodes traded for few expensive ones.

**Disposition:** flag **default-OFF**, `=1` opts in. Kept rather than deleted, per the `DISCOPT_CUT_INHERIT` precedent CLAUDE.md §5 cites for this exact case — the mechanism is sound (an ordinary complete solve of the original model; cert-clean on every panel run, 0 certifications lost), it is tested, and the measurement is recorded. **Re-graduating it requires a wall-clock panel, not a node panel.**

**The standing lesson** (`performance-plan.md` §72), not specific to this flag: *on this solver a node count is not a cost.* Two formulations of the same model can differ ~100× in cost per node, so any comparison **between** formulations must be made in wall clock. Node counts remain the right instrument **within** one formulation — which is what §5's bound-neutral regime uses them for.

This leaves the two separator fixes as the correctness content of this PR. They are independent of the flag and ship regardless.

## What this does NOT fix

Stated rather than dropped:

- **`nvs13`'s 111 % root gap** (637 nodes, 0 uncertified / 0 undecided — visible thanks to the instrumentation). Its lift is never built (x² monomials), and extending the reform to integer squares is argued dead by this PR's own evidence: that mechanism yields an exact root and a hard tree. `nvs13` needs root strength *without* a lift — #196/#208 territory.
- **Residual floating-point exposure in GMI on big-M models.** This removes the O(1) hole; numerically-safe GMI (Cook–Dash–Fukasawa–Goycoolea directed rounding) is the answer to the remainder.
- **The cheap-first gate's value on a wider corpus.** Default-OFF on this evidence; `minlplib.org` and the MINLPLib snapshot were not reachable from this environment, so only 12 in-repo instances were measured.

## Corrections made during this work

Recorded because the instrument errors cost more here than the solver defects did — and the last one invalidated a shipped default:

1. A cut evaluation that dropped the epigraph column and reported 23 phantom violations.
2. A violation gate scaled by `|z*|max = 3.3e8` (threshold −0.86 against a real −0.017) that reported "0 violated" while looking straight at the bad cut — this caused me to **retract a correct finding** and spend a cycle chasing the tree instead.
3. A sense-blind panel check that flagged `syn05hfsg` as a second defect; it is a **MAXIMIZE** model, so `bound ≥ obj` is correct. Re-checked sense-aware, both arms are 0 violations.
4. A cheap-first probe whose guard blocked recursive probing but not the probe adopting the lift itself, so it "certified" with the lift's own 297 nodes.
5. **`root_bound=None` meant three different things** — root region certified empty, root LP undecided, search never reached node 1 — the §6 conflation reproduced *inside* the instrument built to remove it. Now `root_status` names the arm.
6. **A test that breaks when the solver gets faster**: `test_..._budgeted_exit` asserted `status != "optimal"` for `tls2` at 5 s. It now asserts the contract for whichever regime the solve lands in, and counts which arm ran — a skip would be the §6 no-op that reads as a pass.
7. **The node-count panel that graduated item B** (see above). The largest of the set: it measured the winning arm only, and the metric itself was wrong for the comparison being made.

`mir.rs:242` was flagged and is a **false positive**: that skip only chooses whether to complement a column; no term is deleted.

One process note: this container's `mypy` on PATH is 1.19.1 while the repo pins **2.1.0** via pre-commit (which CI runs). 1.19.1 reports a pre-existing `mpec.py` error that 2.1.0 does not — running the unpinned one would have meant chasing a phantom while missing a real error the pinned one caught.

## Caveat on corpus coverage

All panels ran on the **122 in-repo instances** across `minlplib_nl` and `minlplib`. `minlplib.org` is blocked by this environment's egress policy and the MINLPLib snapshot directory is not present, so no panel saw the full corpus or the `syn`/`rsyn` class. That is sufficient for the correctness fixes — each is proven by construction plus a witness — but the item B wall-clock result rests on the 12 instances the gate can change.

https://claude.ai/code/session_01WMEj6Z3EtkQMPu4deuhUd1

---
_Generated by [Claude Code](https://claude.ai/code/session_01WMEj6Z3EtkQMPu4deuhUd1)_


jkitchin:
## Code review — 2 blocking, 4 medium, 4 low

CI is red on `Python fast`, and the failing test is the same defect as finding 5 below:

```
FAILED python/tests/test_1153_budget_monotonicity.py::test_no_unsaturated_role2_carve
assert not [('solver.py', 'budget = _IPX_PROBE_BUDGET_FRACTION * float(time_limit) - float(elapsed)')]
```

**Verified-correct parts** (checked, not defects): the cover-cut capacity charge is a valid relaxation (`sum_items w x <= cap - sum_nonitems min(w x)`), and both driver call sites pass root-box bounds (`ctx.l_w`/`ctx.u_w` are `NodeCtx`'s read-only tree bounds), so the cuts stay globally valid; the convex-kernel site is explicitly node-local with a restore point. The GMI rhs bookkeeping for a dropped term (`rhs -= psi*range`) is algebraically correct in both the `at_lower` and `at_upper` accumulations, and `psi >= 0` holds on every branch. The dive-funnel identity `DiveRuns == hit + abandoned + exhausted` closes — those are the only exits in `try_dive_repair`.

---

## Blocking

### 1. `python/discopt/solver.py:10211` — the cheap-first probe drops every caller option but two, and its result is returned as the final answer

```python
_ipx_probe = _ipx_unlifted_probe(
    model, time_limit, time.perf_counter() - _solve_t0,
    gap_tolerance=gap_tolerance, max_nodes=max_nodes,   # <- everything else dropped
)
if _ipx_probe is not None:
    return _ipx_probe
```

`_ipx_unlifted_probe` (`solver.py:6576`) forwards `**solve_kwargs` straight into a nested `solve_model`, so with only those two keys passed, `lazy_constraints`, `incumbent_callback`, `node_callback`, `abs_gap_tolerance`, `initial_point`/`warm_start`, `nlp_solver`, `solver`, `cuts`/`rlt`, `presolve`, `deterministic`, `incumbent_time_extension` and `**kwargs` all silently vanish.

With `lazy_constraints` set, the probe solves a **relaxation of the user's problem**; if it reports `optimal` and `gap_certified`, that certified-but-wrong result is returned unchanged. A user-tightened `abs_gap_tolerance=1e-9` silently reverts to the route default; `incumbent_callback`/`node_callback` never fire.

This file already has the correct pattern one screen up — `_sub_blocked` (`solver.py:8295`) refuses the substitution presolve outright when `initial_point`/`lazy_constraints`/`incumbent_callback`/`node_callback`/`decomposition_structure` is present. The probe needs the same refusal **plus** full kwarg forwarding.

### 2. `crates/discopt-core/src/lp/gomory.rs:420` — excluding tiny kept coefficients from `min_c` makes the `max_c == 0.0` refusal unreachable

Initialization is `max_c = 0.0`, `min_c = f64::INFINITY` (lines 320-321). `max_c` is updated for **every** kept term (line 409), but `min_c` now only under `if !tiny` (line 421). So a cut whose nonbasic coefficients are all ~1e-12 has `max_c = 1e-12 != 0` and `min_c = INFINITY`, and the guard

```rust
|| (min_c > 0.0 && max_c / min_c > max_dynamism)
```

evaluates `1e-12 / inf = 0` — it passes. Before this change every such term was skipped, `max_c` stayed `0.0`, and `max_c == 0.0` refused the cut.

Scenario: a tableau row whose reduced coefficients are ~1e-12 on columns with `u = 1e20` (the big-M masters this PR targets, where the `u[j] < 1e20` drop branch never fires) emits `sum 2e-12 * xtilde >= 1`. On the convex-kernel path `substitute_slacks` (`convex_kernel.rs:777`) applies **no** coefficient cleanup or dynamism check — it only drops exact zeros — and `select_cuts` will select it (violation ~1 at `sol.x`). That row reaches the node LP, where at a 1e-9 feasibility tolerance it reads as `0 >= 1` and the node is fathomed as infeasible although it contains feasible points: a false prune feeding a certified bound. The driver path is protected only because `substitute_slacks_to_structural` refuses such a cut; the kernel path is not.

---

## Medium

3. **`crates/discopt-core/src/lp/gomory.rs:391` — `tiny` is derived from `abar`, not from `psi`.** `tiny = psi.abs() <= tol || (integrality[j] && snap_eligible)`. When an integer column is pinned at a *fractional* bound — the case lines 350-360 explicitly handle by falling back to the continuous formula (`use_integer == false`) — an integral `abar` gives `psi = |abar|/f0`, which is `>= 1/f0` and can exceed 1000. That coefficient is flagged `tiny` and excluded from `min_c`, so the dynamism gate silently ignores a large coefficient. `snap_eligible` should only be consulted on the `use_integer` branch.

4. **`crates/discopt-core/src/lp/gomory.rs:374-400` — removing the `|psi| <= tol` skip makes GMI cuts dense on exactly the models this PR targets.** Every nonbasic column now gets a nonzero coefficient unless `u[j] < 1e20 && l[j] > -1e20 && psi*range <= tol`. On a big-M master with `u = 1e20` the drop branch never fires, so cuts become fully dense. On the convex-kernel path `substitute_slacks` then expands *every* tiny slack coefficient into its row's full support (`convex_kernel.rs:788-797`), turning each cut into a dense row accumulated with heavy cancellation; `appended_row_cap` bounds the row *count*, not the width. The differential panel measured node counts, not LP time or cut nnz, so this would not have shown up in it.

5. **`python/discopt/solver.py:6598` — the probe budget is a role-2 wall clock that bypasses `_role2_budget`/`_role2_deadline`.** `budget = 0.40*time_limit - elapsed` is precisely the "fraction of `time_limit` handed to a stage" that `_role2_budget` (line 7210) exists to neutralize, and `deterministic` is not forwarded to the nested solve. An instance whose un-lifted arm runs near `0.4*T` flips between the lifted and un-lifted formulation across runs on a loaded machine, changing node count, route and possibly certification — which is what `test_1236_determinism_contract.py` asserts cannot happen for a certifying solve. **This is the CI failure above**; the test names the three acceptable fixes (`solver_tuning.saturate_role2(seconds, frac)`, a constant `min(...)` ceiling, or a categorized `KNOWN` entry).

6. **`python/discopt/solver.py:6598` / `:10211` — `max_nodes` is spent twice and the probe's nodes are never reported.** The caller's full `max_nodes` goes to the probe; if it fails to certify, the lifted solve gets the full `max_nodes` again, so a caller using it as a work budget can get up to 2x the nodes. The returned `SolveResult.node_count` covers only the winning arm, so on every instance where the lift is *kept* (`ex1264`/`ex1265`) total work is under-reported — meaning the PR's "total 4208 -> 3636 (-13.6 %)" figure omits the probe's nodes on exactly those instances. That number should be re-measured before the claim ships.

7. **`python/discopt/solver.py:6556` — `_IPX_CHEAP_FIRST_IN_PROBE` is a module global, not thread-local.** With two `solve_model` calls in threads, thread A setting the flag for its probe causes thread B, on reaching the same block, to take the `_ipx_adopt = False` branch and solve the un-lifted model without ever probing. The formulation choice then depends on an unrelated concurrent solve. A `contextvars.ContextVar` (or threading it through as a parameter) would avoid it.

---

## Low

8. **`crates/discopt-core/src/bnb/spatial_tree.rs:576-580`, `:600-606` — three different root outcomes collapse to `root_bound=None`.** The `EmptyRegion` arm sets `+INFINITY`; the undecided arm sets `parent_bound`, which at the root is `-inf`. `solver.py:1928` maps any non-finite value to `None` — the same value the doc comment reserves for "the root never produced one". So "root region proved infeasible", "root LP decided nothing" and "search exited before node 1" are indistinguishable from Python: the section-6 conflation this instrumentation exists to remove.

9. **`python/tests/test_1236_gmi_drops_no_term.py:98` — uses the relative tolerance the fixture README forbids.** The new `fac2_master0.npz` section of `README.md` states tolerances against this fixture must be **absolute**; the test then computes `slack = 1e-6 * abs(attained)`, about **331.8** absolute. It catches the 7839 error, but it is a scale-relative gate of the kind the README says is how the defect was missed, and it would not catch a regression an order of magnitude smaller.

10. **New `@pytest.mark.smoke` tests run 60-second solves against the 300 s per-test timeout.** `test_all_milp_backends_agree_on_the_captured_master` runs 4 backends at `time_limit=60.0` (up to 240 s in one test); `test_a_certifying_solve_explores_the_same_tree_every_time` runs 3 full solves per parametrization at 60 s; `test_gate_declines_a_lift...` runs 2 solves x 3 instances at 60 s. `pytest -m smoke` is the per-PR gate. Separately, `test_the_suite_does_not_claim_determinism_for_a_budgeted_exit` asserts `r.status != "optimal"` for `tls2` at a 5 s budget — that fails the moment the solver gets fast enough, a self-inflicted future break in the smoke gate.

---

Findings 1, 5 and 6 are all the same helper, so they are one focused change to `_ipx_unlifted_probe`; fixing 5 also clears the red CI job.

jkitchin:
Thank you — this was a high-yield review. Eight findings were real and are fixed; one is falsified by measurement; one was already fixed on a newer head. But **finding 6 is the one that mattered**: chasing it ended with item B's premise falsified and the flag defaulted OFF.

Commits: `9dd0982` (findings 1–3, 6–10), `5a7220e` (a typecheck error the local gate caught), `f53f5dd` (item B). PR body updated with the retraction rather than quietly edited.

---

## The headline: finding 6 killed the feature, not just the number

You flagged that `node_count` covers only the winning arm. Correct, and worse than stated.

**Retraction 1 — the −13.6 % omitted the probe's own nodes.** Measured via the new `solver_stats["cheap_first/probe_nodes"]`: the losing probe spends **2627 nodes on `ex1264`** and **938 on `ex1265`**, against a total credited saving of 572. The published figure inverts: **4208 → 7219, +71.6 %**.

**Retraction 2 — the premise.** Item B exists because the lift "is harmful more often than not". *Every number behind that sentence is a node count.* Re-measured in wall clock on the same 12 instances — the complete population this gate can change — interleaved ON/OFF, 2 reps, 60 s, pooled sd ≤ 0.26 s:

| | gate ON | gate OFF (lift adopted) |
|---|---|---|
| total wall | **81.6 s** | **7.4 s** |
| total nodes | 11474 | 12328 |
| faster on | 1 instance | **11 instances** |
| certifications lost | 0 | 0 |

An **11.1× slowdown for a 6.9 % node saving**. The lift wins by 30.3× (`ex1264`), 33.0× (`prob03`), 24.2× (`prob02`), 16.5× (`ex1263`). The only rows it loses are `nvs02` (3.4×) and `nvs14` (3.7×) — *the very rows in #1236's table that motivated the item*.

`nvs02` is the whole lesson in one row:

```
lifted     297 nodes   0.31 s
un-lifted    3 nodes   1.06 s
```

A node-count panel reads that as a 99 % improvement. It is a 3.4× regression. The lifted model is a pure MILP on the Rust simplex; the un-lifted model pays an NLP relaxation per node. **On this solver a node count is not a cost** — any comparison *between* formulations has to be in wall clock. Written up as `performance-plan.md` §72.

`DISCOPT_IPX_CHEAP_FIRST` is now **default-OFF** (`=1` opts in), kept rather than deleted per the `DISCOPT_CUT_INHERIT` precedent CLAUDE.md §5 cites for exactly this case. Re-graduating it requires a wall-clock panel.

**None of this touches the correctness content.** Both separator fixes are independent of the flag.

---

## Blocking

**1 — probe forwarded 2 of 47 options.** Confirmed and fixed two ways, because a list of names is only as good as its last edit:
- The options it cannot honour (`lazy_constraints`, both callbacks, `cut_callback`, `initial_point`, `warm_start`, `decomposition_structure`) now **decline the probe outright** — pre-#1236 behaviour, never a wrong answer, mirroring `_sub_blocked` as you suggested.
- Everything else is forwarded explicitly, and `test_the_probe_forwards_every_option_that_changes_the_answer` enumerates `solve_model`'s **live signature**, failing on any parameter that is not forwarded, blocked, or exempt *with a stated reason*. Verified non-vacuous: deleting `abs_gap_tolerance` makes it name that parameter.

**2 — `max_c == 0.0` unreachable.** Confirmed exactly as you describe, including that the driver is shielded by `substitute_slacks_to_structural` while the kernel's `substitute_slacks` is not. Fixed by refusing when `!min_c.is_finite()` — precisely "no kept coefficient was above tolerance", which is what the `max_c == 0.0` arm used to mean. The new test asserts the *spec* (whatever is emitted must be enforceable at LP tolerance) rather than a cut count.

## Medium

**3 — `snap_eligible` on the wrong branch.** Confirmed; gated on `use_integer`. Pre-fix the test shows the separator emitting `[0, 2, 2e6]` — a true spread of 1e6 that the gate saw as 1.0 and a `max_dynamism = 1e3` waved through. Note the interaction: once finding 2's refusal exists, this bug would *also* have started refusing good cuts whose only coefficient came from such a column. Two gates, one mislabel, opposite failures.

**4 — cut density: falsified.** Plausible mechanism, wrong on the target instance. Added `SepGomoryCutsEmitted` / `SepGomoryCutNnz` (both — the ratio is meaningless without its denominator) and measured both arms on the `fac2` master, 485 columns, with arm identity asserted *behaviourally* (pre-fix returns the known false optimal 331845337.44, fixed returns 331837498.18, so a wrong `.so` fails before measuring):

| arm | objective (identity) | cuts | mean nnz/cut | % of columns |
|---|---|---|---|---|
| pre-fix | 331845337.44 | 3 | 35.3 | 7.3 % |
| fixed | 331837498.18 | 2 | **34.5** | **7.1 %** |

Marginally *sparser*, two orders of magnitude off dense. The mechanism misses two things: most tiny-ψ columns on a big-M master are **bounded binaries**, still charged to the rhs and dropped; and the driver refuses a cut whose pin is unbounded (`SubstDropUnbounded = 1` on that run). Scope stated rather than implied: this is the driver path, not the kernel — whose exposure is now bounded at the separator by finding 2's fix instead.

**5 — already fixed.** `_IPX_PROBE_BUDGET_CAP_S = 30.0` landed in `3e99f77`; this review ran against `9e9537f`. It is the "constant `min(...)` ceiling" the test names — I declined the KNOWN entry as weakening a guard. `deterministic` not being forwarded was real and is covered by finding 1.

**6 — fixed** (see above). `max_nodes` is now charged once, floored at 1; the losing probe's nodes are surfaced in `solver_stats`.

**7 — `ContextVar`.** Confirmed and fixed. A module global meant the formulation choice could depend on an unrelated concurrent solve.

## Low

**8 — `root_bound=None` meant three things.** Confirmed — the §6 conflation reproduced inside the instrument built to remove it. The kernel now reports `root_status` (`bounded` / `infeasible` / `undecided` / `not_reached` / `propagation_fathom`), surfaced as `solver_stats["root/status"]`.

**9 — relative tolerance the README forbids.** Correct, and it was my own README. Replaced with **measured** absolute slacks: against the witness the backends land at `auto` −1.67e-4, `simplex` −1.67e-4, `highs` −1.64e-4 (all *below* it) and `pounce` +0.32, the worst legitimate excess. Hence 1.0 for the driver and 10.0 cross-backend. **Detection margin 24× → ~800×.** The README now says to state the number in objective units and where it came from.

**10 — partly.** The self-inflicted break is real and fixed: `test_..._budgeted_exit` no longer asserts `status != "optimal"`; it asserts the contract for whichever regime the solve lands in and counts which arm ran (a skip would be the §6 no-op that reads as a pass).

On the timing half, the estimate was worst-case rather than measured: `time_limit=60` is a cap, not a cost. Measured durations — `test_all_milp_backends_agree_on_the_captured_master` **44.6 s** (not 240 s; each backend certifies in ~11 s), the two `ex126x` gate tests ~26 s each, everything else < 5 s. Worst case is 44.6 s against a 300 s timeout, a 6.7× margin, so I've left the limits alone; truncating them would risk a spurious failure for no real gain.

---

**Verification:** `pytest -m smoke` 1653 passed / 0 failed · adversarial suite 19 passed · `cargo test -p discopt-core` **763 passed across 15 binaries** · ruff, ruff-format, and the **pinned mypy 2.1.0** clean. Both new Rust tests confirmed *failing* against the reverted source before passing against the fix.

One process note worth passing on: this container's `mypy` on PATH is 1.19.1 while the repo pins 2.1.0 via pre-commit, and 1.19.1 reports a pre-existing `mpec.py` error that 2.1.0 does not. Running the unpinned one would have had me chasing a phantom while missing the real error it caught.

`mir.rs:242` remains a false positive, as you note.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WMEj6Z3EtkQMPu4deuhUd1)_

---

Issue #1299, opened 2026-09-17
correctness: Sensitivity.objective and SolveResult.gradient have the wrong sign under maximize (#1251)


jkitchin:
> Discovered by the discopt adversary agent (automated solver correctness testing).

## Summary

Under `maximize`, two sensitivity values come back with the **wrong sign**:

| quantity | reported | true |
|---|---|---|
| `Sensitivity.objective` (new in #1251) | **-1.0** | 1.0 (`SolveResult.objective` is correct, 1.0) |
| `SolveResult.gradient(p)` (older code, `_relax/differentiable._compute_sensitivity_at_solution`) | **-1.0** | +1.0 = dV/dp |

`Sensitivity.dobj_dp` is correct (+1.0). So the #1251 docstring's claim that `dobj_dp` matches `gradient` fails under maximize.

```python
import discopt.modeling as dm
m = dm.Model("s"); x = m.continuous("x", lb=-10, ub=10); p = m.parameter("p", value=2.0)
m.maximize(-(x*x) + p*x)          # x* = 1, V(p) = p^2/4 = 1, dV/dp = x* = 1
r = m.solve(sensitivity=True)
r.objective, m.sensitivity().objective, r.gradient(p), m.sensitivity().dobj_dp
# (1.0, -1.0, -1.0, [1.])
```

## Evidence (two methods)

- **Method 1, closed form:** V(p) = p²/4, so V'(2) = 1.
- **Method 2, finite differences of re-solved objectives:** adv-C's `adversary/runs/2026-09-17_pr1251_sens_maximize_chain.py` (5 failed checks) agrees with `dobj_dp` and disagrees with `gradient`.

## Cause

- `python/discopt/sensitivity.py:670` copies `raw.objective`, the solver's internal minimization value, without undoing the sense flip.
- `_compute_sensitivity_at_solution` misses the same flip.

## Expected

Both quantities are reported in the user's sense, so under maximize the values are 1.0 and +1.0.

Issue #1283, opened 2026-09-17
correctness: sum-reduction canonicalization reuses stale id()-memo entries — false certified optimum, nondeterministic (#1256 regression)


jkitchin:
> Discovered by the discopt adversary agent (automated solver correctness testing).

## Summary

The #1256 sum-reduction canonicalization (PR #1265) can return **certified-optimal results with a dual bound above the true optimum**, and different runs give different wrong answers. Found while adversarially testing PR #1265.

## Reproducer

```python
import discopt.modeling as dm
m = dm.Model("epi")
x = m.continuous("x", lb=-1.5, ub=1.5, shape=(3,))
Y = m.continuous("Y", lb=-1, ub=1, shape=(2, 2))
t = m.continuous("t", lb=-100, ub=100)
m.subject_to(dm.sum(0.5**x) + dm.sum(10.0**Y) <= t)
m.minimize(t)
r = m.solve(time_limit=30)
```

- True optimum: `3*0.5**1.5 + 4*0.1 = 1.460660`, at x = 1.5 and Y = -1.
- discopt on main (a5887a4c), three identical runs:

  | run | status | objective | bound | gap_certified |
  |---|---|---|---|---|
  | 0 | optimal | 3.871327 | 3.871327 | True |
  | 1 | optimal | 5.774887 | 5.774887 | True |
  | 2 | optimal | 5.774887 | 5.774887 | True |

- The certified bound is above the true optimum, so this is a false certificate.

## Independent confirmation

1. Analytic optimum 1.460660 (every term is monotone in its own variable).
2. numpy evaluation of `sum(0.5**x) + sum(10**Y)` at the returned points gives 2.2379 and 2.4990. Both are feasible values below the certified bound.
3. The adversary's script prints the canonical term for `10**Y[0,0]` as `exp(2.3026*exp(-0.6931*x[2]))`: a term from the *first* sum substituted into the second. `Y[0,0]` does not appear at all. 3.618 (another run) is exactly the minimum of that wrong function.

## Likely cause

- `_Canonicalizer._memo` in `python/discopt/_relax/canonical_expr.py` is keyed by `id(expr)`. Its comment says this is safe only while the model pins every subtree.
- The new `sum` branch (#1256) canonicalizes **temporary** element expressions produced by `scalar_elements(op)`. Nothing pins them, so they are freed once the branch returns.
[...]

PR #1303, opened 2026-09-17
fix(correctness): adversary batch 3 — #1293–#1297

### LINKED ISSUE #1293: correctness: dm.register_function certifies a kinked body (e.g. -abs) as convex; unsound envelope gives false certified optima (#1275)
> Discovered by the discopt adversary agent (automated solver correctness testing).

## Summary

A function registered with `dm.register_function` whose body contains a kink (`abs`, `sign`) gets a **false curvature verdict**. Its envelope is then unsound, and the solve returns a **certified false optimum**. Introduced with #1275 (#1248 A).

```python
negabs = dm.register_function("negabs", lambda x: -dm.abs(x))
negabs._derived()[2](-1.0, 2.0)   # -> 'convex'  (it is concave; midpoint -0.5 > chord -1.5)
```

## Root cause

- `RegisteredFunction._derived` (`python/discopt/operators.py:156-221`) takes f'' symbolically with `bilevel.symbolic_diff.diff`. That module uses subgradients at kinks (`symbolic_diff.py:181-183`: `abs' = sign`, `sign' = 0`).
- So f'' of `-abs(x)` is identically 0. The interval enclosure of f'' is [0, 0], and `curvature` returns `"convex"` because `enc_lo >= 0`. The Dirac mass at the kink is invisible.
- `_emit_1d` then builds a "convex" envelope (tangents) for a concave function. The same happens for any kinked body, e.g. `1 - abs(x - 0.3)`.

## Evidence (two independent oracles)

Separable concave minimization over a box plus one knapsack row, n=6, seed 3. The instance generator is in `adversary/runs/2026-09-17_pr1275_register_kinked_atom.py`.

| inst | registered `negabs` (certified optimal) | plain `-dm.abs` | oracle A: vertex enumeration | oracle B: min over 2^6 sign-pattern LPs |
|---|---|---|---|---|
| #4 | -10.313636 | -10.686115 | -10.6861 | -10.686115 |
| #7 | -16.244455 | -16.361528 | -16.3615 | -16.361528 |
| #8 | -16.2526 | correct | -16.6245 | – |
| #10 | -11.2453 | correct | -11.8282 | – |

- 4 of 12 registered-arm instances certify a bound above the true optimum. The plain spelling is correct on all 12.
- The root relaxation of `min negabs(x)+negabs(y)+0.3xy` on [-1,2]^2 is LP-infeasible, although the model is feasible with optimum -3.6. The end-to-end solve happens to be right there because the root incumbent is optimal.

Repro:
```
python adversary/runs/2026-09-17_pr1275_register_kinked_atom.py      # 143 checks, exit 2
python adversary/runs/2026-09-17_verify_pr1275_knapsack.py           # LP-enumeration oracle, exit 2
```

## Expected

The verdict must abstain (return `None`) whenever the lowered body contains a nonsmooth node (`abs`, `sign`, `max`/`min`, and fractional powers at 0) whose kink can lie in the box. Alternatively, the curvature rule must add the kink's contribution, e.g. treat `abs` of an affine argument as convex with the kink sign-aware. Abstaining costs tightness only.

### LINKED ISSUE #1294: correctness: continuous(..., ub=None) stores a NaN bound; feasible model certified infeasible
> Discovered by the discopt adversary agent (automated solver correctness testing).

## Summary

`m.continuous(name, ub=None)` (or `lb=None`) stores the bound as **NaN** without complaint. Downstream:

- `ub=None`: a trivially feasible model is reported **`infeasible`**, which is a false certificate.
- `lb=None`: the HiGHS LP route raises `ValueError: standard form xl contains NaN`.

```python
m = dm.Model(); x = m.continuous("x", lb=0.0, ub=None)
x.ub                      # nan
m.subject_to(x >= 1); m.subject_to(x <= 3); m.minimize(x)
m.solve().status          # 'infeasible'   (x = 1 is optimal, objective 1)
```

`None` is outside the annotated type (`float | ndarray`), but it is the natural spelling of "no bound", and `np.asarray(None, dtype=float)` silently yields NaN. Per the correctness policy, a malformed bound must be refused loudly, not turned into a certificate.

## Evidence

- **Oracle 1 (analytic):** x = 1 satisfies every constraint.
- **Oracle 2:** the same model with the default `ub` solves `optimal`, objective 1.

Repro: `python adversary/runs/2026-09-17_verify_s2_ub_none.py`

## Expected

`continuous` / `integer` / `binary` either map `None` to the infinite default or raise `TypeError`/`ValueError`. In addition, reject any NaN in declared bounds at variable creation.

### LINKED ISSUE #1295: correctness: default pure-MILP HiGHS route certifies a false bound on badly scaled MILPs (trusts mip_dual_bound)
> Discovered by the discopt adversary agent (automated solver correctness testing).

## Summary

The default pure-MILP route ("highs-milp: verified HiGHS route (HiGHS kOptimal); milp/bound_provenance=highs-fp") returns **certified optima that are not optimal** on badly scaled big-M MILPs. It appears to trust HiGHS's floating-point `mip_dual_bound`. A1 found 32 of 80 instances wrong with coupling coefficients eps in {1e-5, 1e-7, 1e-9}.

## Evidence (instance #70, seed 11; generator in `adversary/runs/.pr1280_p2_defs.py`)

- discopt returns `optimal`, `gap_certified=True`, objective = bound = **-131.644309**.
- An explicit witness point:
  - row violation **exactly 0** in `fractions.Fraction` arithmetic;
  - bounds and integrality exact;
  - objective c·w = **-131.843773**, below the certified bound by 0.1995.
- Second method: a SciPy HiGHS LP with the integer part fixed to the witness gives -131.8437730540767.

Instance #45 behaves the same way: it certifies -214.1445, while a witness has -220.5510. That witness has a float-rounding row residual of 8.6e-11, which is irrelevant against a gap of 6.4.

The `DISCOPT_LP_MILP_BACKEND=rust` route was correct on the instances examined.

Repro:
```
python adversary/runs/2026-09-17_verify_pr1280_highs_milp.py
python adversary/runs/.pr1280_p2_witness.py 1e-5,1e-7,1e-9 45,70,75
python adversary/runs/.pr1280_p2_e2e_scan.py 1e-10 80
```

## Expected

A certificate on this route must not rest on HiGHS's unverified `mip_dual_bound` when the model is badly scaled. Either verify the bound (safe dual / exact re-check of the final node LPs), or downgrade to `feasible` with an uncertified gap. See `docs/dev/lp-milp-highs-routing-plan.md`.

### LINKED ISSUE #1296: correctness: Rust MILP driver certifies a false optimum with 1e-11 row coefficients; reported objective != c.x of its own incumbent
> Discovered by the discopt adversary agent (automated solver correctness testing).

## Summary

The in-house Rust MILP driver certifies a false optimum on mixed-knapsack MILPs with a row coefficient of **1e-11**. That is below the 1e-10 drop threshold targeted by #1280, so this is not a regression of that fix: every instance at eps ≥ 1e-10 was clean.

The reported objective also **disagrees with c·x of the solver's own returned x**, which points at the objective/bound bookkeeping rather than at the LP.

## Evidence (instance #116 from `adversary/runs/2026-09-17_pr1280_tiny_coef_cuts.py`, seed 20260917)

Instance: big-M shape, `eps=[1e-11, 1e-7]`, `ybig=[1e11, 1e10]`, `y1_ub=1e12`.

| method | value |
|---|---|
| exact `Fraction` witness (rows and bounds exactly feasible, integral) | **-199.694063182027** |
| exhaustive enumeration oracle (adv-A1) | -199.694063182027 |
| `milp_simplex.solve_milp`: status OPTIMAL, objective = bound | **-198.0716445881931** |
| c·x of that call's own returned x | -199.69406318202653 |
| `Model.solve` with `DISCOPT_LP_MILP_BACKEND=rust`: optimal, `gap_certified=True` | **-199.3321943420291** |

Further observations from adv-A1:

- The crossing persists with all cuts disabled, and GMI makes it worse.
- At eps=1e-11, 42 of 201 big-M instances cross the bound.
- With the integers fixed, the Rust LP is correct.
- Four returned incumbents violate a knapsack row (scale-aware residual): #38 1.1e-4, #41 4.0e-3, #89 4.9e-4, #209 5.6e-5.

Repro:
```
python adversary/runs/2026-09-17_verify_pr1280_rust_tiny.py   # exit 2
python adversary/runs/2026-09-17_pr1280_tiny_coef_cuts.py     # full scan
```

Possibly related to #1287 (Rust MILP kernel accepting infeasible points). The symptom here differs: objective ≠ c·x, and a false bound on a correctly bounded model.

## Expected

- The reported objective equals c·x of the returned incumbent, recomputed.
- The bound never exceeds a feasible point's value. When coefficients are too small to be handled soundly, refuse or downgrade the certificate.

### LINKED ISSUE #1297: correctness: feasible MINLP with zero-slack elementwise max row and an integer variable reported infeasible
> Discovered by the discopt adversary agent (automated solver correctness testing).

## Summary

A feasible MINLP with a zero-slack elementwise `max` row and an integer variable is reported **`infeasible`**. The bug predates #1250 and affects every spelling tried: nested two-argument `maximum`, the n-ary form, `max(...) <= v`, and one scalar row per element.

```python
import discopt.modeling as dm
m = dm.Model("I"); v = m.continuous("v", shape=(3,), lb=0, ub=4); s = m.integer("s", lb=0, ub=3)
m.minimize(v.sum() - 1.5*s)
m.subject_to(dm.maximum(dm.maximum(v, s), 1.0) - v <= 0)
m.solve(time_limit=30).status    # 'infeasible'; true optimum 1.5 at v=(1,1,1), s=1
```

## Evidence (two methods)

- **Method 1, analytic:** the row is equivalent to `v_i >= max(s, 1)`. Enumerating s ∈ {0..3} gives optimum `n*max(s,1) - 1.5 s`, which is 1.5 for n=3 and 2.5 for n=4. The point v=1, s=1 is feasible.
- **Method 2, linear form:** the same feasible set written as `v - s >= 0, v >= 1` solves `optimal` with 1.5 and 2.5.

Further observations (adv-C):

- The bug needs n ≥ 3. With n=2 the value is right but no bound is returned.
- It still occurs with `DISCOPT_NLPBB_ROOT_CUTS=0`.
- It disappears if s is continuous or if the right-hand side is loosened to 1e-6.
- Unverified hypothesis: a node NLP failure on the zero-slack nonsmooth rows is being treated as proof of infeasibility.

Repro:
```
python adversary/runs/2026-09-17_verify_advC.py                       # U1
python adversary/runs/2026-09-17_pr1250_nary_elementwise_minlp.py      # N3
```

## Expected

The solve returns `optimal` with objective 1.5 (n=3) and 2.5 (n=4). A node NLP failure must never be converted into an infeasibility certificate without a verified proof.

jkitchin:
Third batch of fixes for the adversarial findings filed against this week's PRs
(#1283-#1299). Stacked on the batch-2 branch (`fix/adversary-2026-09-17-b`, PR
#1302); the diff below is the five commits of this batch only.

| Issue | Fix |
|---|---|
| #1293 | Registered-function curvature abstains on boxes that may contain a kink |
| #1294 | A `None` variable bound means unbounded; NaN bounds are refused |
| #1295 | The HiGHS MILP route stops certifying on unscalable open columns |
| #1296 | The Rust MILP driver withdraws its certificate on noise-floor entries |
| #1297 | Per-evaluator-row OA inputs; nonsmooth MINLPs take the spatial path |

Every fix adds a regression test that fails before it and passes after; each of
those "fails before" claims was checked against a worktree at the parent commit
(and #1293-#1296 against `origin/main` at 108bcf3b) with the built extension
copied in, so the baseline runs the code under test and not the main tree.

#### #1296 in detail

The instance is a knapsack pair whose overflow is bought through `W x - eps*y <=
cap` with `eps = 1e-11` on an open column: below the simplex's equilibration
noise floor, so the node LPs ran as if the entry were absent. `solve_milp`
returned `optimal` with bound -198.07 against an exact (Fraction-arithmetic)
witness of -199.694, and an objective that was not `c.x` of its own point. The
driver now refuses the certificate instead of guessing at the dropped term.
The Rust change means this PR's gate includes `cargo test -p discopt-core`.

Because the fix is in the Rust extension, the local runs used a build made
without touching the venv: `maturin build --release -o <dir>`, then the
`discopt/_rust.abi3.so` from that wheel copied over `python/discopt/`.

#### #1297 in detail

A feasible convex MINLP with an elementwise `max` row returned `infeasible` for
n >= 3 on 108bcf3b. Two independent defects:

1. OA/LOA/AMP built the convexity mask and sense list per `Constraint` *object*,
   while cut generation and the feasibility checks index them per *evaluator
   row* — and an array-valued constraint is one object over several rows.
   `solver="mip-nlp"` and `gdp_method="loa"` raised IndexError on any vector
   row, and AMP's `_check_constraints` raised a broadcast error, rejecting every
   candidate point. The new `_oa_rows()` expands both over
   `evaluator.constraint_row_map()` and raises rather than guessing when the
   lengths disagree.
2. Every gradient route trusted a smooth NLP at the `max` kink, where it stalls.
   Such models now take the spatial B&B's exact envelopes, as nonsmooth
   *continuous* models already did.

Neutrality evidence: of 27 nonsmooth MINLPLib instances every one classifies
nonconvex, so neither the route gate nor the demotion fires on them; over 96
instances the decomposition's mask and senses are identical to the old
per-object lists.

#### Re-run at the branch head (ad78e8ef, after the three test repairs)

The last commit touches a source file (`operators.py`), so the gates were re-run.
These numbers are from `fix/adversary-2026-09-17-d`, which is this branch plus
batch 4 and therefore a superset of what merges here.

* `pytest -m smoke` — **1742 passed**, 2 skipped, 0 failed. The +2 over the run
  below is exactly the two new tests in `test_1248_register_function.py`, which
  carries a module-level `smoke` mark (20 collected in that file, was 18).
* `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — **19 passed**.
* CI's "Python fast" lane reproduced verbatim under `-n 4` xdist —
  **10850 passed, 0 failed**, against `2 failed` on the pre-repair tree.
* CI's "Python correctness (fast subset, serial)" lane reproduced verbatim —
  **465 passed, 0 failed**.
* `cargo test -p discopt-core` was not re-run: no Rust changed after 243da503,
  where it was green at 756 passed.

#### What was run (local, at 243da503)

* `pytest -m smoke` — 1740 passed, 2 skipped.
* `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — 19 passed.
* `cargo test -p discopt-core` — 756 passed, 0 failed (Rust touched by #1296).
* Related OA/AMP/GDP-LOA/mip-nlp/NLP-ipopt/DAE suites — 1202 passed, 1 xfailed,
  2 xpassed.
* The new `test_1297_nonsmooth_vector_minlp.py` — 21 tests; 20 fail on the
  pre-fix tree, all 21 pass here. The one that passes on both is the
  smooth-model control asserting the OA route is still selected.
* `ruff check` and `ruff format --check` clean; `mypy` clean on changed files.

Known unrelated failure: `test_oa.py::TestOAEdgeCases::test_infeasible_model`
(`slow`-marked) asserts `result.x == {}` and gets `None`. It fails identically
on `origin/main` at 108bcf3b and is untouched by this branch.

## #1296's blast radius, and two test repairs it forced

#1296 changes certified bounds, so per CLAUDE.md §5 here is the differential
measurement rather than an adjective. The tiny-entry gate withdraws the
certificate on **5 of the 66** in-repo corpus instances (7.6%) — bchoco06,
bchoco07, bchoco08, contvar, hda — and every one of the five still returns
`status=optimal` with a finite bound. No instance loses a certificate it had.

On hda the gate is an improvement, not a regression, and it is large. Measured
at the root McCormick LP with the control repeated exactly:

| | loose-bound OFF | loose-bound ON |
|---|---|---|
| pre-#1296 (e8746ea8) | -999106100811.8599 | -1378479214.7502906 |
| post-#1296 (2c267183) | -64675.249181289175 | -64675.249181289175 |

The new bound is seven orders of magnitude tighter and stays sound: minlplib's
optimum for hda is -5964.534084, and -64675.25 is below it.

### Retraction (CLAUDE.md §11)

`test_row_filter_loose_bound.py`'s docstring described the #671 loose-bound
trigger as what tightens hda's root bound, and asserted a 50x gap between the
two flag arms. That is no longer true of this tree and the docstring now says so
in writing. #1296's decert routes the node through #671's *failure* trigger —
the one the loose-bound flag does not gate — so the loose re-solve never runs and
both arms return the identical -64675.249181289175. Measured over all 66 corpus
instances, `DISCOPT_RELAX_ROW_FILTER_LOOSE_BOUND` now changes the certified root
bound on **0** of them: it is inert corpus-wide, so there was no other witness to
retarget the old assertion at.

The test was rewritten to discriminate on the row filter as a whole
(`DISCOPT_RELAX_ROW_FILTER`), which still earns its keep on hda: -77002.685837
filtered off vs -64675.249181 on, a 1.19x tightening, with the bar set at 1.05x.
Both arms keep their soundness assertions. Proved non-vacuous — it fails at
1.00x with the filter forced off.

Getting here required retracting a measurement of my own first: I read the
`MilpTinyEntryDecert` counter without `DISCOPT_PROFILE=1`, got 0, and concluded
the gate never fires — falsifying a hypothesis that was correct. `profile::incr`
is gated behind `enabled()`, so the counter was a dead instrument. Re-measured
with it armed: `MilpTinyEntryDecert == 1` on hda.

### `test_benders_soundness.py`

`test_c3_unbounded_recourse_reported` used `ub=None` and passed only by accident:
`np.asarray(None, dtype=float)` is NaN, and the NaN tripped the Benders unbounded
path. #1294 maps `None` to the finite default box, which is the correct behavior
and removes the accident. The test now says `ub=1e30` — at or above the Rust LP
layer's `1e20` INF sentinel, so it is genuinely unbounded — and the docstring
records why.

### `test_1293_registered_kinked_curvature.py` (the xdist-only failure)

Root-caused rather than retried: `clear_registered()` clears the process-global
registry, and two other test modules called it without restoring, unregistering
the atom `test_1293` registers at import time. Nothing raises when that happens —
`atom_of` returns None by design and the model relaxes term-by-term — so the only
symptom is the envelope going untaken, i.e. `use_count == 0`. Reproduced in 2.3 s
with two files and no xdist. Fixed with `operators.registry_snapshot()`; #1293's
assertion is unchanged.

Closes #1293
Closes #1294
Closes #1295
Closes #1296
Closes #1297


jkitchin:
## Review — holding this one

CI is green, but there is a soundness defect plus two gates I think are calibrated too tightly. Details below; I verified each against the PR head (`243da503`) rather than reading the diff alone.

### Blocking: `gdpopt_loa.py:121` did not get the guard `oa.py:1415` got

`#1297` added an `n_rows == 1` gate to the extractor call in `oa.py`:

```python
coeffs = _extract_body_coeffs(c.body, model, n_vars) if n_rows == 1 else None
```

The identical call in `gdpopt_loa.py:121` is ungated. The comment directly above it asserts exactly the belief this PR's own `oa.py` fix disproves:

> `_extract_body_coeffs` returning `(c_vec, offset)` *is* the proof the body is one linear row — it cannot claim more than it produces — so it is the correct and only gate.

An array-valued body is several evaluator rows but comes back as **one** row with the rhs summed. So `c·x <= [-1.0, -2.0]` — true feasible set `c·x <= -2.0` — enters the LOA master (lines 785/808) as `c·x <= -3.0`. That is *tighter* than the true set, so the master stops being a relaxation and its bound can rise above the true optimum. That is an invalid bound, not a lost certificate.

Two call sites of one helper now hold contradictory beliefs about it, and the one this PR fixed is the correct one. The fix looks like the same two lines plus a regression test over `evaluator.constraint_row_map()`.

### Non-blocking, but worth reconsidering before merge

Both of these fail **safe** — they withdraw a certificate, they never issue a false one — so they are coverage regressions rather than soundness bugs. I think the coverage cost is larger than the calibration suggests.

**1. `lp_milp_highs.py:910` — `UNSCALABLE_OPEN_RATIO = 2**-20` (≈9.54e-7) catches ordinary big-M rows.**

A textbook `x - 2e6*y + s = 0` with `x` on the default open box gives ratio `1/2e6 = 5e-7` and loses certification on a model the route solves exactly (obj −3.0, bound −3.0, returned as `feasible` / `gap_certified=False`). Any in-row coefficient range past ~1.05e6 touching an unbounded column is affected.

The docstring's calibration is real and well documented — 97 false certificates at ≤5.3e-7, the 26 `ref/HiGHS/check` MILPs plus i1183 at ≥3.4e-5. But the "real" side is 27 models, and big-Ms at 1e6 are ordinary practice. The evidence is thinnest precisely in the 5.3e-7 → 3.4e-5 gap, and `2**-20` sits near its bottom edge. Worth either widening the real-model corpus or moving the cap up within the gap.

**2. `milp_driver.rs:1312` — `has_unscalable_tiny_entry` trips on a row's own logical column.**

I traced the whole chain: `_relax/std_form.py:182` gives a non-`eq` logical `ub = inf_value`, the 1e20 sentinel, so `u[j] >= INF` makes it *open*; its coefficient is 1.0 and its column's only entry, so `a >= MAX_LINE_RANGE * col_max[j]` holds; but with `MAX_LINE_RANGE = 1e-10`, any row whose max coefficient exceeds 1e10 fails `a >= MAX_LINE_RANGE * row_max[i]`, and the whole MILP is decertified (`Optimal` → `Feasible`, bound `-inf`).

The logical is the row's slack — structural, not noise, and exactly what equilibration is for. Excluding logical columns from this scan looks like the right fix.

### Smaller

- `lp_milp_highs.py:924` — `open_col` uses `np.abs(...) >= READBACK_LIMIT`, dropping the signed test used at 666/727/898-899. A genuinely bounded column such as `xl = 1e16, xu = 1e16+10` reads as open and widens the gate.
- `milp_driver.rs:2783` — the recomputed `c·x` is not written back to `tm`, so `decide_status`'s `tm.gap()` and `res.bound = stats.global_lower_bound` still read the stale incumbent. `bound > obj` at `Optimal` is reachable, clamped only by `_certified_bound` in one caller's `optimal` branch.

### Checked and found sound

The `_kink_sites` walk (the `NotImplementedError` arm is unreachable — `symbolic_diff.diff` raises first; the `positive_only` interval test is a correct exclusion proof); `_oa_rows` row-map expansion and its length assertions; the AMP `skip_reasons` rebuild (report-only, `convex_mask` alone gates generation); `_model_contains_nonsmooth_node`'s `id()` memo; the #1297 convexity demotion against the NLP-BB auto-select and `_convex_constraint_mask`; `_infer_constraint_bounds`'s evaluator delegation; `open_column_coefficient_ratio` not aliasing `sf.A` (StdForm stores CSC, so the CSR conversion copies); and the `decertify` early return for Farkas-proved infeasibility.

The neutrality evidence for #1297 (27 nonsmooth instances all classifying nonconvex; identical mask/senses over 96 instances) is the right shape of evidence and I have no quarrel with it.

---

Issue #1305, opened 2026-09-17
correctness/perf: _RootLP.senses is per-Constraint but indexed per row — root cuts silently skipped on every vector-constraint model


jkitchin:
## Summary

`_RootLP.senses` is built with one entry per `Constraint` **object**, but every use of it indexes an **evaluator/Jacobian row**. An array-valued constraint is one object and several rows, so on any model with a vector-valued constraint the list is shorter than the row count and the constructor raises `IndexError`.

`generate_root_cuts` is called inside a `try: ... except Exception as e: logger.debug("NLP-BB root cuts skipped: %s", e)` (`python/discopt/solver.py:17997`), so the failure is invisible: the entire root cutting-plane stage — graduated default-ON for NLP-BB by the 2026-08-20 panel (#1061/#1062) — is **silently skipped** for these models.

This is the same per-object/per-row class as #1297, found while fixing #1298 in the same file and deliberately kept out of that PR's scope.

## The code

`python/discopt/solvers/_root_cuts.py:227`:

```python
self.senses = [
    c.sense if isinstance(c.sense, str) else c.sense.value for c in model._constraints
]
```

and the three uses, all indexed by row (`i` ranges over `ja.shape[0]`, the Jacobian's rows):

- `_root_cuts.py:261` — `s = self.senses[i]`, choosing whether a linearised row becomes `A_le` or `A_eq`
- `_root_cuts.py:318` — `s = self.senses[row_idx]`
- `_root_cuts.py:329` — `s = self.senses[i]`

## Reproduction

Directly, on the class:

```python
import numpy as np
import discopt.modeling as dm
from discopt._relax.nlp_evaluator import NLPEvaluator
from discopt.solvers._root_cuts import _RootLP

m = dm.Model("rootlp")
x = m.continuous("x", shape=(3,), lb=0.0, ub=1.0)
y = m.continuous("y", lb=0.0, ub=1.0)
m.subject_to(x <= 0.5)      # ONE object, THREE evaluator rows
m.subject_to(y >= 0.2)
m.minimize(x.sum() + y)

ev = NLPEvaluator(m)
_RootLP(m, ev, np.zeros(4), np.ones(4), np.zeros(4, bool), np.zeros(4, bool), False)
```

```
evaluator rows = 4; per-object senses = ['<=', '<='] (len 2)
_RootLP RAISED IndexError: list index out of range
```

[...]

Issue #324, opened 2026-06-25
differentiable_solve crashes on models with mixed-shape (vector) constraints


jkitchin:
## Summary

`discopt._jax.differentiable.differentiable_solve` raises `TypeError` during its envelope-theorem sensitivity step whenever the model contains **vector constraints of differing shapes** (e.g. `x <= a` with `x` shape `(3,)` and `y <= b` with `y` shape `(2,)`). This makes it unusable on any model whose constraints are not all the same shape, including every `discopt.dae` collocation model (which mixes collocation residuals of shape `(nfe, ncp)` with continuity rows of shape `(nfe,)`).

Version: discopt 0.4.0, jax 0.4.x, Python 3.12.

## Minimal reproducer

```python
import discopt.modeling as dm
from discopt._jax.differentiable import differentiable_solve

m = dm.Model("b1")
x = m.continuous("x", shape=(3,), lb=0, ub=1)
y = m.continuous("y", shape=(2,), lb=0, ub=1)
p = m.parameter("p", 1.0)
m.subject_to(x <= 0.5 * p)   # constraint body shape (3,)
m.subject_to(y <= 0.5 * p)   # constraint body shape (2,)
m.minimize(dm.sum(x) + dm.sum(y))

differentiable_solve(m, nlp_solver="ipm")
```

## Traceback

```
File ".../discopt/_jax/differentiable.py", line 447, in differentiable_solve
    sensitivity = np.asarray(grad_lagrangian_p(p_flat))
File ".../discopt/_jax/differentiable.py", line 439, in lagrangian_p
    con_vals = jnp.array([cf(x_star_jax, p_flat_arg) for cf in constraint_fns])
TypeError: Cannot concatenate arrays with shapes that differ in dimensions
other than the one being concatenated: concatenating along dimension 0 for
shapes (1, 3), (1, 2).
```

## Root cause

In `lagrangian_p` (differentiable.py:439) the per-constraint values are combined with `jnp.array([...])`, which tries to **stack** them into a single dense array. That only works when every constraint compiles to the same shape. With heterogeneous constraint shapes, `jnp.array` falls through to `lax.concatenate` and fails.

## Expected behavior

`differentiable_solve` should support models with mixed-shape (vector) constraints. The NLP solve itself succeeds; only the sensitivity assembly fails.

## Suggested fix

Flatten/ravel each constraint value before combining, and weight by the matching multipliers, e.g.

```python
[...]

PR #1306, opened 2026-09-17
fix(correctness): adversary batch 4 — #1298–#1299

### LINKED ISSUE #1298: crash: segfault in root GMI separator (_root_cuts.py getBasicVariables) on a MINLP with no linear rows
> Discovered by the discopt adversary agent (automated solver correctness testing).

## Summary

`Model.solve` **segfaults** (the process dies with SIGSEGV, exit -11) on a tiny convex MINLP with a `max(v, c) - v <= 0` row and an integer variable. The crash happens in the root GMI separator.

```
python -X faulthandler -c "import discopt.modeling as dm; m=dm.Model('D'); v=m.continuous('v',lb=0,ub=4); s=m.integer('s',lb=0,ub=3); m.minimize(v-1.5*s); m.subject_to(dm.maximum(v,1.0)-v<=0); print(m.solve(time_limit=30).status)"
```

faulthandler points at `python/discopt/solvers/_root_cuts.py:417` (`separate_gmi`, `h.getBasicVariables()` in highspy). The call path is `generate_root_cuts`, called from `solver._solve_nlp_bb`.

## Evidence

- **Default settings:** exit -11 with "Fatal Python error" in `_root_cuts.py`. Reproduced in a subprocess by `adversary/runs/2026-09-17_verify_advC.py` (U2).
- **With `DISCOPT_NLPBB_ROOT_CUTS=0`:** exit 0, `optimal` with objective -3.5, which is correct (v=1, s=3).
- Per adv-C, there is no crash if any linear row is added or if `max` is replaced by `exp` or `v**2`. The likely cause is that the root cut LP has **no rows**, and `getBasicVariables` is called on an empty or unsolved basis.

## Expected

The GMI separator returns no cuts, and never calls into HiGHS basis accessors, when the LP has zero rows or no valid basis. A native crash must never take down the user's process.

### LINKED ISSUE #1299: correctness: Sensitivity.objective and SolveResult.gradient have the wrong sign under maximize (#1251)

jkitchin:
Fourth and final batch from this week's adversarial pass over the merged PRs. Two issues, one commit each.

## #1298 — `Model.solve()` died with SIGSEGV in the root GMI separator

A tiny convex MINLP with a `maximum(v, 1.0) - v <= 0` row crashed the interpreter (exit -11, no traceback).

Root cause is a HiGHS precondition, established with a five-line pure-`highspy` reproduction and a systematic sweep over row counts and matrix occupancy: `getBasicVariables()` reads a basis that was never factorized when the LP has rows but **zero matrix entries** — HiGHS returns `kOptimal` after zero simplex iterations and never builds one. Measured: `nnz == 0` crashes at `num_row` 1, 2 and 3; `num_row == 0` is safe; one empty row beside an ordinary row is safe.

`_RootLP` reaches that shape on its own: it classifies a constraint whose Jacobian is constant over its sampled points as *linear*, and this row's coefficients are all zero.

The fix is a precondition check, `_basis_accessors_are_safe(h)`, at the single accessor chokepoint in `separate_gmi`. When it declines, the separator returns no cuts and says why at DEBUG. Declining costs no correctness — GMI cuts are optional, and the B&B is unaffected.

Proved load-bearing: the end-to-end reproduction exits 139 without the guard and 0 with it.

## #1299 — sensitivity objectives and gradients had the wrong sign under `maximize`

Every evaluator and kernel in discopt minimizes, so a MAXIMIZE model is solved as `-f`. Six user-facing sites copied that internal value out without undoing the flip:

- `Sensitivity.objective` (`sensitivity.py`), which took POUNCE's raw objective;
- `differentiable_solve` — both `DiffSolveResult.objective` and the envelope gradient behind `.gradient(p)`;
- `_compute_sensitivity_at_solution`, which backs `SolveResult.gradient` and the fix-and-differentiate integer path;
- `differentiable_solve_l3` — its objective, its L1 envelope gradient (including the `_perturbation_gradient` fallback, which differences that same internal value) and `implicit_gradient`, which differentiates the negated objective function.

The issue named the first three. The L3 path is the same defect; it surfaced because the #324 vector-constraint tests had pinned the wrong convention in their own expectations and so passed throughout.

On `max -(x^2) + p*x` over `x in [-10, 10]` at `p = 2` (so `V(p) = p^2/4 = 1` and `dV/dp = p/2 = 1`):

| | before | after |
|---|---|---|
| `Sensitivity.objective` | -1.0 | **1.0** |
| `DiffSolveResult.objective` | -1.0 | **1.0** |
| `.gradient(p)` | -1.0 | **1.0** |
| `Sensitivity.dobj_dp` | 1.0 | 1.0 |

`dobj_dp` never had the bug — it differentiates the model's own objective expression — which is what makes it the cross-check in the new tests.

Rather than a sixth hand-written sense check, the fix adds one helper, `discopt.modeling.core.objective_sense_sign(model)`, and multiplies by it at each reporting boundary. The internal negation is untouched: the multipliers and the stationarity rows still belong to `-f`, as they must.

### On the #324 test changes

`test_differentiable_vector_constraints.py` is corrected, not weakened. The finite-difference cross-check — the multiplier-alignment property #324 is actually about — is unchanged and still passes. Only three hardcoded min-convention constants change: `-2.1, -2.1, -2.0` become `+2.1, +2.1, +2.0`, which is what those maximize models' value functions actually do.

## Rebased onto the updated batch-3 branch

This branch was rebased onto `fix/adversary-2026-09-17-c` at `ad78e8ef`, so it now
carries batch 3's three test repairs as well. It is stacked on #1303 and should
merge after it; until then its diff against `main` shows both batches.

The three inherited repairs, in order: the Benders C3 test now uses a bound at the
`1e20` INF sentinel instead of `ub=None` (which used to pass only because
`np.asarray(None, dtype=float)` is NaN); the row-filter test discriminates on
`DISCOPT_RELAX_ROW_FILTER` rather than the now-inert loose-bound flag, with the
retraction written into its docstring; and `clear_registered()` no longer drops
other modules' registered atoms, which is what made
`test_1293_registered_kinked_curvature` fail in CI's xdist lane and nowhere else.

Verified on the rebased head: the CI fast lane reproduced locally under `-n 4`,
10850 passed / 0 failed, against `2 failed` on the pre-repair tree.

## Regression tests

Both fixes ship with tests that fail at the parent commit and pass at the head, each verified in a separate worktree with the fix's marker asserted absent.

- `python/tests/test_1298_root_gmi_empty_matrix.py` — 4 tests. Three discriminate on the guard directly (refused with no entries; allowed at `num_row == 0`, with entries, and with one empty row beside a real one; `separate_gmi` declines and logs instead of crashing). The fourth is `@pytest.mark.slow` and runs the reported model in a subprocess asserting `returncode == 0` — since #1297 that model routes to spatial B&B and no longer reaches the separator, so it locks the reported symptom while the guard-level tests carry the discrimination.
- `python/tests/test_1299_maximize_sensitivity_sign.py` — 13 tests over all four entry points. Every case is parametrized max/min against a minimize twin with the same argmin, so a blanket negation fails the controls. Pre-fix: 8 failed, 5 passed. Post-fix: 13/13.

## What was run

- `pytest -m smoke python/tests/` — 1740 passed, 2 skipped, 11106 deselected (0:06:01)
- `pytest -m slow python/tests/test_adversarial_recent_fixes.py` — 19 passed (0:02:24)
- No Rust was touched in this batch, so `cargo test -p discopt-core` was not re-run. (It was run green at 756 passed / 0 failed on the batch-3 head, which is this branch's parent.)
- Targeted: 144 tests across the root-cut and NLP-B&B suites for #1298; 265 across the sensitivity, differentiable, argmin, parametric and convexity suites, plus 438 across every other file referencing `differentiable_solve`, `.gradient(`, `.sensitivity(`, `dobj_dp` or `implicit_gradient`, for #1299.
- `ruff check python/` and `ruff format --check` clean. mypy reports no error in any changed module; the 6 it does report are pre-existing, in `mpec.py`, `binary_multilinear_reform.py`, `surrogate.py` and `direct.py`.

## Related, deliberately not in scope

While working in `_root_cuts.py` for #1298 I found a second defect of the #1297 per-object/per-row class: `_RootLP.senses` is built per `Constraint` object and indexed per Jacobian row, so `_RootLP` raises `IndexError` on any model with an array-valued constraint, and `solver.py`'s enclosing `except Exception` swallows it at DEBUG — the whole root-cut stage is silently skipped. Reproduced both directly and end-to-end against a scalar control. Filed as #1305 rather than expanded into this branch.

Closes #1298
Closes #1299


jkitchin:
## Review — holding this one, and it is larger than it looks

### First: this PR contains all of #1303

The description says "two issues, one commit each", but the base is `main` (`f315471`), not #1303's branch, so the diff carries **seven** commits — `e72d627`, `75b1ba3`, `e8746ea`, `2c26718`, `243da503` (all of #1303, covering #1293–#1297) plus `091e8a0` and `2012e5e`. Merging this merges #1303's content too, including what I've just held #1303 on. Either retarget the base onto `fix/adversary-2026-09-17-c`, or treat the two as one review.

### Blocking: the LOA master can be tighter than the true feasible set

`#1297` added an `n_rows == 1` gate to the extractor call in `oa.py:1415`. The identical call at `gdpopt_loa.py:121` is ungated, and the comment above it asserts the belief that gate exists to disprove — that `_extract_body_coeffs` returning `(c_vec, offset)` "*is* the proof the body is one linear row".

Measured at this head (`2012e5e`), one `Constraint` over an evaluator row map of `[(0, 2)]`:

| rhs | `_extract_body_coeffs` → master row | true feasible set | |
|---|---|---|---|
| `[1.0, 2.0]` | `c·x <= 3.0` | `c·x <= 1.0` | looser — harmless |
| `[-1.0, -2.0]` | `c·x <= -3.0` | `c·x <= -2.0` | **tighter — invalid** |

The extractor collapses both rows into one and sums the rhs. When the rhs entries are negative the master row is strictly tighter than the true set, so the LOA master stops being a relaxation and its bound can rise above the true optimum. The sign of the rhs decides, so it fails silently and data-dependently. `oa.py` is right; `gdpopt_loa.py` needs the same gate plus a regression test over `evaluator.constraint_row_map()`.

### Also blocking-ish: `gdpopt_loa.py:628` drops `==` cuts

Now that real senses reach `_add_oa_cuts`, its `if cut.sense == "<=" / elif ">="` chain has no `==` arm, so an equality cut falls through both branches and is silently discarded. Compare `oa.py:3088` and `oa.py:3567`, which handle `==` explicitly (relaxing to `<=` under `equality_relaxation` while keeping `original_sense` for the global-validity test). Dropping a cut is safe for the bound on its own, but combined with the gate above an affine equality over several rows can end up represented nowhere, and the loop exits `unknown` on a feasible model.

The `>=` arm added in the same hunk is a genuine fix — the old `<=` default emitted the reversed inequality for a concave `>=` row — so that part is right and worth keeping.

### #1295's gate is broader than the commit message says (applies to #1303 too)

The commit message owns the trade: *"every model in the gated class loses its certificate, including big-M rows `y - M z <= 0` with M above ~1e6 on an unbounded y. That trade is deliberate."* The part that looks unintended is **which** columns count as open. `open_column_coefficient_ratio` (`lp_milp_highs.py:924`) tests `np.abs(sf.xl) >= READBACK_LIMIT | np.abs(sf.xu) >= READBACK_LIMIT` over the *standard form*, whose inequality **slack** columns carry the 1e20 sentinel as `ub`. So essentially every inequality row has an open column in it, and the gate degenerates to "some inequality row has a coefficient range past ~1.05e6" — independent of whether any *structural* column is actually unbounded.

Verified end-to-end on this head: the same model at M=1e5 returns `optimal / -48`; at M=1e7 it returns `feasible`, `gap_certified=False`, `bound=-49.99999`.

Two things follow. The measured calibration (97 false certificates at ≤5.3e-7 vs 27 real MILPs at ≥3.4e-5) was taken over a population where this widening was already in effect, so it does not tell us how many *ordinary* models fall in the gate. And the Rust twin `has_unscalable_tiny_entry` does **not** trip on this family, so the two routes now disagree about the same matrix. Restricting the open test to structural columns, and using the signed form already used at lines 666/727/898-899 (a bounded column like `xl = 1e16, xu = 1e16+10` currently reads as open), would bring them back in line.

Separately, `milp_driver.rs:1312` has the mirror-image problem: a row's own **logical** column is "open" because `_relax/std_form.py:182` gives a non-`eq` logical `ub = inf_value`, so any row whose max coefficient exceeds 1e10 decertifies the whole MILP. Excluding logical/slack columns on both sides looks like one fix, not two.

### Minor

`lp_milp_highs.py:1279` — `decertify()` records `milp/decertified_unscalable = 1.0` and rewrites `milp/bound_provenance` *before* the two early returns that withdraw nothing (Farkas-proved infeasible; `unbounded`/`error`). No-op runs are logged as decertified, and `solver.py` warns with a stale or empty reason.

### Cleared

The #1299 sign fix is correct at every boundary I could find: `_compile_parametric_objective`, `nlp_result.objective`, `_perturbation_gradient` → `_solve_model_nlp`, and `argmin._build_layer`'s `last_solve_info["objective"]` are all internal-minimization values, and there is no double-flip in `SolveResult.gradient`, `_differentiable_solve_integer`, `implicit_gradient` or `make_milp_objective_layer`. Using `dobj_dp` as the cross-check — the one quantity that never had the bug — is the right test design, and the #324 constant changes are a correction rather than a weakening.

The #1298 guard is sound: `separate_gmi` is the only basis-accessor site, the guard precedes all three accessors, and `a_matrix_.start_` is never empty (a 0-column 1-row LP gives `start_ = [0]`). The five-line highspy reproduction and the row-count/occupancy sweep are exactly the right evidence, and the note that the end-to-end test no longer reaches the separator — so the guard-level tests carry the discrimination — is the honest way to write it up.

#1293's kink-site walk holds up: attribute names check out against `core.py`, `min`/`max` cannot reach it because `symbolic_diff` refuses them first, and the out-of-domain NaN cases are caught downstream by `_emit_1d`'s `_finite` guard. #1296's Rust side is sound — all `solve_milp*` entry points funnel through the new wrapper, `withdraw_certificate`'s match is exhaustive, and node `lower_bound` includes `obj_const` so the recomputed `cᵀx + obj_const` is consistent.

---

Issue #1323, opened 2026-09-18
abs_gap_tolerance still silently inert on the Rust MILP re-entry run and the convex-MINLP OA route (#1315 follow-up)


jkitchin:
> Found by the discopt adversary agent (round 3: adversarial testing of the fixes merged in #1317 / #1318). Report-only. The reported bounds were valid in every case below: these are silent *inert-option* failures, not false certificates.

#1315 connected `abs_gap_tolerance` to the Rust MILP engine. Two routes still drop it silently.

## 1. The #698 re-entry run of the Rust MILP engine drops `abs_gap_tol`
- `_solve_milp_simplex` calls `solve_milp_csc_py` twice.
- The first call (`solver.py:24248`) passes `abs_gap_tol=`. The re-entry call at `solver.py:24374` does not.

**Repro** (`adversary/runs/2026-09-17_r3_absgap_reentry.py 234 0 30`): a 150-binary × 15-row correlated knapsack, solved with `nlp_solver="simplex"`, `time_limit=30`, `abs_gap_tolerance=234`.
- Run 1 hit the 10 s budget cap with an absolute gap of about 234.7.
- Run 2 was called without `abs_gap_tol` and used the full 19.9 s.
- The final result was `feasible`, `gap_certified=False`, with |obj−bound| = 233.27 ≤ 234.

**Expected:** stop once the absolute criterion is met, and report optimal within the tolerance. **Oracle:** highspy at gap 0.

## 2. The default-ON convex-MINLP OA auto-route (#1059, `mip-nlp/oa`) ignores it
- `solve_mip_nlp` receives only `gap_tolerance` (`solver.py:~9066`).
- `_MIP_NLP_IGNORED_OPTIONS` (`~5662`) does not list `abs_gap_tolerance`, so the option is neither honoured, refused nor warned about.

**Repro** (`adversary/runs/2026-09-17_r3_absgap_routes.py`): a convex MINLP with K=12 using exp and quadratic terms, indicator binaries and a cardinality constraint.
- With `abs_gap_tolerance=1e4`, OA ran the same trace as the baseline and stalled with LB=-220.51, UB=-34.436. That gap is 186, well inside the requested 1e4.
- Only the spatial-B&B fallback honoured the tolerance: 3 nodes versus 241 in the baseline.
- **Oracle:** enumeration over the binaries plus scipy gives an optimum of -34.436.

## Not run, code reading only
[...]

Issue #1321, opened 2026-09-18
correctness: saved_bounds read-only invariant still breaks after load(), inside fixed() scopes, and for GAMS-read models (#1311 follow-up)


jkitchin:
> Found by the discopt adversary agent (round 3: adversarial testing of the fixes merged in #1317 / #1318). Every repro below was reproduced independently on `main` @ 4ede7c85. Report-only.

#1311 made `fix()` push read-only copies, so that one fix/unfix cycle no longer breaks `Model.saved_bounds(copy=False)`'s invariant: an in-place bound write must raise, not silently corrupt the by-reference snapshot. That fix covers a single call site. The invariant still fails on several other paths, and on each of them the corruption changes the solve's answer.

## 1. `load()` brings the fix stack back writable
```python
import discopt.modeling as dm
from discopt.serialize import dumps, loads
m = dm.Model("t"); x = m.continuous("x", lb=0, ub=10); y = m.continuous("y", lb=-5, ub=5)
m.subject_to(x + y <= 8); m.minimize(x + 2*y)
with x.fixed(4.0): s = dumps(m)
mm = loads(s); xx = mm._variables[0]; xx.unfix()
xx.lb.flags.writeable          # True
with mm.saved_bounds(): xx.lb[()] = 7.5
float(xx.lb), mm.solve().objective   # (7.5, -2.5)  -- expected (0.0, -10)
```
- The corrupted value also becomes the declared domain: a later `mm.fixed(x=3)` is refused as "outside declared bounds [7.5, 10]".
- **Cause:** `serialize.py:584-587` rebuilds `_bound_stack` from `_dec_array`, which returns writable arrays.

## 2. While a variable is fixed, its live bounds are writable
```python
with x.fixed(3.0):
    with m.saved_bounds(): x.lb[()] = 1.0    # accepted, persists
    float(x.lb), float(x.ub), m.solve().objective   # (1, 3, -9) -- expected (3, 3, -7)
```
- `fix(where=...)` behaves the same way.
- **Cause:** `core.py:~1453/1457`. `new_lb`/`new_ub` are plain writable `np.array`/`np.where` results.

## 3. The GAMS reader makes declared bounds writable
- For `x.lo=1; x.up=10`, `x.lb.flags.writeable` is True.
- An in-place write inside `saved_bounds()` survives, and the solve returns 5.0 instead of 1.0.
- This predates #1311; it is a sibling path with the same problem.
[...]

Issue #1309, opened 2026-09-17
correctness: DISCOPT_LP_MILP_BACKEND=rust opt-out returns a false certified-infeasible LP; HiGHS MILP root cross-check can be silently skipped


jkitchin:
Found during adversarial testing of #1258 ("route pure LP/MILP to HiGHS with verified certificates").

Extensive fuzzing of the #1258 HiGHS route itself (`lp_milp_highs.py`) held up well: 4000 randomized single-row huge-box trials, 800 multi-row huge-box trials, and 1500 pure-integer MILPs cross-checked against brute force all came back with 0 mismatches. The #1259/#1261 fixes hold. Two problems remain in the surrounding routing/certification logic.

## 1. `DISCOPT_LP_MILP_BACKEND=rust` opt-out returns a false certified-infeasible on a trivially feasible LP

Repro (fresh process):
```python
import discopt.modeling as dm
m = dm.Model("false_infeasible_demo")
v0 = m.continuous("v0", lb=-1e16, ub=1e16)
v1 = m.continuous("v1", lb=5e15, ub=2e18)
v3 = m.continuous("v3", lb=-5e15, ub=1e16)
m.subject_to(5 * v0 - v1 - v3 >= -2, name="r0")
m.minimize(5 * v1 + 2 * v3)
res = m.solve()
print(res.status, res.objective, res.gap_certified)
```
- `DISCOPT_LP_MILP_BACKEND=rust` → `status=infeasible, gap_certified=True`
- default (unset, #1258's HiGHS route) → `status=optimal, objective=1.5e16, gap_certified=True`

Ground truth: v0=-2/5, v1=5e15, v3=-5e15 satisfies the row exactly (`5·v0 - v1 - v3 = -2 ≥ -2`) and every bound, giving objective 1.5e16 — matching HiGHS. The "rust" route's `infeasible, gap_certified=True` is a **false certified infeasible**, violating CLAUDE.md §1 ("never a false optimal/infeasible, full stop").

[...]

Issue #1286, opened 2026-09-17
correctness: pure LP route certifies 'unbounded' on a finite LP — primal_ray_verified relative test misses a 1e-12 coupling (#1258)


jkitchin:
> Discovered by the discopt adversary agent (automated solver correctness testing).

## Summary

The pure LP route returns **`unbounded` for an LP with a finite optimum of 0**, on both the HiGHS route (#1258) and the `DISCOPT_LP_MILP_BACKEND=rust` route.

PR #1258 says a HiGHS answer produced after a kWarning coefficient drop is refused unless discopt verifies it. Here the recession ray is accepted, because `primal_ray_verified` (`python/discopt/solvers/lp_milp_highs.py` ~451) tests `|A d| <= RAY_REL * |A||d|` with `RAY_REL = 1e-9`. That relative test cannot see a row whose only coupling to the ray is a 1e-12 coefficient.

## Reproducer (main a5887a4c)

```python
import discopt.modeling as dm
m = dm.Model("A")
x1, x2, x3 = (m.continuous(n, lb=-1e20, ub=1e20) for n in ("x1", "x2", "x3"))
m.subject_to(x1 + x2 == 0)
m.subject_to(x1 + x2 + 1e-12 * x3 == 0)
m.maximize(x3)
m.solve()   # status 'unbounded', route 'highs-lp: verified HiGHS route (HiGHS kUnbounded)'
```

The rust route also returns `unbounded`. The adversary saw the same result for eps = 1e-12, 1e-13 and 5e-13. The MILP variant returns `error` on HiGHS, which is a sound refusal. Solver stats show `lp/highs_pass_warning=1` and `lp/recession_ray_lp=1`. The accepted ray is d = [1, -1, 1], whose exact residual on row 2 is 1e-12.

## Independent confirmation

1. Exact rational elimination: row 2 minus row 1 gives `1e-12 * x3 = 0`, so x3 = 0 and the optimum is 0.
2. `scipy.optimize.linprog` on the equivalent scaled model (x1+x2=0, x3=0) returns optimal, 0.

## Notes

- A 1e-12 coefficient is below typical LP drop thresholds, so a refusal (`error` / numerical) is an acceptable outcome. A certified `unbounded` is not.
[...]

PR #1327, opened 2026-09-18
fix(correctness): #1319/#1320 stop certifying statuses whose cross-check never held

### LINKED ISSUE #1319: correctness: POUNCE LP/QP route still certifies false infeasible/unbounded on huge-bound problems (#1309 follow-up)
> Found by the discopt adversary agent (round 3: adversarial testing of the fixes merged in #1317 / #1318). Every repro below was reproduced independently on `main` @ 4ede7c85. No code was changed; this is report-only.

On huge-bound problems (bounds ≥ ~1e15) the POUNCE interior-point route can still report a **certified wrong status**. Part 1 is a gap in the #1309 fix, and parts 2–3 are sibling paths that fix did not cover.

## 1. The LP Phase-1 cross-check can certify a false `infeasible` (`DISCOPT_LP_MILP_BACKEND=rust`)

#1309 made Ipopt code 2 depend on the elastic Phase-1 LP, on the grounds that "a positive minimal violation is a Farkas certificate". That only holds if Phase-1 is solved exactly. Here Phase-1 is solved by POUNCE's IPM on a box whose ≥1e15 bounds were relaxed, and `_is_infeasible_violation` scales its threshold only with the RHS size, not with `|A||x|`.

```python
# DISCOPT_LP_MILP_BACKEND=rust
import discopt.modeling as dm
L = 9e14; m = dm.Model("pinch")
x1 = m.continuous("x1", lb=L, ub=3*L); x2 = m.continuous("x2", lb=-3*L, ub=3*L)
m.subject_to(x1 - x2 <= 1); m.subject_to(x2 - x1 <= -1); m.maximize(x1 + x2)
r = m.solve()   # status='infeasible', gap_certified=True
```
- **Expected:** optimal, 5.4e15. `scipy.optimize.linprog(method="highs")` returns status 0 with objective 5.4e15, and the point (L+1, L) is exactly feasible.
- **Default backend:** the same model returns an honest `error`.
- **Path:** raw POUNCE returns UNBOUNDED, then Phase-1 returns a slack sum of 0.053 against a threshold of about 4e-6, so INFEASIBLE is certified.
- **Scope:** it also reproduces at L=2e15 and 5e15.
- **Pointers:** `python/discopt/solvers/lp_pounce.py:516-523` and `:637-662` (`_is_infeasible_violation`). `solver.py:~21116` trusts INFEASIBLE with no check of its own.
- **Minor:** the comment at `lp_pounce.py:292-294` still says code 2 "is a sound INFEASIBLE".

## 2. `qp_pounce.solve_qp` still trusts raw Ipopt code 2, on the **default** route

This is the #1309 LP instance plus a separable convex term. No env var is needed:
```python
m = dm.Model("q")
v0 = m.continuous("v0", lb=-1e16, ub=1e16); v1 = m.continuous("v1", lb=5e15, ub=2e18)
v3 = m.continuous("v3", lb=-5e15, ub=1e16); w = m.continuous("w", lb=0, ub=1)
m.subject_to(5*v0 - v1 - v3 >= -2); m.minimize(5*v1 + 2*v3 + (w - 0.5)**2)
m.solve()   # status='infeasible', gap_certified=True
```
- **Expected:** feasible. The witness v0=-0.375, v1=5e15, v3=-5e15, w=0.5 gives row value -2.0 ≥ -2 exactly.
- Calling `qp_pounce.solve_qp` directly with Q=0 also returns INFEASIBLE.
- **Pointer:** `python/discopt/solvers/qp_pounce.py:173-190`. The `elif certificate and ... INFEASIBLE` branch was left unconverted, although the same branch in `lp_pounce` was converted.

## 3. QP route reports a bounded QP as `unbounded`

```python
L = 1e15; m = dm.Model("q3")
w = m.continuous("w", lb=0, ub=1); y = m.continuous("y", lb=-L, ub=0)
m.subject_to(w + y <= 1); m.minimize((w - 0.5)**2 + y)
m.solve()   # status='unbounded' (gap_certified=False); expected optimal, -1e15 (analytic)
```
- It is correct for L ≤ 9e14.
- The LP route has the #850 `_DeferredUnbounded` guard for bounds the IPM relaxes to infinity. `_solve_qp_matrix` (`solver.py:~21590`, UNBOUNDED arm ~21909) has no such guard and trusts the verdict.

## What "done" looks like
- The Phase-1 infeasibility threshold accounts for the magnitude of `|A||x|`, or an INFEASIBLE produced by an IPM Phase-1 on a relaxed box gets an exact re-verification before it is certified.
- `qp_pounce` gets the same code-2 cross-check as `lp_pounce`.
- The QP route gets the huge-bound unbounded guard.
- Each repro above becomes a regression test.

Probes: `adversary/runs/2026-09-17_r3_pounce_*.py`, `2026-09-17_r3_qp_*.py`. Related: #1309, #850.

### LINKED ISSUE #1320: correctness: HiGHS MILP route certifies when the NS-safe root cross-check runs but is inconclusive (#1309 follow-up)
> Found by the discopt adversary agent (round 3: adversarial testing of the fixes merged in #1317 / #1318). Report-only.

## 1. An *inconclusive* NS-safe root cross-check still certifies

#1309 (part b) downgrades a HiGHS MILP `optimal`/`infeasible` to uncertified when the root cross-check is **skipped** because `remaining() <= 0`. The same principle ("a certificate whose safety net never ran is not a certificate") is not applied when the check **runs but is inconclusive**. With a sliver of budget left:
1. `solve_lp_std` returns `time_limit` or `error`.
2. The code only sets `stats["milp/root_check_inconclusive"] = 1.0` (`python/discopt/solvers/lp_milp_highs.py:1500-1502`).
3. The result is returned unchanged: `optimal`/`infeasible`, `gap_certified=True`, `root_bound=None`.

I confirmed this by reading the code. It was reproduced with the same fake-clock technique as the fix's own test (`test_1309_highs_milp_root_check_skip.py`), with the clock sequence `[0, 1e-6, TL-1e-9, TL-1e-9, TL-1e-9, TL+1, ...]`:
- **Knapsack:** `optimal`, certified.
- **`2x = 1, x ∈ Z`:** `infeasible`, certified.
- **End to end through `Model.solve`:** `status=optimal, gap_certified=True`. Brute force gives 19.

A HiGHS root-LP `error` (`kUnknown`, sentinel-magnitude readback refusals) was seen on natural models. No natural kOptimal instance hitting this arm was found, so the end-to-end repro relies on the mock.

**Expected:** apply the same downgrade as the skip branch:
- kOptimal → uncertified `feasible`, keeping the incumbent
- kInfeasible → `error`

## 2. The downgraded result sends mixed signals

When the skip branch *does* downgrade, the user sees:
- `status="feasible"` (correct)
- `gap=0.0`, `bound_valid=True`, and a route label reading "verified HiGHS route"

Those three fields still say "certified". Downgraded results should report a consistent state: a gap that isn't a proven 0, and no "verified" wording.

Probes: `adversary/runs/2026-09-17_r3_highs_root_starved.py`, `2026-09-17_r3_highs_skip_e2e.py`. Related: #1309, #1295.

jkitchin:
## Summary

Two #1309 follow-ups, both the same failure: a certificate issued where the check
backing it did not actually hold. Plus, added while driving this to green, a fix
for a regression the first version of this PR introduced and the retry that
finally makes #1319's own repros solve.

**#1319 — POUNCE LP/QP route, huge-bound problems.** Three sibling defects:

1. **LP Phase-1 threshold.** #1309 made Ipopt code 2 depend on the elastic Phase-1
   LP because "a positive minimal violation is a Farkas certificate" — true only
   if Phase-1 is solved *exactly*, but it is solved by the same IPM on the same
   relaxed box, and `_is_infeasible_violation` scaled its threshold with the RHS
   alone, never with `|A|·|x|`. On a box reaching ~1e15 those differ by twenty
   orders of magnitude, so pure cancellation noise read as a Farkas proof and a
   feasible LP came back certified `infeasible`.
2. **QP code 2.** `qp_pounce.solve_qp` still trusted raw Ipopt code 2 on the
   *default* route — the `elif certificate and … INFEASIBLE` branch was left
   unconverted when the `lp_pounce` one was fixed.
3. **QP unbounded.** `_solve_qp_matrix` had no #850 huge-bound guard and trusted
   the verdict, calling a QP bounded below by -1e15 `unbounded`.

**Root cause behind all three.** `_FINITE_BOUND_THRESHOLD = 1e15` was four orders
of magnitude more conservative than the engine requires. Measured: POUNCE returns
the exact analytic optimum for every bound magnitude up to **9.9e18** and flips to
UNBOUNDED at exactly **1e19** — Ipopt's `nlp_{lower,upper}_bound_inf`. Because the
declared bound never reached POUNCE, the box it solved was strictly *larger* than
the one posed. This **falsifies the #850 Obs 1 premise** that the barrier "cannot
condition so huge a finite bound" for the whole `[1e15, 1e19)` range.

**#1320 — HiGHS MILP route.** #1309 applied "a certificate whose safety net never
ran is not a certificate" only to the branch where the root cross-check is
*skipped*. When it **runs and is inconclusive** the code only recorded
`milp/root_check_inconclusive` and returned `optimal`/`infeasible` with
`gap_certified=True`. Both branches now share one `decertify_root_check`, and a
downgraded result no longer sends mixed signals.

### Added while driving this to green

4. **A regression this PR's own fix introduced.** Raising the infeasibility bar by
   a row-activity floor left the *reading* of "under the bar" unchanged: the code
   took it as proof of **feasibility**. It is not. A genuine conflict whose
   violation is small in absolute terms sits under that floor whenever the
   Phase-1 point is large, so Phase-1 said "not infeasible", the raw code-3/4
   `UNBOUNDED` survived the ray check, and an **infeasible** problem was certified
   `unbounded` — trading a false `infeasible` for a false `unbounded`. Measured on
   `x - y <= 1` ∧ `x - y >= 10` over `x, y ∈ [1e15, 3e15]`: true violation 9, row
   activity 4e15, floor `1e-12 × 4e15 = 4000`. `main` returns infeasible on that
   input; the pre-fix branch did not. Fixed by making the verdict **tri-state** —
   `_phase1_verdict` returns infeasible / feasible / **undecided**, and both POUNCE
   routes refuse to certify anything on undecided.
5. **The retry that closes #1319.** The defects above were fixed on both flag
   settings, but on the default box two of #1319's own repros still ended at an
   honest `error`: correct, and useless. The route now re-solves once over the box
   the caller declared, at the two entry points where a POUNCE-route answer is
   abandoned.

Closes #1320. **Closes #1319.**

## Correctness

Every change *removes* a certificate or *tightens* the box; none adds a claim.
Honoring the declared box is the sound direction: a bound is a constraint, so
discarding it enlarged the feasible set and the old path solved a superset.

Two guards were explicitly **kept** when a draft fix would have weakened them: a
Phase-1 call with no point still certifies nothing in either direction (reading it
against the rhs alone restores the false `infeasible`; reading it as feasible
restores the false `unbounded`), and the two tests that pin "never `infeasible` on
a feasible QP / never `unbounded` on a bounded one" are untouched — they now carry
the stronger companion that anything *certified* must equal the known optimum.

| repro | before | flag OFF (default) | flag ON |
|---|---|---|---|
| #1319 part 1 (LP) | certified `infeasible` | `optimal 5.4e15` | `optimal 5.4e15` |
| #1319 part 2 (QP) | certified `infeasible` | `optimal 1.5e16` | `optimal 1.5e16` |
| #1319 part 3 (QP) | `unbounded` | `optimal -1e15` | `optimal -1e15` |
| infeasible @ 1e15 | — | not `unbounded` | not `unbounded` |
| #1320 kOptimal | certified `optimal` | uncertified `feasible` | same |
| #1320 kInfeasible | certified `infeasible` | `error` | same |

All three #1319 repros now return the correct certified optimum **with no env var
set**, 0 false certificates. 1.5e16 is exactly the witness #1319 itself gives.

- [x] `pytest -m smoke` → **1723 passed, 14 skipped**
- [x] `ruff check` / `ruff format --check` clean; `mypy` clean on the touched solver modules

Also: the POUNCE/HiGHS route, #850, #1286, #937, #1321 and #1323 suites together
**196 passed** on the merged tree.

## Regression test

`test_1319_pounce_huge_bound_false_certificates.py` and
`test_1320_highs_milp_root_check_inconclusive.py`, **40 tests** together.

- The tri-state fix adds 3 tests; the 2 unit ones fail before / pass after. The
  end-to-end one passed before as well and is labelled as a guard rather than a
  reproduction — on that construction the exact simplex decides it.
- The retry adds 7 tests, all fail before / pass after, including the gate's
  no-op property in both directions (below the window, and with the flag already
  on) and that the scoped override restores.

## Why the retry needs no §5 graduation

The retry runs **only** from a state that was about to report `error`, so no solve
that succeeds today can change. `_declared_box_retry_applies` gates it on the
legacy threshold being in force **and** the model declaring a bound in
`[1e15, 1e19)`; outside that window both arms build a bit-identical box, so it is a
*provable* no-op rather than a measured-neutral one. The feasibility, KKT and
Phase-1 cross-checks are unconditional and still gate whatever comes back.

This is the difference from flipping `DISCOPT_POUNCE_DECLARED_BOX`, which moves the
box for **every** solve in the window — including ones that currently succeed — and
is what §5's net-positive gate exists to measure.

- [x] Default-OFF flag: `DISCOPT_POUNCE_DECLARED_BOX` (unset/`0` = legacy 1e15, byte-identical path)
- [x] Differential-bound test + feasible-point sampling — see the panel below
- [x] **Not graduating — default-off only.** Gate 1 passes; gate 2 is unanswerable on this corpus

**#1319 no longer depends on that vote.** The retry recovers the affected class
without moving the default box, so the flag can stay off indefinitely without
leaving the issue half-fixed. Graduating it remains a separate, optional call.

**Measurement** — `scripts/pounce_declared_box_panel.py`, in-repo MINLPLib corpus,
arms interleaved within each instance, `max_nodes=300` + 60 s backstop:

```
GATE 1 CERT-CLEAN : PASS
GATE 2 NET-POSITIVE: INCONCLUSIVE

affected by the flag    : 2 of 66  (st_miqp3, st_miqp4, both at exactly 1e15)
newly certified by ON   : 0
DEcertified by ON       : 0
```

Gate 2 is inconclusive **by construction**: only 2 of 66 instances declare a bound
in `[1e15, 1e19)` — everything else is either small-bounded or default-boxed at
1e20, above POUNCE's own infinity — which is too few to establish "measurably
helpful broadly" either way. Reporting neutral-as-fail would infer a verdict from
a measurement that could not happen.

## Related defects found and filed, not fixed here

- **#1328** — `gurobi.py:194` relaxes `|b| >= 1e15` to `GRB.INFINITY` while
  `_solve_lp_gurobi`/`_solve_qp_gurobi` pass `relaxes_huge_bounds=False` on the
  premise that Gurobi honors such bounds. Both cannot be true, so the #850 guard is
  disabled on a false premise. Not fixed here: `gurobipy` is unavailable in this
  environment, and blind-patching an unverifiable optional backend is how the
  original guess survived.
- **#1335** — the rust LP backend certifies `optimal` on an infeasible LP at ~1e15,
  returning a point that violates its own row. **Pre-existing on `main`** — verified
  identical against a pristine `origin/main` worktree — so it does not block this
  PR, but it is a live false certificate and may share the magnitude-relative
  tolerance flaw behind #1319.

https://claude.ai/code/session_01PytPusRx2DkEXZqCN1AQmT


jkitchin:
## Round-4 adversarial review: blocking, this PR adds a new false status

Tested at branch head f0a6924d. Each run asserted `discopt.__file__` and the marker `finite_bound_threshold` in `qp_pounce.py`: present on the PR tree, absent on main d4e0bb08. I reproduced finding 1 independently.

### 1. Correctness: `unbounded` reported on an INFEASIBLE LP/QP (main is correct)
```
a, b in [0,10];  a - b <= 0;  a - b >= 1          # contradictory
z in [L,2L], v in [0,2L];  z - v <= 0              # unrelated, satisfiable, L = 1e12
y in ±1e20, w in [-1,1]
min (w-0.5)^2 - y   (QP)   /   min -y   (LP)
```
Oracle: `scipy.optimize.linprog(method="highs")` status 2 (infeasible).

| route | PR (flag OFF and ON) | main |
|---|---|---|
| QP default | **unbounded** | infeasible, gap_certified |
| LP, `DISCOPT_LP_MILP_BACKEND=rust`, nlp_solver=pounce | **unbounded** | infeasible, gap_certified |
| LP, rust, simplex first | **unbounded** | infeasible, gap_certified |
| LP, default HiGHS | infeasible | infeasible |

- This appears at L ≈ 6e11, far below the 1e15 window, so the new flag has no effect.
- Cause: the new noise floor in `_is_infeasible_violation`, `1e-12 * Σ_i (|A||x|)_i`, is summed over all rows. One large satisfiable row lifts the floor above the O(1) violation of a separate well-scaled block. Phase-1 then counts as feasible, and `_settle_ambiguous_unbounded` → `_certify_unbounded_ray` certifies y's ray. That certifier is only sound once feasibility is established.
- This contradicts the claim that the PR only removes certificates.
- Suggested fix: a per-row floor, `slack_i > tol*(1+|rhs_i|) + 1e-12*act_i`. The ray certifier should also require a positive feasibility proof, not merely "Phase-1 did not prove infeasible".
- Probe: `adversary/runs/2026-09-18_r4_d_false_unbounded_min.py` (run with `TREE=pr|main`).

### 2. Regression vs main: correct `infeasible` certificates lost on the POUNCE LP/QP path
- `x1 >= L, x1 - x2 <= 0, x1 - x2 >= 1` for L ≥ 1e12 returns `error` on LP rust+pounce and on QP.
- `K*x1 - K*x2 <= 0, x1 - x2 >= 1` with x1 in [1000,2000] and K ≥ 1e9 returns `error` on the QP default route.
- Main certifies `infeasible` on both. This has the same root cause as finding 1.
- Probes: `_r4_d_infeas_offset.py`, `_r4_d_decoupled.py`.

### 3. Regression vs main: HiGHS MILP loses a correct `infeasible` when the root LP is unbounded
- Repro: `x integer in [0,5]; 2x == 1; y in [0,1e20]; min -y`.
- The PR returns `error` (route "unverified … kInfeasible"). Main returns certified `infeasible`, and brute force agrees.
- This is defensible under the PR's rule, but an unbounded root LP combined with kInfeasible from the MILP could be settled by checking integer feasibility directly. At least document it and add a test.
- Probe: `_r4_d_milp.py`.

### 4. Cosmetic
- The docstring of `_declared_box_relaxed_to_ipm_inf` (solver.py ~21262) says the flag is "on (the default since #1319)", but it is default-OFF.
- The QP `error` warning still says "[1e15, 1e20)" and "Tighten the bounds below 1e15" when the flag is ON, where the window is [1e19, 1e20).

### What held up
- **#1320, forced-inconclusive root check.** `solve_lp_std` was monkeypatched in the probe only, returning time_limit, error, unbounded or feasible-without-bound.
  - Every result was correctly downgraded, with no gap, bound or verified label. Main certified all 8 combinations.
  - The unpatched knapsack is still certified optimal at 51, matching brute force. 22 assertions.
- **Feasible huge-box LP/QP families,** L from 1e10 to 1.5e19, both flags: no false status on the PR tree (130 assertions).
- **A main-only false `unbounded`** on `min (w-.5)^2 + y, y ∈ [-L,0]` for L ≥ 2e15 is fixed by this PR.

Recommendation: do not merge until finding 1 is fixed with a regression test. Findings 2 and 3 should be fixed, or explicitly accepted in the PR body.

jkitchin:
## Round-4 findings addressed — head is now `6cf1848`

Thanks, finding 1 was real and your diagnosis was sharper than mine. I had independently found the same class of bug at `f0a6924` and fixed it in `a7caa6b` by making the Phase-1 verdict tri-state, which removed the false `unbounded` — but I attributed it only to reading "under the floor" as proof of feasibility. I missed that the floor was **summed over rows**, which is the part that makes an O(1) contradiction disappear behind an unrelated 1e12 row at magnitudes nowhere near the declared-box window.

### 1. False `unbounded` on an infeasible LP/QP — **fixed**

Reproduced your case at `f0a6924` and traced it to exactly what you describe. Fixed in `6da50ec` with your suggested per-row form, `slack_i > tol*(1+|rhs_i|) + 1e-12*act_i`:

| route | before | now | main |
|---|---|---|---|
| QP default | **unbounded** | `infeasible`, certified | `infeasible` |
| LP, rust backend | **unbounded** | `infeasible`, certified | `infeasible` |
| LP, default HiGHS | `infeasible` | `infeasible` | `infeasible` |

Both halves are kept, because they fix different things: the per-row floor stops one row's arithmetic excusing another's, and the tri-state stops "under this row's own floor" being read as proof of feasibility (that is what let the ray certifier run at all). Regression tests: `test_large_unrelated_row_is_never_certified_unbounded`, `test_phase1_floor_is_per_row_not_summed`, `test_per_row_floor_restores_the_infeasible_certificate`.

The sweep behind the `1e-12` constant only ever varied one magnitude at a time, so it structurally could not have caught this. That limitation is now recorded beside the constant.

On your second suggestion — the ray certifier requiring a *positive* feasibility proof rather than "Phase-1 did not prove infeasible" — I agree, and it is the right long-term shape. It is not in this PR: see #1336.

### 2. Lost `infeasible` certificates — **partly fixed, remainder accepted and filed**

The per-row floor restores these at L ≈ 1e12 (both your `_r4_d_decoupled` shape and the offset one). At larger magnitudes the cause turned out to be different from the floor entirely — traced:

```
raw _solve_qp_core     -> SolveStatus.INFEASIBLE
_phase1_min_violation  -> None          <- no point, nothing to verify against
FINAL                  -> error
```

Phase-1 does not solve, so a raw Ipopt code 2 has nothing to be cross-checked against. `main` returns `infeasible` by trusting that code 2 unchecked — correct in fact, and exactly what #1319 part 2 exists to stop trusting. I have taken the honest `error` rather than reinstating an unverified certificate, pinned it with `test_unverifiable_infeasible_is_an_error_not_an_unchecked_certificate`, and filed **#1336** with the remedy your review points at: the constraint system is linear even on the QP route, so the exact simplex can supply the positive feasibility proof and would beat both the current `error` and main's unverified label.

### 3. HiGHS MILP integral infeasibility — **accepted, tested, filed**

Confirmed: `error` on this branch, certified `infeasible` on main. Taking your "at least document it and add a test" option — `test_integral_infeasibility_under_an_unbounded_root_lp_is_not_certified` asserts the §1 line (never `optimal`, never `unbounded`) and accepts `infeasible` or `error`, so it stays deliberate. Filed as **#1337**, noting what I think is the cheap fix: the root check here is defeated by the *objective* being unbounded, not by anything about feasibility, so re-running it with the objective dropped would likely decide it outright.

### 4. Cosmetic — **both fixed**

`_declared_box_relaxed_to_ipm_inf`'s docstring no longer claims the flag is "the default since #1319" (it is default-OFF), and the QP warning now interpolates the live threshold instead of hardcoding `[1e15, 1e20)` / "below 1e15".

### Also in this PR since your review

- `1c770a7` closes the rest of #1319: where the route used to dead-end at `error` because it discarded a declared bound, it now re-solves once over the declared box. All three of #1319's repros return the correct certified optimum **with no env var set** (5.4e15 / 1.5e16 / -1e15). It runs only from a state that was already `error`, gated on the model declaring a bound in `[1e15, 1e19)`, so it is a provable no-op elsewhere and needs no §5 graduation — the flag default stays OFF.

### Verification on `6cf1848`

- `pytest -m smoke` → **1723 passed, 14 skipped**
- `pytest -m slow test_adversarial_recent_fixes.py` → **19 passed**
- POUNCE/HiGHS route + #850 + #1286 + #937 + #1321 + #1323 suites → **213 passed**
- `ruff` clean, `mypy` clean on the touched solver modules
- Your finding-1 repro, all three routes: no false `unbounded`, `infeasible` certified

One more, unrelated to your findings and **pre-existing on `main`** (verified against a pristine `origin/main` worktree, so not a blocker here): the rust LP backend certifies `optimal` on an infeasible LP at ~1e15 and returns a point violating its own row — **#1335**. It may share the magnitude-relative-tolerance shape behind #1319.

---

Issue #1337, opened 2026-09-18
HiGHS MILP route loses a correct `infeasible` when the conflict is integral and the root LP is unbounded


jkitchin:
Split out of the round-4 review of #1327 (finding 3), where it is pinned as an accepted trade rather than fixed. Sound, but weaker than `main`.

## What happens

```python
import discopt.modeling as dm
m = dm.Model("f3")
x = m.integer("x", lb=0, ub=5)
m.subject_to(2 * x == 1)      # no integer solution
y = m.continuous("y", lb=0, ub=1e20)
m.minimize(-y)                 # root LP relaxation is unbounded in y
print(m.solve(time_limit=60).status)
```

| tree | result |
|---|---|
| #1327 branch | `error` (route label "unverified … kInfeasible") |
| `main` | `infeasible`, certified — and brute force agrees |

## Why

The infeasibility is purely **integral**. The root LP relaxation is perfectly happy at `x = 0.5`, and here it is additionally *unbounded* (`min -y` over `y ∈ [0, 1e20]`). So the NS-safe root cross-check added for #1295/#1320 comes back `unbounded` — it settles nothing about the MILP's infeasibility — and the decertify rule refuses to pass HiGHS's raw `kInfeasible` through as a certificate.

`main` reports `infeasible` by trusting the tree label unchecked, which is exactly what #1295 and #1320 exist to stop. It is correct here; it is not verified here.

## The gap worth closing

An unbounded root LP combined with `kInfeasible` from the MILP is actually *informative*: it says the relaxation is feasible, so any genuine infeasibility must be integral. That is a decidable question, and deciding it directly would return `infeasible` **with** a proof — strictly better than both the current `error` and main's unverified label.

Possible directions:

- Check integer feasibility of the constraint system directly (a small auxiliary MILP/feasibility problem on the integer rows).
- Re-run the root check with the objective dropped, so an unbounded objective cannot make the relaxation check inconclusive when the question asked is only about feasibility.

The second looks cheap and well-targeted: the root cross-check here is being defeated by the *objective* being unbounded, not by anything about feasibility.

## Scope

- Pinned by `test_integral_infeasibility_under_an_unbounded_root_lp_is_not_certified` in `python/tests/test_1320_highs_milp_root_check_inconclusive.py`, which asserts the §1 line (never `optimal`, never `unbounded`) and accepts `infeasible` or `error`, so the behavior stays deliberate rather than drifting.
[...]

Issue #1336, opened 2026-09-18
POUNCE routes lose a correct `infeasible` when the elastic Phase-1 LP itself fails to solve


jkitchin:
Split out of the round-4 review of #1327 (finding 2), where it is explicitly accepted rather than fixed. Sound, but weaker than `main`.

## What happens

Since #1309/#1319 a raw Ipopt code 2 (`Infeasible_Problem_Detected`) is not trusted on its own — it is cross-checked against the elastic Phase-1 LP, because the barrier method raises code 2 from numerical failure on badly-conditioned problems with no infeasibility behind it.

When `_phase1_min_violation` **returns no point at all**, there is nothing to cross-check against, so the route reports an honest `error`.

Repro (QP default route; the LP route degrades to the exact simplex and is unaffected):

```python
import discopt.modeling as dm
L = 1e14
m = dm.Model("decoupled")
a = m.continuous("a", lb=0, ub=10)
b = m.continuous("b", lb=0, ub=10)
m.subject_to(a - b <= 0)
m.subject_to(a - b >= 1)          # contradicts the row above -> INFEASIBLE
z = m.continuous("z", lb=L, ub=2 * L)
v = m.continuous("v", lb=0, ub=2 * L)
m.subject_to(z - v <= 0)          # unrelated, satisfiable, activity ~1e14
y = m.continuous("y", lb=-1e20, ub=1e20)
w = m.continuous("w", lb=-1, ub=1)
m.minimize((w - 0.5) ** 2 - y)
print(m.solve(time_limit=60).status)
```

Traced on the #1327 branch:

```
raw _solve_qp_core     -> SolveStatus.INFEASIBLE
_phase1_min_violation  -> None          <- no point, nothing to verify against
FINAL                  -> error
```

`main` returns `infeasible`, certified. The problem genuinely *is* infeasible, so main's answer is correct **in fact** — but it is reached by trusting exactly the unverified code 2 that #1319 part 2 exists to stop trusting. The trade is deliberate: an honest `error` over an unchecked certificate.

## Why this is worth fixing properly

The trade costs real certificates on real models, and it does not have to. The constraint system is **linear** even on the QP route, so feasibility is an LP question — and discopt already ships an exact engine that honors the declared box and decides it: the Rust simplex.

[...]

Issue #1335, opened 2026-09-18
correctness: rust LP backend certifies `optimal` on an infeasible LP at ~1e15, returning a point that violates its own row


jkitchin:
Found while investigating a #1319 regression. **Pre-existing on `main`** — reproduced identically on `main` (`d4e0bb0`) and on the #1327 branch, so it is not introduced by that PR, but it is a live false certificate.

## Repro

```python
import discopt.modeling as dm
L = 1e15
m = dm.Model("s")
x = m.continuous("x", lb=L, ub=3 * L)
y = m.continuous("y", lb=L, ub=3 * L)
m.subject_to(x - y <= 1.0)
m.subject_to(x - y >= 10.0)     # conflicts with the row above
m.minimize(x + y)
print(m.solve(time_limit=60))
```

Run with `DISCOPT_LP_MILP_BACKEND=rust`:

```
STATUS optimal  obj 2000000000000005.2  certified True
POINT {'x': array(1.e+15), 'y': array(1.e+15)}
```

The system is infeasible: `x - y` cannot be both `<= 1` and `>= 10`. Two independent things are wrong in that one line:

1. **The status is a false certificate.** `optimal` with `gap_certified=True` on an infeasible problem is the exact failure CLAUDE.md §1 admits no slack on.
2. **The returned point does not satisfy the model.** At `x = y = 1e15`, `x - y = 0`, which violates `x - y >= 10`. The reported objective (`2000000000000005.2`) is not even `x + y` at that point (`2e15` exactly), so the point and the objective disagree with each other as well.

## Scope

| route | verdict |
|---|---|
| `DISCOPT_LP_MILP_BACKEND=rust` | **`optimal`, certified** ← wrong |
| default (HiGHS) | `infeasible`, certified ← correct |

Same on `main` and on the #1327 branch, both backends — verified by running the identical script against a pristine `origin/main` worktree.

Smaller-magnitude versions of the same conflict (`x + y <= 1` with `x + y >= 10` over `[0, 100]`, and over `[0, 1e16]`) are correctly reported `infeasible` on **both** backends, so the defect appears to need the variables *forced* to large magnitude — the conflict lives in a difference of two ~1e15 quantities while each variable's own box is satisfiable.

[...]

Issue #1334, opened 2026-09-18
[Adversary] #1324 follow-up: NaN warm start crashes nlp_bb; any warm start crashes solver=amp on GDP


jkitchin:
> Discovered by the discopt adversary agent (round 4, attacking the #1324 fix 173a75c9). Reproduced independently on main d4e0bb08.

Two warm starts still crash a solve. #1324's goal was that "a warm start can no longer crash a solve". No incorrect incumbent was ever returned in 1149 warm-start assertions; these are loud crashes only.

## 1. nlp_bb: a non-finite warm start crashes the injection block
```python
# y integer in [0,3], z in [0,4], y + z <= 5
r = m.solve(); r.x[...] = nan
m.solve(nlp_bb=True, warm_start=r)
# ValueError: cannot convert float NaN to integer   (round() at solver.py:~18207)
```
A raw `initial_point` containing `inf` raises `OverflowError` instead. The default route drops the same warm start cleanly. Root cause: `primal_point_from_result` accepts NaN (`np.clip` passes it through), but `validate_initial_solution` would reject it. The injection path should filter non-finite points before rounding, or reuse the validator.

## 2. `solver="amp"` on a GDP model: every warm start crashes
```python
x in [0,10]; m.either_or([[x <= 2], [x >= 6]]); m.minimize((x-4)**2)
m.solve(solver="amp", warm_start=r0)          # or initial_solution={x: 6}
# ValueError: AMP initial_point has length 1; expected 3
```
The selector columns added by the reformulation are never completed with `complete_initial_point`. This is the same class #1324 fixed on nlp_bb, on a route it did not cover. The default route solves the model correctly.

Probe: `adversary/runs/2026-09-18_r4_c_ws_crash_min.py` (plus `_r4_c_warmstart.py` for the 1149-assertion panel).

## Minor
`result.x` exposes the internal reformulation auxiliary `_fr_aux_0` to the user.

PR #1341, opened 2026-09-18
fix(correctness): #1339 follow-up — certify an unbounded MILP; revert the QP fallback

jkitchin:
## Summary

One of the two gaps PR #1339 left open is closed. The other was attempted,
measured, and **reverted in this same branch** — it made things worse, and the
commit history keeps the evidence rather than hiding the attempt.

Also included: a `CLAUDE.md` correction found while answering a question about the
skills subsystem.

### Closed — a genuinely unbounded MILP came back `error`

Certifying `unbounded` needs a verified recession ray of the relaxation AND an
integer-feasible point (Meyer: `rec(conv(S)) = rec(P)` for rational data with `S`
nonempty). HiGHS supplies the ray but reports `kUnbounded` with **no incumbent** —
it stopped because the objective ran away, not because the system is empty — so
the witness was missing only because nobody asked for it.

Asking is #1337's move again: drop the objective, keep the rows, the box and the
integrality, and solve for feasibility alone. On the repro that probe returns
`x = [1, 0]` immediately. The witness is re-checked here against the rows, box and
integrality before it may support a certificate, and is now carried on the result
rather than discarded. When the probe instead *proves* the integer system empty,
that resolves `kUnboundedOrInfeasible` the other way as `infeasible`.

### Reverted — the QP fallback (#359)

A HiGHS convex-QP backend was built so a feasible QP would stop returning `error`
when POUNCE's point fails the KKT-stationarity guard. It passed its own repro and
its own 8 tests. It did not survive the certification suite:
`test_bound_beyond_pounce_infinity_is_still_relaxed` (#1319, a QP with a declared
bound of 5e19) either **hangs** or takes the process down with a **floating-point
exception** — pytest exited 136 (128 + SIGFPE) on one run and 124 (timeout) on the
next, same test, same code.

Isolated to a standalone probe, one process per magnitude, on
`min 0.5*2*(w - 0.5)^2 + y` with `y in [-L, 0]`:

```
L=1e3    rc=0    OPTIMAL, obj -1000.25
L=1e12   rc=124  *** HANG ***
L=1e15   rc=124  *** HANG ***
L=1e18   rc=124  *** HANG ***
L=5e19   rc=124  *** HANG ***
```

HiGHS's QP solver is unusable from ~1e12 upward, which is precisely the class the
fallback exists for: POUNCE fails on those models *because* they are ill-scaled,
and handing them to an engine that hangs or dies on the same scaling fixes
nothing. A SIGFPE also cannot be contained from Python — the signal arrives inside
native code, so no timeout, `try`/`except` or validation on this side helps. Under
§1 that is strictly worse than the honest `error` it replaced.

(A second defect in the same code: `_solve_qp` has no `time_limit` parameter, so
the backend ran uncapped. That explains the hangs but not the crash, and fixing it
would not have made the backend safe.)

`_solve_qp` is POUNCE-only again, `solver.py` restored verbatim to its parent.
**The gap #1339 documented is still open and still accurate as written there.**
What a future attempt needs, recorded in the revert commit: a time limit actually
plumbed through, a measured scale envelope for HiGHS's QP solver, and a subprocess
to contain a native crash — and the envelope measured here excludes the models the
fallback was for, so whether it beats POUNCE-only is an open question.

### Docs

`CLAUDE.md` pointed at `.claude/commands/` for the Claude Code skills. That
directory does not exist — it is a *destination* the installer writes to, while
the files live in `python/discopt/skills/{commands,agents}/` and ship inside the
wheel. It also listed 6 of the 8 commands and never mentioned the 17 agent
personas. Replaced with the real paths, the four `install-skills` flags (verified
against `--help`), both inventories, the `skills/__init__.py` discovery API, and
the edit-the-package-not-the-copy rule that `--dev` exists for.

Closes nothing on its own; the QP half of #1339's follow-up remains open.

## Correctness

The MILP probe runs **only** where the route was already about to report `error`,
and adds evidence rather than assumption: the new `unbounded` carries a verified
recession ray plus an integer point re-checked here — strictly more than the
branch required before.

The probe keeps its own `root_check`: without it a `kInfeasible` exit would come
back `gap_certified` with no cross-check at all — the unchecked label #1295/#1320
exist to refuse — and this code would have handed it on as a proof of emptiness.
It also cannot recurse (a zero objective can never be unbounded), pinned by a test
asserting exactly one probe solve.

| repro | before | after |
|---|---|---|
| unbounded MILP, integer-feasible | `error` | **`unbounded` + witness** |
| MILP with a bounded objective (control) | `optimal` | `optimal` |
| integrally-infeasible MILP (#1337) | `infeasible` | `infeasible` |
| feasible convex QP, ill-scaled (#359) | `error` | `error` (unchanged — see above) |

- [x] `pytest -m smoke` → **1755 passed, 14 skipped**
- [x] `ruff check` / `ruff format --check` clean; `mypy python/discopt/` →
      **Success, no issues found in 350 source files**
- [x] Certification suites (#850 / #940 / #952 / #1319 / #1320 / #1334 / #1335 /
      #1336 / #1337 / #1339) → **162 passed, 1 deselected**

Every number above is from a run whose **own** exit code was captured. An earlier
round of this branch reported a passing certification suite that had not passed:
`pytest … | tail -4` returns *tail's* status, so the `exit 0` was the pipe, and the
four captured lines were a crash dump with no summary. That is what let the
reverted backend look green, and it is why the suites here are run unpiped.

## Regression test

`python/tests/test_1339_qp_fallback_and_unbounded_milp.py` — **7 passed**; the
end-to-end and probe-unit tests fail against the pre-fix tree, on the defect
itself. The controls (bounded MILP, #1337's integrally-infeasible MILP) pass
before and are labelled as guards. Its module docstring records the reverted QP
attempt and the measurements that killed it, so the next attempt starts from them.

N/A — no bound, cut, or relaxation is changed. The probe adds a decision where the
route previously refused to answer; it does not move a dual bound, so no §5 flag or
graduation panel applies.

https://claude.ai/code/session_01PytPusRx2DkEXZqCN1AQmT

---
_Generated by [Claude Code](https://claude.ai/code/session_01PytPusRx2DkEXZqCN1AQmT)_


jkitchin:
## Correction: this PR and #1339 both mis-cite #359 (CLAUDE.md §11 retraction)

Retracting a claim in **both** merged PR bodies. The code landed here is unaffected —
the QP fallback was reverted before merge — but the *reason* recorded for it is
wrong, and as written it points a future attempt straight back at the approach
that was already rejected.

### What both PRs claim

- #1339: "A missing-fallback-engine gap (#359), not a cross-check one."
- This PR: "Reverted — the QP fallback (**#359**) … The gap #1339 documented is
  still open."

### What is actually true

**#359 is closed**, completed on 2026-06-30 by **PR #361, "Make the QP path
HiGHS-free and remove qp_highs (issue #359)".**

So `python/discopt/solvers/qp_highs.py` is a module this project *deliberately
deleted*, and the reverted commit recreated it at the exact same path. #359's goal
was to get HiGHS **out** of the QP seam; the code comment "the QP route is
POUNCE-only by design, #359" means precisely what it says — that is the achieved
design, not a gap in it.

That makes the crash the *second* reason the revert was right. Even had the backend
been sound, landing it would have regressed #361.

### Where the misreading came from

I cited #359 from that code comment without reading the issue. Local history could
not have corrected me either — this clone is shallow (oldest commit 2026-09-05), so
June's removal is not in it. Reading the issue first would have.

### What the open question actually is

Not "add a second QP engine". #359 already evaluated that and states why a naive
flip is unsound: POUNCE returns interior-point multipliers rather than vertex
duals, and can drift on ill-conditioned problems while still reporting `OPTIMAL`
(#145). Its own **Option B — a pure-Rust QP (active-set over the existing simplex
basis machinery, or a small dense IPM) returning exact vertex/KKT duals** — is
named there as the clean long-term answer.

So the live question behind the symptom in #1339 (a feasible convex QP returning
`error` when POUNCE's point fails the KKT-stationarity guard) is Option B, or
improving the conditioning POUNCE is handed, or accepting `error` as the honest
outcome. **It is not reintroducing HiGHS**, and the measurements in the revert
commit (HiGHS's QP solver hangs or SIGFPEs from ~1e12 upward — exactly the
ill-scaled class the symptom lives in) are an independent argument against it.

The measurements themselves stand; only the `#359` framing around them is
withdrawn.

---

Issue #1351, opened 2026-09-19
AMP selects zero partition variables when the nonconvexity is concave squares in the objective, so the dual bound never moves off the root relaxation


jkitchin:
## Summary

`solver="amp"` returns `status="feasible"` with a 12.8% gap on a 3-variable concave QP, and **no amount of refinement changes the bound**. Instrumenting the partition-variable selector shows why: AMP picks an **empty** partition set on this model, so there is nothing for the refinement loop to refine — it then iterates to `max_iter` doing work that cannot possibly improve the bound.

No soundness problem. The incumbent is the true global optimum and AMP correctly declines to certify it rather than returning a false `optimal`. This is a bound-quality and wasted-work bug.

Found while auditing the README, where this exact model was the worked example for the **certified-global** path — a reader ran it and got `feasible` with a 12.8% gap. The README has been changed to a bilinear model that does certify (see the `docs(readme)` commit on `claude/kind-meitner-7mnb32`); this issue tracks the underlying behavior, which is unfixed.

## Reproduction

```python
import discopt.modeling as dm

m = dm.Model("concave_qp")
c = [-1.0, 0.5, 1.5]
xs = [m.continuous(f"x{i}", lb=-2.0, ub=2.0) for i in range(3)]
m.subject_to(sum(xs) >= -1.0)
m.subject_to(sum(xs) <= 3.0)
m.minimize(sum(-((xs[i] - c[i]) ** 2) for i in range(3)))  # concave

r = m.solve(solver="amp", rel_gap=1e-4)
print(r.status, r.objective, r.bound, r.gap)
# feasible -23.500000031248363 -26.500000000000103 0.12765955947340224
```

The bound is invariant under every refinement knob — only the runtime changes:

| knobs | status | bound | gap | wall |
|---|---|---|---|---|
| defaults | `feasible` | −26.500000 | 0.127660 | 1.6s |
| `max_iter=1000` | `feasible` | −26.500000 | 0.127660 | **24.9s** |
| `n_init_partitions=8` | `feasible` | −26.500000 | 0.127660 | 1.2s |
| `n_init_partitions=16, max_iter=500` | `feasible` | −26.500000 | 0.127660 | 11.8s |

## Evidence

[...]

PR #1353, opened 2026-09-19
fix(amp): #1351 let partition refinement reach a power with an affine base

jkitchin:
## Summary

Contributes to #1351 (the bound-quality defect; default-OFF, not graduated).

**Root cause.** `_build_power` lowers `w = t**p` into an aux carrying a sound 1-D envelope over the base's enclosure, then registers it in `monomial_map` only when the base is the bare `x_i`. That gate is *correct* — the key `(i, p)` **means** `x_i**p`, so `single_orig_col` must refuse `(2 x_i)**2`, which would claim `x_i**2`. But nothing else picked up a shifted base, so for `(x_i - c)**2` there was no channel at all: `_apply_partition_refinement` had nothing to attach the partition to and was a no-op for the term. AMP then iterated against a relaxation its own refinement could not touch, leaving the dual bound pinned at the root secant — `max_iter=1000` and `n_init_partitions=16` moved it by exactly `0.0`.

Measured, with partitions present and refining (`{0:6, 1:9, 2:6}`):

```
monomial_map for  -(x_i - c_i)^2   : []                        <- nothing to refine
monomial_map for  -(x_i^2) + 2c x  : [(0,2), (1,2), (2,2)]
```

Two models differing only by the constant `-sum c_i^2`. One certified; the other could not move its bound.

**Fix.** Record `(w, var, coeff, const, p)` for a power whose base is affine in ONE original, and refine it with `_add_piecewise_univariate` — the helper that already refines `f(coeff*x + const)` univariate intrinsics and takes the affine argument natively. Nothing is registered in `monomial_map`, so the soundness gate above is untouched. Odd `p` injects the `t == 0` inflection mapped back to x-space (`x = -const/coeff`) so no sub-interval straddles the curvature flip.

| | status | bound | gap |
|---|---|---|---|
| OFF | `feasible` | −26.500000 | 1.28e−01 |
| ON | **`optimal`** | −23.501600 | **6.81e−05** |

Also on this branch: `443f177`, the README correctness pass that PR #1349's merge left behind (that merge took only `7b52e94`, the dependency bump), and `9d558e7`, which registers the flag as a graduation-gate arm so it is actually measurable — `graduation_gate.py`'s own docstring describes unregistered parked flags as the reason "nothing ever graduates".

## Correctness

- [x] Cannot produce a false certificate (or: N/A — no solver-math change) — **this IS a solver-math change** (it tightens a relaxation), so it is argued rather than waived. Soundness rests on three things: the new path makes **no** claim on `monomial_map`, so `single_orig_col`'s gate is untouched; the refinement reuses `_add_piecewise_univariate`, the same sound piecewise envelope already used for univariate intrinsics with an affine argument; and the odd-`p` inflection is injected so no piece straddles the curvature change. Verified empirically below.

- [x] `pytest -m smoke` → **1759 passed, 14 skipped, 0 failed** (1755 before + the 4 new tests; flag default-OFF, so the legacy path is what smoke exercises)
- [x] `cargo test -p discopt-core` → **N/A** — no Rust touched
- [x] `pytest python/tests/test_amp*.py` → **185 passed**

## Regression test

- [x] Added a regression test that fails before / passes after — `python/tests/test_1351_affine_power_partition.py`, 4 tests. Verified fail-before by stashing only the source change: `test_affine_power_partition_certifies_shifted_squares` fails with `expected certification, got feasible`, and passes after. The other three (soundness, default-off, bare-monomial control) correctly pass in both states.

- [x] Default-OFF flag: `DISCOPT_AFFINE_POWER_PARTITION` (flag-OFF path unchanged — smoke is the witness)
- [x] Differential-bound test (new bound ≥ old AND ≤ true box optimum) + feasible-point sampling
- [ ] **Graduation panel evidence — NOT complete.** See below.

**Differential panel, 7 models, 0 violations.** No bound loosened, none above a true optimum, none above its own incumbent. Five controls (bare monomial, bilinear, convex affine square, odd power, quartic) identical ON vs OFF.

```
model                  OFF bound      ON bound     OFF obj      ON obj  verdict
shifted3              -26.500000    -23.501600  -23.500000  -23.500000  ok  [feasible->optimal]
shifted_scaled        -59.250000    -55.250001  -51.250001  -55.250001  ok  [feasible->optimal]
convex_affine_sq        0.000000      0.000000    0.000000    0.000000  ok  [optimal->optimal]
odd_power              -3.375000     -3.375000   -3.375000   -3.375000  ok  [optimal->optimal]
quartic               -33.191407    -33.191407  -33.191407  -33.191407  ok  [optimal->optimal]
bare_control          -20.001600    -20.001600  -20.000000  -20.000000  ok  [optimal->optimal]
bilinear_control      -10.000503    -10.000503  -10.000000  -10.000000  ok  [optimal->optimal]
```

`shifted_scaled` also finds a *better* incumbent (−51.250001 → −55.250001, the true optimum by vertex enumeration) — the OFF arm was returning a worse feasible point.

**Feasible-point / brute-force soundness check on the shapes most likely to break the new path** — odd powers (curvature flips at `t=0`) and **negative** `coeff` (arg interval flips relative to the x interval) — against grid optima on a 1201² grid: **5 cases, 0 violations**, no bound above its true optimum or its own incumbent. `mixed` shows real tightening (−36.625 → −35.938, still below true −28.375).

### Graduation status — explicitly incomplete

The flag is registered as a gate arm (`regime: bound_changing`) but **has not graduated and nothing here argues it should**. Splitting the gate honestly:

- **Cert-panel neutrality — running as of this writing, result to follow in a comment.** The 52-row vendored panel, its oracle and the self-calibrating harness are all in-repo, so this half *is* runnable here. I said earlier it was not; that was wrong and I'm correcting it.
- **Held-out benefit arm — not runnable here.** `generality_sweep` samples the ~4,800-instance MINLPLib snapshot from `~/Dropbox/projects/discopt-minlp-benchmark`, which does not exist on this container (verified). Missing dataset, not an incapable machine. The graduation-gate workflow header documents the same local-full vs CI-subset split.

**One thing the panel already shows, and it is not about this flag.** The OFF *control* run — flag off, i.e. stock behavior — reports **5 violations against the committed `cert-baseline.jsonl`**: `nvs05` objective drift `3.37e-08`, and node regressions on `nvs05` (+55%), `st_e36` (+7%), `tls2` (+210%), `tspn05` (+19%). This reproduces the staleness CLAUDE.md already records for that file, which is why the ON arm must be judged against the OFF control rather than against the committed baseline. Flagging it because a reader could otherwise mistake it for fallout from this PR.

Also unfixed, and still open on the issue: **defect 2** — with a relaxation refinement cannot touch, AMP still runs to `max_iter` instead of stopping when a round moves the bound by zero. Independent of this change.

https://claude.ai/code/session_01PA1kdGjwANiaCgpWwRBuya

---
_Generated by [Claude Code](https://claude.ai/code/session_01PA1kdGjwANiaCgpWwRBuya)_


jkitchin:
## Cert-panel neutrality: PASS, 52/52 rows byte-identical

Both arms run, on a quiet box, sequentially (not concurrently — the harness picks per-instance budgets by calibration, so competing load would make "did this row certify" a fact about the runner, CLAUDE.md §9). Host calibration came out at ×3.33 (OFF) and ×3.29 (ON) of the reference machine, so each row got the reference's *work* allowance rather than its seconds.

Row-level diff of the two arms, all 52 rows parsed from both logs:

```
rows parsed: OFF=52 ON=52 common=52
only in OFF: []
only in ON : []

ROW-LEVEL DIFFS (ON vs OFF): 0
STATUS changes: 0   OBJECTIVE changes: 0
COMPARED_ROWS=52
```

No status change, no objective change, no node-count change. The flag is **cert-neutral** on the panel.

### What this does and does not establish

It establishes **no regression**: turning the flag on loses no `optimal` status and changes no certified objective on the panel. That is the soundness bar the CI subset of the gate enforces.

It does **not** establish benefit, and — more importantly — 0 diffs is also what you would see if the panel simply never exercises the flag's structure. Those are very different readings of the same number, so I am measuring structural prevalence across the in-repo corpus separately rather than letting "0 diffs" stand as if the panel had validated the mechanism. Result to follow; if the prevalence is zero, this neutrality result is *trivially* satisfied and should be read as nothing more than "did no harm".

### The 5 violations in both arms are pre-existing, not this PR

Both arms report the *same* 5 violations against the committed `cert-baseline.jsonl`:

```
nvs05    [objective/soundness]   |Δobj|=3.370e-08 (5.470934075438167 -> 5.4709341091342365)
nvs05    [node_regression/perf]  node_count 149 -> 231 (+55%)
st_e36   [node_regression/perf]  node_count  83 ->  89  (+7%)
tls2     [node_regression/perf]  node_count 153 -> 475 (+210%)
tspn05   [node_regression/perf]  node_count  43 ->  51 (+19%)
```

Identical in the flag-OFF control, which is stock `main` behavior — so none of them are attributable to this change. This reproduces the staleness CLAUDE.md already records for `cert-baseline.jsonl` ("that reference is stale against `main`"), and is exactly why the arms are diffed against **each other** here rather than against the committed file. Both arms exit non-zero for that reason; the meaningful comparison is the 0-diff table above, not the exit code.

Oracle bracket held in both arms: 51 of 52 rows checked against the true optimum (`bound <= opt <= incumbent`), the 52nd (`tspn05`) having no oracle entry. No oracle crossing in either arm.

### Graduation status unchanged

Still **not graduated**, and this comment does not argue it should be. The held-out benefit arm needs the ~4,800-instance MINLPLib snapshot at `~/Dropbox/projects/discopt-minlp-benchmark`, which is absent on this container (verified). Cert-clean half: done, above. Net-positive half: not run.

---

jkitchin:
## Non-vacuity check: the panel DOES exercise the flag — and that is the bad news

Following up on the previous comment, which deliberately refused to let "52/52 identical" stand as evidence until I knew whether the panel exercises the structure at all.

**Structural prevalence, all 66 in-repo corpus instances scanned** (flag ON, counting instances whose lowered relaxation registers ≥1 affine-base power atom; 0 errors/skips):

```
instances scanned : 66  (errors/skips: 0)
instances registering an affine-base power atom: 4
    ex1222  (1 atom-registrations)
    nvs03   (4 atom-registrations)
    nvs04   (6 atom-registrations)
    nvs08   (9 atom-registrations)
```

**All 4 are in the 52-row cert panel** (checked against `cert-baseline.jsonl`: 4 of 4 present).

So the neutrality result is **non-vacuous** — the flag genuinely engaged on four panel rows and changed nothing. That is a real cert-clean result, not an accident of the panel missing the structure.

### The finding that matters for graduation

The flag engaged on 4 real instances and produced **bit-identical status, objective and node count on every one**. So:

> **Benefit is demonstrated only on synthetic models. On every real corpus instance that carries the structure, this flag is a no-op.**

This is exactly the pattern CLAUDE.md §"Working on an issue" cites as the #727 RLT lesson — *"a mechanism validated only on a synthetic proxy can be a no-op on the real class (synthetic root-gain 0.68, real gain 0.0)."* The synthetic gain here is dramatic (`feasible`/−26.5 → `optimal`/−23.5016, a 12.8% gap closed); the measured gain on `ex1222`/`nvs03`/`nvs04`/`nvs08` is **0.0**.

I am recording this against the flag rather than leaving the synthetic numbers to speak for it.

### What that does and does not change

- It does **not** undermine the root cause. The mechanism is real and specifically diagnosed: partitions present and refining, `monomial_map` empty, bound pinned at the root secant, and lifting the base by hand fixes it. That diagnosis stands on its own.
- It does **not** block this PR, which parks the flag default-OFF. The flag-OFF path is unchanged (smoke is the witness), and the README pass and the clippy/MSRV fix are independent of it.
- It **does** block graduation, harder than "we haven't run the held-out arm yet". A flag that is cert-clean but neutral-or-unmeasured on real instances is precisely the `DISCOPT_CUT_INHERIT` case CLAUDE.md §5 names: *sound ≠ helpful*, and such a flag **stays OFF**, with the measurement recorded.
- The honest next step for graduation is not "run the panel on a bigger machine" but **find whether any real class benefits at all**. If the 4 carriers in-repo are representative, the answer may be no — in which case the right outcome is to keep the flag parked (or remove it) and say so, not to chase a default-ON.

### Where that leaves the four checks

| gate half | status |
|---|---|
| cert-clean (panel neutrality, oracle bracket, `incorrect_count`) | **PASS**, and non-vacuous (4/4 carriers in-panel) |
| net-positive (benefit on real instances) | **NOT DEMONSTRATED** — 0 of 4 carriers improved |
| held-out arm (~4,800-instance snapshot) | not runnable here (dataset absent) |

Graduation checkbox stays unchecked, and on this evidence it should stay unchecked until a real class is shown to benefit.

---

